PR #80, opened 2021-11-10
Extracted WebSocketApplication.kt from Application.kt

dkovacevic:
By splitting this monster class Application we can have App that only sends messages and if we want to also receive them then WebSockApplication

### DIFF
--- a/src/main/java/com/wire/kalium/helium/LoginClient.kt
+++ b/src/main/java/com/wire/kalium/helium/LoginClient.kt
@@ -18,9 +18,11 @@
 package com.wire.kalium.helium
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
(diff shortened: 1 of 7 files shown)


src/main/java/com/wire/kalium/helium/LoginClient.kt
```diff
@@ -198,7 +200,7 @@ open class LoginClient(client: Client) {
             return response.readEntity(NotificationList::class.java)
         } else if (status == 401) {   //todo nginx returns text/html for 401. Cannot deserialize as json
             response.readEntity(String::class.java)
-            throw AuthException(status)
+            throw HttpException(message = null, code = status, label = "ohh")
```
MohamadJaara, line 203: the AuthException class was in xenon and got deletedbecouse is not used over there
i maby now it make sence to add it agian

Issue #111, opened 2021-11-18
split asset api


MohamadJaara:
split the following functions from the API.kt class using ktor as the http client:
- [x] #137
- [x] #169

PR #140, opened 2021-11-29
feat: add AssetsApi interface & implementation classes

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Implements issue #111 and adds the backbone of the `uploadAsset` method.
### Causes (Optional)

----
##### References


tmspzz:
> Error: Pull request has only one commit and it's not semantic; this may lead to a non-semantic commit in the base branch (see https://github.community/t/how-to-change-the-default-squash-merge-commit-message/1155). Amend the commit message to match the pull request title, or add another commit.

tmspzz:
Nitpick: 
> answers the question: If merged, this PR will: ... ³

So the title needs to be changes to something like `feat: add AssetsApi interface & implementation classes`

src/main/java/com/wire/kalium/api/assets/AssetsApi.kt
```diff
@@ -0,0 +1,12 @@
+package com.wire.kalium.api.assets
+
+import com.wire.kalium.api.KaliumHttpResult
+import io.ktor.client.statement.*
+
+interface AssetsApi {
+
+    suspend fun downloadAsset(assetKey: String, assetToken: String?): KaliumHttpResult<HttpResponse>
```
vitorhugods, line 8: Why wrap a `HttpResponse`?

Shouldn't it be something like `KaliumHttpResult<Asset>` where Asset contains mimetype and byte array, or similar?

Passing the raw HTTP result forward and exposing Ktor to whoever uses the lib is probably not wanted.

vitorhugods (commented):
A question about what should `downloadAsset` actually return.

src/main/java/com/wire/kalium/api/assets/AssetsApi.kt
vitorhugods: A more philosophical question is:

Suppose the asset is an image, how do users of this library can use it?

Something along these lines maybe? 

```kotlin
val asset = api.downloadAsset(...).assetStuff

asset.data // bytearray
asset.type // mpeg, jpeg, gif, png, pdf
```

gongracr: Yes you are right about the lack of specification of the response object type. However this was intended as the documentation in Swagger from this endpoint is incomplete and I couldn't get any insightful info from the Backend team. So I decided to leave it as an HttpResponse object in case we want to debug it and access to the information manually from the `CLIApplication` client

gongracr: So I guess this can still be considered as WIP until we make more sense out of the Swagger docs or the backend team replies back with a thorough explanation on how to execute this endpoint and what kind of object we should expect from it on the response

vitorhugods (approved):

PR #151, opened 2021-12-03
chore: add NetworkResponse to substitute KaliumKtorResult class

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In the legacy the `KaliumKtorResult` was not flexible enough to be able to wrap the execution of the current http client without leaking details of the Ktor implementation. This is fixed with the current approach.

As you can see in the code, the `wrapKaliumResponse` method will be the one mapping the result of the KtorHttpResult into the right KaliumException, removing the necessity of knowing about the Ktor library.  
This PR is just an attempt to minimize this friction as we evolve Kalium into an atomic library that can be consumed externally. It is therefore not a perfect solution because even though the new `wrapKaliumResponse` will "swallow" all possible exceptions thrown by the server, using wrongly the extensions methods from the `NetworkResponse` could also throw a `ClassCastException` when accessing a success value when an error was thrown instead.

----

----
##### References


network/src/commonMain/kotlin/com/wire/kalium/network/utils/NetworkUtils.kt
```diff
@@ -0,0 +1,50 @@
...
+import kotlinx.serialization.SerializationException
+
+sealed class NetworkResponse<out T> {
+    data class Success<out T : Any>(val response: HttpResponse, val body: T) : NetworkResponse<T>()
+    data class Error<out E : KaliumException>(val kException: KaliumException) : NetworkResponse<E>()
+}
+
+fun <T> NetworkResponse<T>.successValue(): T = (this as NetworkResponse.Success).body
+fun <T> NetworkResponse<T>.isSuccessful(): Boolean = this is NetworkResponse.Success
+fun <T> NetworkResponse<T>.responseHeaders(): String = (this as NetworkResponse.Success).response.headers.toMap().toString()
```
MohamadJaara, line 17: i think the headrs are something like Map<String, List<String>>

network/src/commonMain/kotlin/com/wire/kalium/network/exceptions/KaliumException.kt
```diff
@@ -1,27 +1,8 @@
...
-
-    constructor(
-        message: String?,
-        code: Int,
-        label: String?
-    ) : super(message, code, label)
+sealed class KaliumException : Exception() {
+    class RedirectError(code: Int, throwable: Throwable?): KaliumException()
+    class InvalidRequestError(code: Int, throwable: Throwable?): KaliumException()
+    class ServerError(code: Int, throwable: Throwable?): KaliumException()
```
MohamadJaara, line 6: server errors return something like this in the body for error code 401 - 499:
{
  "code": 403,
  "message": "The user has no verified identity (email or phone number).",
  "label": "no-identity"
}
so preserving this info is usefull
there is an ErrorResponse data class in legacy that you can use

network/src/commonMain/kotlin/com/wire/kalium/network/api/message/MessageApiImp.kt
```diff
@@ -36,13 +36,13 @@ class MessageApiImp(private val httpClient: HttpClient) : MessageApi {
         parameters: MessageApi.Parameters.DefaultParameters,
         conversationId: String,
         option: MessageApi.MessageOption
-    ): KaliumHttpResult<SendMessageResponse> {
+    ): NetworkResponse<SendMessageResponse> {
 
         suspend fun performRequest(
```
MohamadJaara, line 41: performRequest here can throw ktor Exceptions

network/src/commonMain/kotlin/com/wire/kalium/network/api/KtorHttpClient.kt
```diff
@@ -2,26 +2,17 @@ package com.wire.kalium.network.api
...
+import io.ktor.client.engine.*
+import io.ktor.client.features.*
+import io.ktor.client.features.auth.*
+import io.ktor.client.features.auth.providers.*
+import io.ktor.client.features.json.*
+import io.ktor.client.features.json.serializer.*
+import io.ktor.client.features.websocket.*
+import io.ktor.client.request.*
+import io.ktor.client.statement.*
+import io.ktor.http.*
```
MohamadJaara, line 15: i think the .editorconfig file is not working for some reason 
make sure it's enabled in settings -> Code Style -> Enable EditorConfig support

MohamadJaara (changes requested):

network/src/commonMain/kotlin/com/wire/kalium/network/api/KtorHttpClient.kt
vitorhugods: I think a different setting is msising in `EditorConfig`.

IntelliJ has these annoying settings by default:

grafik

When I've added the `.editorconfig`, IntelliJ didn't generate one that disabled them. I'll raise a PR with the missing `.editorconfig` property to fix this.

network/src/commonMain/kotlin/com/wire/kalium/network/api/message/MessageApiImp.kt
gongracr: But I thought we wanted to "swallow" these exceptions? Or at least map them to `KaliumExceptions` like objects, so that Ktor implementation details could be hidden to anyone that will consume Kalium

gongracr: I changed it to throw a KaliumException instead

network/src/commonMain/kotlin/com/wire/kalium/network/utils/NetworkUtils.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.wire.kalium.network.api.ErrorResponse
+import com.wire.kalium.network.exceptions.KaliumException
+import io.ktor.client.call.*
+import io.ktor.client.features.*
+import io.ktor.client.statement.*
+import io.ktor.util.*
+import kotlinx.serialization.SerializationException
+
+sealed class NetworkResponse<out T> {
+    data class Success<out T : Any>(val response: HttpResponse, val body: T) : NetworkResponse<T>()
```
vitorhugods, line 12: The name of this `response` value leads to `networkResponse.successValue().response`:

```kotlin
val networkResponse = loginApi.loginWithEmail("user@example.org", "password")
networkResponse.successValue().response
```

A response inside a response.

vitorhugods: Also, we should not expose Ktor's internals, like `HttpResponse`.

Let's assume a cookie or header is being accessed by `Android` and `iOS`  clients:

```kotlin
val response = loginApi.loginWithEmail("user@example.org", "password")
val zuidCookie = response.successValue().response.cookie["zuid"]
```

Now, we change our dependency on Ktor to Ktor 2.0, and the cookie API changes.

Both `Android` and `iOS`, that were directly accessing Ktor in order to get the cookie, suddenly have a breaking change, that wasn't directly made by us, but introduced by us.

I suggest encapsulating whatever we want to expose by ourselves and commit to it, for example:

```kotlin
sealed class NetworkResponse<out T>(val httpStatusCode: Int, val headers: Map<String, String>) {
    
    data class Success<out T : Any>(httpStatusCode: Int, val headers: Map<String,String>, val response: HttpResponse, val body: T) : NetworkResponse<T>(httpStatusCode, headers)
    data class Error<out E : KaliumException>(httpStatusCode: Int, val headers: Map<String,String> , val kException: KaliumException) : NetworkResponse<E>(httpStatusCode, headers)
}

```

After this, we commit on supporting and making these fields (httpStatusCode and headers, in this _example_) stable and there's no way someone could access Ktor directly. If new fields are needed, then we can update the library with new features.

network/src/commonMain/kotlin/com/wire/kalium/network/utils/NetworkUtils.kt
```diff
@@ -0,0 +1,54 @@
...
+    data class Error<out E : KaliumException>(val kException: KaliumException) : NetworkResponse<E>()
+}
+
+fun <T> NetworkResponse<T>.successValue(): T = (this as NetworkResponse.Success).body
+fun <T> NetworkResponse<T>.isSuccessful(): Boolean = this is NetworkResponse.Success
```
vitorhugods, line 17: Amazing opportunity to use Contracts:

```suggestion

@OptIn(ExperimentalContracts::class)
fun <T> NetworkResponse<T>.isSuccessful(): Boolean {
    contract {
        returns(true) implies (this@isSuccessful is NetworkResponse.Success)
        returns(false) implies (this@isSuccessful is NetworkResponse.Error)
    }
    return this@isSuccessful is NetworkResponse.Success
}

```

This allows the compiler to infer the type and unlocks amazing auto-completion in the IDE:

```kotlin
        if(response.isSuccessful()){
            response.body //Because of the contract, the compiler already knows that this must be a success 
        }else{
            response.kException //Because of the contract, the compiler already knows that this must be an error
        }
```

network/src/commonMain/kotlin/com/wire/kalium/network/utils/NetworkUtils.kt
vitorhugods: I just saw the functions below with the `asHttpResponseCode()` and `responseHeaders()`, so I guess we can use those and make `response` internal.

network/src/commonMain/kotlin/com/wire/kalium/network/utils/NetworkUtils.kt
```diff
@@ -0,0 +1,54 @@
...
+fun <T> NetworkResponse<T>.asHttpResponseCode(): Int = (this as NetworkResponse.Success).response.status.value
```
vitorhugods, line 18: Maybe a bit more of consistency in the naming? the prefix `as` is applied in here, but not in the `responseHeaders` for example.

Not sure if maintaining or removing would be the best. Personally, I would love to remove all of these casts and force the library user to check the response.

network/src/commonMain/kotlin/com/wire/kalium/network/utils/NetworkUtils.kt
vitorhugods: If we use this, maybe we could remove the `successValue()` and `errorValue()` casting functions. For quickly prototyping these functions are really useful, but they are a shortcut for mistakes, casting without knowing what's inside, and in any production application we would need to check every API call anyway.

network/src/commonMain/kotlin/com/wire/kalium/network/utils/NetworkUtils.kt
```diff
@@ -0,0 +1,54 @@
+package com.wire.kalium.network.utils
+
+import com.wire.kalium.network.api.ErrorResponse
+import com.wire.kalium.network.exceptions.KaliumException
+import io.ktor.client.call.*
+import io.ktor.client.features.*
+import io.ktor.client.statement.*
+import io.ktor.util.*
```
vitorhugods, line 8:
```suggestion
import io.ktor.client.call.receive
import io.ktor.client.features.ClientRequestException
import io.ktor.client.features.RedirectResponseException
import io.ktor.client.features.ResponseException
import io.ktor.client.features.ServerResponseException
import io.ktor.client.statement.HttpResponse
import io.ktor.util.toMap
```

vitorhugods (changes requested):
Overall, I love the idea of wrapping the exceptions and making it all strongly typed and guiding whoever is using this lib so everything is clearly either Success or Failure. So this PR is from my point of view very welcome!
Just some topics that I opened for discussion about what informations the `NetworkResponse` can expose and how can we model it so casting exceptions and the like can be prevented.

network/src/commonMain/kotlin/com/wire/kalium/network/utils/NetworkUtils.kt
gongracr: That was one of he inconveniences of using this kind of approach vs parsing manually the ktor result. However I LOVE these Contracts. I was not aware of their existance till now 🙏

gongracr: Now I'm wondering / playing around on how to integrate them when calling also `successValue()` and `errorValue()`

vitorhugods: I think something like this can work, but I don't see much value:

```kotlin

fun <T> NetworkResponse<T>.successValue(): Boolean {
    contract {
        returns() implies (this@isSuccessful is NetworkResponse.Success)
    }
    return this as NetworkResponse.Success
}


fun <T> NetworkResponse<T>.errorvalue(): Boolean {
    contract {
        returns() implies (this@isSuccessful is NetworkResponse.Error)
    }
    return this as NetworkResponse.Error
}
```

Basically: if it returns, it implies. The compiler will know about what it means, etc.
But if it doesn't return: 
# 💥 Boom

So we still need to check the value anyway... I think removing `successValueI()` or `errorValue()` for now isn't that bad.

`isSuccessful()` delivers a lot of value now and it's probably for the best that we don't provide unsafe casts in core functions of our library.

vitorhugods: The only way to prevent an `IllegalCastException` is by testing it. 

`errorValue()` and `successValue()` are not tests, they are assertions. So it will crash if it's not what the user expects.

`isSuccessful()` is one way of testing it. And by adding the contracts, the casts are done automatically by the compiler.

Instead of manually checking the type:
```kotlin
if(response is NetworkResponse.Success){ 
    showDataToUser(response.body) // Compiler knows about what this is
} else { 
    logAndHandleException(response.kException) 
}
```

We can make it like this:
```kotlin
if (response.isSuccessful()) {
    showDataToUser(response.body) // Compiler knows about what this is too, even though it's hidden behind another function
} else {
    logAndHandleException(response.kException)
}
```

gongracr: Hmm, I'm not entirely sold about the idea of adding yet another layer of checks. If we want to make sure calling `successValue/errorValue` will not throw yet another Exception (which was one of the main motivations for this change), I would suggest then to declare the return type as nullable `fun <T> NetworkResponse<T>.successValue(): T?`. 
That way the user will need to make sure of the validity of value returned before using the body of the response.

cli/src/main/kotlin/com/wire/kalium/cli/ConversationsApplication.kt
```diff
@@ -19,15 +20,22 @@ class ConversationsApplication : CliktCommand() {
...
-
-        println("Your conversations:")
-        conversations.forEach {
-            println("ID:${it.id}, Name: ${it.name}")
+        )
+
+        if (!loginResult.isSuccessful()) {
+            println("There was an error on the login :( check the credentials and the internet connection and try again please")
+        } else {
+            credentialsLedger.onAuthenticate(loginResult.value.accessToken, "") //TODO extract refresh token from cookie response
```
vitorhugods, line 28: This is only a comment, not a request for change.

This piece of code reminded me of one extra possibility using the contracts in `isSuccessful()`: Guard clauses.

```kotlin
if(!loginResult.isSuccessful()) {
    println("Failure to authenticate")
    return
}
loginResult.value.accessToken // The compiler allows it! If it was an error, this code would be reachable
```

We don't need to nest checks 🤩

network/src/commonTest/kotlin/com/wire/kalium/api/tools/json/api/prekey/PrekeyApiTest.kt
```diff
@@ -25,13 +25,13 @@ class PrekeyApiTest : ApiTest {
...
                 assertJson()
-                assertPathEqual(PATH_PREYKEY)
+                assertPathEqual(PATH_PREKEYS)
             }
         )
         val preKeyApi: PreKeyApi = PreKeyApiImpl(httpClient)
 
         val response = preKeyApi.getUsersPreKey(VALID_GET_USERS_PREKEY_REQUEST.serializableData)
-        assertEquals(response.resultBody, VALID_GET_USERS_PREKEY_RESPONSE.serializableData)
+        assertEquals(response.value(), VALID_GET_USERS_PREKEY_RESPONSE.serializableData)
```
vitorhugods, line 34: That does not compile.

This does:
```suggestion
        assertTrue(response.isSuccessful())
        assertEquals(response.value, VALID_GET_USERS_PREKEY_RESPONSE.serializableData)
```

vitorhugods (changes requested):
It seems that some new endpoints covered in#164 and #167 are still using `KaliumHttpResult`.

Also, tests are broken. A lot of them were asserting that some requests would fail, `SelfApiTest` for example, has this piece here:
```kotlin
        val error = assertFailsWith<ClientRequestException> { selfApi.getSelfInfo() }
        assertEquals(error.response.receive<ErrorResponse>(), ERROR_RESPONSE)
```

But now this needs to be replaced with:
```
        val result = selfApi.getSelfInfo()
        assertFalse(result.isSuccessful())
        assertEquals(result.kException.errorResponse, ERROR_RESPONSE)
```

cli/src/main/kotlin/com/wire/kalium/cli/ConversationsApplication.kt
gongracr: Yes I saw that. I nested it because I didn't want to return / halt the execution of the application, just in case we came up later with a mechanism of retry the login in case it fails. But can remove the if else if you prefer it that way. Not very opinionated tbh

vitorhugods: Not very opinionated for this one too. Just wanted to highlight that we do have this possibility going forward.

I'll mark this as resolved :)

network/src/commonMain/kotlin/com/wire/kalium/network/exceptions/KaliumException.kt
```diff
@@ -1,27 +1,16 @@
...
-
-@Serializable
-class AuthException : NetworkException {
-    constructor(message: String?, code: Int) : super(_message = message, _code = code)
-    constructor(code: Int) : super(_code = code)
+sealed class KaliumException(val errorCode: Int) : Exception() {
+    class RedirectError(val errorResponse: ErrorResponse, throwable: Throwable?) : KaliumException(errorCode = errorResponse.code)
+    class InvalidRequestError(val errorResponse: ErrorResponse, throwable: Throwable?) : KaliumException(errorCode = errorResponse.code)
+    class ServerError(val errorResponse: ErrorResponse, throwable: Throwable?) : KaliumException(errorCode = errorResponse.code)
+    class GenericError(val errorResponse: ErrorResponse?, throwable: Throwable?) : KaliumException(errorCode = errorResponse?.code ?: 400)
```
vitorhugods, line 10: These `Throwable`s are not being used.

For non-generic errors, like lack of connection, 
I don't think they are actually needed for non-generic errors.

But for Generic ones, I guess they are always caused by an underlying exception, so we should always add the throwable as cause and don't accept a `null` cause. Generic errors can be bad on their own, at least we need to make sure that we can know what caused them.

So, my suggestion:

```suggestion
sealed class KaliumException(val errorCode: Int) : Exception() {
    val errorCode: Int

    constructor(errorCode: Int) : super() {
        this.errorCode = errorCode
    }
    constructor(errorCode: Int, cause: Throwable) : super(cause) {
        this.errorCode = errorCode
    }

    class RedirectError(val errorResponse: ErrorResponse) : KaliumException(errorCode = errorResponse.code)
    class InvalidRequestError(val errorResponse: ErrorResponse) : KaliumException(errorCode = errorResponse.code)
    class ServerError(val errorResponse: ErrorResponse) : KaliumException(errorCode = errorResponse.code)
    class GenericError(val errorResponse: ErrorResponse?, cause: Throwable) : KaliumException(errorCode = errorResponse?.code ?: 400, cause = cause)
```

vitorhugods (changes requested):

network/src/commonMain/kotlin/com/wire/kalium/network/exceptions/KaliumException.kt
gongracr: Yep, sounds about right. I didn't want to get too extensive on this, but good point on taking our time now. Let me know if it looks good enough now

network/src/commonMain/kotlin/com/wire/kalium/network/exceptions/KaliumException.kt
```diff
@@ -4,10 +4,11 @@ import com.wire.kalium.network.api.ErrorResponse
...
 
 sealed class KaliumException(val errorCode: Int) : Exception() {
-    class RedirectError(val errorResponse: ErrorResponse, throwable: Throwable?) : KaliumException(errorCode = errorResponse.code)
-    class InvalidRequestError(val errorResponse: ErrorResponse, throwable: Throwable?) : KaliumException(errorCode = errorResponse.code)
-    class ServerError(val errorResponse: ErrorResponse, throwable: Throwable?) : KaliumException(errorCode = errorResponse.code)
-    class GenericError(val errorResponse: ErrorResponse?, throwable: Throwable?) : KaliumException(errorCode = errorResponse?.code ?: 400)
+    class RedirectError(val errorResponse: ErrorResponse) : KaliumException(errorCode = errorResponse.code)
+    class InvalidRequestError(val errorResponse: ErrorResponse) : KaliumException(errorCode = errorResponse.code)
+    class ServerError(val errorResponse: ErrorResponse) : KaliumException(errorCode = errorResponse.code)
+    class GenericError(val errorResponse: ErrorResponse?) : KaliumException(errorCode = errorResponse?.code ?: 400)
```
vitorhugods, line 10: Shouldn't we have the original exception for Generic errors, with a stacktrace and message, etc.?

network/src/commonMain/kotlin/com/wire/kalium/network/exceptions/KaliumException.kt
```diff
@@ -4,10 +4,11 @@ import com.wire.kalium.network.api.ErrorResponse
...
+    class NetworkUnavailableError(val errorResponse: ErrorResponse?, cause: Throwable?) : KaliumException(errorCode = errorResponse?.code ?: 400)
```
vitorhugods, line 11: This `cause` is not being used.

vitorhugods (changes requested):

vitorhugods (approved):

PR #173, opened 2021-12-14
feat: core logic module

### LINKED ISSUE #156: Create a module to contain the business logic

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Closes #156 

### Solutions

A new module that puts `cryptography` and `network` together in order to make it fly. The good old `2 + 2 = 5`.

Thinking about the issues with multiple user sessions, I tried to address it first, and create a base skeleton of how this business logic module can be expanded.

Here is an example of the proposed consuming-facing API:

```kotlin
val clientLabel = "Kalium Sample App - Google Pixel 6 Pro on Android 12"
val rootProteusDirectoryPath = context.getDir("proteus", MODE_PRIVATE).path
val core = CoreLogic(clientLabel, rootProteusDirectoryPath)

// Inside the authenticationScope, you can manage and authenticate user sessions
val newUserSession = core.authenticationScope {
    // Logs in and stores a new user session
    val loginResult = loginUsingEmail("username@example.com", "password", shouldPersistClient = false)

    if (loginResult !is LoginUsingEmailUseCase.Result.Success) {
        // Handle errors
    }

    // Gets a list of authenticated user sessions in storage
    getSessions()

    loginResult.userSession // new user session will be returned to the outside of this scope
}

// With the new user session, you can get a sessionScope to operate on a specific user
core.sessionScope(newUserSession) {
    val firstConversation = conversations.getConversations().first()

    messages.sendTextMessage(conversationList.first().id, "Hello everyone")

    // Some other operations can take place here:
    user.getProfile()
    user.updateProfile(name = "John Doe")
    contacts.getClients("contactUserId")
    preferences.isDarkThemeOn()
}

```

### Testing

Not done yet

----

----
##### References


typfel:
Trying to understand the concepts involved here:

**CoreLogic**

Top level entry point for accessing top level scopes `AuthenticationScope` and `UserSessionScope`

**Scope**

A collection of use cases or sub scopes.

- This is public API from the library

**UseCase**

High level use cases which the the consumer of the library will invoke. This is done by combining one or more methods from repositories.

Example:  send a message in a conversation.

- This is public API from the library

**Repository**

Example: ConversationRepository

- Uses the network API 
- Maps from network layer entities to model layer entities
- Integrates with persistence layer

vitorhugods:
> Trying to understand the concepts involved here:
> 
> **CoreLogic**
> 
> Top level entry point for accessing top level scopes `AuthenticationScope` and `UserSessionScope`
> 
> **Scope**
> 
> A collection of use cases or sub scopes.
> 
> - This is public API from the library
> 
> **UseCase**
> 
> High level use cases which the the consumer of the library will invoke. This is done by combining one or more methods from repositories.
> 
> Example:  send a message in a conversation.
> 
> - This is public API from the library
> - Contains business logic
> 
> **Repository**
> 
> Example: ConversationRepository
> 
> - Uses the network API 
> - Maps from network layer entities to model layer entities
> - Integrates with persistence layer

You nailed it and I couldn't have described it better in so few words.

Good that my code managed to get that across!

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/GetSessionsUseCase.kt
```diff
@@ -0,0 +1,9 @@
+package com.wire.kalium.logic.feature.auth
+
+import com.wire.kalium.logic.data.session.SessionRepository
+
+class GetSessionsUseCase(private val sessionRepository: SessionRepository) {
+
+    suspend operator fun invoke(): List<AuthSession> = sessionRepository.getSessions() //TODO Any kind of failure possible?
```
typfel, line 7: Might contain expired sessions? But I guess this will generate an error later when there's an attempt to use it.

logic/src/commonMain/kotlin/com/wire/kalium/logic/AuthenticatedDataSourceSet.kt
```diff
@@ -0,0 +1,9 @@
+package com.wire.kalium.logic
+
+import com.wire.kalium.cryptography.ProteusClient
+import com.wire.kalium.network.AuthenticatedNetworkContainer
+
+class AuthenticatedDataSourceSet(
```
typfel, line 6: This will later also hold a reference to an on disk persistence client?

vitorhugods: Something like that I guess.


I didn't think exactly how persistence can be packed, I guess one local persistence client/manager is necessary to perform DB migrations, provide independent scope per session, etc.


At some point we inject those into the repositories and have fun.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/GetSessionsUseCase.kt
vitorhugods: Could be. 

I didn't plug in session expiration yet.

How that will work is up to us I guess. Maybe we can filter it out and call it `GetValidSessionsUseCase`, or remove them a
completely once we detect they're expired.

typfel:
Do you have ideas already how deal requirements like:
- We only want to fetch incoming events after we've slow synced all users, conversations, etc...
- We only want to send messages after we've synced & processed all incoming events.

Should these rules be enforced:
- Inside the UseCases with some shared object which manages this global state?
- Does it make sense to have a scopes like `SynchronizingScope` `OnlineScope`, where `SynchronizingScope` doesn't expose the function for sending messages but `OnlineScope` does.

vitorhugods:
Very nice questions.


>     * We only want to fetch incoming events after we've slow synced all users, conversations, etc...
> 
>     * We only want to send messages after we've synced & processed all incoming events.
> 
> 
> How should these rules be enforced?
> 
>     * Inside the UseCases with some shared object which manages this global state?
 
For these three topics, I guess we could have a `SyncManager` (or similarly named) that UseCases can call and check Sync status before doing operations, or just enqueue operations to be done after sync.

Something like this, maybe?

```kotlin
class CalculateTwoPlusTwoUseCase(private val syncManager: SyncManager) {
    suspend operator fun invoke(): Result {

        // syncManager allows you to get the current sync status
        syncManager.syncStatus // IDLE, RUNNING, or DONE? Maybe get specific with Quick/Full Sync?

        // It also exposes a function that blocks and runs an action after sync is DONE. If IDLE, can start Sync.
        val calculationSyncResult = syncManager.onSyncCompleted {
            2 + 2
        } // The result can be Success, or failure in case Sync fails (Internet drops for example?).
        return if (calculationSyncResult.isSuccess()) {
            Result.Success(calculationSyncResult.value)
        } else { // Sync fails? Like internet going down during sync?
            Result.Failure("Could not complete sync in order to fetch foo")
        }
    }
}
```

>     * Does it make sense to have a scopes like `SynchronizingScope` `OnlineScope`, where `SynchronizingScope` doesn't expose the function for sending messages but `OnlineScope` does.

It's something we all can discuss, on what's the responsibility of the core business logic.
My gut feeling is that these scopes should not be exposed to whoever is consuming this lib. Sync, for example, is part of the business logic and our use cases can care about it, trigger it and respond in case of failure.

If a use-case needs sync done in order to succeed, then we can program it with this in mind, not the consumers of this API.

typfel:
> Very nice questions.
> 
> > ```
> > * We only want to fetch incoming events after we've slow synced all users, conversations, etc...
> > 
> > * We only want to send messages after we've synced & processed all incoming events.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > How should these rules be enforced?
> > ```
> > * Inside the UseCases with some shared object which manages this global state?
> > ```
> 
> For these three topics, I guess we could have a `SyncManager` (or similarly named) that UseCases can call and check Sync status before doing operations, or just enqueue operations to be done after sync.
> 
> Something like this, maybe?
> 
> ```kotlin
> class CalculateTwoPlusTwoUseCase(private val syncManager: SyncManager) {
>     suspend operator fun invoke(): Result {
> 
>         // syncManager allows you to get the current sync status
>         syncManager.syncStatus // IDLE, RUNNING, or DONE? Maybe get specific with Quick/Full Sync?
> 
>         // It also exposes a function that blocks and runs an action after sync is DONE. If IDLE, can start Sync.
>         val syncResultThatMayOrMayNotContainTheCalculatedValue = syncManager.onSyncCompleted {
>             2 + 2
>         } // The result can be Success, or failure in case Sync fails (Internet drops for example?).
>         return if (syncResultThatMayOrMayNotContainTheCalculatedValue.isSyncComplete()) {
>             Result.Success(syncResultThatMayOrMayNotContainTheCalculatedValue.value)
>         } else { // Sync fails? Like internet going down during sync?
>             Result.Failure("Could not complete sync in order to fetch foo")
>         }
>     }
> }
> ```

I think this seems like a reasonable solution. 👍 

>  > ```
> > * Does it make sense to have a scopes like `SynchronizingScope` `OnlineScope`, where `SynchronizingScope` doesn't expose the function for sending messages but `OnlineScope` does.
> > ```
> 
> It's something we all can discuss, on what's the responsibility of the core business logic. My gut feeling is that these scopes should not be exposed to whoever is consuming this lib. Sync, for example, is part of the business logic and our use cases can care about it, trigger it and respond in case of failure.
> 
> If a use-case needs sync done in order to succeed, then we can program it with this in mind, not the consumers of this API.

Yep agree, agree exposing this to the consumer facing API probably not a good idea.

typfel (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/GenericFailure.kt
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * Server internal error, or we can't parse the response,
+     * or anything API-related that is out of control from the user.
+     * Either fix our app or our backend.
+     */
+    object ServerMiscommunication : GenericFailure()
+
+    class UnknownFailure(val rootCause: Throwable?): GenericFailure()
+
+}
```
gongracr, line 19: Could we somehow reuse the `GenericError` class from the network module?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUsingEmailUseCase.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    sealed class Result {
+        class Success(val userSession: AuthSession) : Result()
+
+        sealed class Failure : Result() {
+            object InvalidCredentials : Failure()
+            class Generic(val genericFailure: GenericFailure) : Failure()
+        }
+    }
```
gongracr, line 26: Does it make sense to create again an equivalent of the already created class `NetworkResponse` or should we reuse it?

network/src/commonMain/kotlin/com/wire/kalium/network/api/RefreshToken.kt
```diff
@@ -0,0 +1,5 @@
+package com.wire.kalium.network.api
+
+object RefreshTokenProperties {
+    const val COOKIE_NAME = "zuid"
```
gongracr, line 4: Shouldn't the key value still be `cookie` according to the BE docs?

Screenshot 2021-12-17 at 16 17 57

android/src/main/kotlin/com/wire/kalium/presentation/MainActivity.kt
```diff
@@ -1,65 +1,65 @@
...
-        val bob = SampleUser("bobId", "Bob")
-        val bobFile = File(rootProteusDir, bob.id)
-        bobFile.mkdirs()
-        val bobClient = ProteusClient(rootProteusDir.absolutePath, bob.id)
-        val bobSessionId = CryptoSessionId(UserId(bob.id), CryptoClientId("bobClient"))
-        bobClient.open()
+        setContent {
+            val conversations = runBlocking {
+                val rootProteusDirectoryPath = getDir("proteus", Context.MODE_PRIVATE).path
+                val core = CoreLogic(clientLabel, rootProteusDirectoryPath)
```
gongracr, line 25: I know we are still in exploratory phase, but should we move this core logic singleton instance creation to the `AndroidApplication` level so that it´s visible everywhere?

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -0,0 +1,46 @@
...
+            val proteusClient = ProteusClient(rootProteusDirectoryPath, session.userId)
+            AuthenticatedDataSourceSet(networkContainer, proteusClient).also {
+                userScopeStorage[session] = it
+            }
+        }
+        return UserSessionScope(session, dataSourceSet)
+    }
+
+    suspend fun <T> authenticationScope(action: suspend AuthenticationScope.() -> T)
+            : T = getAuthenticationScope().action()
```
gongracr, line 41: Nicee ✨

logic/src/commonMain/kotlin/com/wire/kalium/logic/GenericFailure.kt
vitorhugods: We could. Personaly, I don't think we should.

`GenericError` from network module are exactly that, generic errors from **network** module.

If, for example, we have generic DB errors (like a missing migration, or corrupted db file), the **network** errors have nothing to do with that.

And even if we forget about DB and other kinds of data sources, some of the `BadRequest` and `Serialization` errors in `Network` can be treated as `generic` from of the business logic. So some kind of mapping is needed anyway. For example:

Serialization error? Bad Request? Server returning 500? Our users don't know, they don't care. They all are `ServerMiscommunication` and they all mean something along the lines of "Our services are not working properly now, please try again later".

From what I see, this is business logic, this is how we should treat users. And the logic for errors should be completely detached from the network or any other data source.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUsingEmailUseCase.kt
vitorhugods: Related to the topic above.

network/src/commonMain/kotlin/com/wire/kalium/network/api/RefreshToken.kt
vitorhugods: The the cookie we need to read when receiving the credentials is called `zuid`. So the name here is for unpacking the cookie from the response.

If we want to use the cookie later for some requests, it's called `cookie` 🤷🏼‍♂️

android/src/main/kotlin/com/wire/kalium/presentation/MainActivity.kt
vitorhugods: I'll change it to go this way.

Maybe later if we plug some DI into this, we can change it again.

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
vitorhugods: I have a thing for domain-specific language 🤷🏼‍♂️

android/src/main/kotlin/com/wire/kalium/presentation/MainActivity.kt
vitorhugods: @MohamadJaara is doing this in the sample app already, so I'll leave it to him 🙈

PR #177, opened 2021-12-20
feat: add Either classes for internal usage in `logic` module

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

For splicit success/failure return values and easy mapping of results, the something like `Either` can be employed, at least internally in the `logic` module.
It can assist us to tie the different pieces / data sources together in a strongly typed way.

### Solutions

Brought up the `Either` class from Android Reloaded, with a small tweak: `isLeft` and `isRight` are not functions and benefit from the kotlin contracts.

### Testing

Included

----

----
##### References

### DIFF
--- a/android/src/main/kotlin/com/wire/kalium/presentation/MainActivity.kt
+++ b/android/src/main/kotlin/com/wire/kalium/presentation/MainActivity.kt
@@ -10,7 +10,7 @@ import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import com.wire.kalium.logic.CoreLogic
 import com.wire.kalium.logic.data.conversation.Conversation
-import com.wire.kalium.logic.feature.auth.LoginUsingEmailUseCase
+import com.wire.kalium.logic.feature.auth.AuthenticationResult
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.runBlocking
 
@@ -28,7 +28,7 @@ class MainActivity : ComponentActivity() {
                         email = "username@example.com", //TODO Form or something to get user input
                         password = "password",
                         shouldPersistClient = false
-                    ) as LoginUsingEmailUseCase.Result.Success // assuming success
+                    ) as AuthenticationResult.Success // assuming success
                 }.userSession
 
                 val conversationList = core.sessionScope(session) {
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreFailure.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreFailure.kt
@@ -1,19 +1,20 @@
 package com.wire.kalium.logic
 
-sealed class GenericFailure {
+sealed class CoreFailure {
     /**
      * Failed to establish a connection with the necessary servers in order to pull/push data.
      * Caused by weak - complete lack of - internet connection.
      */
-    object NoNetworkConnection : GenericFailure()
+    object NoNetworkConnection : CoreFailure()
 
     /**
      * Server internal error, or we can't parse the response,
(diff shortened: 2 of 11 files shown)


MohamadJaara (approved):

gongracr:
Even though it seems with the `Either` class we cover a lot of error handling ground, I still want to point a couple of minor nits about it:
- What would happen when a request takes longer than usual, i.e. the result can't be directly mapped to an `Error`- `Success` state but rather to a a Loading state (Idle, inProgress, Done)? Would we then have to add an extra layer of state abstraction on top of the `Right` element?
- So atm, the normal flow of execution vs mapping type of results is: 
 
```
Api Request --mapping to-> NetworkResponse<BodyType>
       --- isSuccesful? --
              No -- mapping from KaliumException to-> CoreFailure--> mapping to Either<CoreFailure, Nothing>
              Yes -- mapping to-> Either<Nothing,BodyType>
```

I feel that somehow this whole abstraction is a bit too complex to grasp at first (at least for me that I was not familiar with the `Either` concept) and maybe could benefit of an easier way of encapsulating these results. Or what do you think?

In any case, great job on this Core Logic module so far :)

vitorhugods:
> Even though it seems with the `Either` class we cover a lot of error handling ground, I still want to point a couple of minor nits about it:
> 
>     * What would happen when a request takes longer than usual, i.e. the result can't be directly mapped to an `Error`- `Success` state but rather to a a Loading state (Idle, inProgress, Done)? Would we then have to add an extra layer of state abstraction on top of the `Right` element?
> 
>     * So atm, the normal flow of execution vs mapping type of results is:
> 
> 
> ```
> Api Request --mapping to-> NetworkResponse<BodyType>
>        --- isSuccesful? --
>               No -- mapping from KaliumException to-> CoreFailure--> mapping to Either<CoreFailure, Nothing>
>               Yes -- mapping to-> Either<Nothing,BodyType>
> ```
> 
> I feel that somehow this whole abstraction is a bit too complex to grasp at first (at least for me that I was not familiar with the `Either` concept) and maybe could benefit of an easier way of encapsulating these results. Or what do you think?
> 
> In any case, great job on this Core Logic module so far :)

@gongracr Yeah, it's a pain in the ass that I don't like too much too. And I am more than happy to discuss.
Here's my view on it:

I try to picture it in a way that each module establishes a contract with whoever is consuming, and it says "These are the possible outcomes. That's it.".

By trying to not expose underlying dependencies and issues that may surface on these underlying layers, this mapping of failures is created.

The `CoreFailure` and `Either` part is to help to tie it all together for the `core` library. The internal mechanisms can fail (as `CoreFailure`) and we can handle these using `Either` internally, but when it comes to expose that to the outside world (to the app consuming the core), we – as core maintainers – can branch it down to a subset of possibilities for each use-case.

So, maybe, a way of visualising this is: 

Untitled Diagram drawio

Where the exceptions are emitted on top and cascade down, being handled and mapped as needed before reaching (or not) the UI.

Some UseCases can have the same possible outputs and exceptions. For example `SendTextMessage`, `SendPingMessage`, etc. and other kinds of messages: the same mapping can be done for all of them. Same for "fetch resource" kind of use-case, like "Get Conversation Details" and "Get User Profile", the same exception mapping and handling can be used.

gongracr (approved):
Ok, fair enough. This is not a trivial project, and it's true that there can be heaps of different useCases that will require different ways of handling internal and external errors. I was just trying to find a way to simplify / smooth some edges on the way these mechanisms are wired internally, but I guess that can be done on a individual basis, once we have all these tools in place. 

As I said, gr8 job auf jeden Fall! 🍰 🌞

PR #178, opened 2021-12-22
feat(core logic): add register client usecase

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

#173 

### Testing

WIP

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
```diff
@@ -0,0 +1,85 @@
...
+        lastKey = preyKeyMapper.toPreKeyDTO(param.lastKey),
+        label = param.label,
+        deviceType = toDeviceTypeDTO(param.deviceType),
+        type = toClientTypeDTO(param.type),
+        capabilities = param.capabilities?.let { it.map { toClientCapabilityDTO(it) } } ?: run { null },
+        model = param.model,
+        preKeyDTOS = param.preKeys.map { preyKeyMapper.toPreKeyDTO(it) },
+    )
+
+    fun fromRegisterClientResponse(response: RegisterClientResponse): Client = Client(
```
tmspzz, line 28: would this make sense and extension function on RegisterClientResponse?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.wire.kalium.network.api.user.client.DeviceTypeDTO
+import com.wire.kalium.network.api.user.client.RegisterClientRequest
+import com.wire.kalium.network.api.user.client.RegisterClientResponse
+
+class ClientMapper(
+    private val preyKeyMapper: PreyKeyMapper,
+    private val locationMapper: LocationMapper
+) {
+
+    fun toRegisterClientRequest(param: RegisterClientParam): RegisterClientRequest = RegisterClientRequest(
```
tmspzz, line 17: would this make sense as an extensions function on RegisterClientParam ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientModel.kt
```diff
@@ -0,0 +1,72 @@
...
+    LegalHold;
+
+    override fun toString(): String {
+        return this.name.lowercase()
+    }
+}
+
+enum class ClientCapability {
+    LegalHoldImplicitConsent {
+        override fun toString(): String {
```
tmspzz, line 68: In general, not specifically in this case, I'm confused about the use of `toString`. It this some sort of string representation of a client capability? 

In Swift for example there isn't a `toString` method. But rather:

- `description: String`. Used for [describing an object](https://developer.apple.com/documentation/swift/customstringconvertible) as a String and called when String interpolation takes place
- `debugDescription: String`. [This is self explanatory](https://developer.apple.com/documentation/swift/customdebugstringconvertible) 
- `as? String` to cast something, to String. The cast will fail if the object is not a String.

There are also other ways that achieve [slightly](https://developer.apple.com/documentation/swift/losslessstringconvertible/) different [things]((https://developer.apple.com/documentation/swift/expressiblebystringliteral))

None of these to, are "magically" added to any object.

tmspzz (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientModel.kt
vitorhugods: @tmspzz `toString` is kind of in a no-man's-land. It is something Kotlin inherited from Java and it may be considered harmful for some people.

Following Kotlin's documentation on it:
> Returns a string representation of the object. Can be called with a null receiver, in which case it returns the string "null".

There's no specifically-set purpose for it. But I'd say it's closer to swift's `description`, and some people will use it in a `debugDescription` way. 

The default implementation of `toString()` in Kotlin:
* For classes will return something like: `MyClass@1eed786` where `1eed786` is the memory address where the object is stored. Of course you can inherit the `toString` from parent classes.
* For primitives: their value,
* For `enum`s: their name as written in code.

The `as? String` is the same in Kotlin as in Swift. It will crash if the thing you're trying to cast isn't a String.

vitorhugods: For this case, I understand that `toString` is being overriden to match the API and work the (de)serialization, which for an `enum` created solely for this, seems reasonable.

BUT. I think making it clear why the `toString` is being overriten is always great, and overwriting it for each possible enum value can be too verbose.

Instead of:

```kotlin
enum class Make {
    AlfaRomeo {
        override fun toString(): String = "alfa-romeo"
    },
    Ferrari {
        override fun toString(): String = "ferrari"
    },
    Fiat {
        override fun toString(): String = "fiat"
    }
}
```

This can be done:

```kotlin
@Serializable
enum class Make {
    @SerialName("alfa-romeo") AlfaRomeo,
    @SerialName("ferrari") Ferrari,
    @SerialName("fiat") Fiat
}
```

No need to alter how the default `toString()` works, less code, and makes it clear why the different name exists.

tmspzz: `toString` definitely harmful in my opinion

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
vitorhugods: I have mixed feelings about both approaches.

When dealing with mapping of data on Reloaded, we used to test that:
1. Mapper does it's job as a Unit (throw different inputs to it, assert the outputs)
2. Repository called the mapper (or didn't, in case of failures when fetching from data source)
3. Mapper received the correct values from the repository
4. The repository used/returned the mapped values

----

If we want to keep doing that, I see that we have the options:

A. We could mock and stub the extension functions as @tmspzz is suggesting. But there's a compromise: we cannot mock things on iOS. So our mapper-related tests would need to run only on `jvm` and `android` for example.


| Pro                                                                                                  | Con                                                                                     |
|------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------|
| Easier to use in code                                                                                | A bit harder to test, it's possible to mock static/extension calls. Not impossible tho. |
|  | Only testable in JVM and Android                                                        |



B. Just using `ClientMapper` as a concrete class isn't much of an improvement on this regard, as we would need to extend it and create our own fakes in order to test it on all targets. 

| Pro                | Con                              |
|--------------------|----------------------------------|
| Easier to test     | More boiler-plate in code        |
| Clear dependencies | Only testable in JVM and Android |

C. Have a `ClientMapper` interface, and a `ClientMapperImpl`. We can implement fakes of `ClientMapper` during tests for all platforms.

| Pro                     | Con                                                                                    |
|-------------------------|----------------------------------------------------------------------------------------|
| Testable in all targets | Probably the most annoying to test, with more boiler-plate and extensive lines of code |
| Clear dependencies      | A bit more annoying to implement too, as it needs Interface and `Impl` classes         |

D. ? Other ideas ?
----

Between A and B I don't see much of a difference. Neither would allow us to test on all targets. 

If we want to test repositories (as unit, separate from the mappers) and mappers like we used to, C is a necessary evil.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -21,6 +26,14 @@ class UserSessionScope(
...
         get() = MessageRepository(authenticatedDataSourceSet.authenticatedNetworkContainer.messageApi)
 
+    private val preyKeyMapper: PreyKeyMapper get() = PreyKeyMapper()
+    private val locationMapper: LocationMapper get() = LocationMapper()
+    private val clientMapper: ClientMapper get() = ClientMapper(preyKeyMapper, locationMapper)
+
+    private val clientRepository: ClientRepository
+        get() = ClientRepository(authenticatedDataSourceSet.authenticatedNetworkContainer.clientApi, clientMapper)
+
+    val clientScope: ClientScope get() = ClientScope(clientRepository)
```
vitorhugods, line 36: Can we rename this to `client`, or rename the other two to `conversationScope` and `messageScope` ?

When using DSL, adding the `scope` suffix sounds a bit redundant.

Maybe it could even be removed from the `sessionScope(userSession) { ... }` too and make it:

```
core.session(userSession) {
    client.register()
}
```

vitorhugods (changes requested):
Just continued on @tmspzz's discussion and tiny change suggestions regarding naming and serialization name.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
tmspzz: If it's customary to have separate *Mapper* classes, don't mind my comments. In Swift I would have probably just made `extensions` in the appropriate files.

Does option B. Take into account that we can subclass it and override whatever we need to in our fakes?

vitorhugods: Option B was making only one Mapper class, and having it final (which is the default in Kotlin). Mocking libraries allow us to stub them anyway, even if the classes are final. But Mocking libraries = no iOS.

We could, perhaps, do an option D, which would be open Mapper classes instead of final. No need for the `Interface`+`Impl` classes, and we can write our own fakes without mocking libraries.

Makes implementation a bit simpler, but tests stay verbose 🤷🏼‍♂️

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientModel.kt
MohamadJaara: using only '''@SerialName("alfa-romeo")''' will not work in our serlization tests, since when creating the json string kinda manually e.g. ''' "make": "${Make.AlfaRomeo}" ''' the result string will be ''' "make": "AlfaRomeo" ''' and it will differ from the  '''@SerialName("alfa-romeo")'''. 
so we need the '''toString''' of an enum, which is it's name in kotlin, to match the '''@SerialName'''.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
MohamadJaara: make sense 👍🏼
i will change it to client

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientModel.kt
vitorhugods: I guess I shot myself on the foot a few of weeks ago when I did this `JsonProvider` thing 😿 

Definitely an improvement point to think about regarding `JsonProvider`.

MohamadJaara: we can remove the `toString` from the enums and change the json tests to  to something like `"make": "alfa-romeo"`

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
gongracr: If we want to take logic as the only entry point for external consumers of Kalium, we should be quite vigilant on the visibility methods we qualify/declare our network objects

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
gongracr: I think for the sake of simplicity, we should unit test then these mapper classes in isolation inside the `commonTest` folder (1st bullet point). Any other interaction of this mapper class with repositories that we want to test, should live under `AndroidTest` where we can make use of mocking mechanisms.

gongracr (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/RegisterClientUseCase.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.wire.kalium.logic.failure.AuthenticationFailure
+import com.wire.kalium.logic.functional.Either
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+
+
+class RegisterClientUseCase(
+    private val clientRepository: ClientRepository
+) {
+    suspend operator fun invoke(param: RegisterClientParam): Flow<RegisterClientResult> = flow {
```
vitorhugods, line 14: For now, I think this is good.

But I understand that if I want to use `coreLib` in an app, I should not need to worry about things like `preKey`, `lastPreKey`, etc.

One thing we need to always keep in mind, is what `core` is responsible for, and how one should use it.

This UseCase + Repository combo doens't filter out any complexity or implementation details, it's just a proxy to the network layer.

The current utilization is something like:

```kotlin
coreLogic.userSession(currentUserSession){
    val prekeys = ??
    val lastKey = ??
    val deviceType = ??
    val type = ??
    val capabilities = ??
    val model = ??
    val registerParams = RegisterClientParam(password, preKeys, lastKey, deviceType, type, label, capabilities, model)
    client.registerClient(registerParams)
}
```

I'd expect the CoreLib to be able to get device info by itself on multiple platforms (easily achieved by `expect/actual`), use the `cryptography` module to create prekeys, and only needing needing things that are actual specific to the app/UI. So the consumer would do something like this:

```kotlin
coreLogic.userSession(currentUserSession){
    val registerParams = RegisterClientParam(password, type, clientName)
    client.registerClient(registerParams)
}
```


So, again: For now, I think this is good. This unblocks the sample app. 
But as the core lib gets thicker and implements other repositories and usecases, we can start to make it smarter.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/prekey/PreyKeyMapper.kt
```diff
@@ -0,0 +1,9 @@
+package com.wire.kalium.logic.data.prekey
+
+import com.wire.kalium.network.api.prekey.PreKeyDTO
+
+class PreyKeyMapper {
```
typfel, line 5:
```suggestion
class PreKeyMapper {
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
typfel: Do you have any existing examples in reloaded of these mapping tests?

I would guess mocking of the mapping classes would be necessary to implements tests of type 2 and 3. Looking at this from the outside, this kind of testing seems like it's testing private logic: correct parameters are forwarded, a mapping is happening or not which would also be covered by tests of type 4.

Besides that I think worth considering what Gonzo suggested. Does all tests need to run on all platforms or can some platform independent logic be tested only by common tests.

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/prekey/PreyKeyMapper.kt
vitorhugods: The existence of a `PreyKeyMapper` implies there's a `HunterKeyMapper` 😅

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/RegisterClientUseCase.kt
MohamadJaara: i added the `ClientConfig` class for client type, model and so on

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
```diff
@@ -0,0 +1,87 @@
...
+        return when (clientCapability) {
+            ClientCapability.LegalHoldImplicitConsent -> ClientCapabilityDTO.LegalHoldImplicitConsent
+        }
+    }
+
+    private fun fromClientCapabilityDTO(clientCapabilityDTO: ClientCapabilityDTO): ClientCapability {
+        return when (clientCapabilityDTO) {
+            ClientCapabilityDTO.LegalHoldImplicitConsent -> ClientCapability.LegalHoldImplicitConsent
+        }
+    }
```
vitorhugods, line 68:
```suggestion
    private fun toClientCapabilityDTO(clientCapability: ClientCapability): ClientCapabilityDTO = when (clientCapability) {
        ClientCapability.LegalHoldImplicitConsent -> ClientCapabilityDTO.LegalHoldImplicitConsent
    }

    private fun fromClientCapabilityDTO(clientCapabilityDTO: ClientCapabilityDTO): ClientCapability = when (clientCapabilityDTO) {
        ClientCapabilityDTO.LegalHoldImplicitConsent -> ClientCapability.LegalHoldImplicitConsent
    }
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    private fun fromDeviceTypeDTO(deviceTypeDTO: DeviceTypeDTO): DeviceType {
+        return when (deviceTypeDTO) {
+            DeviceTypeDTO.Phone -> DeviceType.Phone
+            DeviceTypeDTO.Tablet -> DeviceType.Tablet
+            DeviceTypeDTO.Desktop -> DeviceType.Desktop
+            DeviceTypeDTO.LegalHold -> DeviceType.LegalHold
+        }
+    }
```
vitorhugods, line 86:
```suggestion
    private fun toDeviceTypeDTO(deviceType: DeviceType): DeviceTypeDTO = when (deviceType) {
        DeviceType.Phone -> DeviceTypeDTO.Phone
        DeviceType.Tablet -> DeviceTypeDTO.Tablet
        DeviceType.Desktop -> DeviceTypeDTO.Desktop
        DeviceType.LegalHold -> DeviceTypeDTO.LegalHold
    }

    private fun fromDeviceTypeDTO(deviceTypeDTO: DeviceTypeDTO): DeviceType = when (deviceTypeDTO) {
        DeviceTypeDTO.Phone -> DeviceType.Phone
        DeviceTypeDTO.Tablet -> DeviceType.Tablet
        DeviceTypeDTO.Desktop -> DeviceType.Desktop
        DeviceTypeDTO.LegalHold -> DeviceType.LegalHold
    }
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/RegisterClientUseCase.kt
```diff
@@ -0,0 +1,28 @@
...
+        emit(RegisterClientResult.Loading)
```
vitorhugods, line 15: I'd say that `Loading` is not a result, it doesn't add much information to whoever is using, and it may need customization by whoever is using it if multiple flows are used:

```kotlin
fooUseCase().collect {
    // Handle possible progress bar, but try not to interfere with the other usecase
}
```

Could be:
```kotlin
showProgressBar()
fooUseCase().collect{}
hideProgressBar()
```

----

On more complicated cases it can get messed up:
```kotlin
fooUseCase().collect {
    // Handle possible progress bar, but try not to interfere with the other usecase
}

barUseCase().collect {
    // Handle possible progress bar, but try not to interfere with the other usecase
}
```

And we could solve by adding the jobs to another supervisor job and changing a LiveData as they finish: 

```kotlin
class UserProfileViewModel {

    private val isLoading = MutableLiveData<Boolean>(false)
    private val loadingJob = SupervisorJob()
    private val loadingScope = CoroutineScope(Dispatchers.IO + loadingJob) //TODO: Handle cancellation on destroy

    private fun loadingAction(action: (suspend CoroutineScope.() -> Unit)) {
        val newJob = loadingScope.launch { action() }
        uploadLoadingStatus() // Immediately update the LoadingStatus
        newjob.launchOnCompletion { updateLoadingStatus() } // Update again when completing this job
    }

    /**
     * Checks if there is any loading job that is still running updating the [isLoading] LiveData accordingly
     */
    private fun updateLoadingStatus() {
        isLoading.postValue(
            loadingJob.children.any { it.isActive }
        )
    }

    fun loadUserProfile() {
        // Both actions run in parallel:
        loadingAction {
            getUserProfileData().collect {
                // handle result only. No need to handle loading state
            }
        }
        loadingAction {
            getUserClientData().collect {
                // handle result only. No need to handle loading state
            }
        }
    }
}
```



TL;DR: Maybe we could leave loading status to whoever is using this library to figure out, not us :shrug: ? What do you think?

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -11,7 +12,8 @@ import com.wire.kalium.network.LoginNetworkContainer
 class CoreLogic(
     private val clientLabel: String,
     private val rootProteusDirectoryPath: String,
-    private val authSessionMapper: AuthSessionMapper = AuthSessionMapper()
+    private val authSessionMapper: AuthSessionMapper = AuthSessionMapper(),
+    private val clientConfig: ClientConfig
```
vitorhugods, line 16: Can we leave the optional parameter to the end? Just to avoid naming parameters unnecessarily.

vitorhugods (changes requested):
Great one!

Just a small sugar coating on some syntax and a question about `Loading` for use cases.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
MohamadJaara: In my opinion A, B are no go since testing on ios /js is necessary.
C can be a pain in the ass but i don't see it as an issue.
we can mock intefaces in commonTest using something like [this](https://github.com/mockative/mockative) (not really battle tested or fully featuredbut but still a cool concept).

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
MohamadJaara: I am thinking about a different approach, like making `CoreLogic` platform dependent and the android class will require `Context` since context is needed in many different other classes E.g. DataStore

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/RegisterClientUseCase.kt
MohamadJaara: make sense, i will do the changes

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
vitorhugods: That can also work, but maybe it grows this PR too big? I'd suggest changing these platform-specific things in another PR.

In any case, I would not expect anyone using `CoreLogic` to actually create a `AuthSessionMapper`. Especially when it's a closed class.

Currently something like this is needed:
```kotlin
val core = CoreLogic("Kalium Sample App", context.file("proteusRoot"), clientConfig = ClientConfig())
```

The `clientConfig = ` is something I'm trying to avoid, because `AuthSessionMapper` gets in the way. We don't provide good discoverability to whoever would be using this code.

vitorhugods (approved):

logic/src/androidMain/kotlin/com/wire/kalium/logic/configuration/ClientConfig.kt
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    actual fun deviceModelName(): String = "${Build.MANUFACTURER} ${Build.MODEL}"
+
+    actual fun deviceName(): String = Settings.Secure.getString(context.contentResolver, bluetoothName) ?: run {deviceModelName()}
+
+    actual fun clientType(): ClientType = ClientType.Permanent
+
+    private companion object {
+        private const val bluetoothName = "bluetooth_name"
```
gongracr, line 27: Maybe default constant to uppercase?

typfel (approved):
Approving with expectation that `RegisterClientUseCase` updated in separate PR to not expose `PreKey` to public.

gongracr (approved):
Looks good enough to me. However I'm not super sold on the idea of only making the Core/Logic lib consumable from the outside as a lot of boilerplate code will need to be written. If we do so, then we need to make sure we limit the visibility of all methods and classes from Network and Cryptography, as there is a number of elements and methods that are still public and therefore accessible.

In any case I imagine this belongs to a lengthier conversation and I will not block anymore this PR.

GJ with all the mapping classes nevertheless 🍄

PR #184, opened 2022-01-11
feat: preferences storage and session repo

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

1. Managing and storing different users sessions
2. Previusly there was no way to inslises object from classes  that are platform dependent E.g. `CoreLogic.kt` cann't have an object of `ClientConfig.kt` since it's a platform dependent and needs `Context` on android, so it was just passed to the constructor wich is not a good solution
```kotlin 
class CoreLogic(
    ...
    private val clientConfig: ClientConfig
) {
```

### Solutions

1. Implement a secure encrypted `DataStore` storage in android (if something already exist that is kmp please let me know)
2. I tried to solve this issue by making `CoreLogic.kt` an `abstract class` and having an `expect class` that implement it and can initialize the platform dependent objects
```kotlin

expect class CoreLogic: CoreLogicCommon

abstract class CoreLogicCommon(
    ....
) {
    ...
    abstract fun getAuthenticationScope(): AuthenticationScope
    abstract fun getSessionScope(session: AuthSession): UserSessionScopeCommon

``` 
the same solution can work also with `var, val and fun` since in kotlin we can have abstract variables.

### Testing

to be added in a later pr

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -1,48 +1,42 @@
...
-        val dataSourceSet = userScopeStorage[session] ?: run {
-            val networkContainer = AuthenticatedNetworkContainer(authSessionMapper.toSessionCredentials(session))
-            val proteusClient = ProteusClient(rootProteusDirectoryPath, session.userId)
-            AuthenticatedDataSourceSet(networkContainer, proteusClient).also {
-                userScopeStorage[session] = it
-            }
-        }
-        return UserSessionScope(session, dataSourceSet, clientConfig)
-    }
+    abstract fun getSessionScope(session: AuthSession): UserSessionScopeCommon
```
gongracr, line 35: Would we need to make this whole method abstract if we had already the `ClientConfig` value injected?

persistence/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+android {
+    compileSdk = Android.Sdk.compile
+    sourceSets["main"].manifest.srcFile("src/androidMain/AndroidManifest.xml")
+    defaultConfig {
+        minSdk = Android.Sdk.min
+        targetSdk = Android.Sdk.target
+    }
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
```
gongracr, line 18: Shouldn't we also extract this to a common file inside the root `buildSrc` folder?

persistence/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+kotlin {
+    jvm {
+        compilations.all {
+            kotlinOptions.jvmTarget = "1.8"
```
gongracr, line 25: Same than above

persistence/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+    android()
+
+    sourceSets {
+        val commonMain by getting {
+            dependencies {
+                // coroutines
+                implementation(Dependencies.Coroutines.core)
+                implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.3.0")
```
gongracr, line 39: Let's use version vars instead of hardcoding it

persistence/src/androidMain/kotlin/com/wire/kalium/persistence/util/SecurityUtil.kt
```diff
@@ -0,0 +1,60 @@
...
+import android.security.keystore.KeyProperties.KEY_ALGORITHM_AES
+import android.security.keystore.KeyProperties.PURPOSE_DECRYPT
+import android.security.keystore.KeyProperties.PURPOSE_ENCRYPT
+import java.security.KeyStore
+import javax.crypto.Cipher
+import javax.crypto.KeyGenerator
+import javax.crypto.SecretKey
+import javax.crypto.spec.GCMParameterSpec
+
+class SecurityUtil {
```
gongracr, line 15: Shouldn't these security methods be responsibility of the `cryptography`module?

gongracr (commented):
Awesome work, just a bit too long of a PR for me to follow up correctly on how all pieces are connected 🙈 

So I'm not sure I understood the solution proposed correctly. If the only platform specific object is `ClientConfig`, therefore why don't we inject it into the `CoreLogic` class on via a `GetCurrentClientConfigurationUseCase` or even provide this info on an early activation in the Application class? 
That way we would avoid having to rewrite the (already complex enough) `CoreLogic`class as well as other critical classes on each different platform.

I feel that the most implementation divergence between platforms we allow on these early phases, the more chances we have it will grow too complex too soon, biting us in the tail eventually.

MohamadJaara:
@gongracr 
This pr is to avoid rewriting stuff for diffrent targets the idea is simply you write all the logic inside `CoreLogicCommon.kt` and leave any variable/function that need platform spicific stuff as `abstract`
E.g. 
```kotlin
// some class in androidMain
actual class Foo (androidClass: AndroidClass) {
...
}



//  in commonMain

// only the declaration on platform specific classes
expect class Clazz: ClazzBase 

// logic
abstract class ClazzBase {

    abstract val foo: Foo

  // here you can still have all of the common stuf
 // and even use foo normaly
  ...
}

// and in androidMain
// only dicleration no logic
actual class Clazz {
   override val foo = Foo(androidClass)
}
```


the alternative is as we did before which is to leave the declaration of `Foot` to the app E.g.
```kotlin
// some  class in androidMain
actual class Foo (androidClass: AndroidClass) {
...
}

//  in commonMain
class Clazz (private val foo: Foo){
 ..
}

// and in the consumer app
private val foo: Foo(androidClass)
val clazz = Clazz(foo)
```
 and i don't really like that for the following reasons:

1. this issue is present but not limeted to  `ClientConfig.kt`, as we expand the library we will encounter this problem alot (wanna use `ConnectivityManager, DataStore` or you need `ApplicationContext`), adding a new variable every time to the constrector is no option.
2. by passing `ClientConfig.kt` in the constrector of`CoreLogig.kt` means that an object of `ClientConfig` will stay in momory even when it is used one time.

persistence/src/androidMain/kotlin/com/wire/kalium/persistence/util/SecurityUtil.kt
MohamadJaara: not really this is not related to `Proteus`, `SecurityUtil` is simply and encryption layer on top off `DataStore` (dataStore files are not encripted by default) and moving this to `cryptography` will add an un unnecessary complexity

persistence/build.gradle.kts
MohamadJaara: i didn't really get what you mean, this is the gradle to build the `persistence` module

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
MohamadJaara: `UserSessionScope` in android need `Context`

persistence/build.gradle.kts
gongracr: I meant to centralise the java version of all the module `build.gradle.kts` files into one variable inside the shared `Dependencies` file
```suggestion
        sourceCompatibility = Versions.javaVersion
        targetCompatibility = Versions.javaVersion
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
gongracr: Not really, we only need the `Context` to populate the `ClientConfig` class, so we could pass an instance of this as a parameter instead no?

persistence/build.gradle.kts
MohamadJaara: it's done this way in all kalium modules, it make sense to be changed and unified across the app, so i will add a ticket and will change it in a later PR (i have enough changes in this one)

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
typfel: Would another alternative be to introduce a factory which contains all the platform specific parts

```
// android
actual CoreLogicFactory(private applicationContext: Context) {
	
}

expect CoreLogicFactory {

	val clientConfig: ClientConfig
	val databaseDriverFactory: DatabaseDriverFactory
	val coreLogic: CoreLogic
}

CoreLogic(clientConfig: ClientConfig, databaseDriverFactory) {
	
}
```

typfel (approved):

PR #191, opened 2022-01-28
feat: ConversationDAO

typfel:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

Add persistence of conversations and its members

### Solutions

Introduce ConversationDAO

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #192, opened 2022-01-28
feat: slow sync

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We only want to fetch resources like conversations and users only once after login and store them in a local database, subsequent updates are received over the event stream. 

### Solutions

- Update ConverationRespository to fetch conversations from the database instead of doing network requests.
- Introduce SyncManager which starts and keep track of which resources have been fetched. The `GetConversationsUseCase` queries the sync manager to not attempt to fetch the conversation list until the conversation resource has been fetched.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/kalium/pull/191

### Testing

Not testing yet.

### Notes 

- The slow sync is currently repeated on every launch, we would need to start persisting the last fetched even ID in order to  determine if can skip the slow sync.
- The sync is currently started from the main activity, in the future it should run in the work manager / be encapsulated in a background task.

----

----

----
##### References

### DIFF
--- a/android/src/main/kotlin/com/wire/kalium/KaliumApplication.kt
+++ b/android/src/main/kotlin/com/wire/kalium/KaliumApplication.kt
@@ -1,5 +1,28 @@
 package com.wire.kalium
 
 import android.app.Application
+import androidx.work.*
+import com.wire.kalium.logic.CoreLogic
+import com.wire.kalium.logic.sync.WrapperWorkerFactory
+import java.io.File
 
-class KaliumApplication: Application()
+class KaliumApplication: Application(), Configuration.Provider {
+
+    lateinit var coreLogic: CoreLogic
+
+    override fun onCreate() {
+        super.onCreate()
+
+        val rootProteusDir = File(this.filesDir, "proteus")
+        coreLogic = CoreLogic(applicationContext, "kalium", rootProteusDir.absolutePath)
+    }
+
+    override fun getWorkManagerConfiguration(): Configuration {
+        val myWorkerFactory = WrapperWorkerFactory(coreLogic)
+
+        return Configuration.Builder()
+            .setWorkerFactory(myWorkerFactory)
+            .build()
+    }
+
+}
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncManager.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncManager.kt
@@ -1,6 +1,10 @@
 package com.wire.kalium.logic.sync
 
+import com.wire.kalium.logic.CoreFailure
 import com.wire.kalium.logic.data.conversation.ConversationRepository
+import com.wire.kalium.logic.feature.conversation.SyncConversationsUseCase
+import com.wire.kalium.logic.functional.Either
+import com.wire.kalium.logic.functional.onSuccess
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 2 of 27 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncManager.kt
```diff
@@ -0,0 +1,53 @@
...
+}
+
+class SyncManager(private val conversationRepository: ConversationRepository) {
+
+    private val internalSyncState = MutableStateFlow(SyncState.WAITING)
+    private var syncJob: Job? = null
+
+    suspend fun performSlowSync() {
+        internalSyncState.update { SyncState.FETCHING_CONVERSATIONS }
+        conversationRepository.fetchConversations()
```
typfel, line 26: Was looking at the slow sync implementation for the android reloaded and saw that each slow sync phase was implemented as separate use case instead of calling repositories directly. I guess it makes sense to adopt the same  approach here.

typfel:
@vitorhugods @MohamadJaara this is ready be reviewed now.

android/src/main/kotlin/com/wire/kalium/KaliumApplication.kt
```diff
@@ -1,5 +1,28 @@
...
+        coreLogic = CoreLogic(applicationContext, "kalium", rootProteusDir.absolutePath)
+    }
+
+    override fun getWorkManagerConfiguration(): Configuration {
+        val myWorkerFactory = WrapperWorkerFactory(coreLogic)
+
+        return Configuration.Builder()
+            .setWorkerFactory(myWorkerFactory)
+            .build()
+    }
```
vitorhugods, line 26: I wonder if we could add this implementation to `CoreLogic/androidMain` so it would be easier to use plug-in into an Android app outside the box.

Not needed now, but something to keep in mind.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncManager.kt
```diff
@@ -0,0 +1,39 @@
...
+    private fun startSlowSyncIfNotAlreadyCompletedOrRunning() {
+        val syncState = internalSyncState.updateAndGet {
+            when (it) {
+                SyncState.WAITING -> SyncState.SLOW_SYNC
+                else -> it
+            }
+        }
+
+        if (syncState == SyncState.SLOW_SYNC) {
+            workScheduler.schedule(SlowSyncWorker::class, SlowSyncWorker.name)
```
vitorhugods, line 36: Android's Workers won't necessarily start immediately. 
Maybe we could try to run things without scheduling, and scheduling for the future only in case of failure.

Maybe the `WorkScheduler` can change a bit depending on the platform. It's a pain in the ass.

Nothing I would fix or change now. Just to keep in mind.
As we proceed with Android Reloaded we can fine tune this.

vitorhugods (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncManager.kt
typfel: What about configuring the work tasks as expedited?https://developer.android.com/topic/libraries/architecture/workmanager/how-to/define-work#expedited

vitorhugods: It does improve yes. 

I guess we can do that for now without any costs. But the lack of `100% confidence` that the work will start now is always mind bogging.

PR #193, opened 2022-01-31
feat: Add session-scoped storage for event processing (AR-915)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Even though we already have a global preferences/storage to store the multiple sessions, we also need one storage PER session. 
For example, when dealing with processing of events, we need to store info about the latest event that was processed.

### Solutions

Expanded on @MohamadJaara's work for settings/preferences and created one instance of `KaliumSettings` per `UserScope`.
Based on it, created an `EventInfoStorage` that can hold an ID of the latest processed event.

### Testing

Added tests for the new `EventInfoStorage`.

----

----
##### References

### DIFF
--- a/logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
+++ b/logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
@@ -28,6 +28,6 @@ actual class CoreLogic(
                 userScopeStorage[session] = it
             }
         }
-        return UserSessionScope(applicationContext, dataSourceSet)
+        return UserSessionScope(applicationContext, session, dataSourceSet)
     }
 }
--- a/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
+++ b/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
@@ -3,6 +3,8 @@ package com.wire.kalium.logic.feature
 import android.content.Context
 import com.wire.kalium.logic.AuthenticatedDataSourceSet
 import com.wire.kalium.logic.configuration.ClientConfig
+import com.wire.kalium.logic.feature.auth.AuthSession
+import com.wire.kalium.persistence.kmm_settings.EncryptedSettingsHolder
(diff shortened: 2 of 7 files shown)


typfel:
I'm a bit cautious about storing the last event id information in the preferences DB which is backed by the keychain on iOS. Because items in keychain are not always deleted together with the app which uses them. This means that we could potentially get into a situation like this:

1. installs the Wire app and logs in
1. User deletes the Wire app and re-installs it again
2. Last processed event ID is not null
3. We think we don't need to perform a slow sync and end up in a broken state.

I think it would safer to store this information in the main DB or in the separate event DB or we would need some additional flag to know whether the app has been re-installed.

vitorhugods:
> I'm a bit cautious about storing the last event id information in the preferences DB which is backed by the keychain on iOS. Because items in keychain are not always deleted together with the app which uses them. This means that we could potentially get into a situation like this:
> 
>     1. installs the Wire app and logs in
> 
>     2. User deletes the Wire app and re-installs it again
> 
>     3. Last processed event ID is not null
> 
>     4. We think we don't need to perform a slow sync and end up in a broken state.
> 
> 
> I think it would safer to store this information in the main DB or in the separate event DB or we would need some additional flag to know whether the app has been re-installed.

Is there any Keychain data we would like to keep when having the app reinstalled?

Could we detect a reinstall and nuke the keychain?

typfel:
> Could we detect a reinstall and nuke the keychain?

We could potentially had a hack which nukes the keychain by storing a flag like "keychain_has_been_reset" in the UserDefaults after installing the app. Keys stored in the user defaults doesn't survive a re-install.

typfel:
> For now, especially as there's no iOS target in Core to customise this, I'd like to keep it the way it is.

Sure we can keep it like this for now.


> Once we have an implementation of `UnencryptedStorage`, or a way to customize `CoreLogic` data sources for iOS, we can work around this more easily.

The default version of https://github.com/russhwolf/multiplatform-settings is backed by `UserDefaults` on iOS.

typfel (approved):

PR #194, opened 2022-01-31
refactor: make network backend agnostic (AR-841)

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

`Network` module need to be agnostic to the backend it's connected to in order to be able to switch backends

### Solutions
creating `BackendConig` class that is needed to initialize `Network`
```kotlin 
data class BackendConfig(
    val apiBaseUrl: String,
    val accountsUrl: String,
    val webSocketUrl: String
)
```
and in `BackendType` inside `logic`
```kotlin
sealed class BackendType {
    object Production : BackendType()
    object Staging : BackendType()
    data class Custom(val networkConfig: NetworkConfig) : BackendType()
}
```
to point `network` to the correct backend links

----

----
##### References


build.gradle.kts
```diff
@@ -29,8 +29,8 @@ tasks.withType<Test> {
 
 allprojects {
     repositories {
-        maven(url = "https://oss.sonatype.org/content/repositories/snapshots")
         google()
         mavenCentral()
+        maven(url = "https://oss.sonatype.org/content/repositories/snapshots")
```
vitorhugods, line 34: 👍🏻

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/backend_config/BackendTypeMapper.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    private val production: BackendConfig
+        get() =
+            BackendConfig(
+                """"https://prod-nginz-https.wire.com"""",
+                """"https://account.wire.com"""",
+                """"https://prod-nginz-ssl.wire.com""""
+            )
+
+    fun toBackendConfig(buildType: BuildType): BackendConfig = when (buildType) {
```
vitorhugods, line 23: My only pet peeve I find here is about this `BuildType` -> `BackendConfig` mapping.

I'm not sure if Kalium should say what backend you should use.

Maybe iOS, Web or even Android will need different logics. It's not Kalium/CoreLogic to decide how our product builds will work.

So I'd say to leave this outside, maybe  add `BackendConfig` static values to the class or make it a Sealed Class or something like that. Example:

```kotlin
data class BackendConfig(
    val apiBaseUrl: String,
    val accountsUrl: String,
    val webSocketUrl: String
){ 
    companion object {
        val PRODUCTION = BackendConfig(
                """"https://prod-nginz-https.wire.com"""",
                """"https://account.wire.com"""",
                """"https://prod-nginz-ssl.wire.com""""
            )
        val STAGING = BackendConfig(
            """"https://staging-nginz-https.zinfra.io"""",
            """"https://wire-account-staging.zinfra.io"""",
            """"https://staging-nginz-ssl.zinfra.io"""
        )
    }
}

``` 

And then, when creating the `CoreLogic` whoever is using can just pass `BackendConfig("my", "custom", "values")` or `BackendConfig.PRODUCTION`, for example..

vitorhugods (changes requested):
Nicely done. Just one different idea on who should have the responsibility of deciding which what are build types and how they should behave.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/backend_config/BackendTypeMapper.kt
typfel: +1 for this.

For federation for example, we added the option to switch between testing federation backends in the iOS client.

network/src/commonMain/kotlin/com/wire/kalium/network/AuthenticatedNetworkContainer.kt
```diff
@@ -32,14 +32,12 @@ import io.ktor.client.plugins.websocket.WebSockets
 import io.ktor.http.URLProtocol
 
 class AuthenticatedNetworkContainer(
+    backEndConfig: BackendConfig,
```
typfel, line 35:
```suggestion
    backendConfig: BackendConfig,
```

network/src/commonMain/kotlin/com/wire/kalium/network/HttpClientProvider.kt
```diff
@@ -13,16 +14,18 @@ import io.ktor.client.plugins.logging.SIMPLE
 import io.ktor.client.request.header
 import io.ktor.http.URLProtocol
 import io.ktor.serialization.kotlinx.json.json
-import kotlinx.serialization.json.Json
 
 internal fun provideBaseHttpClient(
     engine: HttpClientEngine,
     isRequestLoggingEnabled: Boolean = false,
+    backEndConfig: BackendConfig,
```
typfel, line 21:
```suggestion
    backendConfig: BackendConfig,
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/backend_config/BackendTypeMapper.kt
MohamadJaara: the `BuildType` is suplied by the cosumer app of `kalium` and it's not part of kalium logic e.g.
```kotlin
class CoreLogicCommon(
    protected val buildType: BuildType
) {
```
and the android app will pass this bit of information.
I wanted to make kalium aware of the `BuildType` because it will not only be used for the BEConfig but also can be used for logging (if we want add it in the future)
so my idea was the `BuildType` will ditermin which BE to use, similer to the old implementation in reloaded
```kotlin
object NetworkConfig {
    val properties = mapOf(

        //Config field values for DEBUG Build Type
        BuildTypes.DEBUG to mapOf(
            ConfigFields.API_BASE_URL to ConfigFields.API_BASE_URL.defaultValue,
            ConfigFields.ACCOUNTS_URL to ConfigFields.ACCOUNTS_URL.defaultValue,
            ConfigFields.WEB_SOCKET_URL to ConfigFields.WEB_SOCKET_URL.defaultValue
        ),
        //Config field values for RELEASE Build Type
        //TODO: Certificate pinning, change backend based on flavour
        BuildTypes.RELEASE to mapOf(
            ConfigFields.API_BASE_URL to """"https://prod-nginz-https.wire.com"""",
            ConfigFields.ACCOUNTS_URL to """"https://account.wire.com"""",
            ConfigFields.WEB_SOCKET_URL to """"https://prod-nginz-ssl.wire.com/await?client=""""
        )
    )
}
```

typfel: OK, I see your point but I think this this is a but inflexible for example in the  use case I mentioned above we added support for running the regular iOS DEV build against a federation backend. In the system you propose the iOS app would need to supply `Build.Custom` which means it loses the ability collect debug logs.

Would it make sense to initialise `CoreLogic` with a `BuildType` and an optional `BackendConfig` which if omitted can default to staging or release depending on the BuildType. This way the client is free to run against production with logging enabled.

vitorhugods: @typfel I had a discussion with @MohamadJaara and @mchenani about this.

We agreed to leave the concept of `BuildType` out of Kalium and leave that to the apps to figure out.

Also, we had an idea on how to tackle custom backends (and switching backends using deeplinks).

Not 100% correct, but the rough idea is:
```kotlin
// Initialize CoreLogic with ClientInfo and whatever else is needed by the platform.
// ClientInfo will have backend information, URLs, etc.

val core = CoreLogic(someClientInfo)

// All operations here will use the default backend, that was setup before
val sessionAtDefaultBackend = core.authenticationScope {
    loginUsingEmail(email, password).session
}

// The session contains tokens AND information about the backend (urls, etc)


// When creating the userScope, CoreLogic will create HttpClients based on the session info
core.userScope(sessionAtDefaultBackend) {
    conversation.sendTetMessage(params)
}


// User clicks on a deeplink to login into a custom backend

val sessionAtCustomBackend = core.authenticationScope(backendOverwriteData = backendData) {
    // All operations here will use this backend overwrite
    loginUsingEmail(email, password).session
}

core.userScope(sessionAtCustomBackend) {
    // This will use the custom backend, without the need to do anything else.
    conversation.sendTetMessage(params)
}
```

vitorhugods: @MohamadJaara @typfel thinking about it, what if we just get rid of the default backend info and make the `backendData` mandatory for when getting a `authenticationScope`?

This would make the CoreLogic API easier to understand when dealing with any backend, maybe?

MohamadJaara: Already did it in a similar way, so when invoking `LoginUseCase` `ServerConfig` must be provided regardless if it's on premise or wire BE both are treated the same by `CoreLogic` and after a successful login the user session + `ServerConfig` (the one provided when logging in) will be stored and used by `UserSessionScope` (`CoreLogic` will fetch it internaly from user prefrences).
we can also store the name on the `ServerConfig` since the json data will provide a title as well (in case there is a need to know the BE type).
This approche will allow us to support multible acounts connected to diffrent BE out of the box

MohamadJaara:
As of this time when invoking `LoginUseCase` `ServerConfig` must be provided regardless if it's on premise or wire BE both are treated the same by `CoreLogic` and after a successful login the user session + `ServerConfig` (the one provided when logging in) will be stored and used by `UserSessionScope` (`CoreLogic` will fetch it internally from account preferences).
we can also store the name on the `ServerConfig` since the json data will provide a title as well (in case there is a need to know the BE type).
This approach will allow us to support multiple acounts connected to diffrent BE out of the box

the storing of `ServerConfig` is **not** yet implemented i left TODOs where the code will be changed for the sake of not making this PR any bigger

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/auth/login/LoginRepository.kt
```diff
@@ -13,23 +14,23 @@ import com.wire.kalium.network.utils.httpResponseCookies
 import com.wire.kalium.network.utils.isSuccessful
 
 interface LoginRepository {
-    suspend fun loginWithEmail(email: String, password: String, shouldPersistClient: Boolean): Either<CoreFailure, AuthSession>
-    suspend fun loginWithHandle(handle: String, password: String, shouldPersistClient: Boolean): Either<CoreFailure, AuthSession>
+    suspend fun loginWithEmail(email: String, password: String, shouldPersistClient: Boolean, serverConfig: ServerConfig): Either<CoreFailure, AuthSession>
```
typfel, line 17: Why are we passing the `serverConfig` for every method which makes a request now? Wasn't the idea to pass the network config once when creating a `AuthenticationScope` and then forward it to `LoginNetworkContainer`?

typfel:
@MohamadJaara is this PR ready to be reviewed now?

MohamadJaara:
> @MohamadJaara is this PR ready to be reviewed now?

it's now 🤘🏼

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/auth/login/LoginRepository.kt
MohamadJaara: this is **only** for login since kalium hold no info about the BE before calling `LoginUseCase`

typfel: But why not hold the information about the BE? I'm not saying we should persist it, just hold it in memory for the duration off the `AuthenticationScope`.

vitorhugods: @typfel IIRC, I discussed this with @MohamadJaara. One problem with this would be dependency injection in the apps. The ViewModels would not be able to just get a `LoginUseCase`, they would actually need to get `CoreLogic` and use it directly to get a `AuthenticationScope`.


One way to fix this would be by wrapping the `UseCase` in the app.

```kotlin
class MyAppLoginUseCase(private val coreLogic: CoreLogic) {

    suspend fun invoke(backendStuff: BackendSettings, email: String, password: String) : Result {
        coreLogic.authenticationScope(backendStuff) {
            loginUsingEmailAndPassword(email, password)
        }
    }
}
```

A bit more annoying for all the use cases inside `AuthenticationScope`.


I didn't find have any favourites in this.

There might be some UseCases inside `AuthenticationScope`  that don't require backend at all. Like `getListOfCurrentSessions` or `getCurrentActiveSession` and `setSessionAsActive`.

Maybe we can find another way to unify and reduce the duplication for the ones where backend settings are required. Like making them extend the same `BaseLoginUseCase` that takes backend config, etc.

vitorhugods (approved):

PR #197, opened 2022-02-03
feat: persist client id after registration

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We need the ClientId for a lot of operations, and after successfully calling `registerClient`, we don't store its value anywhere.

### Solutions

* Created key/value `ClientRegistrationStorage` for persist the client id.
* Perform storage after registering in `RegisterClientUseCase`

### Testing

Full coverage for all the layers changed: 
* `ClientRegistrationStorage`
* `ClientRepository`
* `RegisterClientUseCase` (WIP)

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
@@ -28,7 +28,7 @@ import com.wire.kalium.persistence.event.EventInfoStorage
 import com.wire.kalium.persistence.kmm_settings.EncryptedSettingsHolder
 import com.wire.kalium.persistence.kmm_settings.KaliumPreferencesSettings
 
-expect class UserSessionScope : UserSessionScopeCommon
+expect class UserSessionScope: UserSessionScopeCommon
 
 abstract class UserSessionScopeCommon(
     private val session: AuthSession,
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
@@ -22,10 +22,9 @@ import com.wire.kalium.logic.feature.auth.AuthSession
 import com.wire.kalium.logic.feature.client.ClientScope
 import com.wire.kalium.logic.feature.conversation.ConversationScope
 import com.wire.kalium.logic.feature.message.MessageScope
+import com.wire.kalium.logic.sync.SyncManager
 import com.wire.kalium.persistence.client.ClientRegistrationStorage
 import com.wire.kalium.persistence.client.ClientRegistrationStorageImpl
-import com.wire.kalium.logic.sync.WorkScheduler
-import com.wire.kalium.logic.sync.SyncManager
 import com.wire.kalium.persistence.db.Database
 import com.wire.kalium.persistence.event.EventInfoStorage
 import com.wire.kalium.persistence.kmm_settings.EncryptedSettingsHolder
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
@@ -28,7 +28,7 @@ import com.wire.kalium.persistence.event.EventInfoStorage
 import com.wire.kalium.persistence.kmm_settings.EncryptedSettingsHolder
 import com.wire.kalium.persistence.kmm_settings.KaliumPreferencesSettings
 
-expect class UserSessionScope: UserSessionScopeCommon
+expect class UserSessionScope : UserSessionScopeCommon
 
 abstract class UserSessionScopeCommon(
     private val session: AuthSession,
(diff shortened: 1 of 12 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/RegisterClientUseCase.kt
```diff
@@ -2,25 +2,26 @@ package com.wire.kalium.logic.feature.client
...
-            is Either.Right -> emit(RegisterClientResult.Success(result.value))
+class RegisterClientUseCase(private val clientRepository: ClientRepository) {
 
-            is Either.Left -> {
-                if (result.value is AuthenticationFailure)
-                    emit(RegisterClientResult.Failure.InvalidCredentials)
-                else
-                    emit(RegisterClientResult.Failure.Generic(result.value))
+    suspend operator fun invoke(param: RegisterClientParam): RegisterClientResult = suspending {
+        //TODO Should we fail here if the client is already registered?
```
typfel, line 12: For sure it doesn't make sense to register the client again but then again wouldn't the BE respond with an error if you attempt register an already registered client?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -22,11 +22,13 @@ import com.wire.kalium.logic.feature.auth.AuthSession
...
 import com.wire.kalium.logic.feature.conversation.ConversationScope
 import com.wire.kalium.logic.feature.message.MessageScope
+import com.wire.kalium.persistence.client.ClientRegistrationStorage
+import com.wire.kalium.persistence.client.ClientRegistrationStorageImpl
 import com.wire.kalium.persistence.event.EventInfoStorage
 import com.wire.kalium.persistence.kmm_settings.EncryptedSettingsHolder
 import com.wire.kalium.persistence.kmm_settings.KaliumPreferencesSettings
 
-expect class UserSessionScope: UserSessionScopeCommon
+expect class UserSessionScope : UserSessionScopeCommon
```
typfel, line 31: Intentional?

typfel (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
vitorhugods: Definitely not. Good catch!

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/RegisterClientUseCase.kt
vitorhugods: The backend allows you to use the same authentication token to register multiple clients.

We just send the information needed to create/register a new client. And it either registers it - responding with a new ClientId and so on - or fails is some data is not as it expects.

It's another philosophical question. Should we allow multiple clients for the same user session in this SDK?
Is that a valid use case? I kinda think no.

So if the session has an ClientId, we could just fail the registration and return `wtf dude?`.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -62,8 +64,10 @@ abstract class UserSessionScopeCommon(
             clientMapper
         )
 
+    private val clientRegistrationStorage: ClientRegistrationStorage = ClientRegistrationStorageImpl(userPreferencesSettings)
```
MohamadJaara, line 67: if `clientRegistrationStorage` meant to be a singleton, can it be lazy initialized ?
```suggestion
    private val clientRegistrationStorage: ClientRegistrationStorage by lazy { ClientRegistrationStorageImpl(userPreferencesSettings) }
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/RegisterClientUseCase.kt
MohamadJaara: should the check if an account already exist in the app be done here, after login (since login without client registration is possible) or why not both ?

 I will assume that BE returns a new `ClientId` after client register, so is it possible to compare clientIds ?

MohamadJaara (commented):
great work 🚀, just a small comment about creating singletons inside `CoreLogic`

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/RegisterClientUseCase.kt
vitorhugods: @MohamadJaara Yes, it would be a new `ClientId`. But comparing after registering is too late, either the user created an unnecessary client, which is annoying for cryptographic reasons and might reach the max clients limit, or this could fail unnecessarily.

Considering that the user needs a password in order to register a client, the app asking again for registration is a bit weird and probably a bug. So we checking this in Kalium may prevent a buggy app from re-registering without necessity.

I feel like this check doesn't need to be added right now, but it's a good reminder we can have a say and be opinionated on how the projects relying on this SDK should behave.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
vitorhugods: I don't know what I had in my mind 😄. An oversight from my part, absolutely. Nice catch!

PR #209, opened 2022-02-14
feat: sync users (AR-1030)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In order to have information about conversation members, we need to sync them after conversations.

### Solutions

- Add another step to `SyncManager` that fetches information about all known users.
- Rework `fetchUserDetails` API so it works with more than `64` users at a time (backend limit is `64`)

### Testing

W.I.P

### Notes (Optional)

About reworking `fetchUserDetails`. I thought it was something that the `network` module should worry about. Or should we leave it to logic?

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -69,6 +69,8 @@ abstract class UserSessionScopeCommon(
             database.userDAO,
             database.metadataDAO,
             authenticatedDataSourceSet.authenticatedNetworkContainer.selfApi,
+            authenticatedDataSourceSet.authenticatedNetworkContainer.userDetailsApo,
```
MohamadJaara, line 72:
```suggestion
            authenticatedDataSourceSet.authenticatedNetworkContainer.userDetailsApi,
```

logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -26,7 +26,7 @@ actual class CoreLogic(
         val dataSourceSet = userScopeStorage[session] ?: run {
             val networkContainer = AuthenticatedNetworkContainer(
                 sessionCredentials = sessionMapper.toSessionCredentials(session),
-                backendConfig = serverConfigMapper.toBackendConfig(serverConfig = session.serverConfig)
+                backendConfig = serverConfigMapper.toBackendConfig(serverConfig = session.serverConfig), isRequestLoggingEnabled = true
```
MohamadJaara, line 29: maybe add a TODO so we can remove `isRequestLoggingEnabled = true` later when we have a better solution for logging

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -40,6 +48,22 @@ class UserDataSource(
         }
     }
 
+    override suspend fun fetchKnownUsers(): Either<CoreFailure, Unit> {
+        val ids = userDAO.getAllUsers().first().map { userEntry ->
+            idMapper.toApiModel(idMapper.fromDaoModel(userEntry.id))
+        }.distinct()
```
MohamadJaara, line 54: are user ids unique in the database ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -40,6 +48,22 @@ class UserDataSource(
...
+
+        val usersRequestResult = userApi.getMultipleUsers(ListUserRequest.qualifiedIds(ids))
+        if (!usersRequestResult.isSuccessful()) {
+            usersRequestResult.kException.printStackTrace()
+            return Either.Left(CoreFailure.ServerMiscommunication)
+        }
+
+        val usersToBePersisted = usersRequestResult.value.map(userMapper::fromApiModelToDaoModel)
+        userDAO.insertUsers(usersToBePersisted)
+        return Either.Right(Unit)
```
gongracr, line 64: What if the insertion on the DB fails? there is no way to catch it, or we just simply confirm that the NetworkResponse was successful?

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SlowSyncWorker.kt
```diff
@@ -10,10 +12,16 @@ class SlowSyncWorker(userSessionScope: UserSessionScope) : UserSessionWorker(use
 
         val result = userSessionScope.users.syncSelfUser()
             .flatMap { userSessionScope.conversations.syncConversations() }
+            .flatMap { userSessionScope.users.syncContacts() }
             .onSuccess { userSessionScope.syncManager.completeSlowSync() }
 
         when ( result ) {
             is Either.Left -> {
+                val failure = result.value
+                println("SYNC FAILED $failure")
```
gongracr, line 21: Is this `println(...)` getting printed on the logcat? I'm not 100% positive, but maybe we want to encapsulate it with `Log.e(...)` in case we want to have more control on how to filter these messages against others with more verbosity?

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SlowSyncWorker.kt
```diff
@@ -10,10 +12,16 @@ class SlowSyncWorker(userSessionScope: UserSessionScope) : UserSessionWorker(use
 
         val result = userSessionScope.users.syncSelfUser()
             .flatMap { userSessionScope.conversations.syncConversations() }
+            .flatMap { userSessionScope.users.syncContacts() }
             .onSuccess { userSessionScope.syncManager.completeSlowSync() }
 
         when ( result ) {
```
gongracr, line 18: Can we apply some automated code formatting to this file?

network/src/commonMain/kotlin/com/wire/kalium/network/AuthenticatedNetworkContainer.kt
```diff
@@ -59,6 +61,8 @@ class AuthenticatedNetworkContainer(
 
     val selfApi: SelfApi get() = SelfApi(authenticatedHttpClient)
 
+    val userDetailsApo: UserDetailsApi get() = UserDetailsApiImp(authenticatedHttpClient)
```
gongracr, line 64: Check typo
```suggestion
    val userDetailsApi: UserDetailsApi get() = UserDetailsApiImp(authenticatedHttpClient)
```

network/src/commonMain/kotlin/com/wire/kalium/network/api/user/details/UserDetailsApiImp.kt
```diff
@@ -1,20 +1,83 @@
...
+        return results.fold(results.first()) { acc, item ->
+            item as NetworkResponse.Success
+            acc.mapSuccess { currentValue ->
+                currentValue + item.value
+            }
+        }
+    }
+
+    private suspend fun performChunkedRequests(chunkedRequests: List<ListUserRequest>): List<NetworkResponse<List<UserDetailsResponse>>> =
+        coroutineScope {
```
gongracr, line 48: Just dumb question, why do need to spcify this `coroutineScope` before the api call? doesn't it come implicitly within the suspended function bounds?

MohamadJaara (changes requested):

gongracr (commented):
Just small comments, nothing blocker. 

Should we also add some tests for the new API call and use case?

logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
vitorhugods: Oh shit. You found my debugging!

I'll remove this for now.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
vitorhugods: Yeaah. It could throw an Exception and crash the app.

I'll add a TODO for now.

I was thinking of having a `wrapDbOperation` function so our repositories would use them and map to `Right/Left`.

Buuuuuuuut, the problem is that right now is a bit annoying to catch only SQLite-related-Exceptions, as the different classes are platform-dependent.

For example: `SQLException` for JVM/Android. I am not sure what are the classes for JS and iOS for example.

logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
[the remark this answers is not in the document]
vitorhugods: We could do that for `androidMain` I guess.

But we still need a mechanism for other targets.

network/src/commonMain/kotlin/com/wire/kalium/network/api/user/details/UserDetailsApiImp.kt
vitorhugods: The problem is the `Collections.map{}` call.
It doesn't take a `suspend` lambda.

This way, we can open a `coroutineScope`, perform the async requests in parallel and wait for them to complete side-by-side. It's actually faster this way too 🏎️ 🚀

typfel (changes requested):

network/src/commonMain/kotlin/com/wire/kalium/network/api/user/details/UserDetailsApiImp.kt
```diff
@@ -1,20 +1,83 @@
...
+
+    private fun separateUserIdsRequest(users: ListUserRequest): List<QualifiedUserIdListRequest> {
+        users as QualifiedUserIdListRequest
+        return users.qualifiedIds.chunked(MAX_USERS_PER_REQUEST).map { chunk ->
+            QualifiedUserIdListRequest(chunk)
         }
     }
 
     private companion object {
+        const val MAX_USERS_PER_REQUEST = 64
```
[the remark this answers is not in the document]
vitorhugods: I fucked up.

I was reading Sync docs and mixed the old and new endpoints 😄 

grafik


I'll remove this overkill thing.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConverationDAOImpl.kt
```diff
@@ -63,12 +67,16 @@ class ConversationDAOImpl(private val conversationQueries: ConverationsQueries,
     }
 
     override suspend fun insertMember(member: Member, conversationID: QualifiedID) {
-        memberQueries.insertMember(member.user, conversationID)
+        memberQueries.transaction {
+            userQueries.insertUser(member.user, null, null)
```
typfel, line 71: So if the user already exists the name and handle will get nulled out since the insert query will overwrite any existing if it already exists. So maybe we shouldn't do an `upsert` in this case.

typfel (changes requested):

network/src/commonMain/kotlin/com/wire/kalium/network/api/user/details/UserDetailsApiImp.kt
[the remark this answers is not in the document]
typfel: Well they might still have a limit so might make sense to keep but just increase the max chunk.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConverationDAOImpl.kt
vitorhugods: Oversight on my side. I just thought order of sync would be enough, but this is too loose.

I added a new query to `userQueries` to insert or ignore.

typfel (approved):

MohamadJaara (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
gongracr: Why don't we just wrap it as a generic `Exception` for now and add a comment to rewrite it to a more concrete Exception once all the targets have been written?

logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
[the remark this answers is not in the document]
gongracr: Fair enough. Still if we leave it like this, it will still log error messages even for release versions of Kalium

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
gongracr: Or even better, add another type of `KaliumException` to inform about a persistence/DB related error

vitorhugods: > Why don't we just wrap it as a generic Exception for now and add a comment to rewrite it to a more concrete Exception once all the targets have been written?

The problem is that at some point we need:
```kotlin
try{
    doThing()
} catch (e: /** Exception? **/) {
    multiplatformLogger.logException(e)
    Either.Left(CoreLogic.DatabaseException(cause = e))
}
```
We can only catch Exception/Throwable for now. Or we do per-platform Exception handling/mapping.

I don't want to explore proof of concepts like this in this branch.

So for now I'll just keep the TODO.
After we get sending and receiving in place. I think it's a good time to re-visit the code and make it bullet-proof based on our experience while developing it.

The DB exception handling is my top issue so far.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAOImpl.kt
```diff
@@ -22,8 +23,11 @@ class MemberMapper {
     }
 }
 
-class ConversationDAOImpl(private val conversationQueries: ConverationsQueries,
-                          private val memberQueries: MembersQueries): ConversationDAO {
+class ConversationDAOImpl(
+    private val conversationQueries: ConverationsQueries,
```
gongracr, line 27: Late to the party but maybe to keep in mind for next PR 👓 🙊 
```suggestion
    private val conversationQueries: ConversationsQueries,
```

vitorhugods: Yeaaaaaaaaah hahahah

It's an old typo that went through review and is being corrected little by little 😆

PR #211, opened 2022-02-14
feat: UpdateCurrentSessionUseCase and CurrentSessionUseCase

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
providing CurrentSession and updating the session

### Causes (Optional)

### Solutions

### Testing

unit test for UpdateCurrentSessionUseCase and CurrentSessionUseCase

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/functional/Either.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/functional/Either.kt
@@ -54,18 +54,17 @@ sealed class Either<out L, out R> {
      * @see Left
      * @see Right
      */
-    fun <T> fold(fnL: (L) -> T?, fnR: (R) -> T?): T? =
-        when (this) {
-            is Left -> fnL(value)
-            is Right -> fnR(value)
-        }
+    fun <T> fold(fnL: (L) -> T, fnR: (R) -> T): T = performFold(fnL, fnR)!!
+
 
     /**
      * Applies fnL if this is a Left or fnR if this is a Right.
      * @see Left
      * @see Right
      */
-    fun <T> safeFold(fnL: (L) -> T, fnR: (R) -> T): T =
+    fun <T> nullableFold(fnL: (L) -> T?, fnR: (R) -> T?): T? = performFold(fnL, fnR)
+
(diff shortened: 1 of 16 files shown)


yamilmedina (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/functional/Either.kt
```diff
@@ -59,6 +59,17 @@ sealed class Either<out L, out R> {
...
+    /**
+     * Applies fnL if this is a Left or fnR if this is a Right.
+     * @see Left
+     * @see Right
+     */
+    fun <T> safeFold(fnL: (L) -> T, fnR: (R) -> T): T =
+        when (this) {
+            is Left -> fnL(value)
+            is Right -> fnR(value)
+        }
```
yamilmedina, line 72: Just a question, I didn't try it. Could we achieve the same using the `fold` function from above?, and just forcing the return to value to be non-nullable, ie:

```suggestion
        fold(fnL, fnR)!!
```

vitorhugods: Nice suggestion @yamilmedina!

Food for thought: 
Personally I am not a fan of the current `fold` function and I would try to rename `safeFold` to `fold` and remove the current implementation.

I think that nowhere `fold` isn't being used with nullable values at the moment.

If the most common one is called `safeFold`, then we should make it the standard. `nullableFold` should be the alternative, not the default.

vitorhugods (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/functional/Either.kt
MohamadJaara: i did change it to this
```kotlin 
    fun <T> fold(fnL: (L) -> T, fnR: (R) -> T): T = performFold(fnL, fnR)!!

    fun <T> nullableFold(fnL: (L) -> T?, fnR: (R) -> T?): T? = performFold(fnL, fnR)

    private inline fun <T> performFold(fnL: (L) -> T?, fnR: (R) -> T?): T? =
        when (this) {
            is Left -> fnL(value)
            is Right -> fnR(value)
        }
```

PR #214, opened 2022-02-15
feat: Get recent messages from DB UseCase (AR-970)

borichellow:
[Task-970](https://wearezeta.atlassian.net/browse/AR-970)
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Added `Message` table, `DAO` and `UseCase`s for getting messages by conversationId and saving message into BD

----

----

----
##### References


persistence/src/commonMain/sqldelight/com/wire/kalium/persistence/db/Messages.sq
```diff
@@ -0,0 +1,42 @@
...
+import java.lang.Long;
+
+CREATE TABLE Message (
+    qualified_id TEXT AS QualifiedID NOT NULL PRIMARY KEY,
+    content TEXT,
+    conversation_id TEXT AS QualifiedID,
+    timestamp INTEGER AS Long,
+    sender_id TEXT AS QualifiedID,
+    status TEXT
+);
```
vitorhugods, line 11: A couple of things to keep in mind:

## Timezone

If we persist messages as a long without having information on timezone, we will lose information and display wrong information when the user changes timezones.

We receive a String with date, time and timezone from the backend. I'd suggest we keep this information instead of reducing to Long.

## ID of messages

Messages are attached to a `Conversation`, which already have a `QualifiedID`.
Because of this, they don't have qualified IDs.

So we just need a simple `TEXT` to hold the message ID.

## Nullability

Some fields here are completely mandatory, so we can make them `NOT NULL`.


## Sender ID (user and client)

We need to have information not only about the user that is sending the message (`QualifiedID`), but also the device that sent the message (`ClientID`).

## Tight everything together

We can't afford to have loose data. If we attempt to insert something that could lead to a silent bug, let's make sure that we get an exception. Especially now that we are starting with the project and we can catch the bugs early.
Let's take advantage of [Foreign Keys](https://www.sqlite.org/foreignkeys.html) to make sure that the data belongs somewhere.

#### Exception: Deleted users and clients
Because users and clients can be deleted and the conversation may still exist, I guess we can keep the users/clients *without* them being foreign keys.

## Status can have only so many values

At least now, I can think of `PENDING`, `SENT`, `READ`. So I suggest we create an enum for that. [SQLDelight supports enums](https://cashapp.github.io/sqldelight/multiplatform_sqlite/types/).

-----

With all of this in mind, my suggestion is:

```suggestion
CREATE TABLE Message (
    id TEXT NOT NULL,
    content TEXT,
    conversation_id TEXT AS QualifiedID NOT NULL,
    timestamp TEXT NOT NULL,
    sender_user_id TEXT AS QualifiedID NOT NULL,
    sender_client_id TEXT NOT NULL,
    status TEXT AS Message.Status NOT NULL,
    FOREIGN KEY (conversation_id) REFERENCES Conversation(qualified_id),
    PRIMARY KEY (id, conversation_id)
);
```

vitorhugods (changes requested):
`yarn.lock` scared the shit out of me with `+2,535` lines to review 😆 

Welcome to SQLDelight 🚀 
Just some suggestions on modeling the DB.

persistence/src/commonMain/sqldelight/com/wire/kalium/persistence/db/Messages.sq
typfel: > So we just need a simple TEXT to hold the message ID.

And this ID is created by the sender of the message and we refer to this ID as a nonce. The nonce is prevents us sending duplicated message when the network connection is bad and it allows to refer to the message later on.

The nonce is a UUID.

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendTextMessageUseCase.kt
```diff
@@ -1,11 +1,22 @@
...
-import com.wire.kalium.network.api.ConversationId
+import com.wire.kalium.persistence.dao.QualifiedID
+import com.wire.kalium.persistence.dao.message.Message
 
 class SendTextMessageUseCase(private val messageRepository: MessageRepository) {
 
-    suspend operator fun invoke(conversationId: ConversationId, text: String) {
-        TODO("Send a message!")
+    suspend operator fun invoke(conversationId: QualifiedID, text: String) {
+        //TODO
```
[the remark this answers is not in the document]
borichellow: it's about to set a propper data into `Message` as some of it is still hardcoded

vitorhugods (approved):
👨🏼‍🍳🤌🏼

PR #215, opened 2022-02-15
feat: create usecase to upload user picture as a public asset  [AR-1095]

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- This creates the use case and the respective call to the endpoint, for uploading the Avatar (user profile picture) to the `Assets` endpoint, returning an Asset Id with the reference to it.
- In the next PR, I will be taking this newly created Asset Id, and updating the user picture via self endpoint, also will have to save locally this reference and the image ideally.

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetMapper.kt
```diff
@@ -0,0 +1,11 @@
+package com.wire.kalium.logic.data.asset
+
+import com.wire.kalium.network.api.model.Asset
+import com.wire.kalium.network.api.model.AssetMetadata
+import com.wire.kalium.network.api.model.AssetRetentionType
+
+fun UploadAssetMetadata.toApiModel(md5: String): AssetMetadata {
```
MohamadJaara, line 7: it's not a bad idea to use extension functions for mapper classes, but we can't have anyway to mock it for Unit testing, so changing it to somthing like
```kotlin
interface AssetMapper {
    fun toApiModel(uploadAssetMetadata :UploadAssetMetadata): AssetMetadata
    fun toDomainModel(asset: Assrt): UploadAssetId
}

class AssetMapperImpl: AssetMapper {
    // override
}
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/Asset.kt
```diff
@@ -1,5 +1,21 @@
...
-data class PublicAsset(val key: String) : Asset()
+/**
+ * On creation of this model, the use case should "calculate" the logic.
+ * For example, rules that apply for an eternal/public asset
+ */
+data class UploadAssetMetadata(
+    val mimeType: String,
+    val isPublic: Boolean,
+    val retentionType: RetentionType
+)
```
MohamadJaara, line 13: is there any issue by having data and metadata of an asset at the same place

```suggestion
data class UploadAssetData(
    val data: ByteArray,
    val mimeType: AssetType, // where AssetType is an enum type of supported assets
    val isPublic: Boolean,
    val retentionType: Int
) {
     var md5: String
        private set
    init {
        md5 = calcMd5(data)
    }
    
    // auto generated code by the IDE to calculate the equals and hashCode 
    // because of the ByteArray

    override fun equals(other: Any?): Boolean {
        if (this === other) return true
        if (other == null || this::class != other::class) return false

        other as UploadAssetMetadata

        if (!data.contentEquals(other.data)) return false
        if (mimeType != other.mimeType) return false
        if (isPublic != other.isPublic) return false
        if (retentionType != other.retentionType) return false
        if (md5 != other.md5) return false

        return true
    }

    override fun hashCode(): Int {
        var result = data.contentHashCode()
        result = 31 * result + mimeType.hashCode()
        result = 31 * result + isPublic.hashCode()
        result = 31 * result + retentionType
        result = 31 * result + md5.hashCode()
        return result
    }
}

```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetMapper.kt
yamilmedina: Ok makes sense, solved it

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/Asset.kt
yamilmedina: Great suggestion, I updated with a proposal and a sealed class for mimetypes.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/Asset.kt
```diff
@@ -1,5 +1,60 @@
...
+ */
+data class UploadAssetData(
+    val data: ByteArray,
+    val mimeType: AssetType,
+    val isPublic: Boolean,
+    val retentionType: RetentionType
+) {
+
+    var md5: String = calcMd5(data)
+        private set
```
MohamadJaara, line 19:
```suggestion
    val md5: String = calcMd5(data)
```

MohamadJaara (approved):
nicely done 🍾

PR #216, opened 2022-02-16
fix: remove line breaks from PreKeys Base64 encoding/decoding on Android

vitorhugods:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

When encoding PreKeys on Android to Base64, a `\n` is being added at the end of the string. Backend does not accept prekeys with this format.

### Causes (Optional)

Use of `DEFAULT` instead of `NO_WRAP`

### Solutions

`NO_WRAP` 😃 

### Testing

Not yet in place.

----
#### PR Post Merge Checklist for internal contributors

 - [X] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #217, opened 2022-02-16
feat: receive events (AR-915)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In order to receive messages (and other updates), we need to handle the notifications api and websocket events.

### Solutions

- Merged `NotificationsApi` and `EventsApi` into one, as they give the same data.
- Fixed some issues with the current WebSocket implementation
- Added a `ListenToEventsUseCase` which is not 100% done and will be explored further in another PR with the consuming of `NewMessage` events.

### Dependencies

This branch contains the same fix present in #216.

Needs releases with:

- [ ] #216 

### Testing

Partially done. We don't have a way of mocking `Api` classes completely at the moment and this blocked some of the testing.
I've tested manually by registering a client and getting notifications.
Both the Rest API and the Websocket are working on Android once the client is registered.

### Notes

- There's no recovery/retry mechanism for when the internet connection drops in order to attempt to begin receiving events again.
- There's no protection against calling `ListenToEventsUseCase` twice.

There are multiple ways of doing both of the above, and I would like to explore them later as they might be app/platform dependent. For now, I would say that this is just the basic functionality. The consumers of Kalium can attempt to retry/recover manually.

----

----
##### References


logic/src/commonTest/kotlin/com/wire/kalium/logic/data/event/EventRepositoryTest.kt
```diff
@@ -0,0 +1,59 @@
...
+//        given(notificationApi)
+//            .suspendFunction(notificationApi::getAllNotifications)
+//            .whenInvokedWith(any(), any())
+//            .then {
+////                NetworkResponse.Success() TODO: Can't mock Network Response!
+//            }
+
+//        eventRepository.events()
+
+    }
```
ohassine, line 58: It's ignored, still WIP ?

network/src/commonMain/kotlin/com/wire/kalium/network/api/notification/EventContentDTO.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.wire.kalium.network.api.UserId
+import com.wire.kalium.network.api.notification.conversation.MessageEventData
+import com.wire.kalium.network.api.notification.user.NewClientEventData
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class EventResponse(
+    @SerialName("id") val id: String,
+    @SerialName("payload") val payload: List<EventContentDTO>?
```
ohassine, line 13: Maybe we can add `transient` field now to prevent refactoring later ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
```diff
@@ -0,0 +1,88 @@
...
+) : EventRepository {
+
+    override suspend fun events(): Flow<Either<CoreFailure, Event>> = suspending {
+        clientRepository.currentClientId().coFold({
+            flowOf(Either.Left(CoreFailure.MissingClientRegistration))
+        }, { clientId ->
+            val pendingEventsFlow = pendingEventsFlow(clientId)
+            val liveEventsFlow = liveEventsFlow(clientId)
+            flowOf(pendingEventsFlow, liveEventsFlow).flattenConcat()
+        })!!
```
gongracr, line 42: Wouldn't this result be always non-nullable?

MohamadJaara (approved):
great job 🍾

network/src/commonMain/kotlin/com/wire/kalium/network/HttpClientProvider.kt
```diff
@@ -51,6 +52,7 @@ internal fun provideBaseHttpClient(
             level = LogLevel.ALL
         }
     }
+    install(WebSockets)
```
gongracr, line 55: Wouldn't this cause some sort of exception on iOS targets as Ktor only support Websockets for Android and JVM? For now it is not a problem at all, but maybe we should write some comments and think on how to inject these websockets on a platform targetted mode (perhaps using the expect-actual KMP way?).

gongracr (commented):
Just added some small thinking-out-loud comments but all in all amazing job with this huuge PR adding a lot of strong glue! 👏 🎩

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private suspend fun liveEventsFlow(clientId: ClientId): Flow<Either<CoreFailure, Event>> =
+        notificationApi.listenToLiveEvents(clientId.value)
+            .map {
+                println("Mapping eventResponse from LiveFlow ${it}")
```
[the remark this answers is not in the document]
vitorhugods: Yes!

There are a few things I am unhappy about the project and I want to revisit next week. Logging is one of them.

network/src/commonMain/kotlin/com/wire/kalium/network/api/notification/EventContentDTO.kt
vitorhugods: From transient docs:

> A boolean value. If true, the notification will only be received over a websocket, and not be stored on the backend for later retrieval. 

I wonder if it makes sense for us to ever handle it.

I'll add it anyway for now.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
vitorhugods: Yes. I was based on the old implementation of `fold` and `coFold`. I can remove the double-bang now.

network/src/commonMain/kotlin/com/wire/kalium/network/HttpClientProvider.kt
vitorhugods: Yes!
As soon as we add `iOS` as a target, this will break.

I had some ideas about creating an `expect class WebsocketClient` which allows us to pass an URL and get an `incoming` flow.
Then we can implement per platform (most will just use the current one), but iOS for example would have its own magic.

I'll add a comment for now.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventMapper.kt
```diff
@@ -0,0 +1,32 @@
+package com.wire.kalium.logic.data.event
+
+import com.wire.kalium.logic.data.conversation.ClientId
+import com.wire.kalium.logic.data.id.IdMapper
+import com.wire.kalium.network.api.notification.EventContentDTO
+import com.wire.kalium.network.api.notification.EventResponse
+
+class EventMapper(private val idMapper: IdMapper) {
```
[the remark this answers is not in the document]
vitorhugods: For events, in particular, I think it's tricky.

Their structure is quite odd as they come from the backend:
```
Notification [ id, List<Payload>]

Payload [type, other stuff depending on type]
```

The structure of `Payload` is very inconsistent. Some events have unnecessary inner data structures with `date` inside, and some have `date` at the root level.

It's quite a mess, really. So mapping to something easier to use is already a pro.


We might want to store these Events in a database for processing later too in order to support At-Rest encryption, something iOS has and Android probably will have in the future too. So this flatter structure also helps with this.

network/src/commonMain/kotlin/com/wire/kalium/network/api/notification/EventContentDTO.kt
ohassine: > When receiving an event via websocket, the event can be marked as transient. A transient event will not be present in the notification stream, therefore it should not be stored as the last known notification.

https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/15566871/Incremental+synchronization#Storing-last-known-notification

If I understand well, transient events should not be pushed to notification stream

vitorhugods: @ohassine exactly.

So we will never receive them through `GET /notifications`. But we can receive through the WebSocket.

The thing is: If we are receiving them in real-time, then we process it aleady 😛.

We might add a storage for these events later and process them in the future. So I guess there's where the `transient` field comes in handy.

PR #218, opened 2022-02-16
fix: proteus client not initialized

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

proteus client not initialized 

### Solutions

call open after proteus client is creatd in `CoreLogic`

Needs releases with:

### Testing

manually

### Notes (Optional)

it's still debatable where to close the protuse client

----

----
##### References


logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -30,7 +31,10 @@ actual class CoreLogic(
                 backendConfig = serverConfigMapper.toBackendConfig(serverConfig = session.serverConfig),
                 isRequestLoggingEnabled = BuildConfig.DEBUG //TODO: Multi-platform logging solution!
             )
+
             val proteusClient: ProteusClient = ProteusClientImpl(rootProteusDirectoryPath, session.userId)
+            runBlocking { proteusClient.open() }
```
vitorhugods, line 36: I don't quite remember what's the importance of closing Cryptobox 😞 

I'd like to gather @typfel's experiene here.

Can't we just do this on `ProteusClientImpl$init`?

```kotlin
actual class ProteusClientImpl(constructorStuff){
    
    init {
        runBlocking { open() }
    }
    // ...
}

MohamadJaara: yep we totally can, the issue here is can we do the same on JS and and IOS since open() is a suspend function (it need to be suspend only because of JS)

typfel: Yeah, @MohamadJaara and I discussed this yesterday. We need to revisit this and potentially implicitly `open()` in the init but then also revise the API so that it's compatible with JS where most things are async.

The importance of closing cryptobox would potentially be to free up any locks which might be taken. But they might anyway be released when he application terminates. I think it's even more important guarantee that we only ever open one cryptobox per session.

typfel (approved):

PR #221, opened 2022-02-17
chore: prepare sessions for encryption (AR-971)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When sending a message, we need to make sure we have cryptographic session with all the recipients. If we don't, then we need to fetch PreKeys for each missing session and establish them.

### Solutions

Created a `OutgoingMessageRecipientsRetriever` which needs a better name.

### Dependencies

Needs `PreKeyRepository`, which is not in place yet. 

* #223 

### Testing

[W.I.P]

----

----
##### References


cryptography/src/jvmMain/kotlin/com/wire/kalium/cryptography/ProteusClientImpl.kt
```diff
@@ -45,6 +45,10 @@ actual class ProteusClientImpl actual constructor(rootDir: String, userId: Strin
         return wrapException { box.newPreKeys(from, count).map { toPreKey(it) } as ArrayList<PreKey> }
     }
 
+    override suspend fun doesSessionExist(sessionId: CryptoSessionId): Boolean {
+        TODO("Nope. Can't do this using cryptobox4j.")
```
vitorhugods, line 49: I think we can discuss with @LukasForst and see if we can turn [CryptoBox#getSession](https://github.com/wireapp/cryptobox4j/blob/4af990ad1874f08c42d306e0b2d0e7646f72f027/src/main/java/com/wire/bots/cryptobox/CryptoBox.java#L334) public and publish a new version of Cryptobox4j.

LukasForst: Sure, I can do that 👍

Or I can add function for checking if the session exists (preferred), the way I see it, Cryptobox4j was designed in a way that consumers of its API won't have access to session directly. Is that enough for you @vitorhugods? Something like `Cryptobox.doesSessionExist(sessionId: String): Boolean`?

LukasForst: `doesSessionExist` is now in release `1.1.1` - https://github.com/wireapp/cryptobox4j/blob/master/src/main/java/com/wire/bots/cryptobox/CryptoBox.java#L347

vitorhugods: @LukasForst I just went for a nap and you solved it already 😍
This is enough, yes! Thank you very much.

typfel (approved):

MohamadJaara (approved):

PR #223, opened 2022-02-17
chore: prepare repository for getting prekeys (AR-1105) 

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

For sending messages, we need preKeys to establish session with all the recipients

### Solutions

Created a repository that returns clients preKeys

### Testing

unit tested

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/prekey/PreKeyMapper.kt
```diff
@@ -1,16 +1,39 @@
 package com.wire.kalium.logic.data.prekey
 
-import com.wire.kalium.cryptography.PreKey
```
vitorhugods, line 3: This import change broke the `toPreKeyDTO` signature, which was expecting a `cryptography` `PreKey`.

This function was being used by `ClientMapper`.

We need to change `RegisterClientParam` to use the new `PreKey` class.

vitorhugods (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/prekey/PreKeyMapper.kt
ohassine: yes I just finished from that now

vitorhugods (approved):
👨🏼‍🍳🤌🏼

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/prekey/PreKey.kt
```diff
@@ -0,0 +1,6 @@
+package com.wire.kalium.logic.data.prekey
+
+data class PreKey(
```
typfel, line 3: Why do we need a `PreKey` type in the logic module? Do we want expose prekeys in the public interface? Otherwise wouldn't it be enough with a PreKey type in cryptography and in the network module?

ohassine: @typfel yes it could be enough..But I wanted that the Crypto module stay independent  from any other module except the logic

MohamadJaara: i also had the same suggestion since there is only 2 ways preyKeys are needed 
1- get some from BE and forword to crypto
2- generate some in crypto and send to the network
we can keep both seperated by having 2 models `PreKeyCrypto` and `PreKeyDTO` and a mapper in `logic`
in this case nor crypto or network will be aware of the other

PR #224, opened 2022-02-17
feat: added get Conversation Details UseCase

borichellow:
# What's new in this PR?

- added `GetConversationDetailsUseCase` for getting details for single Conversations by ID
- fixed IDs problem in `MessageScope`

### Issues

1. prev. `MessageRepository.getMessagesForConversation` was accepting `QualifiedID` from `persistence` as a param. 
But that QualifiedID is not accessible from the other modules.
2. there was no option to get just single Conversations details by ID

### Causes (Optional)

1. I didn't know there are so many `QualifiedID` :) 

### Solutions

1. made `MessageRepository.getMessagesForConversation`to  accept `QualifiedID` from `logic` module as a param and added `idMapper`.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
@@ -2,6 +2,7 @@ package com.wire.kalium.logic.data.conversation
 
 import com.wire.kalium.logic.CoreFailure
 import com.wire.kalium.logic.data.id.IdMapper
+import com.wire.kalium.logic.data.id.QualifiedID
 import com.wire.kalium.logic.data.user.UserId
 import com.wire.kalium.logic.failure.ResourceNotFound
 import com.wire.kalium.logic.functional.Either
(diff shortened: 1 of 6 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -11,13 +12,15 @@ import com.wire.kalium.network.api.user.client.ClientApi
...
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 
 interface ConversationRepository {
     suspend fun fetchConversations(): Either<CoreFailure, Unit>
     suspend fun getConversationList(): Flow<List<Conversation>>
     suspend fun getConversationDetails(conversationId: ConversationId): Either<CoreFailure, Conversation>
+    suspend fun getConversationDetailsFromDB(conversationId: QualifiedID): Flow<Conversation>
```
vitorhugods, line 23:
```suggestion
    suspend fun observeConversationDetails(conversationId: ConversationId): Flow<Conversation>
```
To whoever is consuming the Repository, It shouldn't matter where it comes from. Also, it's better to use `ConversationId`, in case it ever changes in the future. Like it recently did (from single string to qualified).


On a side note, I think `getConversationDetails` shouldn't be the way it is. I don't think we will fetch things from remote sources all the time as we rely heavily on sync and backend events to keep data up-to-date.
Maybe the refactor spirits will give it a go soon when nobody is expecting it 🙏🏼

vitorhugods (changes requested):

vitorhugods (approved):
👨🏼‍🍳🤌🏼

PR #227, opened 2022-02-18
feat: link uploaded asset to profile picture (AR-1077) (AR-1096)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

https://wearezeta.atlassian.net/browse/AR-1077
https://wearezeta.atlassian.net/browse/AR-1096
https://wearezeta.atlassian.net/browse/AR-1119
https://wearezeta.atlassian.net/browse/AR-1120

### Solutions

This PR is the work of the second part of uploading a user avatar (profile picture), consisting in calling the **PUT** `self/` endpoint with the linked asset. Also, to achieve this, I had to map the user data fields from the **GET** `self/` endpoint, an outline of what is included:

- Map and persist the current user data with the new fields to the database
- Call the endpoint self to update the user, the work is ready to update also the name, accentId and email (hopefully) as the endpoint to update is the same but with different data
- Also had to fix some tests, so take the opportunity to extract some common stub logic about the user

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
```diff
@@ -2,36 +2,88 @@ package com.wire.kalium.logic.data.user
...
+    ): UserUpdateRequest {
+        return UserUpdateRequest(
+            id = user.id.value,
+            qualifiedId = idMapper.toApiModel(user.id),
+            name = newName ?: user.name,
+            accentId = newAccent ?: user.accentId,
+            assets = if (newAssetId != null) {
+                listOf(
+                    UserAssetRequest(newAssetId, ImageSize.Complete),
+                    UserAssetRequest(newAssetId, ImageSize.Preview)
```
typfel, line 70: Missing TODO?

The preview image should a separate downscaled image of the complete image.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
```diff
@@ -2,36 +2,88 @@ package com.wire.kalium.logic.data.user
...
+    override fun fromModelToUpdateApiModel(
+        user: SelfUser,
+        newName: String?,
+        newAccent: Int?,
+        newAssetId: String?
+    ): UserUpdateRequest {
+        return UserUpdateRequest(
+            id = user.id.value,
+            qualifiedId = idMapper.toApiModel(user.id),
+            name = newName ?: user.name,
```
typfel, line 65: I suppose we don't want to reset the name to `null` if the `newName` is `null`?

How is this serialized to JSON? Is name property not included in the JSON object if it's null?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -74,8 +76,19 @@ class UserDataSource(
         }
     }
 
+    override suspend fun updateSelfUser(newName: String?, newAccent: Int?, newAssetId: String?): Either<CoreFailure, Unit> {
+        val user = getSelfUser().firstOrNull() ?: return Either.Left(CoreFailure.ServerMiscommunication) // TODO: replace for a DB error
+
+        val updateRequest = userMapper.fromModelToUpdateApiModel(user, newName, newAccent, newAssetId)
+        val updatedSelf = selfApi.updateSelf(updateRequest)
+        return if (updatedSelf.isSuccessful()) {
+            Either.Right(Unit)
```
typfel, line 85: Shouldn't we update the persisted `UserEntity` on success?

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAO.kt
```diff
@@ -11,17 +11,22 @@ data class QualifiedID(
...
+data class UserEntity(
     val id: QualifiedID,
     val name: String?,
-    val handle: String?
+    val handle: String?,
+    val email: String?,
+    val phone: String?,
+    val accentId: Int,
+    val team: String?
+    // val picture: List<UserAsset>
```
typfel, line 22: Since we only have two types of assets II think we can unroll the list of UserAssets into:
`preview_asset_id: String?`
`complete_asset_id: String?`

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
yamilmedina: exactly that was the idea, but after checking we don't serialize null so removed the logic for that handling

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAO.kt
yamilmedina: I added a TODO comment, that will be addressed on upcoming PR, when for example we have an `assets` table to map this field with 👍

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
yamilmedina: This part I had some doubts, specially looking at the current app code and having previews for the avatar or not. 
I will revisit on the next PR when we will be saving these assets into the upcoming `Assets` table for sure, at least this enables us to start working with the profile picture from the Reloaded app =)

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAO.kt
typfel: OK 👍 

My suggestion was that we don't necessarily need an `assets` table and we could include the asset data directly in the user table. But we can discuss further in the upcoming PR then.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
```diff
@@ -2,36 +2,102 @@ package com.wire.kalium.logic.data.user
...
+    ): UserUpdateRequest {
+        return UserUpdateRequest(
+            id = user.id.value,
+            qualifiedId = idMapper.toApiModel(user.id),
+            name = newName,
+            accentId = newAccent,
+            assets = if (newAssetId != null) {
+                listOf(
+                    UserAssetRequest(newAssetId, ImageSize.Complete),
+                    UserAssetRequest(newAssetId, ImageSize.Preview)
```
typfel, line 72:
```suggestion
                    UserAssetRequest(newAssetId, ImageSize.Preview) // TODO should be downscaled image of original asset
```

typfel: Ok let's add a TODO and then I'm good.

typfel (approved):

PR #226, opened 2022-02-18
refactor: refactor CoreFailure part-1 (network)

MohamadJaara:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

the way API, DB, and Crypto errors are handled in logic is a bit chaotic which lead to different implementation of error handling by repositories

### Solutions

there will be a series of PRs to unify error Handling
changes in this PR:
1 - `wrapApiRequest` to handle network calls
2 - rewriting `NetworkResponse` to not expose ktor class `HttpResponse` ffor testing purposes
3 - `LoginApi.login()` now return a `SessionDTO` with the refresh Token (no need to extract the token from the headrs in logic)

Needs releases with:

- [ ] GitHub link to other pull request

### Testing

unit test for `wrapApiRequest`

----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #230, opened 2022-02-22
feat: logout use case

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

atm logout does nothing

### Solutions

implementing `LogoutUseCase` that does the following: 
- [x] make logoutApi request to delete the session remotly
- [x] clear user DB
- [x] clear user Preferences
- [x] delete the session locally
- [x] update the current session

Needs releases with:

- [x] #226 

### Testing

WIP

----

----
##### References


mchenani (approved):
🔥✌️

MohamadJaara:
@ohassine what is logout behaviour in the current app

ohassine:
> @ohassine what is the logout behaviour in the current app

In the current app, a new database is created for every new  logged in user.
we don't delete nor clear the database after user logout

vitorhugods:
> > @ohassine what is the logout behaviour in the current app
> 
> In the current app, a new database is created for every new logged in user. we don't delete nor clear the database after user logout

We just delete the "session"? Without deleting anything else? Like client, etc. etc. ?
That kinda sounds doable.
But depending on how long it has been since the last time you were authenticated, it could mean a lot of events to process. And if there's a bug that "logging out and logging in again" could fix... maybe this wouldn't be possible. 🤔

MohamadJaara:
> There is a corner case that we need to care about: when the user have two logged in accounts 😐

in kalium we create a different DB for each user, that means deleting the DB file for user A will not effect user B

ohassine:
> @MohamadJaara and @ohassine
> 
> Can we scope this down and treat the "recover some data like messages and the like when the user logs in again on the same device" as a different feature?
> 
> That is: for now, we erase everything and we make sure that Logout works.
> 
> If we want to be fancy, we can do it later. That's a different feature.

Alright, we can do it later as different feature.
but this may cost us extra refactoring later.
I have a suggestion: only revoke user session and do not delete the whole database
what do you think ?

MohamadJaara:
@ohassine for now, I will merge the PR as it's and any fancy features we need (e.g. reset user session) can be added in the future. Just to make sure that everything is working as expected.

PR #234, opened 2022-02-23
fix: added serialization/deserialization methods on QualifiedID class

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
Added `JsonSerializable` interface to allow Json serialization/deserialization of simple objects

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/id/QualifiedId.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/id/QualifiedId.kt
@@ -1,16 +1,10 @@
 package com.wire.kalium.logic.data.id
 
-import com.wire.kalium.network.tools.KtxSerializer
+import com.wire.kalium.logic.util.JsonSerializable
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.decodeFromString
-import kotlinx.serialization.encodeToString
-import kotlinx.serialization.json.Json
 
 @Serializable
 data class QualifiedID(
     val value: String,
     val domain: String
-) {
-    fun toJson(): String = Json.encodeToString(this)
-    fun fromJson(jsonString: String): QualifiedID = KtxSerializer.json.decodeFromString(jsonString)
-}
+) : JsonSerializable
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/util/JsonSerializable.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/util/JsonSerializable.kt
@@ -1,0 +1,10 @@
+package com.wire.kalium.logic.util
+
+import com.wire.kalium.network.tools.KtxSerializer
+import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.encodeToString
+
+interface JsonSerializable {
+    fun toJson(): String = KtxSerializer.json.encodeToString(this)
+    fun fromJson(jsonString: String): Any = KtxSerializer.json.decodeFromString(jsonString)
+}


MohamadJaara (commented):
is there a reason why logic models need to be Serializable ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/id/QualifiedId.kt
```diff
@@ -1,8 +1,16 @@
...
+
+@Serializable
 data class QualifiedID(
     val value: String,
     val domain: String
-)
-
-typealias ConversationId = QualifiedID
+) {
+    fun toJson(): String = Json.encodeToString(this)
```
mchenani, line 14: Thanks @gongracr, I have a suggestion, what do you think of making these functions as extensions and generic, so we can access them for all of data model? since we need them in most of models that are playing as arguments between our screens.

vitorhugods, line 14:
```suggestion
    fun toJson(): String = KtxSerializer.json.encodeToString(this)
```

Better to use the same serializer to encode and to decode.

vitorhugods: > So we can access them for for all of data model

If it is generic, how could we be sure that the class we are calling `.toJson()` is actually annotated with `@Serializable`?

From all the data that `logic` exposes, just a handful of classes have this need of being serialised from/to JSON. Mainly IDs I guess.

vitorhugods (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/util/JsonSerializable.kt
```diff
@@ -0,0 +1,10 @@
+package com.wire.kalium.logic.util
+
+import com.wire.kalium.network.tools.KtxSerializer
+import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.encodeToString
+
+interface JsonSerializable {
+    fun toJson(): String = KtxSerializer.json.encodeToString(this)
+    fun fromJson(jsonString: String): Any = KtxSerializer.json.decodeFromString(jsonString)
```
vitorhugods, line 9: How can this be called? It would be good to have some tests to see the usage. 

I can't quite understand the way it currently is.

For `toJson` OK:

```kotlin

val originalQualifiedID = QualifiedID("value", "domain")
val myJsonString = qualifiedID.toJson()
// So far so good. Seems to work well
```


But for `fromJson`:

```kotlin

// I don't have a Qualified ID and I want to call fromJson in order to get one.
// I need an instance of QualifiedID in order to call fromJson a QualifiedID

val dumbID = QualifiedID("", "")
val originalQualifiedID = dumbID.fromJson(myJsonString)
```

vitorhugods (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/util/JsonSerializable.kt
vitorhugods: Maybe something like this can work:

```kotlin


    interface JsonSerializable{
        companion object
    }

    @Serializable
    class Foo(
        @SerialName("bar") val bar: String
    ) : JsonSerializable
    
    fun JsonSerializable.toJsonString() = KtxSerializer.json.encodeToString(this)

    inline fun <reified T : JsonSerializable> fromJsonString(jsonString: String) =
        KtxSerializer.json.decodeFromString<T>(jsonString)


    fun givenOriginalFoo_whenSerializingAndDeserializing_thenTheResultShouldMatchTheOriginal() {
        val originalFoo = Foo("someData")

        val fooString = originalFoo.toJsonString()
        val result = fromJsonString<Foo>(fooString)

        assertEquals(foo, result)
    }
```

gongracr:
Closing as this was still an unclear PR and achieved the arguments passing between screens with a different serialization strategy

PR #238, opened 2022-02-23
feat: send message (AR-912)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Let's communicate, shall we?

### Solutions

`MessageSender` does EXACTLY the thing you would expect it to do.

### Testing

WIP

### Notes

* The only retry situation is for failure due to new clients being added.
* This doesn't handle scheduling or any fancy queueing shenanigans (yet)

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
@@ -36,6 +36,7 @@ import com.wire.kalium.logic.data.prekey.PreKeyMapperImpl
 import com.wire.kalium.logic.data.prekey.PreKeyRepository
 import com.wire.kalium.logic.data.prekey.remote.PreKeyListMapper
 import com.wire.kalium.logic.data.prekey.remote.PreKeyRemoteDataSource
+import com.wire.kalium.logic.data.prekey.remote.PreKeyRemoteRepository
 import com.wire.kalium.logic.data.user.UserDataSource
 import com.wire.kalium.logic.data.user.UserMapperImpl
 import com.wire.kalium.logic.data.user.UserRepository
@@ -138,7 +139,7 @@ abstract class UserSessionScopeCommon(
     private val preKeyListMapper: PreKeyListMapper
         get() = PreKeyListMapper(preyKeyMapper)
 
-    private val preKeyRemoteDatabase: PreKeyRemoteDataSource
+    private val preKeyRemoteDatabase: PreKeyRemoteRepository
         get() = PreKeyRemoteDataSource(authenticatedDataSourceSet.authenticatedNetworkContainer.preKeyApi, preKeyListMapper)
 
     private val preKeyRepository: PreKeyRepository
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
@@ -110,7 +110,7 @@ abstract class UserSessionScopeCommon(
         get() = ClientRegistrationStorageImpl(userPreferencesSettings)
 
     private val clientRepository: ClientRepository
-        get() = ClientDataSource(clientRemoteRepository, clientRegistrationStorage)
+        get() = ClientDataSource(clientRemoteRepository, clientRegistrationStorage, database.clientDAO, userMapper)
 
     private val assetMapper: AssetMapper get() = AssetMapperImpl()
     private val assetRepository: AssetRepository
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
@@ -30,8 +30,12 @@ import com.wire.kalium.logic.data.message.MessageRepository
 import com.wire.kalium.logic.data.message.ProtoContentMapper
 import com.wire.kalium.logic.data.message.SendMessageFailureMapper
 import com.wire.kalium.logic.data.message.SendMessageFailureMapperImpl
+import com.wire.kalium.logic.data.prekey.PreKeyDataSource
 import com.wire.kalium.logic.data.prekey.PreKeyMapper
 import com.wire.kalium.logic.data.prekey.PreKeyMapperImpl
+import com.wire.kalium.logic.data.prekey.PreKeyRepository
+import com.wire.kalium.logic.data.prekey.remote.PreKeyListMapper
+import com.wire.kalium.logic.data.prekey.remote.PreKeyRemoteDataSource
 import com.wire.kalium.logic.data.user.UserDataSource
 import com.wire.kalium.logic.data.user.UserMapperImpl
 import com.wire.kalium.logic.data.user.UserRepository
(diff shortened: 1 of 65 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -26,18 +29,44 @@ class MessageSenderImpl(
...
+        }
+
+        when (val failure = sendResult.value) {
+            is SendMessageFailure.ClientsHaveChanged -> {
+                messageSendFailureHandler.handleClientsHaveChangedFailure(failure).flatMap {
+                    getRecipientsAndAttemptSend(conversationId, messageUuid)
+                }
+            }
+            SendMessageFailure.NoNetworkConnection -> Either.Left(CoreFailure.NoNetworkConnection)
+            else -> Either.Left(failure)
```
typfel, line 69: In these cases we should set message status to `FAILED` or try again? I think we need to different strategies for different kind of messages. User created text messages for example we want to fail and notify the user but for non user generated messages we probably want to retry longer.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendTextMessageUseCase.kt
```diff
@@ -11,31 +12,37 @@ import com.wire.kalium.logic.functional.Either
...
 
+            clientRepository.currentClientId().flatMap { currentClientId ->
                 val message = Message(
-                    id = "someUUID",
+                    id = generatedMessageUuid,
                     content = MessageContent.Text(text),
                     conversationId = conversationId,
-                    date = "25 Jan 2022 13:30:00 GMT",
+                    // TODO: Do we care about clock sync? Should we get server time?
+                    date = Clock.System.now().toString(),
```
typfel, line 38: AFAIK this date is only for our own internal ordering of messages. On iOS we don't care about clock sync except when dealing with calling messages but that happens inside AVS we just provide the server time.

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
vitorhugods: I didn't come up with a full-fledged solution for this yet but it's something I would like to explore when creating the scheduling and retry mechanisms, also when expanding on different message types.

Maybe this part could do that (or delegate to an specialised entity to figure out). Based on the message type and kind of failure we can decide if we:
- Retry
- Give up (marking the message as `FAILED`)
- Just Give Up 

At the moment, this is the current approach:
image

I'd say we can expand later.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -26,18 +29,44 @@ class MessageSenderImpl(
             syncManager.waitForSlowSyncToComplete()
 
             // TODO: make this thread safe (per user)
+            getRecipientsAndAttemptSend(conversationId, messageUuid)
+        }
+
+    private suspend fun getRecipientsAndAttemptSend(conversationId: ConversationId, messageUuid: String): Either<CoreFailure, Unit> =
```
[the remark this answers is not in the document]
vitorhugods: Oooh 🤔.

Interesting problem to solve! I planned this solely thinking : "insert into DB, schedule for sending later using the ID".

But perhaps we can have two ways of attempting to send, one using ID (for regular messages), and one that jumps this step and is something like a `sendTransientMessage`.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
typfel: Sounds like plan!

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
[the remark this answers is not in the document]
typfel: Yep, that should work.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -131,6 +135,15 @@ abstract class UserSessionScopeCommon(
             eventMapper
         )
 
+    private val preKeyListMapper: PreKeyListMapper
+        get() = PreKeyListMapper(preyKeyMapper)
+
+    private val preKeyRemoteDatabase: PreKeyRemoteDataSource
+        get() = PreKeyRemoteDataSource(authenticatedDataSourceSet.authenticatedNetworkContainer.preKeyApi, preKeyListMapper)
```
MohamadJaara, line 142:
```suggestion
    private val preKeyRemoteRepository: PreKeyRemoteRepository
        get() = PreKeyRemoteDataSource(authenticatedDataSourceSet.authenticatedNetworkContainer.preKeyApi, preKeyListMapper)
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -131,6 +135,15 @@ abstract class UserSessionScopeCommon(
...
+
+    private val preKeyRepository: PreKeyRepository
+        get() = PreKeyDataSource(preKeyRemoteDatabase)
```
MohamadJaara, line 145:
```suggestion
        get() = PreKeyDataSource(preKeyRemoteRepository)
```

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
```diff
@@ -0,0 +1,35 @@
...
+         * As seen here: https://github.com/ktorio/ktor/issues/1127
+         * Ktor is pretty has `application/protobuf` but not `application/x-protobuf`
+         * and it doesn't support custom content types!
+         */
+        addInterceptor(object : Interceptor {
+            override fun intercept(chain: Interceptor.Chain): Response {
+                val currentRequest = chain.call().request()
+                val currentHeaders = chain.call().request().headers
+                // When sending a byteArray, Ktor will put `octet-stream`. We can identify this and replace with protobuf
+                val isProtoBuf = currentHeaders["Content-Type"] == "application/octet-stream"
```
MohamadJaara, line 22: what will happen when do need to send `application/octet-stream` hint upload asset will BE complain about having an `application/x-protobuf` in the headers in that case?

vitorhugods: The same that would happens when we want to send `application/x-protobuf` and Ktor+OkHttp sends an `application/octet-stream`: The server returns a `404`, saying it can't find an endpoint (that takes this content).

There are currently no endpoints that take an `application/octet-stream`. So this is a problem I would leave for the older and more experienced versions of ourselves to solve. Hopefully Ktor will have a better support for this in the future.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
vitorhugods: Ooh, indeed!

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
MohamadJaara: [post:provider/assets](https://staging-nginz-https.zinfra.io/api/swagger-ui/#/default/post_provider_assets)
> A body with content type multipart/mixed body. The first section's content type should be application/json. The second section's content type should be always be application/octet-stream. Other content types will be ignored by the server.

MohamadJaara: ok i did try it and the upload was successful, so maybe BE does not care 
anyway i would keep an eye for this one

vitorhugods: Oooooooh I should be more careful. I only saw the main request (`multipart/mixed`).

Indeed!

I've tested it manually here, and because [we just put stuff into the body](https://github.com/wireapp/kalium/blob/70e8d4f499f1ba3084dbde832880dea037cec064/network/src/commonMain/kotlin/com/wire/kalium/network/api/asset/AssetApiImpl.kt#L54), this interceptor doesn't interfere.

So everything works 😎

MohamadJaara (approved):
✉️

typfel (approved):

PR #241, opened 2022-02-25
feat: add user profile picture persistence as a local asset and maps into users field (AR-1122) (AR-1096)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

https://wearezeta.atlassian.net/browse/AR-1122
https://wearezeta.atlassian.net/browse/AR-1096

- This PR adds an assets table to save the assets, initially required for the user avatar (focus of this PR).
- Also, this updates the newly uploaded asset with the user, remotely first and then updating the later entity to have the user profile picture synced
- Also unified some DTOs that were mapping users assets in different places, so created a `UserAssetDTO` to be reused, because it's the same structure for all usages
- Down scaling the preview asset will be addressed later onto the final PR with some final improvements, like; deleting assets, down scaling and chance to return a Bitmap if possible

### Attachments

----

----

----
##### References


typfel (changes requested):

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/asset/AssetDAO.kt
```diff
@@ -0,0 +1,50 @@
+package com.wire.kalium.persistence.dao.asset
+
+data class AssetEntity(
+    val key: String,
+    val domain: String,
+    val token: String?,
+    val name: String?,
+    val encryption: String?,
```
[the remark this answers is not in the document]
yamilmedina: I think it makes sense, I tried your suggestion with the following refactor  (shown below), as `AssetMetadataRequest` is what gets serialized for Ktor.
The only thing that is stopping us, it's that `:cryptography` module is not declared on `:network` as a dependency, and coming back to `UploadAssetData` class, was the way it allows us to transport data between modules.

Do you think we could/should add crypto dependency on network ? 

```suggestion
class AssetMetadataRequest(
    val encryption: String?,
    val mimeType: String,
    val public: Boolean,
    val retentionType: AssetRetentionType,
    @Transient
    val data: ByteArray
) {
    val md5: String = com.wire.kalium.cryptanalysis.utils.calcMd5(data)
}
```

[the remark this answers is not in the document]
typfel: I don't think we should add crypto as a dependency on network. 

I suggest to calculate the md5 in the mapper in :logic which do have access to cryptography module. Maybe we can make it an internal property if decide to create a `util` module for base64, md5, etc.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/asset/AssetDAO.kt
```diff
@@ -0,0 +1,50 @@
...
+    val mimeType: String?,
+    val sha: ByteArray?,
+    val size: Long,
+    val downloaded: Boolean
```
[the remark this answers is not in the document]
yamilmedina: With the current approach, the downloaded field gets updated when we actually download the asset (public) from the server using the `asset-key` and persist it to the db.

For example:

When we login or open the app, we fetch contacts `SyncContactsUseCase` and each contact have an `asset-key` reference, that we should persist when persisting users `UserRepository.fetchUsersByIds`, at that point I'm persisting only the key with `downloaded` field on false. Other case is when I update my avatar from another client, and we have to download the image on the current device.

How this get downloaded? We, later on demand (ie: usecase fetch, work on PR-243) update the `content`, `size` and `downloaded` field. Also, maybe there could be a worker to download pending assets ? Handle that. This worker that I bring it's not ready, and potentially it isn't a good idea 🙈 . 

It will be helpful if we can do a call to discuss this, as this will be needed the correct approach for the next ticket addressing sending/receiving files https://wearezeta.atlassian.net/browse/AR-933

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/asset/AssetDAO.kt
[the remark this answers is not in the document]
yamilmedina: Solved! Doing it in the mapper from now, I can add the `util` (or `commons` 🤔 ) module at the end, sounds a good idea

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/asset/AssetDAO.kt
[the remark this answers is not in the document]
yamilmedina: @typfel I'll re-request the review, after we talked earlier this morning, I think we covered that for profile picture this is ok and downloaded field now it makes more sense.

We'll have to address with asset messages, the refactor or extension of this `assets` table or an intermedius one for messages, we'll discuss it tomorrow with Gonzalo again =)

yamilmedina:
@typfel @gongracr I applied the changes after our interesting discussion on this asset's topic, so I'll be glad to receive your blessings :D, and you take a look at the changes.

Also, the other part (downloading) it's updated on https://github.com/wireapp/kalium/pull/243 accordingly to what we talked, so it's also ready for review 🤝  🚀

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -67,7 +70,10 @@ class UserDataSource(
             }.coFold({
                 Either.Left(it)
             }, {
-                userDAO.insertUsers(it.map(userMapper::fromApiModelToDaoModel))
+                val usersToBePersisted = it.map(userMapper::fromApiModelToDaoModel)
+                // Save (in case there is no data) a reference to the asset id (profile picture)
+                assetRepository.saveUserPictureAsset(mapAssetsForUsersToBePersisted(usersToBePersisted))
```
typfel, line 75: I think you can delete this. We don't need to insert an empty asset.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/asset/AssetDAO.kt
```diff
@@ -0,0 +1,38 @@
+package com.wire.kalium.persistence.dao.asset
+
+data class AssetEntity(
+    val key: String,
+    val domain: String,
+    val mimeType: String?,
+    val rawData: ByteArray?,
```
typfel, line 7: Can't `rawData` be non optional?

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
yamilmedina: Oh, yes absolutely ! I took care of this in the next PR. You could see more at https://github.com/wireapp/kalium/pull/243/commits/eee3ed0726b6983307dedbede2f740244d43f892#diff-f768a46d2d5299842bd48f5a3c654e69fb0ed6712d749b4e82925eebecfeb709L55 #243

typfel: 👍 saw that later

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/asset/AssetDAO.kt
yamilmedina: Thanks for the catch @typfel , done

typfel (approved):

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/asset/AssetDAO.kt
```diff
@@ -0,0 +1,38 @@
...
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = key.hashCode()
+        result = 31 * result + domain.hashCode()
+        result = 31 * result + mimeType.hashCode()
+        result = 31 * result + rawData.contentHashCode()
+        result = 31 * result + downloadedDate.hashCode()
```
gongracr, line 30: I assume the `31` multiplier factor is a widespread hashing convention that be kept internally to each model object??

yamilmedina: Yep, that's right, this it's related to having a `ByteArray` in the class, the IDE suggest implementing the `hashcode` and `equals` functions, so these are the auto-generated functions by the convention using the prime number 31, we should be ok with that implementation tough

PR #243, opened 2022-02-28
feat: add use case to download assets by its asset key/id (AR-1094)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
https://wearezeta.atlassian.net/browse/AR-1094

This change includes the following aspects of downloading a public asset (user's avatar)

- Use case to map the asset by its ID
- Persist locally the binary data of the image
- Return to the consumer application the image as a ByteArray

### Dependencies (Optional)

This PR needs to be merged after 
https://github.com/wireapp/kalium/pull/241

Needs releases with:

- [X] GitHub link to other pull request

### Attachments (Optional)

----

----

----
##### References


yamilmedina:
@typfel @gongracr I made the changes here after our conversation, if you see specifically the last commit (https://github.com/wireapp/kalium/pull/243/commits/eee3ed0726b6983307dedbede2f740244d43f892) you'll see that now the sync it's happening in a one shot execution 🙌  

My only concern it is with the slow sync (now it's actually slower), since the first time we have to download every contact avatar asset, it's a cost we have to assume with this approach. Anyway, after that first sync we don't have a performance degradation from what I could test, since the assets are already present and no download it's made after that.
Happy to receive your PR review guys 😄

typfel:
> My only concern it is with the slow sync (now it's actually slower), since the first time we have to download every contact avatar asset, it's a cost we have to assume with this approach. Anyway, after that first sync we don't have a performance degradation from what I could test, since the assets are already present and no download it's made after that.
> Happy to receive your PR review guys

But why we are downloading the pictures during the slow sync? Why don't we download the profile pictures the first time we want display them in the UI.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -71,8 +71,9 @@ class UserDataSource(
                 Either.Left(it)
             }, {
                 val usersToBePersisted = it.map(userMapper::fromApiModelToDaoModel)
-                // Save (in case there is no data) a reference to the asset id (profile picture)
-                assetRepository.saveUserPictureAsset(mapAssetsForUsersToBePersisted(usersToBePersisted))
+                // not sure about this, because on first sync it's taking ages =/
+                // options, doing it in 2 steps like before and downloading on demand or doing it somehow in a non.blocking fashion
+                assetRepository.downloadUsersPictureAssets(mapAssetsForUsersToBePersisted(usersToBePersisted))
```
typfel, line 76: Yes, we shouldn't do this here but on demand.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetPublicAssetUseCase.kt
```diff
@@ -0,0 +1,16 @@
+package com.wire.kalium.logic.feature.asset
+
+import com.wire.kalium.logic.CoreFailure
+import com.wire.kalium.logic.data.asset.AssetRepository
+import com.wire.kalium.logic.functional.Either
+import com.wire.kalium.logic.functional.suspending
+
+interface GetPublicAssetUseCase {
+    suspend operator fun invoke(assetKey: String): Either<CoreFailure, ByteArray>
```
typfel, line 9: Can't we have `assetKey` be of type `AssetId` or `UserAssetId`?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
typfel: You already have everything for this though?

GetPublicAssetUseCase(user. previewPicture)

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
yamilmedina: I see an issue with this tough, because we have a `FOREIGN` KEY constraint on `Users` Table pointing to the `Assets` table. So the first time we need to persist in the assets table first, then the user (this was the reason of having the empty asset insert 😓 )

The option to achieve the on demand execution, that I also agree, will be removing the constraint and have the relation only at the Kotlin Model level 🤔  Would that make sense ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetPublicAssetUseCase.kt
```diff
@@ -0,0 +1,24 @@
...
+            PublicAssetResult.Failure
+        }) {
+            PublicAssetResult.Success(it)
+        }
+    }
+}
+
+sealed class PublicAssetResult {
+    class Success(val asset: ByteArray) : PublicAssetResult()
+    object Failure : PublicAssetResult()
```
gongracr, line 23: Should we maybe include the core failure as a value here?
```suggestion
    class Failure(val coreFailure: CoreFailure) : PublicAssetResult()
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
yamilmedina: Applied the changes, now we are not using anymore the FK constraint, and we are able to do the sync on demand later via usecase

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetPublicAssetUseCase.kt
yamilmedina: Thanks, changes applied, completely agree that using this parameter suits it better

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetPublicAssetUseCase.kt
yamilmedina: Thanks @gongracr ! Applied the suggestion, and indeed we could need it ie, user feedback purposes 👍

gongracr (approved):
Looking in great shape now. GJ!! 🍙 🥢

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/UploadUserAvatarUseCase.kt
```diff
@@ -29,7 +30,7 @@ class UploadUserAvatarUseCaseImpl(
             .flatMap { asset ->
                 userDataSource.updateSelfUser(newAssetId = asset.key)
             }.fold({
-                UploadAvatarResult.Failure
+                UploadAvatarResult.Failure(it)
             }) {
                 UploadAvatarResult.Success
```
gongracr, line 35: Could we at least return on the use case response the assetId generated after the upload?
```suggestion
                UploadAvatarResult.Success(asset.key)
```

typfel (approved):

typfel:
🚀

PR #251, opened 2022-03-04
feat: multi-platform log solution

alexandreferris:
----

- The **PR Title**
  - [X] contains a reference JIRA issue number like `AR-1203`

- The **PR Description**
----

# What's new in this PR?

Issue on JIRA: [AR-1203](https://wearezeta.atlassian.net/browse/AR-1203)

### Issues

We didn't had a logging solution that could be used across all platforms.

### Solutions

Add [Kermit from touchlab](https://github.com/touchlab/Kermit) in a wrapper class: `KaliumLogger`

### Testing

To test this you will need to receive an event in the app (new message for example).
And then in the Logcat you will see the new logging. (check image below)

### Notes (Optional)

To use the new Logger, you will need to set up:
- Severity level `(VERBOSE, DEBUG, INFO, WARN, ERROR, DISABLED)` // All docs are in the Logger class.
- TAG (To have an easier way to detect from which module/container/place a log is coming from. Default currently is: `KaliumLogger`

### Attachments (Optional)

----
##### References
1. https://github.com/touchlab/Kermit


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ListenToEventsUseCase.kt
```diff
@@ -8,33 +9,36 @@ import com.wire.kalium.logic.functional.suspending
...
     suspend operator fun invoke() {
         syncManager.waitForSlowSyncToComplete()
+
         eventRepository.events()
             //TODO: Handle retry/reconnection
             .collect { either ->
                 suspending {
                     either.map { event ->
-                        println("Event received: $event")
+                        //TODO: Multiplatform logging
```
typfel, line 27:
```suggestion
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ListenToEventsUseCase.kt
```diff
@@ -8,33 +9,36 @@ import com.wire.kalium.logic.functional.suspending
...
+                        kaliumLogger.i(message = "Event received: $event")
                         when (event) {
                             is Event.Conversation -> {
                                 conversationEventReceiver.onEvent(event)
                             }
                             else -> {
                                 //TODO: Multiplatform logging
```
typfel, line 34:
```suggestion
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ListenToEventsUseCase.kt
```diff
@@ -8,33 +9,36 @@ import com.wire.kalium.logic.functional.suspending
...
-                                println("Unhandled event id=${event.id}")
+                                kaliumLogger.i(message = "Unhandled event id=${event.id}")
                             }
                         }
                     }
                 }.onFailure {
                     //TODO: Multiplatform logging
```
typfel, line 40:
```suggestion
```

logger/src/commonMain/kotlin/com/wire/kalium/logger/KaliumLogger.kt
```diff
@@ -0,0 +1,109 @@
+package com.wire.kalium.logger
+
+import co.touchlab.kermit.Severity
+import co.touchlab.kermit.StaticConfig
+import co.touchlab.kermit.Logger as KermitLogger
+
+/**
+ * LoggerType in order from lowest to the highest severity level.
+ */
+enum class LoggerType {
```
typfel, line 10: Isn't this often called LogLevel?

logger/src/commonMain/kotlin/com/wire/kalium/logger/KaliumLogger.kt
```diff
@@ -0,0 +1,109 @@
...
+                minSeverity = config.severityLevel()
+            ),
+            tag = config.tag
+        )
+    }
+
+    val severity = config.severity
+
+    @Suppress("unused")
+    fun v(message: String, throwable: Throwable? = null) =
```
typfel, line 60: Let's use the whole word instead of an abbreviation? Same thing goes for the rest of the methods.

```suggestion
    fun verbose(message: String, throwable: Throwable? = null) =
```

typfel (changes requested):
Nice work!

logger/src/commonMain/kotlin/com/wire/kalium/logger/KaliumLogger.kt
typfel: I'm OK with it if that's the standard in the kotlin world :)

logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -23,16 +23,22 @@ actual class CoreLogic(
...
-) : CoreLogicCommon(clientLabel, rootProteusDirectoryPath) {
+    kaliumLoggerConfig: KaliumLogger.Config
+) : CoreLogicCommon(clientLabel, rootProteusDirectoryPath, kaliumLoggerConfig) {
+
     override fun getAuthenticationScope(): AuthenticationScope =
-        AuthenticationScope(clientLabel = clientLabel, applicationContext = applicationContext)
+        AuthenticationScope(
+            clientLabel = clientLabel,
+            applicationContext = applicationContext,
+            kaliumLogger = kaliumLogger
```
gongracr, line 33: Why do we need to pass around the instance of `kaliumLogger`? Can't we get it as a Singleton instance created on the WireApplication level?

logger/src/commonMain/kotlin/com/wire/kalium/logger/KaliumLogger.kt
alexandreferris: So, thought on that but followed with the abbreviation due to all the logging/Loggers having it abbreviates (Log.i/Log.id) and also from the Kermit lib itself, but it would indeed make more sense having it as the whole word.

Will change it!

vitorhugods (changes requested):

logger/src/commonMain/kotlin/com/wire/kalium/logger/KaliumLogger.kt
alexandreferris: Decided to rename it to: `KaliumLogLevel` as `LogLevel` was ambiguous due to Kermit lib.

logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
alexandreferris: New module-wide `[Module]Logger` object was added to hold an instance instead of passing it down to where it needed.

You can check: `CoreLogger` specifically, for more info on how it was implemented, and also `ListenToEventsUseCase` on how it is now being used.

MohamadJaara (commented):
Small comment about adding iOS as a target.
And I wonder if we can extend it to add the name of the function where the logging is happing, e.g.
```kotlin
    private fun getPrefix(): String {
        val prefix: String
        try {
            throw NullPointerException()
        }
        catch (e: NullPointerException) {
            val stackTrace = e.stackTrace[2]
            prefix = stackTrace.className + " @ " + stackTrace.methodName + "(): "
        }
        return prefix
    }
    
    fun w(message: String, throwable: Throwable? = null) =
        throwable?.let {
            kermitLogger.w(getPrefix() + message, throwable)
        } ?: kermitLogger.w(getPrefix() + message)
```
The issue is the code above works only on android/JVM and I did not find a way to do the same thing on JS nor iOS

vitorhugods:
> Small comment about adding iOS as a target. And I wonder if we can extend it to add the name of the function where the logging is happing, e.g.
> 
> ```kotlin
>     private fun getPrefix(): String {
>         val prefix: String
>         try {
>             throw NullPointerException()
>         }
>         catch (e: NullPointerException) {
>             val stackTrace = e.stackTrace[2]
>             prefix = stackTrace.className + " @ " + stackTrace.methodName + "(): "
>         }
>         return prefix
>     }
>     
>     fun w(message: String, throwable: Throwable? = null) =
>         throwable?.let {
>             kermitLogger.w(getPrefix() + message, throwable)
>         } ?: kermitLogger.w(getPrefix() + message)
> ```
> 
> The issue is the code above works only on android/JVM and I did not find a way to do the same thing on JS nor iOS


Indeed [Throwable.stackTrace](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/stack-trace.html) is only available on JVM targets right now.

We could simplify this a bit by doing this:

```kotlin
private fun getPrefix(): String {
    val stackEntry = Throwable().stackTrace[2]
    return "${stackEntry.className}@${stackEntry.methodName}(): "
}
```


----

As this isn't supported on all platforms _yet_. I guess we can leave this aside and maybe implement on a future PR.

We need at least a working logger so we can start using it properly. After it's in place we can look for improvements.

logger/build.gradle.kts
```diff
@@ -0,0 +1,50 @@
+plugins {
+    Plugins.androidLibrary(this)
+    Plugins.multiplatform(this)
+    Plugins.serialization(this)
```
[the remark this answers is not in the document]
vitorhugods:
```suggestion
```

logger/build.gradle.kts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    android()
+    js(IR) {
+        browser {
+            commonWebpackConfig {
+                cssSupport.enabled = true
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
vitorhugods: Indeed!
We have iOSX64 in `persistence` module.

@alexandreferris could you add a `PersistenceLogger` in the `persistence` module?

This way we have logging in all lib modules.

vitorhugods (approved):
Weeelcome 🚀🧑‍🚀

PR #255, opened 2022-03-09
ci: run tests on all targets

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

iOS, JVM and JS tests were not being executed in CI.
This lead to some tests being broken, which I fixed in this PR.

### Causes

`./gradlew test` was being called, which only run Android Unit tests. DB Tests are not `Android Unit tests` but rather Instrumented tests.

### Solutions

* `./gradlew check` instead. Which run on all targets.
* Disable JS tests for persistence, as the DB is not yet implemented.

### Testing

PR is all about tests!

----

----
##### References


vitorhugods:
I have the feeling that this will not pass CI right now, as it will attempt to run JVM tests, which rely on `cryptobox` and `libsodium`.

So we still need to run `make` and pass the libs as arguments when running the test commands.

.github/workflows/gradle-run-tests.yml
```diff
@@ -20,13 +20,17 @@ jobs:
           cache: gradle
       - name: Validate Gradle wrapper
         uses: gradle/wrapper-validation-action@e6e38bacfdf1a337459f332974bb2327a31aaf4b
+      - name: Build Native dependencies
+        run: make
       - name: Test with Gradle
-        run: ./gradlew test
+        run: ./gradlew check
```
yamilmedina, line 26: I think if we just want to unit test on this step, we could manage to shave some minutes 👀  if we just run the test tasks directly, ie: `./gradlew :logic:test --max-workers 2 --parallel && ./gradlew :persistence:test --max-workers 2 --parallel` and so on. When we run the check task, I think it will also run for example linting and things that we could run on another gh action.

https://github.com/wireapp/kalium/runs/5494226737?check_suite_focus=true#step:6:663

vitorhugods: I found out the hard way that `test` will only run Android Unit tests.

The idea here is to actually run tests on all targets.

For example: run DB, crypto, network and logic tests on iOS, JVM, JS and on Android.

----

If we want to reduce tests to avoid having simple tests running on all targets, we could start moving some of the tests cases from `commonTest` (runs on all targets) to `jvmTest` (only on JVM).

----

To remove linting and other possible checks, I guess I'll try `testAll` instead.

.github/workflows/gradle-run-tests.yml
```diff
@@ -20,13 +20,17 @@ jobs:
...
-        run: ./gradlew test
+        run: ./gradlew allTests
       - name: Android Instrumentation Tests
         uses: reactivecircus/android-emulator-runner@v2
         with:
           api-level: 29
-          script: ./gradlew cryptography:connectedAndroidTest
+          script: |
+            ./gradlew cryptography:connectedAndroidTest
+            ./gradlew persistence:connectedAndroidTest
```
MohamadJaara, line 33: Just a suggestion
```suggestion
          run: |
            ./gradlew cryptography:connectedAndroidTest
            ./gradlew persistence:connectedAndroidTest
```

vitorhugods: No can do.

`script` is an argument taken by the [`android-emulator-runner`](https://github.com/ReactiveCircus/android-emulator-runner) action.

It is not a regular `run` command from GitHub Actions.

PR #257, opened 2022-03-09
feat:  teams repository usecase creation

alexandreferris:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

[AR-1221](https://wearezeta.atlassian.net/browse/AR-1221)

This PR contains only the addition of:
- Team Repository
- Team Mapper
- `fetchTeamById` is now being used in `fetchSelfUser`, due to the need of having the user teamId. (explanation in details in comments in `fetchSelfUser`)
- `fetchTeamById` is already using the new API path and not the deprecated anymore, thus now returning only a single team and not a list of teams.

### Dependencies

Needs releases with:
- [X] #254 (Already reviewed and merged into the epic)

### Notes (Optional)

This is only the seconds PR, more subsequent PRs will `probably` appear pointing to the epic.
Before reviewing this PR, I suggest reviewing the [first PR](https://github.com/wireapp/kalium/pull/254).

PR #254, opened 2022-03-09
feat: teams database and DAO creation

alexandreferris:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

[AR-1221](https://wearezeta.atlassian.net/browse/AR-1221)

This PR contains only the addition of:
- Team queries
- Team DAO / Team DAO Implementation
- TeamDAO usage in `Database.kt` through the persistence module.

### Notes (Optional)

This is only the first PR, more subsequent PRs will appear pointing to the epic.

PR #270, opened 2022-03-11
test: team fetch/sync

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

[AR-1221](https://wearezeta.atlassian.net/browse/AR-1221)

Added tests for:
- TeamMapper
- TeamRepositoryTest
- TeamDAO

### Dependencies (Optional)

Needs releases with:

- [X] #254 
- [X] #257


MohamadJaara (approved):
lgtm 🚀

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/team/TeamRepositoryTest.kt
```diff
@@ -0,0 +1,80 @@
...
+    fun setUp() {
+        teamRepository = TeamDataSource(
+            teamDAO = teamDAO,
+            teamMapper = teamMapper,
+            teamsApi = teamsApi
+        )
+    }
+
+    @Test
+    fun givenSelfUserExists_whenGettingTeamInfo_thenTeamInfoShouldBeReturned() = runTest {
```
vitorhugods, line 42: We could add another test and assert if `teamDAO` was called to insert the mapped team info.

Something like `givenTeamApiReturnsSuccessfully_whenGettingTeamInfo_thenTeamInfoIsPersisted`.

Also, another couple of tests to make sure that errors in the API or DAO are propagated correctly would be nice:

* `givenTeamApiFails_whenGettingTeamInfo_thenTheFailureIsPropagated`
* `givenTeamDAOFails_whenGettingTeamInfo_thenTheFailureIsPropagated`

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/team/TeamRepositoryTest.kt
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+        given(teamMapper)
+            .function(teamMapper::fromApiModelToDaoModel)
+            .whenInvokedWith(oneOf(team))
+            .then { teamEntity }
+
+        given(teamDAO)
+            .coroutine { insertTeam(team = teamEntity) }
+            .then { } // returns Unit
```
vitorhugods, line 75: You can initialise `teamDAO` like this, so you don't need to stub methods that return `Unit`:

```kotlin
    @Mock
    private val teamDAO = configure(mock(classOf<TeamDAO>())){
        stubsUnitByDefault = true
    }
```

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/team/TeamRepositoryTest.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        val result = teamRepository.fetchTeamById(teamId = "teamId")
+        assertEquals(Either.Right(Unit), result)
```
vitorhugods, line 78: This tests only asserts success in the end, so we might as well just assert the success.

```suggestion
        result.shouldSucceed()
```

This way, if we ever change what `fetchTeamById` returns, maybe from `Either<Failure, Unit>` to `Either<Failure, TeamStuff>`, this test won't break unnecessarily.

vitorhugods (changes requested):
Nicely done. But it would be cool to have a couple more tests here or there.

PR #276, opened 2022-03-15
feat: Add SyncSelfTeamUseCase

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

[AR-1220](https://wearezeta.atlassian.net/browse/AR-1220)

Moved the logic to fetch self user team information from `UserRepository` to a proper UseCase to be used in `SlowSyncWorker`


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -42,8 +41,7 @@ class UserDataSource(
     private val userApi: UserDetailsApi,
     private val idMapper: IdMapper,
     private val userMapper: UserMapper,
-    private val assetRepository: AssetRepository,
-    private val teamRepository: TeamRepository
+    private val assetRepository: AssetRepository
```
vitorhugods, line 44: The amount of dependencies here really made it clear that it was doing too much stuff.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/team/SyncSelfTeamUseCase.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.wire.kalium.logic.data.team.TeamRepository
+import com.wire.kalium.logic.data.user.UserRepository
+import com.wire.kalium.logic.functional.Either
+import kotlinx.coroutines.flow.first
+
+interface SyncSelfTeamUseCase {
+    suspend operator fun invoke(): Either<CoreFailure, Unit>
+}
+
+class SyncSelfTeamUseCaseImpl(
```
vitorhugods, line 13:
```suggestion
internal class SyncSelfTeamUseCaseImpl(
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/team/SyncSelfTeamUseCase.kt
```diff
@@ -0,0 +1,22 @@
+package com.wire.kalium.logic.feature.team
+
+import com.wire.kalium.logic.CoreFailure
+import com.wire.kalium.logic.data.team.TeamRepository
+import com.wire.kalium.logic.data.user.UserRepository
+import com.wire.kalium.logic.functional.Either
+import kotlinx.coroutines.flow.first
+
+interface SyncSelfTeamUseCase {
```
vitorhugods, line 9: I guess we can make it internal.

```suggestion
internal interface SyncSelfTeamUseCase {
```

These sync UseCases are not really supposed to be exposed to whoever is consuming Kalium, but rather for Kalium workers to orchestrate.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/team/TeamScope.kt
```diff
@@ -0,0 +1,14 @@
+package com.wire.kalium.logic.feature.team
+
+import com.wire.kalium.logic.data.team.TeamRepository
+import com.wire.kalium.logic.data.user.UserRepository
+
+class TeamScope(
+    private val userRepository: UserRepository,
+    private val teamRepository: TeamRepository
+) {
+    val syncSelfTeamUseCase: SyncSelfTeamUseCase get() = SyncSelfTeamUseCaseImpl(
```
vitorhugods, line 10:
```suggestion
    internal val syncSelfTeamUseCase: SyncSelfTeamUseCase get() = SyncSelfTeamUseCaseImpl(
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/team/SyncSelfTeamUseCase.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class SyncSelfTeamUseCaseImpl(
+    private val userRepository: UserRepository,
+    private val teamRepository: TeamRepository
+) : SyncSelfTeamUseCase {
+
+    override suspend fun invoke(): Either<CoreFailure, Unit> =
+        userRepository.getSelfUser().first().team?.let { teamId ->
+            teamRepository.fetchTeamById(teamId = teamId)
+        } ?: Either.Right(Unit)
```
vitorhugods, line 21: Fail silently?

_AT LEAST_ a warning log.

This could be an `UnknownFailure(IllegalStateException("SyncSelfTeam cannot be called before SyncSelfUser has succeeded")`

vitorhugods (changes requested):
Some tiny suggestions.

Overall it's a step in the right direction.
`fetchSelfUser` was doing too much, and `syncSelfTeamUseCase` can also handle fetching the members of the team.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/team/SyncSelfTeamUseCase.kt
alexandreferris: Actually a user could not have a team, thus just keep the flow going, if there is an error on fetching the user team, then an exception is thrown.

Otherwise if a User doesn't have a team, then we continue the flow as it should normally (following to syncing contacts).

Does it makes sense? That's how I read the SyncFlow we are having based on users that have and do not have a team.

vitorhugods: Oh shoot!

Nevermind my past comment. I got it wrong :P 

I was thinking about the possibility of `getSelfUser()` returning null, but it returns a `Flow`. So this won't happen.

In any case, I guess a logging saying `Skipping team sync because user doesn't belong to a team` would be a good to have.

alexandreferris: 🙈  Sure! I will be adding it in the next PR as I have already added some more things to this UseCase hehe

vitorhugods (approved):

PR #279, opened 2022-03-16
feat: Fetch and Store Team members

alexandreferris:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

[AR-1220](https://wearezeta.atlassian.net/browse/AR-1220)

Addition of fetching members by teamId in `TeamRepository` and its flow, to be used in `SyncSelfTeamUseCase`.

The mapping function was made in `UserMapper` due to it being related to a User, and it is created/mapped with default/null values because most of it will be replaced in the next Slow Sync steps and thus having it populated later on in the process.

### Dependencies (Optional)

Needs releases with:

- [X] #276

PR #285, opened 2022-03-16
test: fetch TeamMembers

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

[AR-1220](https://wearezeta.atlassian.net/browse/AR-1220)

Added tests for:
- TeamRepositoryTest
- And added a new test file: `UserMapperTest`, which it is only testing the TeamMember to UserEntity mapping function.

### Dependencies (Optional)

Needs releases with:

- [X] #276 
- [X] #279
- [X] #284

### DIFF
--- a/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/team/TeamRepositoryTest.kt
+++ b/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/team/TeamRepositoryTest.kt
@@ -116,7 +116,6 @@ class TeamRepositoryTest {
             legalHoldStatus = LegalHoldStatusResponse.NO_CONSENT,
             createdBy = "nonQualiefiedUserId1",
             permissions = TeamsApi.Permissions(copy = 1, own = 1)
-
         )
 
         val teamMembersList = TeamsApi.TeamMemberList(
--- a/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/team/TeamRepositoryTest.kt
+++ b/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/team/TeamRepositoryTest.kt
@@ -8,13 +8,17 @@ import com.wire.kalium.network.api.AssetId
 import com.wire.kalium.network.api.ErrorResponse
 import com.wire.kalium.network.api.model.TeamDTO
 import com.wire.kalium.network.api.teams.TeamsApi
+import com.wire.kalium.network.api.user.LegalHoldStatusResponse
 import com.wire.kalium.network.exceptions.KaliumException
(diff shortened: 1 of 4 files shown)


logic/src/commonTest/kotlin/com/wire/kalium/logic/data/team/TeamRepositoryTest.kt
```diff
@@ -103,4 +107,72 @@ class TeamRepositoryTest {
...
+
+    @Test
+    fun givenTeamIdAndUserDomain_whenFetchingTeamMembers_thenTeamMembersShouldBeSuccessful() = runTest {
+        val teamMember = TeamsApi.TeamMember(
+            nonQualifiedUserId = "teamMember1",
+            createdAt = "01011970",
+            legalHoldStatus = LegalHoldStatusResponse.NO_CONSENT,
+            createdBy = "nonQualiefiedUserId1",
+            permissions = TeamsApi.Permissions(copy = 1, own = 1)
+        )
```
yamilmedina, line 119: Would it be a good idea to extract this duplicated stub to an external class/function, so we can reuse it ?

yamilmedina (commented):
Just a minor suggestion about a possible component we could reuse for testing, but apart from that, everything looks good! =)

typfel (approved):

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/team/TeamRepositoryTest.kt
alexandreferris: I was thinking about something like having another module called: test-utils and only have [Data]Generator.kt classes, where we could generate/create any type of classes we need

vitorhugods: @alexandreferris `test-utils` would depend on `network` to know what `TeamsApi.TeamMember` is. And `network` wouldn't be able to import `test-utils`. Only `logic` would be able to import it.

So it's as good as just extracting to another file right now.
At least we reduce redundancy inside `logic`.

alexandreferris: Great point @vitorhugods !

I guess then we will only be able to move forward on having the `test-utils` once we also have a specific domain model classes module.

For now I have added `TeamGenerator.kt` to create both a TeamDTO and a TeamMember (based on the scope of this ticket).

yamilmedina (approved):
looks great! 🚀  🌔

PR #294, opened 2022-03-17
feat: implement upload private asset api

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Description

- Modified `uploadAndPersistPublicAsset` and added `uploadAndPersistPrivateAsset` into the asset repository. 
- Added `UploadPrivateAssetUseCase` 
- Added `UploadPrivateAssetUseCaseTest`

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/UploadUserAvatarUseCase.kt
```diff
@@ -14,21 +12,20 @@ interface UploadUserAvatarUseCase {
...
 internal class UploadUserAvatarUseCaseImpl(
     private val userDataSource: UserRepository,
     private val assetDataSource: AssetRepository
 ) : UploadUserAvatarUseCase {
 
-    override suspend operator fun invoke(mimeType: String, imageData: ByteArray): UploadAvatarResult = suspending {
-        assetDataSource
-            .uploadAndPersistPublicAsset(UploadAssetData(imageData, ImageAsset.JPG, true, RetentionType.ETERNAL))
+    override suspend operator fun invoke(imageData: ByteArray): UploadAvatarResult = suspending {
+        val mimeType = ImageAsset.JPEG
```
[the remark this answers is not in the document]
gongracr: I think the idea was to basically use JPEG as the defined format for compression and storage optimisation reasons.

trOnk12 (commented):
missing lots of context but regarding clean code, looks good to me

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/UploadPrivateAssetUseCase.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import com.wire.kalium.logic.CoreFailure
+import com.wire.kalium.logic.data.asset.AssetRepository
+import com.wire.kalium.logic.data.asset.AssetType
+import com.wire.kalium.logic.data.asset.UploadedAssetId
+import com.wire.kalium.logic.data.user.UserRepository
+import com.wire.kalium.logic.functional.Either
+import com.wire.kalium.logic.functional.suspending
+
+interface UploadPrivateAssetUseCase {
```
typfel, line 11: What's the plan for how to use this use case?

I would have expected use cases like

SendImageUseCase()
SendVideoUseCase()

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/UploadPrivateAssetUseCase.kt
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Function allowing the upload of a user profile picture (avatar)
+     * This first will upload the data as an asset and then will link this asset with the [User]
+     *
+     * @param mimeType mime type of the asset to be sent
+     * @param assetEncryptedData binary data of the actual asset to be sent
+     * @return an [Either] tuple containing [CoreFailure] in case of an error or the [UploadedAssetId] containing the identifier of the
+     * successfully uploaded asset
+     */
+    suspend operator fun invoke(mimeType: AssetType, assetEncryptedData: ByteArray): Either<CoreFailure, UploadedAssetId>
```
typfel, line 21: `assetEncryptedData` - that would mean we let the UI deal with encryption which I don't think we should do.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/UploadPrivateAssetUseCase.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.wire.kalium.logic.data.asset.AssetType
+import com.wire.kalium.logic.data.asset.UploadedAssetId
+import com.wire.kalium.logic.data.user.UserRepository
+import com.wire.kalium.logic.functional.Either
+import com.wire.kalium.logic.functional.suspending
+
+interface UploadPrivateAssetUseCase {
+    /**
+     * Function allowing the upload of a user profile picture (avatar)
+     * This first will upload the data as an asset and then will link this asset with the [User]
```
typfel, line 14: Copy paste error?

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/UploadPrivateAssetUseCase.kt
vitorhugods: Yep.

If it is private, public, whatever, whatever, it should not concern the apps (who is consuming the UseCase) to decide.

I guess this should not be exposed to consuming apps. It's only a function that the repository should have.



From lower to higher level (please disconsider possible naming mismatches):

### `AssetApi`

Will take some data in the REST API format and upload to the backend. An almost completely dumb interface in which we can interact with the API through Kotlin code.

### `AssetRepository`

Will map business logic data, maybe provide some utilities like writing and reading `AssetDAO` for caching, and and access `AssetApi` when the time comes to upload it.

**`uploadPrivateAsset` could go in here**

### `SendImageUseCase`, `SendVideoUseCase`, `SendAudioUseCase`, etc.

These can take bytearrays (or similar in order to make Platform-agnostic) into them, and a conversation id.
If needed, they will call the `AssetRepository` to insert assets into DB for sending in the future.
Maybe they can wait for sync to be done, or things like this.
This can try to send a message now, or schedule messages to be sent in the future.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/UploadPrivateAssetUseCase.kt
gongracr: Good catch 🙈

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/UploadPrivateAssetUseCase.kt
gongracr: Thanks for the comments! And yes, that sounds about right. This shouldn't be wrapped as a public consumption use case. So imo there now a few ways of proceeding now in this PR:

1. Refactoring `UploadPrivateAssetUseCase` to something like the proposed use cases above (`SendImageUseCase`, `SendVideoUseCase`, `SendAudioUseCase`, etc.). However these use cases would have to be incomplete for now as there are still some missing pieces required there, like encryption flow, add Asset Message model objects, etc..
2. Removing the `UploadPrivateAssetUseCase` and just leaving the `assetRepository.uploadAndPersistPrivateAsset` method for now. Then later on, once we have the asset encryption methods in place we can try to add the above mentioned use cases.
3. Declaring the `UploadPrivateAssetUseCase` as internal, so that it is not consumable outside of kalium logic module but can be reused later one by a public use case that requires this private asset upload.

After reading your comments I assume this PR should be about doing something similar like point 2, but I just don't want to jump into the refactor wagon without having a clearer picture of it.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/UploadPrivateAssetUseCase.kt
gongracr: This UseCase should definitely not be publicly visible. Check the comment above. However I imagine the asset raw data should be still passed around as a ByteArray right?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/UploadUserAvatarUseCase.kt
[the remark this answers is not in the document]
MohamadJaara: Even if it's compressed, the Type should not be hard codded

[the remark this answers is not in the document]
gongracr: Does that mean that we want to allow users to upload raw image formats or not so lossy ones like .raw or .png?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/UploadPrivateAssetUseCase.kt
gongracr: removed it

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetRepository.kt
```diff
@@ -22,19 +24,28 @@ internal class AssetDataSource(
...
+        val uploadAssetData = UploadAssetData(assetData, mimeType, true, RetentionType.ETERNAL)
+        return doUploadAndPersistAsset(uploadAssetData)
+    }
+
+    override suspend fun uploadAndPersistPrivateAsset(mimeType: AssetType, assetData: ByteArray): Either<CoreFailure, UploadedAssetId> {
+        val uploadAssetData = UploadAssetData(assetData, mimeType, false, RetentionType.PERSISTENT)
+        return doUploadAndPersistAsset(uploadAssetData)
+    }
+
+    private suspend fun doUploadAndPersistAsset(uploadAssetData: UploadAssetData): Either<NetworkFailure, UploadedAssetId> = suspending {
```
typfel, line 37:
```suggestion
    private suspend fun uploadAndPersistAsset(uploadAssetData: UploadAssetData): Either<NetworkFailure, UploadedAssetId> = suspending {
```


I would drop the `do` prefix, `upload` is already a verb.

typfel (approved):
Some naming nit picking otherwise looks good to me.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/UploadUserAvatarUseCase.kt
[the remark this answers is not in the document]
gongracr: Let's just stick to this JPEG hardcoded version of the mimetype for now and revisit it in case there are any future issues or we decide to limit the avatar image size.

PR #296, opened 2022-03-18
feat: add hide/delete message [AR-1123] [AR-1127]

mchenani:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?
Add delete and hide message

### Issues
[AR-1123] 
[AR-1127]

### Testing

Tests added for protobuf encoder/decoder

----
##### References

[AR-1123]: https://wearezeta.atlassian.net/browse/AR-1123
[AR-1127]: https://wearezeta.atlassian.net/browse/AR-1127

### DIFF
--- a/logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
+++ b/logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
@@ -2,8 +2,9 @@ package com.wire.kalium.logic.data.message
 
 import com.waz.model.Messages
 import com.waz.model.Messages.GenericMessage
+import com.wire.kalium.logic.data.id.ConversationId
 
-class ProtoContentMapperImpl: ProtoContentMapper {
+class ProtoContentMapperImpl : ProtoContentMapper {
 
     override fun encodeToProtobuf(protoContent: ProtoContent): PlainMessageBlob {
         val (messageUid, messageContent) = protoContent
@@ -17,6 +18,24 @@ class ProtoContentMapperImpl: ProtoContentMapper {
                     .build()
                 builder.text = text
             }
+            is MessageContent.DeleteMessage -> {
+                val deleted = Messages.MessageDelete.newBuilder()
+                    .setMessageId(messageContent.messageId)
+                    .build()
+                builder.deleted = deleted
+            }
+            is MessageContent.HideMessage -> {
+                val qualifiedConversationId = Messages.QualifiedConversationId.newBuilder()
+                    .setId(messageContent.conversationId.value)
+                    .setDomain(messageContent.conversationId.domain).build()
+                val hidden = Messages.MessageHide.newBuilder()
+                    .setMessageId(messageContent.messageId)
+                    //based on the documentation this conversation id is deprecated, but somehow it crashed when not passing it!
+                    .setConversationId(messageContent.conversationId.value)
+                    .setQualifiedConversationId(qualifiedConversationId)
+                    .build()
+                builder.hidden = hidden
+            }
             else -> {
                 throw IllegalArgumentException("Unexpected message content type: $messageContent")
             }
@@ -28,10 +47,14 @@ class ProtoContentMapperImpl: ProtoContentMapper {
         val genericMessage = GenericMessage.parseFrom(encodedContent.data)
 
         //TODO Handle other message types
-        val content = if (genericMessage.hasText()) {
-            MessageContent.Text(genericMessage.text.content)
-        } else {
-            MessageContent.Unknown
+        val content = when {
+            genericMessage.hasText() -> MessageContent.Text(genericMessage.text.content)
+            genericMessage.hasDeleted() -> MessageContent.DeleteMessage(genericMessage.messageId)
+            genericMessage.hasHidden() -> MessageContent.HideMessage(
+                genericMessage.messageId,
+                ConversationId(genericMessage.hidden.qualifiedConversationId.id, genericMessage.hidden.qualifiedConversationId.domain)
+            )
+            else -> MessageContent.Unknown
         }
         return ProtoContent(genericMessage.messageId, content)
     }
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
@@ -22,7 +22,7 @@ class DeleteMessageUseCase(
     private val messageSender: MessageSender
 ) {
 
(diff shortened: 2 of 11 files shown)


logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
```diff
@@ -17,6 +18,24 @@ class ProtoContentMapperImpl: ProtoContentMapper {
...
+                    .build()
+                builder.deleted = deleted
+            }
+            is MessageContent.HideMessage -> {
+                val qualifiedConversationId = Messages.QualifiedConversationId.newBuilder()
+                    .setId(messageContent.conversationId.value)
+                    .setDomain(messageContent.conversationId.domain).build()
+                val hidden = Messages.MessageHide.newBuilder()
+                    .setMessageId(messageContent.messageId)
+                    //based on the documentation this conversation id is deprecated, but somehow it crashed when not passing it!
```
typfel, line 33: It's deprecated because it's superseded by `QualifiedConversationId` but we can't remove it because:
1. it's marked as `required` in the protobuf definition and therefore it must always be included
2. We need be compatible with older clients

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
```diff
@@ -0,0 +1,57 @@
...
+
+class DeleteMessageUseCase(
+    private val messageRepository: MessageRepository,
+    private val userRepository: UserRepository,
+    private val clientRepository: ClientRepository,
+    private val syncManager: SyncManager,
+    private val messageSender: MessageSender
+) {
+
+    suspend operator fun invoke(conversationId: ConversationId, messageId: String, deleteForEveryOne: Boolean): Either<CoreFailure, Unit> {
```
typfel, line 25:
```suggestion
    suspend operator fun invoke(conversationId: ConversationId, messageId: String, deleteForEveryone: Boolean): Either<CoreFailure, Unit> {
```

typfel, line 25: Can we have a typealias for `messageId`?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
```diff
@@ -0,0 +1,57 @@
...
+                        MessageContent.DeleteMessage(messageId)
+                    else
+                        MessageContent.HideMessage(messageId, conversationId),
+                    conversationId = conversationId,
+                    date = Clock.System.now().toString(),
+                    senderUserId = selfUser.id,
+                    senderClientId = currentClientId,
+                    status = Message.Status.PENDING
+                )
+                messageSender.getRecipientsAndAttemptSend(conversationId, message)
```
typfel, line 45: Might be worth to creating a helper method for constructing messages since only `content` and `conversationId` are relevant the others are always constructed in the same way.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
```diff
@@ -0,0 +1,57 @@
...
+            }.flatMap {
+                messageRepository.deleteMessage(messageId, conversationId)
+            }.onFailure {
+                println(it)
```
typfel, line 49: Use the logger?

typfel (changes requested):
I don't see how we delete messages when we receive a `DeleteMessage` is that not implemented yet?

mchenani:
> I don't see how we delete messages when we receive a `DeleteMessage` is that not implemented yet?

My bad, I didn't specify that in the title. Yes this PR is only adding send delete/hid message and receiving those events are in another wip branch.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
mchenani: I'm afraid not getting your point!

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
mchenani: Agree, actually thinking of that. but at the moment I'd like to see at least one or two events then we can create a better structure because all of these are kinds of actions/utils/extensions around the Message class.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
mchenani: Sure, missed that one! thanks 👍

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
```diff
@@ -0,0 +1,57 @@
...
+
+            clientRepository.currentClientId().flatMap { currentClientId ->
+                val message = Message(
+                    id = generatedMessageUuid,
+                    content = if (deleteForEveryOne)
+                        MessageContent.DeleteMessage(messageId)
+                    else
+                        MessageContent.HideMessage(messageId, conversationId),
+                    conversationId = conversationId,
+                    date = Clock.System.now().toString(),
```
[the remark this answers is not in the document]
typfel: The timestamp string should be in the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators) format in UTC+0 (designated with a Z at the end).

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
typfel: I was thinking we could introduce a typealias for the `messageId` in the same way we do for the clientID to get some better type checking.
```
typealias ClientId = PlainId
```

But I see now that we deal with strings everywhere so I'm fine with doing tackling this in another PR.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/DeleteMessageUseCase.kt
typfel, line 45: The `HideMessage` shouldn't be sent in the conversation of the message, because we only want to send the `HideMessage` to your own devices and therefore we should send it to the "self conversation". The "self conversation" is special conversation which is created by the backend and is of type `self`.

I I think you need to wait for https://github.com/wireapp/kalium/pull/280 to implement this though. If you don't want to wait for it add a TODO.

PR #297, opened 2022-03-18
feat: add "shouldStoreSession" to login and register

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There was a possibility of user logging in or creating account but not registering his client - we could fix that by either checking at the app launch if the client is registered and force user to do so if not or just by simply making both actions atomic (login + register client / create account + register client).

### Causes (Optional)

### Solutions

Added `shouldStoreSession` function parameter to determine if everything went well with the following request and this session should be stored.

### Testing

Added tests to `LoginUseCaseTest` and `RegisterAccountUseCaseTest`.

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
@@ -24,7 +24,8 @@ interface LoginUseCase {
         userIdentifier: String,
         password: String,
         shouldPersistClient: Boolean,
-        serverConfig: ServerConfig
+        serverConfig: ServerConfig,
+        shouldStoreSession: suspend (AuthSession) -> Boolean = { true }
     ): AuthenticationResult
 }
 
@@ -38,7 +39,8 @@ internal class LoginUseCaseImpl(
         userIdentifier: String,
         password: String,
         shouldPersistClient: Boolean,
-        serverConfig: ServerConfig
+        serverConfig: ServerConfig,
+        shouldStoreSession: suspend (AuthSession) -> Boolean
     ): AuthenticationResult = suspending {
         // remove White Spaces around userIdentifier
         val cleanUserIdentifier = userIdentifier.trim()
@@ -57,8 +59,10 @@ internal class LoginUseCaseImpl(
                 is NetworkFailure.NoNetworkConnection -> AuthenticationResult.Failure.Generic(it)
             }
         }, {
-            sessionRepository.storeSession(it)
-            sessionRepository.updateCurrentSession(it.userId)
+            if(shouldStoreSession(it)) {
+                sessionRepository.storeSession(it)
+                sessionRepository.updateCurrentSession(it.userId)
+            }
             AuthenticationResult.Success(it)
         })
     }
--- a/logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCaseTest.kt
+++ b/logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCaseTest.kt
@@ -98,7 +98,7 @@ class LoginUseCaseTest {
         }
 
     @Test
-    fun `given LoginUseCase is invoked, When ValidateEmailUseCase, loginWithEmail and storeSession return success, then returns success`() =
+    fun givenLoginUseCaseIsInvoked_whenValidateEmailUseCase_loginWithEmailAndStoreSessionReturnSuccess_thenReturnSuccess() =
         runTest {
 
             given(validateEmailUseCase).invocation { invoke(TEST_EMAIL) }.then { true }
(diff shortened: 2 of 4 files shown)


vitorhugods:
Do we have an plan on how to persist the session later?

saleniuk:
> Do we have an plan on how to persist the session later?

I think the same way we do it right now, these changes don't affect the way how the session is managed, it just decides whether it should be initially stored after the login or sign up, because for instance in situation when there is an error during client registration then we also don't want the user to be logged in so we simply shouldn't store the session.

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCaseTest.kt
```diff
@@ -123,7 +123,43 @@ class LoginUseCaseTest {
...
+            verify(sessionRepository)
+                .coroutine { storeSession(TEST_AUTH_SESSION) }
+                .wasNotInvoked()
+            verify(sessionRepository)
+                .coroutine { updateCurrentSession(TEST_AUTH_SESSION.userId) }
+                .wasNotInvoked()
+        }
+
+    @Test
+    fun givenLoginUseCaseIsInvoked_whenValidateEmailUseCaseReturnsFalseAndValidateUserHandleUseCaseLoginWithEmailAndStoreSessionReturnSuccess_thenReturnSuccess() =
```
vitorhugods, line 162: Maybe we don't need to be soooooooooooooo discriptive with the test names?

```suggestion
    fun givenValidateEmailFails_butValidateUserHandleAndEverythingElseSucceeds_whenLoggingIn_thenReturnSuccess() =
```

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCaseTest.kt
```diff
@@ -123,7 +123,43 @@ class LoginUseCaseTest {
...
+                .wasInvoked(exactly = once)
+            verify(validateUserHandleUseCase)
+                .function(validateUserHandleUseCase::invoke).with(any())
+                .wasNotInvoked()
+            verify(loginRepository)
+                .coroutine { loginWithEmail(TEST_EMAIL, TEST_PASSWORD, TEST_PERSIST_CLIENT, TEST_SERVER_CONFIG) }
+                .wasInvoked(exactly = once)
+            verify(loginRepository)
+                .suspendFunction(loginRepository::loginWithHandle).with(any(), any(), any(), any())
+                .wasNotInvoked()
```
vitorhugods, line 152:
```suggestion
```
Considering that this test seems to aim on testing how `shouldStoreSession` affects the UseCase, we can simplify and take some assertions out.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
```diff
@@ -24,7 +24,8 @@ interface LoginUseCase {
         userIdentifier: String,
         password: String,
         shouldPersistClient: Boolean,
-        serverConfig: ServerConfig
+        serverConfig: ServerConfig,
+        shouldStoreSession: suspend (AuthSession) -> Boolean = { true }
```
vitorhugods, line 28: Why a lambda?

Do we have a need to solve this as it goes in Reloaded?
What kind of criteria would one check here?

I wonder if it doesn't lead to some spaghettification. The consumers of this will have to worry about the session persistence multiple times.

```kotlin
val authResult = loginUsingEmail(handle, password) { authSession ->
    authSession.userId == "lorem Ipsum"
}

if(authResult is AuthenticationResult.Success){
    // We don't know what happened until we check again.
    val isSessionPersisted = authResult.userSession.userId == "lorem Ipsum"
}
```

----

### Another Approach

I think that if you need a conditional storing or not of a session, you could just do:


```kotlin
// DO NOT persist by default
val authResult = loginUsingEmail(handle, password, shouldPersistSession = false)

// Only persist after checking
if(authResult is AuthenticationResult.Success){
    val shouldPersistSession = authResult.userSession.userId == "lorem Ipsum"
    if(shouldPersistSession) {
         saveSessionUseCase(authResult.userSession)
    }
}
```

vitorhugods (changes requested):
Great work. I just wonder if a lambda is really useful and wouldn't promote code repetition to whoever is consuming Kalium.

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCaseTest.kt
saleniuk: I hope now the names are short enough 😄

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCaseTest.kt
saleniuk: Sure, I removed unnecessary ones.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
saleniuk: I wanted it to be simple, but actually it may be more difficult to understand and prepared specifically for Reloaded. You made me realize that we already have separate usecases for this. 😅 So I turned it into a regular Boolean.

vitorhugods (approved):
🚀🚀

PR #301, opened 2022-03-18
chore: add pbandk protobuf (AR-1303)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We need a multiplatform protobuf solution that covers all our usecases

### Solutions

Add [PBandK](https://github.com/streem/pbandk).

How it is setup:
Two new modules:
* `protobuf-codegen` which is a "Java Library" that includes the `.proto` files. It will call Google's Gradle Protobuf plugin to generate the code, with the PBandK add-on.
* `protobuf` which is a bit of a hack, as Google's Protobuf plugin doesn't work with Kotlin Multiplatform yet. This module depends on `protobuf-codegen` and will trigger a code generation there, adding the generated files to the `commonMain` SourceSet. This module also contains the PBandK in `commonMain`, which allows all targets to encode and decode to/from byte arrays.

The idea was taken from [a sample in pbandk's repo](https://github.com/streem/pbandk/tree/master/examples/browser-js). As [suggested in this issue/discussion](https://github.com/streem/pbandk/issues/33).

### Testing

Add a simple test to serve as a sample and sanity check.

### Notes

- This current implementation replaces the current Protobuf utilization to encode/decode messages, but doesn't replace the creation of envelopes yet. This should be trivial and tackled in a next PR to make this one smaller.
- The 475+ lines of `.proto` files are copy/pasted from our [generic-message-proto](https://github.com/wireapp/generic-message-proto/tree/master/proto) repo.
- Both `messages.proto` and `otr.proto` contained a `QualifiedUserID`. So I've extracted it to its own dedicated file called `qualified_user_id.proto`.
- Naming is an issue and we probably want to solve this before merging. Especially for `ClientId` and `UserId`, which can be misleading, making discoverability through autocompletion a bit annoying.

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
```diff
@@ -1,9 +1,66 @@
...
+            is GenericMessage.Content.Image -> MessageContent.Unknown
+            is GenericMessage.Content.Knock -> MessageContent.Unknown
+            is GenericMessage.Content.LastRead -> MessageContent.Unknown
+            is GenericMessage.Content.Location -> MessageContent.Unknown
+            is GenericMessage.Content.Reaction -> MessageContent.Unknown
+            null -> {
+                kaliumLogger.w("Null content when parsing protobuf. Message UUID = $genericMessage.")
+                MessageContent.Unknown
+            }
+        }
```
vitorhugods, line 62: Exhaustive `when` for protobufs!

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -148,8 +154,11 @@ object Dependencies {
     }
 
     object Protobuf {
+        @Deprecated("A new implementation is avaliable. Use the protobuf project instead.")
```
typfel, line 157: Why aren't we deleting this? The implementation has been deleted, no?

build.gradle.kts
```diff
@@ -19,6 +19,7 @@ buildscript {
         classpath("app.cash.sqldelight:gradle-plugin:$sqlDelightVersion")
         classpath("com.wire:carthage-gradle-plugin:0.0.1")
         classpath("org.jetbrains.dokka:dokka-gradle-plugin:$dokkaVersion")
+        classpath("com.google.protobuf:protobuf-gradle-plugin:0.8.18")
```
alexandreferris, line 22: variable on plugin version

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -148,8 +154,11 @@ object Dependencies {
...
         const val wireJvmMessageProto = "com.wire:generic-message-proto:${Versions.wireJvmMessageProto}"
+        @Deprecated("A new implementation is avaliable. Use the protobuf project instead.")
```
typfel, line 159: Also delete?

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapperProvider.kt
```diff
@@ -1,3 +1,2 @@
 package com.wire.kalium.logic.data.message
 
-expect fun provideProtoContentMapper(): ProtoContentMapper
```
alexandreferris, line 3: Why do we still have this class/file if it is not being used anymore? Can't we delete it? 🤔

alexandreferris (commented):
Great work!

I just have two comments 🤔

protobuf-codegen/src/main/proto/messages.proto
```diff
@@ -0,0 +1,356 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+syntax = "proto2";
+
+import "qualified_user_id.proto";
+
+message GenericMessage {
```
typfel, line 24: The proto files are copied into the project now? This makes it hard to know which of the protobuf file we are using in Kalium. 

Can we pull them GitHub only define the release tag in the project. If this is not easy we could do it in a separate PR  but I think we should do it.

typfel (changes requested):

buildSrc/src/main/java/Dependencies.kt
vitorhugods: Still used in Network for envelope serialisation.

I want to do that in another PR.

buildSrc/src/main/java/Dependencies.kt
vitorhugods: Same as above. Still used for envelope creation.

protobuf-codegen/src/main/proto/messages.proto
vitorhugods: Not 100% copy pasted. As I mentioned in the PR description, the QualifiedUserID, for example, needed to be extracted to a common file.

Some of the Java options supported by Protoc (like outer class name or adding a suffix) are not respected when using PBandK.

I'd like to streamline the process and checkout from the otr-proto repo, but it's far from easy to do so right now. Maybe it would help by extracting the QualifiedUserID in the original otr-proto repo, but that would be a breaking change.

I don't have any solutions now.

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapperProvider.kt
vitorhugods: Oops hahahah

I'll delete it tomorrow when I'm back.

build.gradle.kts
vitorhugods: It's inside the `buildScript` block, so we can't do it 😔

The `buildScript` block is VERY special.

typfel (approved):

protobuf-codegen/src/main/proto/messages.proto
typfel: The `otr.proto` and `messages.proto` are supposed to live in different namesakes doesn't PBandK respect that? See https://github.com/wireapp/generic-message-proto/pull/74#pullrequestreview-680786841

typfel: So yeah, seems like https://github.com/streem/pbandk/issues/179 doesn't respect `package` but also the `option java_package` is different so I don't understand why the clash.

typfel: Right so it clashes because package name inside the proto files are override by the plugin option:
`option("kotlin_package=com.wire.kalium.protobuf")`

vitorhugods: TBH I felt like PBandK doesn't respect a lot of these options yet. I tried playing with them before submitting the PR and in the end I just removed them, as they didn't have any effect on the generated code.
I didn't want to leave `package` and `option`s in there, so they would suddenly kick-in when we bump PBandK's version or someone spends time trying to understand these lines of code in the future.

It's a trade-off for now. We increase a bit of maintenance with `.proto` definitions, but as a pro: we have a multi-platform solution. In my opinion, it's worth it. But I'd like a second confirmation before we move on.

Hopefuly these will be supported Soon™.

typfel: Yes, it's definitely worth it! Just looking for way to avoid having to copy & paste the files and edit them. But like I said earlier I'm not pushing for solving it in this PR.

vitorhugods: > Right so it clashes because package name inside the proto files are override by the plugin option: `option("kotlin_package=com.wire.kalium.protobuf")`

Indeed, the `kotlin_package` was overwriting it 🤦🏼‍♂️. Removing that made the `package` work.

Still, the original `messages.proto` doesn't have a `package`. So we still need to modify the files a tiny bit when getting from the original repo. But the changes are one-liner.

build.gradle.kts
alexandreferris: ayayay didn't see that, sorry 🙈

alexandreferris (approved):

PR #314, opened 2022-03-22
feat: part-4 refresh session: injecting userId for UserSessionScope and WorkScheduler

MohamadJaara:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

`UserSessionScope` and `WorkScheduler` need `AuthSession` which make it hard to dynamically get the updated session after every refresh 

### Solutions

`SessionManager` + userId

Needs releases with:

- [x] #302 
- [x] #304 
- [x] #306 

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #306, opened 2022-03-21
feat: part-3 refresh session:  user session manager impl

MohamadJaara:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

Implementing user session manager

### Dependencies (Optional)
#302 
#304 

----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #304, opened 2022-03-21
feat: part-2 refresh session: network UserSessionManager

MohamadJaara:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

there is no way to get/update user network configration 

### Solutions
in network `UserSessionManager` interface 
```kotlin 
interface UserSessionManager {
    fun userConfig(): Pair<SessionDTO, BackendConfig>
    fun updateSession(accessToken: AccessTokenDTO, refreshTokenDTO: RefreshTokenDTO?): SessionDTO
}
```
that will be implemented by logic

### Dependencies (Optional)
#302 

----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #315, opened 2022-03-23
feat: AR-1167 handle auth token refresh

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

not refreshing the Tokens when expires

this PR contains: 
 
- [x] #302 
- [x] #304 
- [x] #306 
- [x] #314


typfel (approved):

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
```diff
@@ -3,33 +3,42 @@ package com.wire.kalium.network
...
+/**
+ * This is a hack
+ *
+ * As seen here: https://github.com/ktorio/ktor/issues/1127
+ * Ktor is pretty has `application/protobuf` but not `application/x-protobuf`
+ * and it doesn't support custom content types!
+ */
+private fun Request.handleProtoHeader(): Request =
+    when (headers["Content-Type"] == "application/octet-stream") {
+        true -> newBuilder().header("Content-Type", "application/x-protobuf").build()
```
gongracr, line 29: I might not be reading it correctly, but wouldn't this overwrite any other headers that the original might contain??

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
```diff
@@ -3,33 +3,42 @@ package com.wire.kalium.network
...
+/**
+ * Ktor need "WWW-Authenticate" to be set by BE in-order for the tokens refresh to work
+ * see issue https://youtrack.jetbrains.com/issue/KTOR-2806
+ * BE does not set "WWW-Authenticate"
+ *
+ * checks for 401 response -> add WWW-Authenticate header
+ */
+private fun Response.handleUnauthorizedResponse(): Response =
+    when (this.code == 401) {
+        true -> this.newBuilder().addHeader("WWW-Authenticate", "Bearer").build()
```
gongracr, line 42: Same than above

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
MohamadJaara: Yes, this will overwrite the Content-Type header in case it's octet-stream
the issue is when sending encrypted blob of data ktor will set the Content-Type to octet-stream, but BE needs x-protobuf.

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
MohamadJaara: Ktor need to know the type of auth method for the token refresh to work properly.
BE does not set the WWW-Authenticate, so this interceptor will detect 401 Unauthorized responses and add the needed header, I can do this a bit better and add the header only if it's missing.

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
gongracr: I get that, but what about other headers that the original request has? Wouldn't `newBuilder().header("Content-Type", "application/x-protobuf").build()` just create a new request with just one content type header??

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
gongracr: Maybe we can add some tests that contain other type of headers, and see if they survive and reach the backend after a token refresh interception?

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
MohamadJaara: Maybe the style is a bit confusing since it's an extension function, but the `newBuilder().header(......` is equivalent to `currentRequest.newBuilder().header(.......` 
OkHttp Docs:
>   * Sets the header named [name] to [value]. If this request already has any headers
     * with that name, they are all replaced.
    open fun header(name: String, value: String) = apply {
      headers[name] = value
    }

so no headers lost in the process

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
MohamadJaara: The Devs at square already done that for us 😅
https://github.com/square/okhttp/blob/master/okhttp/src/commonTest/kotlin/okhttp3/HeadersTest.kt

PR #324, opened 2022-03-25
fix: remove content-type hack

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We are sending our messages as protobuf and the server requires the content type to be set to `application/x-protobuf`. Ktor however doesn't respect overriding the content-type on a request and will always send binary data as `application/octet-stream`.

We hacked around this on the Android/JVM by installing a http interceptor which re-writes the content type. This has some drawbacks: 
1. It only fixes the problem on Android,
2. The moment we need to send another binary content type the hack doesn't work anymore.

And that moment is now because I need to send binary data with content type `message/mls` when I send MLS messages.

### Solutions

Install the `ContentNegotiation` plugin which can be used to serialise/deserialise requests based on the content-type. We'll only use it to register our missing binary content-types with a dummy converter which simply passes the binary data through as is.

----

----

----
##### References


vitorhugods (approved):
A lot of people would like this one:
https://github.com/ktorio/ktor/issues/1127

PR #327, opened 2022-03-28
feat: add `SendImageMessageUseCase` [AR-1307] 

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
Sending an Image message successfully required several steps to be considered and implemented, namely:
- Being able to create and keep a symmetric `AES256` secret key with which encrypt the raw asset data. 
- Being able to hash the encrypted data with `SHA256` for retrocompatibility with legacy versions.
- Upload the asset encrypted data and get an allocated `UploadAssetId` that will be used to link the asset data with the asset message.
- Persist message in local DB.
- Create an asset message wrapper and send it.

### Solutions

This first PR will only focus on defining the skeleton of the Use Case, as well as the intermediate `AssetProtoContent` object, needed for the mapping of the persisted `MessageEntity` and the final `Message` object to be sent.

### Testing

WIP to be finished by the following PR

### Attachments (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/AssetProtoContent.kt
```diff
@@ -0,0 +1,99 @@
+package com.wire.kalium.logic.data.message
+
+import com.wire.kalium.logic.data.conversation.LegalHoldStatus
+
+data class AssetProtoContent(
```
typfel, line 5: This is reproducing the whole protobuf definition I think we can simplify it a lot.

1. Ignore the uploaded status and only map assets which contain remote data

3. Ignore the preview

4. Ignore legal hold (for now)

5. Ignore expectsReadConfirmation (for now)

Maybe something like this:

```kotlin
data class Asset(
   val size: Int,
   val name: String? = null,
   val metadata: Metadata,
   val remoteData: RemoteData
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
```diff
@@ -2,8 +2,14 @@ package com.wire.kalium.logic.data.message
...
-    data class Calling(val value: String): MessageContent()
+    data class Text(val value: String) : MessageContent()
 
-    object Unknown: MessageContent()
+    data class Calling(val value: String) : MessageContent()
+
+    sealed class AssetContent : MessageContent() {
+        data class ImageAsset(val value: AssetProtoContent) : AssetContent()
+        data class FileAsset(val value: AssetProtoContent) : AssetContent()
+    }
```
typfel, line 12: What do we gain by this nesting? Can we flatten it to:
```kotlin
data class ImageAsset(val value: AssetProtoContent) : MessageContent()
data class FileAsset(val value: AssetProtoContent) : MessageContent()
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/AssetProtoContent.kt
typfel: Also I don't think we should use `Proto` in our mapped class names

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
gongracr: I thought differentiating between `files` and `images` assets will give us an extra degree of redundancy. However since even the protobuff current definition deprecated the `ImageAsset` object in favor of the base `Asset` one, I think we can also treat both as assets on Kalium level.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/AssetProtoContent.kt
gongracr: I thought about adding the whole protobuf definition and extend it to the SQL model so that we have already all the elements in place. Like that, if we want to expand our list of message types we don't have to add these fields. They are initialised to null in any case, so that shouldn't be a problem right?

typfel: The protobuf definition is over complicated we don't need all the fields. IMO we should add the bare minimum and then expand it as necessary.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetMapper.kt
```diff
@@ -46,4 +50,32 @@ class AssetMapperImpl : AssetMapper {
...
+                caption = null // TODO: add a caption field to the MessageEntity DB table
+            ),
+            preview = null,
+            uploadStatus = null
+        )
+    }
+
+    private fun getAssetMessageMetadata(mimeType: String?): AssetProtoContent.AssetMetadata {
+       return when {
+            mimeType?.contains("image/") ?: false -> Image(width = 0, height = 0, tag = null)
```
yamilmedina, line 72: I was wondering if this could be simplified since we are just returning an `Image` `AssetMetadata` on all cases, or add like a `AssetMetadata.Unknown` type to return in case not mapping could be made on the else block.
 
Still, I think this is WIP, so if you think it isn't necessary to address it now, we can do it on the upcoming PRs if you agree

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/SendImageUseCase.kt
```diff
@@ -0,0 +1,120 @@
...
+import com.wire.kalium.logic.data.user.UserRepository
+import com.wire.kalium.logic.feature.message.MessageSender
+import com.wire.kalium.logic.functional.Either
+import com.wire.kalium.logic.functional.suspending
+import com.wire.kalium.logic.kaliumLogger
+import io.ktor.utils.io.core.toByteArray
+import kotlinx.coroutines.flow.first
+import kotlinx.datetime.Clock
+
+fun interface SendImageUseCase {
```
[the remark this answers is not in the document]
gongracr: Nothing of a real impactful modifier, but it helps to identify that it's an interface meant to have just one function member:
https://kotlinlang.org/docs/fun-interfaces.html

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetMapper.kt
gongracr: Ahh yeah sorry, this will be overriden or removed by the following PR, so don't pay too much attention to it. But thanks for the catch anyways 🦅

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAO.kt
```diff
@@ -3,27 +3,89 @@ package com.wire.kalium.persistence.dao.message
...
 interface MessageDAO {
     suspend fun deleteMessage(id: String, conversationsId: QualifiedIDEntity)
     suspend fun deleteAllMessages()
-    suspend fun insertMessage(message: MessageEntity)
-    suspend fun insertMessages(messages: List<MessageEntity>)
-    suspend fun updateMessage(message: MessageEntity)
-    suspend fun getAllMessages(): Flow<List<MessageEntity>>
-    suspend fun getMessageById(id: String, conversationId: QualifiedIDEntity): Flow<MessageEntity?>
-    suspend fun getMessageByConversation(conversationId: QualifiedIDEntity, limit: Int): Flow<List<MessageEntity>>
+    suspend fun insertMessage(message: BaseMessageEntity)
```
typfel, line 85: How about enforcing correct types by adding an `insert` per type.

```kotlin
suspend fun insertTextMessage(message: TextMessageEntity)
suspend fun insertAssetMessage(message: AssetMessageEntity)
```

typfel: Or maybe we can do a similar thing like you do in the logic data model and use a sealed class.

```kotlin
data class MessageEntity {
   val id: String,
   ...
   ...
   val content: MessageContent
}

sealed class MessageContent {
   data class TextContent {}
   data class AssetContent {}
}

**Review by typfel — CHANGES_REQUESTED** [review]

File: persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAO.kt [hunk 6]
```diff
@@ -3,27 +3,89 @@ package com.wire.kalium.persistence.dao.message
...
+    open val conversationId: QualifiedIDEntity,
+    open val date: String,
+    open val senderUserId: QualifiedIDEntity,
+    open val senderClientId: String,
+    open val status: Status,
+
+    // Text message specific field
+    open val content: String? = null,
+
+    // Asset message specific fields
```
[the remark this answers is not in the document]
gongracr: Chose the sealed class suggestion

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/SendImageUseCase.kt
```diff
@@ -0,0 +1,126 @@
...
+            size = dataSize,
+            name = "",
+            mimeType = ImageAsset.JPEG.name,
+            metadata = AssetContent.AssetMetadata.Image(0, 0),
+            remoteData = AssetContent.RemoteData(
+                otrKey = otrKey.data,
+                sha256 = sha256.toByteArray(),
+                assetId = assetId.key,
+                encryptionAlgorithm = AssetContent.RemoteData.EncryptionAlgorithm.AES_CBC,
+                assetDomain = null,
```
typfel, line 116:
```suggestion
                assetDomain = null, // TODO: fill in the assetDomain, it's returned by the BE when uploading an asset.
```

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
```diff
@@ -32,39 +57,84 @@ class MessageDAOImpl(private val queries: MessagesQueries) : MessageDAO {
...
+                    asset_id = if (message.content is AssetMessageContent) message.content.assetId else null,
+                    asset_token = if (message.content is AssetMessageContent) message.content.assetToken else null,
+                    asset_domain = if (message.content is AssetMessageContent) message.content.assetDomain else null,
+                    asset_encryption_algorithm = if (message.content is AssetMessageContent) message.content.assetEncryptionAlgorithm else null,
+                    conversation_id = message.conversationId,
+                    date = message.date,
+                    sender_user_id = message.senderUserId,
+                    sender_client_id = message.senderClientId,
+                    status = message.status
                 )
```
typfel, line 110: Can we extract this duplicated code into a shared method?

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAO.kt
```diff
@@ -5,16 +5,43 @@ import kotlinx.coroutines.flow.Flow
 
 data class MessageEntity(
     val id: String,
-    val content: String?,
+    val contentType: ContentType,
```
typfel, line 8: Do we need to expose the `contentType`? From what I can see we never read it and always do a `when (content) {}`

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
```diff
@@ -32,39 +57,84 @@ class MessageDAOImpl(private val queries: MessagesQueries) : MessageDAO {
...
-            message.date,
-            message.senderUserId,
-            message.senderClientId,
-            message.status
+            id = message.id,
+            text_body = when (message.content) {
+                is TextMessageContent -> message.content.messageBody
+                else -> null
+            },
+            content_type = message.contentType,
```
typfel, line 65: Can't we do a `when` on `message.content`  to map it to the `ContenType`?

typfel (changes requested):
Looking good! Just some smaller comments. I think we can keep the `contentType` internal to the persistence module.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
gongracr: but it maps exactly the object type on both vars. What would you do on that `when` clause exactly?

typfel: I suggestion we remove `contentType` from the `MessageEntity` and do something like this instead:

```kotlin
content_type = contentTypeOf(message.content)


fun contentTypeOf(content: MessageEntityContent): ContentType {
    when(content) {
        is TexMessageContent -> ContentType.TEXT
        is AssetMessageContent -> ContentType.ASSET
    }
}
```

gongracr: This mapping is already done on the `MessageMapper` class

gongracr: But you mean leaving the `ContentType` enum class within the `MessageEntity` class?

typfel: > This mapping is already done on the `MessageMapper` class

Yes it's done in the mapper in logic right now which is IMO is unnecessary because the `contentType` is only relevant when mapping from `SQLDelightMessage` to `MessageEntity`. I therefore propose to make the `ContentType` internal to the persistence module.

typfel: As it is now we store the same data in two fields: content and contentType, which is bad since we can theoretically construct invalid messages like:

content = MessageContent.Asset
contentType = ContentType.TEXT

typfel: 👍

gongracr: Ok, now I left only the contentType on the SQLDelightMessage level, and persisted it only to map it manually to the correct MessageContent type.

typfel (approved):
🧑‍🍳 💯 

Thanks for being patient with me but I think we have a nice foundation for adding more content types now!

PR #322, opened 2022-03-25
chore: call manager

alexandreferris:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [X] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

[AR-1327 - Request Call Configs when AVS needs it](https://wearezeta.atlassian.net/browse/AR-1327)
[AR-1326 - Receive Calling messages and pass them to AVS on Android](https://wearezeta.atlassian.net/browse/AR-1326)

### Issues

There was no way of creating a CallManager for different sessions.

### Solutions

This PR adds a Global Call Manager to handle the initialisation and the creation of a CallManager for AVS to handle calling.

### Dependencies (Optional)

_If there are some other pull requests related to this one (e.g. new releases of frameworks), specify them here._

Needs releases with:

Other parts will come to the epic

### Testing

Manually tested and tests will be done in another PR

1. branch kalium to this branch in android-reloaded project
2. delete and install the app
3. log in
4. from another account in web stage call the account logged in the app
5. in the logs you should see the callbacks being called with the log prefix: `CallManager - { ... }`

PR #328, opened 2022-03-29
chore: expose list of ongoing calls

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With now having a GlobalCallManager providing each session with its instance, we are now able to expose ongoing calls from AVS to all the way to the UI.

### Dependencies (Optional)

This is Part 2 of the epic, tests will come in a later PR.

Needs releases with:

- [X] #322 

### Attachments (Optional)

https://user-images.githubusercontent.com/5890660/160573622-02a40532-947e-4e6c-9a17-95ccc2eb3bfd.mov

### DIFF
--- a/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt
+++ b/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt
@@ -10,6 +10,7 @@ import com.wire.kalium.logic.data.call.CallRepository
 import com.wire.kalium.logic.data.client.ClientRepository
 import com.wire.kalium.logic.data.conversation.ClientId
 import com.wire.kalium.logic.data.id.asString
+import com.wire.kalium.logic.data.id.toConversationId
 import com.wire.kalium.logic.data.message.Message
 import com.wire.kalium.logic.data.message.MessageContent
 import com.wire.kalium.logic.data.user.UserId
@@ -60,16 +61,16 @@ actual class CallManager(
         deferredHandle = startHandleAsync()
     }
 
-    private fun updateCallStatusById(conversationId: String, status: String) {
+    private fun updateCallStatusById(conversationId: String, status: CallStatus) {
         _calls.update {
             mutableListOf<Call>().apply {
                 addAll(it)
 
-                val callIndex = it.indexOfFirst { call -> call.conversationId == conversationId }
+                val callIndex = it.indexOfFirst { call -> call.conversationId.asString() == conversationId }
                 if (callIndex == -1) {
                     add(
                         Call(
-                            conversationId = conversationId,
+                            conversationId = conversationId.toConversationId(),
                             status = status
                         )
                     )
@@ -103,28 +104,28 @@ actual class CallManager(
                 kaliumLogger.i("startHandleAsync -> incomingCallHandler")
                 updateCallStatusById(
                     conversationId = conversationId,
-                    status = "Incoming"
+                    status = CallStatus.INCOMING
                 )
             },
             missedCallHandler = { conversationId: String, messageTime: Uint32_t, userId: String, isVideoCall: Boolean, arg: Pointer? ->
                 kaliumLogger.i("startHandleAsync -> missedCallHandler")
                 updateCallStatusById(
                     conversationId = conversationId,
-                    status = "Missed"
+                    status = CallStatus.MISSED
                 )
             },
             answeredCallHandler = { conversationId: String, arg: Pointer? ->
                 kaliumLogger.i("startHandleAsync -> answeredCallHandler")
                 updateCallStatusById(
                     conversationId = conversationId,
-                    status = "Answered"
+                    status = CallStatus.ANSWERED
                 )
             },
             establishedCallHandler = { conversationId: String, userId: String, clientId: String, arg: Pointer? ->
                 kaliumLogger.i("startHandleAsync -> establishedCallHandler")
                 updateCallStatusById(
                     conversationId = conversationId,
-                    status = "Established"
+                    status = CallStatus.ESTABLISHED
                 )
             },
             closeCallHandler = { reason: Int, conversationId: String, messageTime: Uint32_t, userId: String, clientId: String,
@@ -132,7 +133,7 @@ actual class CallManager(
                 kaliumLogger.i("startHandleAsync -> closeCallHandler")
                 updateCallStatusById(
                     conversationId = conversationId,
-                    status = "Close"
+                    status = CallStatus.CLOSED
                 )
             },
             metricsHandler = { conversationId: String, metricsJson: String, arg: Pointer? ->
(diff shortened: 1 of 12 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt
```diff
@@ -2,8 +2,17 @@ package com.wire.kalium.logic.feature.call
...
 expect class CallManager {
 
     suspend fun onCallingMessageReceived(message: Message, content: MessageContent.Calling)
+    val allCalls: StateFlow<List<Call>>
+}
+
+val CallManager.ongoingCalls get() = allCalls.map {
+    it.filter { call ->
+        call.status in listOf("Incoming", "Answered", "Established")
+    }
```
alexandreferris, line 17: This is currently not being used, but will be in the near future as we want to properly expose each types of calls, such as:
- Ongoing calls
- Missed / Closed calls
- and so on..

typfel (changes requested):
Looking good! Just some questions.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/Call.kt
```diff
@@ -0,0 +1,6 @@
+package com.wire.kalium.logic.feature.call
+
+data class Call(
+    val conversationId: String,
+    val status: String
```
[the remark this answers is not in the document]
ohassine: And conversationId shouldn't be a String, it should be `ConversationId` type to support federation

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
```diff
@@ -7,6 +7,7 @@ import com.wire.kalium.logic.data.user.UserRepository
 expect class GlobalCallManager {
 
     fun getCallManagerForClient(
+        userId: String,
```
ohassine, line 10: It's a UserId not a String 🙂
As we are passing the userId, do we need the userRepository ?

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt
```diff
@@ -53,6 +60,28 @@ actual class CallManager(
         deferredHandle = startHandleAsync()
     }
 
+    private fun updateCallStatusById(conversationId: String, status: String) {
```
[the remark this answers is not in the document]
alexandreferris: Indeed! Our main focus now is making it work for Android and later on when implementing for other targets we can properly look at all the logic we have and what can and can not be extracted to the common package instead of being duplicated for all targets as you mentioned

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
alexandreferris: It is actually a `NonQualifiedUserId` from `network.api` package, but I have already changed it from String.

We are passing `UserId` as `NonQualifiedUserId` to use it as an initial key for holding the `CallManager` so we don't have it created multiple times for one single user.

The `userRepository` we are using to get the correct `UserId` with `QualifiedID`, thus the necessity to have both as they are exercising different things later on.

TLDR: NonQualifiedUserId for holding CallManager instances, userRepository to get user QualifiedID.

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt
```diff
@@ -72,19 +101,39 @@ actual class CallManager(
             incomingCallHandler = { conversationId: String, messageTime: Uint32_t, userId: String, clientId: String, isVideoCall: Boolean,
                                     shouldRing: Boolean, conversationType: Int, arg: Pointer? ->
                 kaliumLogger.i("startHandleAsync -> incomingCallHandler")
+                updateCallStatusById(
+                    conversationId = conversationId,
```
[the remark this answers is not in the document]
alexandreferris: It could be based on federation as far as I understood, it is now being handled in `String.toConversationId()` where it verifies if it contains a federation domain or not and it maps down to `ConversationId`

logic/src/jvmMain/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
```diff
@@ -3,10 +3,12 @@ package com.wire.kalium.logic.feature.call
 import com.wire.kalium.logic.data.call.CallRepository
 import com.wire.kalium.logic.data.client.ClientRepository
 import com.wire.kalium.logic.data.user.UserRepository
+import com.wire.kalium.network.api.NonQualifiedUserId
 
 actual class GlobalCallManager {
 
     actual fun getCallManagerForClient(
+        userId: NonQualifiedUserId,
```
typfel, line 11: You can make this a `QualifiedID` not that https://github.com/wireapp/kalium/pull/335 is merged.

typfel: Don't have to be in this PR.

alexandreferris: If you don't mind I will be doing this on the EPIC branch as I will need to probably solve the conflicts and thus already changing this to UserId

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/id/QualifiedId.kt
```diff
@@ -1,10 +1,23 @@
...
+
 data class QualifiedID(
     val value: String,
     val domain: String
 )
 
 typealias ConversationId = QualifiedID
 
-fun QualifiedID.asString() = "$value@$domain"
+fun QualifiedID.asString() = "$value$VALUE_DOMAIN_SEPARATOR$domain"
```
ohassine, line 12: We need to check if the domain is empty or not as well.
If we are not in federated environment, the domain will be empty
```
if(domain.isEmpty) return "$value"
else "$value$VALUE_DOMAIN_SEPARATOR$domain"
```

ohassine (changes requested):
lgtm 👌 I just left one small comment

ohassine (approved):

logic/src/jvmMain/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
alexandreferris: changed it in this commit: https://github.com/wireapp/kalium/commit/833d363cef28612e61abf6184248aae6dbc24031

PR #332, opened 2022-03-29
feat: add use case to check current client id

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently there is no way to get the currently registered client id for the user or even to check if it has been already registered. We will need that option to check at the app launch if we need to open the "register client" screen.

### Solutions

Implemented CurrentClientIdUseCase and added to ClientScope.

### Testing

Added CurrentClientIdUseCaseTest.
----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/CurrentClientIdUseCase.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import com.wire.kalium.logic.data.client.ClientRepository
+import com.wire.kalium.logic.data.conversation.ClientId
+import com.wire.kalium.logic.functional.suspending
+
+interface CurrentClientIdUseCase {
+    suspend operator fun invoke(): CurrentClientIdResult
+}
+
+class CurrentClientIdUseCaseImpl(private val clientRepository: ClientRepository) : CurrentClientIdUseCase {
```
typfel, line 11: Do you actually need the clientID? My first though is that the consumer shouldn't deal with that and we can instead have a `NeedsToRegisterClientUseCase` which returns a boolean.

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/CurrentClientIdUseCase.kt
saleniuk: That's a good point, but we already have `SelfClientsUseCase` so the consumer can already get even more than just `ClientId` 🤔 @typfel let me know if I should still change it to `NeedsToRegisterClientUseCase` 😉

typfel: Well depends, do you intend to use the client ID where you use the use case or will you just throw it away and look for the error? If it's the latter I think it cleaner and more readable to have a `NeedsToRegisterClientUseCase`.

saleniuk: You're right, I changed it to `NeedsToRegisterClientUseCase` which returns a boolean.

typfel (approved):

PR #334, opened 2022-03-30
fix: include slow sync when setting the user handle

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The situation with "set user handle" isn't one-dimensional and it depends on the state of slow sync, this situation can lead to race condition, because right after the account creation the data isn't yet synced so we don't have a right db and we can't update the user handle locally (it can end up with an error). On other situations, like where we change the user handle after logging in or at app launch if needed, the sync may be already completed so we should update also the data stored locally. 

### Solutions

In SetUserHandleUseCase we need to check whether the sync is ongoing and if so then wait until it's completed and if it has been completed only then we can update the data stored locally.

### Testing

New tests added to SetUserHandleUseCaseTest.

----

----
##### References


alexandreferris (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/SetUserHandleUseCase.kt
```diff
@@ -19,19 +21,29 @@ sealed class SetUserHandleResult {
...
+            true -> userRepository.updateSelfHandle(handle).coFold(
+                {
+                    if (it is NetworkFailure.ServerMiscommunication && it.kaliumException is KaliumException.InvalidRequestError) {
+                        handleSpecificError(it.kaliumException)
+                    } else {
+                        SetUserHandleResult.Failure.Generic(it)
+                    }
+                }, {
+                    if (syncManager.isSlowSyncOngoing())
+                        syncManager.waitForSlowSyncToComplete()
```
typfel, line 39: Can't this be moved out to the top like we do in other use cases?

typfel (changes requested):

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/user/SetUserHandleUseCaseTest.kt
```diff
@@ -45,6 +49,15 @@ class SetUserHandleUseCaseTest {
         given(userRepository)
             .coroutine { updateSelfHandle(handle) }
             .then { Either.Right(Unit) }
+        given(userRepository)
+            .coroutine { updateLocalSelfUserHandle(handle) }
+            .then { }
```
vitorhugods, line 54: ```suggestion```

You can delete stubs of `Unit` functions if you define the Mock like this:

```kotlin
    @Mock
    private val userRepository = configure(mock(classOf<UserRepository>())) { stubsUnitByDefault = true }
```

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/user/SetUserHandleUseCaseTest.kt
```diff
@@ -56,6 +69,82 @@ class SetUserHandleUseCaseTest {
...
+        given(syncManager)
+            .coroutine { syncManager.isSlowSyncOngoing() }
+            .then { true }
+        given(syncManager)
+            .coroutine { syncManager.isSlowSyncCompleted() }
+            .then { true }
+        given(syncManager)
+            .coroutine { syncManager.waitForSlowSyncToComplete() }
+            .then { }
```
vitorhugods, line 98: Same here:

```suggestion```

You can delete stubs of `Unit` functions if you define the Mock like this:

```kotlin
    @Mock
    private val syncManager = configure(mock(classOf<SyncManager>())) { stubsUnitByDefault = true }
```

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/user/SetUserHandleUseCaseTest.kt
```diff
@@ -56,6 +69,82 @@ class SetUserHandleUseCaseTest {
...
+        given(validateHandleUseCase)
+            .function(validateHandleUseCase::invoke)
+            .whenInvokedWith(any())
+            .then { true }
+        given(userRepository)
+            .coroutine { updateSelfHandle(handle) }
+            .then { Either.Right(Unit) }
+        given(userRepository)
+            .coroutine { updateLocalSelfUserHandle(handle) }
+            .then { }
```
vitorhugods, line 89: ```suggestion```

vitorhugods (changes requested):
Other than what @typfel commented. Tiny changes to clean-up some test code.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/SetUserHandleUseCase.kt
saleniuk: Ok so I moved it to the top, but I think `if (syncManager.isSlowSyncOngoing())` is worth keeping in this case because when we set the user handle right after the account creation, we don't have to start the sync and wait until it finishes, it can be done after the user handle is set. It's mainly for safety in case the user is already logged in and has not yet set the user handle so in the meantime the sync may have already been started.

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/user/SetUserHandleUseCaseTest.kt
saleniuk: Nice one, thanks! I changed it 🚀

typfel (approved):

vitorhugods (approved):

PR #341, opened 2022-03-31
feat: update ValidateUserHandleUseCase to return also handleWithoutInvalidChars

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Right now ValidateUserHandleUseCase returns boolean which determines whether the specific handle is valid, but in order to execute the "shake" animation when the user enters invalid character and not allow the invalid ones to be added to the text field, we need to also get the handle string with invalid chars filtered out.

### Solutions

Update ValidateUserHandleUseCase to return ValidateUserHandleResult which contains both boolean and string with invalid chars filtered out.

### Testing

Added tests to ValidateUserHandleUseCaseTest and updated others that use this usecase.

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/ValidateUserHandleUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/ValidateUserHandleUseCase.kt
@@ -1,24 +1,21 @@
 package com.wire.kalium.logic.feature.auth
 
 interface ValidateUserHandleUseCase {
-    operator fun invoke(handle: String): Boolean
+    operator fun invoke(handle: String): ValidateUserHandleResult
 }
 class ValidateUserHandleUseCaseImpl: ValidateUserHandleUseCase  {
-    override operator fun invoke(handle: String): Boolean {
-        return when {
-            isHandleTooShort(handle) -> false
-            !validateHandle(handle) -> false
-            else -> true
-        }
+    override operator fun invoke(handle: String): ValidateUserHandleResult {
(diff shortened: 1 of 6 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/ValidateUserHandleUseCase.kt
```diff
@@ -1,24 +1,21 @@
...
-        return when {
-            isHandleTooShort(handle) -> false
-            !validateHandle(handle) -> false
-            else -> true
-        }
+    override operator fun invoke(handle: String): ValidateUserHandleResult {
+        val handleWithoutInvalidChars = handle.replace(Regex(HANDLE_FORBIDDEN_CHARACTERS_REGEX), "")
+        val hasValidChars = handle == handleWithoutInvalidChars
+        val hasValidLength = handleWithoutInvalidChars.length in HANDLE_MIN_LENGTH..HANDLE_MAX_LENGTH
+        return ValidateUserHandleResult(hasValidChars && hasValidLength, handleWithoutInvalidChars)
```
MohamadJaara, line 11: If we do the check on every char entered by the user, `hasValidChars && hasValidLength` will always be false  for the first char, and it will trigger the shake animation, is this the intended behavior?

saleniuk: For the input I just wanted to compare the text that the user entered with `handleWithoutInvalidChars` and if both aren't the same then it means that the "shake" animation should be started, but maybe you're right and that's not clear enough, so I changed the `ValidateUserHandleResult` to be sealed class and thus be more versatile.

MohamadJaara (approved):

PR #342, opened 2022-03-31
chore: create a ConversationDetails model (AR-1177, AR-881)

vitorhugods:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [X] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

Currently there's no way to fetch details of a conversation without fetching Conversations and then fetching members, and figuring it out yourself.

### Solutions

This PR aims to add a model that will represent results for a "Give me a conversation, but please fill it in with some details, independently of the type it is" use case.

I've also renamed `PublicUser` to `OtherUser`.
The existence of a `PublicUser` implied that there's a `PrivateUser` or some other of "visibility" for users. Which even if it's true (for example custom backends could be seen as "not public"), isn't exactly what this represents.
The now-renamed `PublicUser` class represents any other user that isn't `SelfUser`, so `OtherUser` is a name that lacks creativity, but tells the thruth.

### Testing

No "logic" is added in this PR. Only a new data class

### Notes

This PR is the first of a few. Note that I'm not requesting to merge to `origin/develop`. But `origin/conversation-details/epic` instead.

This branch originally had some other changes that were merged in #280.

----
#### PR Post Merge Checklist for internal contributors

 - [X] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #343, opened 2022-03-31
feat: allow observation of conversation details (AR-1177, AR-881)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We need to expose 1:1 conversation details and other richer details to consumers of Kalium.

### Solutions

We can return `ConversationDetails` (introduced in #342). When the conversation is a 1:1, we can fetch the details of the "other" conversation member and pass it forward.

### Dependencies

Needs releases with:

- [ ] #342 

### Testing

To be done in the next PR.

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -66,7 +70,43 @@ class ConversationDataSource(
...
+        when (conversation.type) {
+            Conversation.Type.SELF -> flowOf(ConversationDetails.Self(conversation))
+            Conversation.Type.GROUP -> flowOf(ConversationDetails.Group(conversation))
+            Conversation.Type.ONE_ON_ONE -> {
+                suspending {
+                    val selfUserId = userRepository.getSelfUser().map { it.id }.first()
+                    getConversationMembers(conversation.id).map { members ->
+                        members.first { itemId -> itemId != selfUserId }
+                    }.coFold({
+                        // TODO: How to Handle failure when dealing with flows?
```
typfel, line 92: So from what I've read we throw exceptions and they can then we caught by the consumer and show an error message.

But in this case it would only throw this error if we can't read from the database or the member data is corrupted for some reason. So perhaps is fine to throw an `IOException` let the application crash if it happens?

typfel (approved):

PR #359, opened 2022-04-05
feat: answer incoming call

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We were able to receive calls, but we couldn't answer them 😞  BUT NOW WE CAN 🎉 

### Dependencies (Optional)

This is the first PR, another PR with the tests will follow.

Needs releases with:

- [ ]

### Testing
Tests will be done in the follow up PR.


logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt
```diff
@@ -184,9 +194,19 @@ actual class CallManager(
...
-            kaliumLogger.d("onCallingMessageReceived -> Passed through")
+            kaliumLogger.d("$TAG - onCallingMessageReceived")
         }
 
+    actual suspend fun answerCall(conversationId: ConversationId) = withCalling {
+        kaliumLogger.d("$TAG -> answerCall")
+        calling.wcall_answer(
+            inst = deferredHandle.await(),
+            conversationId = conversationId.asString(),
+            callType = CallType.CALL_TYPE_NORMAL,
```
vitorhugods, line 205: What does `calLType` do here?

Could we document this?

vitorhugods (approved):

calling/src/androidMain/kotlin/com/wire/kalium/calling/Calling.kt
```diff
@@ -22,6 +22,17 @@ import com.wire.kalium.calling.types.Uint32_t
 // A magic number used to initialize AVS (required for all mobile platforms).
 const val ENVIRONMENT_DEFAULT = 0
 
+/**
+ * WCALL_CALL_TYPE_NORMAL          0
+ * WCALL_CALL_TYPE_VIDEO           1
+ * WCALL_CALL_TYPE_FORCED_AUDIO    2
+ */
+object CallType {
```
vitorhugods, line 30:
```suggestion
internal object CallType {
```

If this is only used in `calling`, we can hide it.

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt
alexandreferris: As talked with Oussama, I have only two explanations for `CallType`:
```
CALL_TYPE_NORMAL => Only Audio
CALL_TYPE_VIDEO => Video/Audio
CALL_TYPE_FORCED_AUDIO = No explanation yet.
```

Those will be added in `Calling.kt` on top of `object CallType` as a doc.

calling/src/androidMain/kotlin/com/wire/kalium/calling/Calling.kt
alexandreferris: It is not actually, it is also being used in `CallManger` in the android package, where it accepts an incoming call with its type, can [see here](https://github.com/wireapp/kalium/blob/d55c4ddb3db8898ab3ff21329f4cb09a25261496/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt) on line 205.

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt
typfel: `CALL_TYPE_FORCED_AUDIO` was used for legacy groups calls (before conference calls). If the number of participants were above a certain limit you weren't allowed join with video.

calling/src/androidMain/kotlin/com/wire/kalium/calling/Calling.kt
```diff
@@ -22,6 +22,17 @@ import com.wire.kalium.calling.types.Uint32_t
 // A magic number used to initialize AVS (required for all mobile platforms).
 const val ENVIRONMENT_DEFAULT = 0
 
+/**
+ * CALL_TYPE_NORMAL = 0 => Only Audio
+ * CALL_TYPE_VIDEO = 1 => Video/Audio
+ * CALL_TYPE_FORCED_AUDIO = 2 => No explanation yet.
+ */
+object CallType {
```
typfel, line 30: Can't this be an enum?

typfel (changes requested):

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManager.kt
vitorhugods: So I think these kind of "deprecated" things can have their definitions inside platform specific-code if necessary.

But I think we can try our best to avoid making them visible to consumers of `logic/commonMain`.
That is: If it isn't supported, they _should_ not be visible.

If I understood correctly, this code here follows this principle, as `CallType` is defined in `:calling`.

typfel (approved):

PR #363, opened 2022-04-05
fix: unresolved references when importing classes generated from proto

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There was a "unresolved references" issue with classes generated from proto. It wasn't breaking the build but making it difficult to use them in code.

image

### Causes (Optional)

It probably was because the same path that was supposed to be the source for `protobuf` module was also the "generated source root" for `protobuf-codegen` and probably Android Studio doesn't like that.

image

### Solutions

Fortunately there is a parameter `generatedFilesBaseDir` that can be used to determine the directory where the generated files are placed. It had to be outside of the `build` directory, because it looks like the pbandk searches for such directory in `build` and marks it as a "generated source root" for the module so separate `generated` folder is created, which is ignored by git, and access to the generated classes is easier.

### Testing

Tested manually.

#### How to Test

Check if the generated classes are accessible in the `kalium` and there are no "unresolved references".

----

----
##### References

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,6 @@ libs/
 
 # CLI working diretory
 .proteus
+
+# kotlin files generated from proto
+generated/
--- a/protobuf-codegen/build.gradle.kts
+++ b/protobuf-codegen/build.gradle.kts
@@ -16,6 +16,7 @@ group = "com.wire.kalium"
 version = "0.0.1-SNAPSHOT"
 
 protobuf {
+    generatedFilesBaseDir = "$projectDir/generated"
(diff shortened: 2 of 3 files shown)


gongracr (approved):
Such a simple yet powerful way of fixing this annoying issue 🎩 👏  well cooked 👨‍🍳

vitorhugods (approved):
I think I thought of 3 or 4 different solutions to this issue, and they either didn't work or were more more complex than this 🤣 

Nicely done!

PR #366, opened 2022-04-06
feat: add use case for creating groups

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We want to able to create group conversations and more specifically I want to create MLS group conversations.

#### persistance

Add `groupId`, `groupState` and `protocol` to the `ConversationEntity`.

groupId: MLS specific identifier for MLS groups provided by the BE.
groupState: Since MLS groups are not created atomically we need to keep a local state which tracks if they "established" r not.
protocol: If the conversation is intended to use the MLS or the Proteus protocol.

#### logic

Add a `createGroupConversation()` method on the `ConversationRepository` which creates a new group conversation on the BE and persists new conversation. The methods takes an option argument for configuring the group and specifically it allows you specify if the conversation should use Proteus or MLS as it's protocol.

Add `MLSConversationRepository` which establishes a MLS group by creating a local group with the MLSClient and broadcasting the welcome message to the members.

Add `CreateGroupConversationUseCase`

#### Test Coverage (Optional)

#### How to Test

Can be tested via CLI but It needs persistence.

### Attachments (Optional)

----

----

----
##### References


persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAOImpl.kt
```diff
@@ -66,6 +79,10 @@ class ConversationDAOImpl(
         )
     }
 
+    override suspend fun updateConversationGroupState(groupState: ConversationEntity.GroupState, groupId: String) {
```
mchenani, line 82: is it safe and enough that we update the groupState only based on the groupId? or is it better to add the conversationId as well?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -141,6 +165,57 @@ class ConversationDataSource(
...
+        wrapApiRequest {
+            conversationApi.createNewConversation(
+                CreateConversationRequest(
+                    if (options.protocol == ConverationOptions.Protocol.PROTEUS) members.map { idMapper.toApiModel(it.id) } else emptyList(),
+                    name,
+                    options.access.toList().map(conversationMapper::toApiModel),
+                    options.accessRole.toList().map(conversationMapper::toApiModel),
+                    userRepository.getSelfUser().firstOrNull()?.team?.let { ConvTeamInfo(false, it) },
+                    null,
+                    if (options.readReceiptsEnabled) 1 else 0,
```
mchenani, line 178: I think if we can replace this part with enum would be great! or at least have some doc comments to make sure we all know the meaning of these numbers.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -141,6 +165,57 @@ class ConversationDataSource(
     override suspend fun deleteMember(conversationID: QualifiedIDEntity, userID: QualifiedIDEntity): Either<CoreFailure, Unit> =
         wrapStorageRequest { conversationDAO.deleteMemberByQualifiedID(conversationID, userID) }
 
+    override suspend fun createGroupConversation(name: String, members: List<Member>, options: ConverationOptions): Either<CoreFailure, Conversation> = suspending {
+        wrapApiRequest {
+            conversationApi.createNewConversation(
+                CreateConversationRequest(
+                    if (options.protocol == ConverationOptions.Protocol.PROTEUS) members.map { idMapper.toApiModel(it.id) } else emptyList(),
```
mchenani, line 172: I think since we have some values that are not clear to the assigned parameter, it would be better to use the named arguments to make the code more readable.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -141,6 +165,57 @@ class ConversationDataSource(
...
+                    name,
+                    options.access.toList().map(conversationMapper::toApiModel),
+                    options.accessRole.toList().map(conversationMapper::toApiModel),
+                    userRepository.getSelfUser().firstOrNull()?.team?.let { ConvTeamInfo(false, it) },
+                    null,
```
mchenani, line 177: I think since we have some values that are not clear to the assigned parameter, it would be better to use the named arguments to make the code more readable.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -141,6 +165,57 @@ class ConversationDataSource(
...
+                }.flatMap {
+                    wrapStorageRequest {
+                        if (options.protocol == ConverationOptions.Protocol.PROTEUS) {
+                            conversationDAO.insertMembers(memberMapper.fromApiModelToDaoModel(conversationResponse.members), conversationEntity.id)
+                        } else {
+                            val selfUserId = userRepository.getSelfUserId()
+                            val selfMember = Member(selfUserId)
+                            conversationDAO.insertMembers((members + selfMember).map(memberMapper::toDaoModel), conversationEntity.id)
+                            Either.Right(Unit)
+                        }
```
mchenani, line 204: Is not setting the Either result in the first part of the if is intentionally?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -141,6 +165,57 @@ class ConversationDataSource(
     override suspend fun deleteMember(conversationID: QualifiedIDEntity, userID: QualifiedIDEntity): Either<CoreFailure, Unit> =
         wrapStorageRequest { conversationDAO.deleteMemberByQualifiedID(conversationID, userID) }
 
+    override suspend fun createGroupConversation(name: String, members: List<Member>, options: ConverationOptions): Either<CoreFailure, Conversation> = suspending {
```
trOnk12, line 168: that is one of the hell long method :P could we perhaps split this into smaller methods ? or perhaps there is some code that could be extracted to a usecase ?

typfel: 👍  I'll try to split out some parts to private methods.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
typfel: The backend decided to have an `Int` for this boolean property 🤷‍♂️ I added a comment for this property in the network module.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAOImpl.kt
typfel: It's safe since we only need deal with the `groupState` for MLS conversations which always have a groupId. But it doesn't really matter I can switch to conversationId as well.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -4,27 +4,50 @@ import com.wire.kalium.logic.CoreFailure
...
+
+    enum class AccessRole {
+        TEAM_MEMBER, NON_TEAM_MEMBER, GUEST, SERVICE
+    }
+
+    enum class Access {
+        PRIVATE, INVITE, LINK, CODE
+    }
+}
+
```
vitorhugods, line 50: Maybe extract these to their own file?

`ConversationRepository` is growing.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
```diff
@@ -8,7 +8,7 @@ import com.wire.kalium.logic.data.user.SelfUser
 import com.wire.kalium.logic.data.user.User
 import com.wire.kalium.logic.data.user.UserId
 
-data class Conversation(val id: ConversationId, val name: String?, val type: Type, val teamId: TeamId?) {
+data class Conversation(val id: ConversationId, val name: String?, val type: Type, val teamId: TeamId?, val groupId: String?) {
```
vitorhugods, line 11: The name `groupId` makes me think of `Conversation.Type.GROUP`.

Is it specific to MLS? Could we name it that way?

Will `Conversation` have a `val protocol: ConvProtocol` too?

What about something like:

```kotlin

data class Conversation(
    val id: ConversationId,
    val name: String?,
    val type: Type,
    val teamId: TeamId?,
    val encryptionInfo: ProtocolInfo
)

sealed class ProtocolInfo{ 
    object Proteus: ProtocolInfo()
    data class MLS(val groupId: String, val groupState: GroupState): ProtocolInfo()
}
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
```diff
@@ -0,0 +1,78 @@
...
+                }
+            }
+        }
+    }
+
+    private suspend fun getConversationMembers(conversationId: ConversationId): Either<StorageFailure, List<UserId>> = wrapStorageRequest {
+        conversationDAO.getAllMembers(idMapper.toDaoModel(conversationId)).first().map { idMapper.fromDaoModel(it.user) }
+    }
+
+}
```
trOnk12, line 78: not an kalium expert but I kind of feel that we got some buisness logic here that could be extracted to a use case, that way we avoid reimplementing things like getConversationMembers, that piece of code could live inside ConversationRepository for example,

Just a suggestion as I am getting more familiar with Kalium

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -141,6 +166,60 @@ class ConversationDataSource(
...
+                        persistMembersFromConversationResponseMLS(conversationResponse, members)
+                    }
+                }.flatMap {
+                    if (options.protocol == ConverationOptions.Protocol.PROTEUS) {
+                        Either.Right(conversation)
+                    } else {
+                        mlsConversationRepository.establishMLSGroup(conversation).flatMap {
+                            Either.Right(conversation)
+                        }
+                    }
```
trOnk12, line 197: perhaps this establishMLSGroup thingy could be moved to a use case ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
typfel: I can't move this code to the `ConversationRepository` and call it from here since then I would create a circular dependency. IMO I would leave like this for now and refactor the conversation repositories when we actually have some more substantial code duplication.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
typfel: `establishMLSGroup()` is an internal thing we need to do before we can send messages in a MLS conversation, but the user of Kalium shouldn't have to care about that so shouldn't expose it as a use case.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
trOnk12: I was actually thinking of combing user repo and this repo inside a use case when we move this code but your call on this one :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
trOnk12: I do not  mean expose it as a use case but handel it inside the body of the usecase for example mlving it to mlsrepo and use it inside repo

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
typfel: I did this at the persistence level and choose to not expose any of this at the logic level for now.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
trOnk12: Just out of curiousity but where do you see that risk?

typfel: In that if I split creating a group into two steps which are connected in the CreateGroupUseCase, there's a risk that the next time someone needs to create a group in another use case he/she might forget to do the second step.

trOnk12: He could reuse that use case if he needs it in another use case :p i think I am used to putting more stuff into the use cases than repos :p anyway i got your point if you think this stuff belongs to repo ill Mark it resolved

vitorhugods (approved):
🚀 Fly me to the mooooon 🌕 🎶

trOnk12 (approved):

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAOImpl.kt
mchenani: That would be enough if we are assured of it.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
mchenani: But we can have an enum in our side that have the int values inside of it.

typfel: Ok, done

mchenani (approved):
✅ 🚀

PR #369, opened 2022-04-06
fix: custom BE url parsing

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

connect the app  to a custom BE ⇾ network is kaput

### Causes (Optional)

the string for production/staging BE url is stored as host only without the protocol, E.g. `apiBaseUrl = """prod-nginz-https.wire.com"""`, but custom BE has the protocol included `https://KnusprigerkeksDerVerdammnis.de` and when connecting to BE we were considering that string as the host without doing any URL parsing 
```kotlin
httpClient.post(PATH) {
                host = apiBaseUrl // here we take the stored BE url and slap it as the host
                url.protocol = URLProtocol.HTTPS
            }
``` 
as a result, when `apiBaseUrl=https://KnusprigerkeksDerVerdammnis.de` the result target of that request will be `https://https://KnusprigerkeksDerVerdammnis.de` ⇾ unknown host exception aka. kaput

### Solutions

parsing the URLs for network 
```kotlin
data class ServerConfigDTO(
    val apiBaseUrl: Url,
    val accountsBaseUrl: Url,
    val webSocketBaseUrl: Url,
    val blackListUrl: Url,
    val teamsUrl: Url,
    val websiteUrl: Url,
    val title: String
)
```
so, we can use it as URL and not a plain string 

```kotlin
            httpClient.post(PATH) {
                host = Url(apiBaseUrl).host // nice looking url host
                url.protocol = URLProtocol.HTTPS
            }
```

### Dependencies (Optional)

Needs releases with:

### Testing

Connected to custom BE and it works now
Any suggestion for tests that can prevent this from happening in the future are welcome.

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/ServerConfig.kt
```diff
@@ -47,10 +48,10 @@ interface ServerConfigMapper {
 class ServerConfigMapperImpl : ServerConfigMapper {
     // TODO: url validation check e.g. remove https:// since ktor will control the http protocol
     override fun toDTO(serverConfig: ServerConfig): ServerConfigDTO =
-        with(serverConfig) { ServerConfigDTO(apiBaseUrl, accountsBaseUrl, webSocketBaseUrl, blackListUrl, teamsUrl, websiteUrl, title) }
+        with(serverConfig) { ServerConfigDTO(Url(apiBaseUrl), Url(accountsBaseUrl), Url(webSocketBaseUrl), Url(blackListUrl), Url(teamsUrl), Url(websiteUrl), title) }
 
     override fun fromDTO(serverConfigDTO: ServerConfigDTO): ServerConfig =
-        with(serverConfigDTO) { ServerConfig(apiBaseUrl, accountsBaseUrl, webSocketBaseUrl, blackListUrl, teamsUrl, websiteUrl, title) }
+        with(serverConfigDTO) { ServerConfig(apiBaseUrl.toString(), accountsBaseUrl.toString(), webSocketBaseUrl.toString(), blackListUrl.toString(), teamsUrl.toString(), websiteUrl.toString(), title) }
```
mchenani, line 54: I see many url to String and String to Url casting, do you think we can keep only the url and get rid of all the toString stuff?

MohamadJaara: the `Url` class is ktor class and using it in each module means we need to include ktor in all the modules and make it something we depends on outside network
Kotlin does have a [URL](https://kotlinlang.org/api/latest/jvm/stdlib/org.w3c.dom.url/-u-r-l/) that target only Js

mchenani: Ok! Thanks for the clarification.

PR #374, opened 2022-04-07
feat: added notification flag in message db

borichellow:
# What's new in this PR?

- added `shouldBeNotified` flag into the Message entity. 
- added queries and UseCase for selecting Messages that the user should be notified about 
- added queries and UseCase for marking Messages that the user was notified about it
- added `DbNotificationConversation` which is base data class for the conversation Notifications and could be re-used for other platforms

#### what is `shouldBeNotified` for: 

Basically it's a flag that determines if user should be notified about that message or not. 
The problem in Android is that there is no option to add messages to existing notification, only to replace it with a new one. 
In designs: each conversation is 1 notification, that contains all the new messages from that conversation. 
So for the case when user has notification about some conversation and a new message comes there, we need to know all the previously displayed messages to upend them by the new one.
This is what this `shouldBeNotified` is for, to know which messages should be added into notification.

#### How to Test

I've made changes in android project to test if all the data comes correctly and all the queries are done correctly. 
It will be available int the PR in android project


typfel:
Can you give some context how `shouldBeNotified` is supposed to be used?

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/BooleanAdapter.kt
```diff
@@ -0,0 +1,15 @@
+package com.wire.kalium.persistence.dao
+
+import app.cash.sqldelight.ColumnAdapter
+
+class BooleanAdapter : ColumnAdapter<Boolean, Long> {
```
borichellow, line 5: this is not used, remove it

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq
```diff
@@ -49,6 +51,21 @@ UPDATE Message
 SET visibility = ?, text_body = ?
 WHERE id = ? AND conversation_id = ?;
 
+updateNotificationFlagByConversation:
+UPDATE Message
+SET shouldNotify = 0
+WHERE conversation_id = ? AND shouldNotify = 1;
+
+updateNotificationFlagForAll:
```
borichellow, line 59: `updateNotificationFlagForAll` and `updateNotificationFlagByConversation` are not used, cause it may be "too big" operations.
remove it

typfel:
Ok, from your description I read it like `shouldBeNotified` used as a read marker for messages so you can only include new unread messages in the notifications.

Assuming this is correct I think we need to align a bit on how we read messages in Wire and how that read state is synchronised among your clients.

Generally we don't mark individual messages as read instead we maintain a "last read" timestamp for each conversation which is updated when the user reads messages. A message is considered unread if its timestamp is newer than the "last read" timestamp of the conversation.

The last read state is synchronised among all your devices by sending a `LastRead` protobuf message all your other devices when you read a message.

But I think easier if sync 1:1 on this.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
```diff
@@ -0,0 +1,92 @@
...
+import com.wire.kalium.logic.data.notification.DbNotificationMessageAuthor
+import com.wire.kalium.logic.data.publicuser.model.OtherUser
+import com.wire.kalium.logic.data.user.UserRepository
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flatMapMerge
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.take
+
+class GetNotificationsUseCase(
```
yamilmedina, line 19: I think we could extract this to an interface and mark this implementation as internal, so we can mock it and test it later on android-reloaded

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/DbNotification.kt
```diff
@@ -0,0 +1,25 @@
+package com.wire.kalium.logic.data.notification
+
+import com.wire.kalium.logic.data.id.ConversationId
+
+data class DbNotificationConversation(
```
yamilmedina, line 5: I was wondering if these `Db` prefixes felt right, or maybe would create some kind of confusion, but I don't have an answer either 🤔  😅 

I was thinking in something like `Local` instead of `Db`, let me know if it makes sense to you =)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MarkMessagesAsNotifiedUseCase.kt
```diff
@@ -0,0 +1,14 @@
+package com.wire.kalium.logic.feature.message
+
+import com.wire.kalium.logic.CoreFailure
+import com.wire.kalium.logic.data.id.ConversationId
+import com.wire.kalium.logic.data.message.MessageRepository
+import com.wire.kalium.logic.functional.Either
+
+class MarkMessagesAsNotifiedUseCase(private val messageRepository: MessageRepository) {
```
yamilmedina, line 8: I think here we could also extract an interface and have the implementation internal, so we can benefit from scoping the implementation and later mock the interface on reloaded for tests :D

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MarkMessagesAsNotifiedUseCase.kt
```diff
@@ -0,0 +1,14 @@
...
+
+    suspend operator fun invoke(conversationId: ConversationId? = null): Either<CoreFailure, Unit> {
```
yamilmedina, line 10: Maybe we could return a Result type here that we can define here at this class level. If I'm not wrong, we are trying to move away from exposing `Either` to reloaded, also because generics could cause problems in other platforms like: iOS, web

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
```diff
@@ -0,0 +1,92 @@
...
+                            )
+                        }
+                }
+            }
+    }
+
+    private data class MessageWithSenderData(val message: Message, val user: OtherUser?)
+}
+
+private suspend fun <A, B> flatMapFromIterable(
```
yamilmedina, line 77: What do you think of moving this cool generic mapping function to a utility class, so we can reuse it later? Something like we have with `Either` or `SuspendableEitherScope`  for example with the `coFold` functions on functional package :D

yamilmedina (commented):
Great work @borichellow  ! 👏   I added some comments to the code that should not be affected by the discussion that we had earlier today (`shouldBeNotified` field)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
borichellow: i was going to do that, but couldn't find a propper place, thanks :)

yamilmedina (commented):
Great work! Looks superb, the only thing that it's missing before we can merge, it's fixing the broken tests.

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Converations.sq
```diff
@@ -8,7 +9,9 @@ CREATE TABLE Conversation (
     team_id TEXT,
     mls_group_id TEXT,
     mls_group_state TEXT AS ConversationEntity.GroupState NOT NULL,
-    protocol TEXT AS ConversationEntity.Protocol NOT NULL
+    protocol TEXT AS ConversationEntity.Protocol NOT NULL,
+    has_unnotified_messages INTEGER AS Boolean DEFAULT 0,
```
typfel, line 13: I would consider replacing `has_unnotified_messages` with a `last_modified_date`. This you would also update this date whenever we persist a new message but it's more generic (can be used to sort the conversation list). You could then retrieve the conversations with un-notified message by `last_modified_date  > last_notification_date`. Another advantage of this approach is that you don't need "reset" the `last_modified_date` like you do for the `has_unnotified_messages`.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -54,6 +55,10 @@ interface ConversationRepository {
     suspend fun deleteMember(conversationID: QualifiedIDEntity, userID: QualifiedIDEntity): Either<CoreFailure, Unit>
     suspend fun getOneToOneConversationDetailsByUserId(otherUserId: UserId): Either<CoreFailure, ConversationDetails.OneOne?>
     suspend fun createGroupConversation(name: String? = null, members: List<Member>, options: ConverationOptions = ConverationOptions()): Either<CoreFailure, Conversation>
+    suspend fun getConversationsForNotifications(): Flow<List<Conversation>>
+    suspend fun setConversationAsNonNotified(qualifiedID: QualifiedID): Either<StorageFailure, Unit>
+    suspend fun setConversationAsNotified(qualifiedID: QualifiedID, date: String): Either<StorageFailure, Unit>
```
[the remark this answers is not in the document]
typfel: Actually if you introduce the `lastModifiedDate` you don't need pass the date at all, because then you can do: `last_notified_date = last_modified_date`

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAO.kt
```diff
@@ -11,7 +11,8 @@ data class MessageEntity(
     val senderUserId: QualifiedIDEntity,
     val senderClientId: String,
     val status: Status,
-    val visibility: Visibility = Visibility.VISIBLE
+    val visibility: Visibility = Visibility.VISIBLE,
+    val shouldNotify: Boolean = true
```
typfel, line 15: Does this work as expected? Can't we remove this and compute it on fly based on if the sender the self user or not?

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq
```diff
@@ -85,10 +86,13 @@ SET status = ?
...
+SELECT * FROM Message ORDER BY date DESC LIMIT ? OFFSET ?;
 
 selectById:
 SELECT * FROM Message WHERE id = ? AND conversation_id = ?;
 
 selectByConversationId:
-SELECT * FROM Message WHERE conversation_id = ? ORDER BY date DESC LIMIT ?;
+SELECT * FROM Message WHERE conversation_id = ? ORDER BY date DESC LIMIT ? OFFSET ?;
+
+selectByConversationIdAndDate:
```
typfel, line 97:
```suggestion
selectMessagesByConversationIdAfterDate:
```

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq
```diff
@@ -85,10 +86,13 @@ SET status = ?
 WHERE id = ? AND conversation_id = ?;
 
 selectAllMessages:
-SELECT * FROM Message;
+SELECT * FROM Message ORDER BY date DESC LIMIT ? OFFSET ?;
 
 selectById:
 SELECT * FROM Message WHERE id = ? AND conversation_id = ?;
 
 selectByConversationId:
```
typfel, line 94:
```suggestion
selectMessagesByConversationId:
```

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Converations.sq
```diff
@@ -38,3 +59,6 @@ SELECT * FROM Conversation WHERE qualified_id = ?;
 
 selectByGroupId:
 SELECT * FROM Conversation WHERE mls_group_id = ?;
+
+selectConversationsWithNotification:
```
typfel, line 63:
```suggestion
selectConversationsWithUnnotifiedMessages:
```

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
```diff
@@ -141,8 +142,14 @@ class MessageDAOImpl(private val queries: MessagesQueries) : MessageDAO {
             .mapToOneOrNull()
             .map { msg -> msg?.let(mapper::toModel) }
 
-    override suspend fun getMessageByConversation(conversationId: QualifiedIDEntity, limit: Int): Flow<List<MessageEntity>> =
-        queries.selectByConversationId(conversationId, limit.toLong())
+    override suspend fun getMessageByConversation(conversationId: QualifiedIDEntity, limit: Int, offset: Int): Flow<List<MessageEntity>> =
```
typfel, line 145:
```suggestion
    override suspend fun getMessagesByConversation(conversationId: QualifiedIDEntity, limit: Int, offset: Int): Flow<List<MessageEntity>> =
```

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
```diff
@@ -141,8 +142,14 @@ class MessageDAOImpl(private val queries: MessagesQueries) : MessageDAO {
...
+        queries.selectByConversationId(conversationId, limit.toLong(), offset.toLong())
+            .asFlow()
+            .mapToList()
+            .map { entryList -> entryList.map(mapper::toModel) }
+
+    override suspend fun getMessageByConversationAndDate(conversationId: QualifiedIDEntity, date: String): Flow<List<MessageEntity>> =
```
typfel, line 151:
```suggestion
    override suspend fun getMessagesByConversationAfterDate(conversationId: QualifiedIDEntity, date: String): Flow<List<MessageEntity>> =
```

typfel (changes requested):
Good work! I have some suggestions on how we can simplify this and make somethings more generic.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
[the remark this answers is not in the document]
borichellow: what is the `lastModifiedDate` ? 
if it's a date when the changes to the DB were done, then it doesn't help me. Cause it will be updated every time we insert a new message into DB, so there will never be notifications.

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Converations.sq
```diff
@@ -18,7 +21,15 @@ deleteConversation:
 DELETE FROM Conversation WHERE qualified_id = ?;
 
 insertConversation:
-INSERT OR REPLACE INTO Conversation(qualified_id, name, type, team_id, mls_group_id, mls_group_state, protocol) VALUES(?, ?, ?, ?, ?, ?, ?);
+INSERT INTO Conversation(qualified_id, name, type, team_id, mls_group_id, mls_group_state, protocol)
```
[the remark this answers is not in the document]
borichellow: yes, with `OR REPLACE` `last_notification_date ` was set to `null` on every replacing

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAO.kt
borichellow: ou, actually this is something that i forgot to remove! 

It was used for the prev. implementation

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -54,6 +55,10 @@ interface ConversationRepository {
     suspend fun deleteMember(conversationID: QualifiedIDEntity, userID: QualifiedIDEntity): Either<CoreFailure, Unit>
     suspend fun getOneToOneConversationDetailsByUserId(otherUserId: UserId): Either<CoreFailure, ConversationDetails.OneOne?>
     suspend fun createGroupConversation(name: String? = null, members: List<Member>, options: ConverationOptions = ConverationOptions()): Either<CoreFailure, Conversation>
```
gongracr, line 57:
```suggestion
    suspend fun createGroupConversation(name: String? = null, members: List<Member>, options: ConverationOptions = ConversationOptions()): Either<CoreFailure, Conversation>
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/LocalNotification.kt
```diff
@@ -0,0 +1,25 @@
+package com.wire.kalium.logic.data.notification
+
+import com.wire.kalium.logic.data.id.ConversationId
+
+data class LocalNotificationConversation(
+    val id: ConversationId,
+    val name: String,
```
gongracr, line 7: What's this? name of the notification? name of the message sender? name of the group/1on1 conversation?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/LocalNotification.kt
```diff
@@ -0,0 +1,25 @@
+package com.wire.kalium.logic.data.notification
+
+import com.wire.kalium.logic.data.id.ConversationId
+
+data class LocalNotificationConversation(
```
gongracr, line 5: Can we add some documentation about what local means? What I understand is that are Kalium self generated notifications, independent from platform push up notifications?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
```diff
@@ -0,0 +1,80 @@
...
+                            LocalNotificationConversation(
+                                id = conversationId,
+                                name = conversationName,
+                                messages = messages,
+                                isOneToOneConversation = isOneToOneConversation
+                            )
+                        }
+                    }
+            }
+    }
```
gongracr, line 64: Can we add some inline documentation to understand better what is each step/operation doing? Otherwise it takes a bit to understand the internal logic of the use case

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MarkMessagesAsNotifiedUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlinx.coroutines.flow.Flow
+
+interface MarkMessagesAsNotifiedUseCase {
+    suspend operator fun invoke(conversationId: ConversationId?, date: String): Result
+}
+
+sealed class Result {
+    object Success : Result()
+    data class Failure(val storageFailure: StorageFailure) : Result()
+}
```
gongracr, line 17: Just a little ocd remark, but can we move this to the end of the use case? 😅

logic/src/commonMain/kotlin/com/wire/kalium/logic/functional/Flow.kt
```diff
@@ -0,0 +1,21 @@
...
+        this@flatMapFromIterable.forEach { a ->
+            block(a)
+                .take(1)
+                .collect { b ->
+                    result.add(b)
+                    if (result.size == this@flatMapFromIterable.size) emit(result)
+                }
+        }
+    }
+}
```
gongracr, line 21: Niceee method. Small question, doesn't `take(1)` do the same than `first()` ?

gongracr (commented):
Really impressive work 👏 🎩  Just added some minor comments about code documentation, but lookin good to be consumed by AR otherwise. I will wait until the discussion with Jacob about notifications and last modified dates is sorted to approve it. 
Also, are you planning to add some tests for the newly created use cases?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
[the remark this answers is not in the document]
typfel: I explain it another comment.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/LocalNotification.kt
borichellow: i thought that as far as the data class is called "LocalNotification`Conversation`" it's obviously the conversation name. But I'll rename it

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
borichellow: good point, done

logic/src/commonMain/kotlin/com/wire/kalium/logic/functional/Flow.kt
borichellow: good question :) 
from what i got: `take(1)` returns `Flow<T>`, so it will take 1 emit of that flow (when it comes), pass it to the `collect{}` and abort the Flow. 
On the other hand:`first()` returns not `Flow<T>` but `T`, so it kinda waiting for the original Flow to emit. 

So in current example: if I need to fetch list of 1000 elements: 
- `take(1)` won't wait for each element to be fetched, but request it, and **when** it comes - add it to the `result` List
- `first()` will wait for each fetching to add it to the `result` List and only after that request for the next element 

If I'm right (which I'm not 100% sure, as have not so deep knowledge in Coroutines) `take(1)` will work faster OR with the same speed as `first()`.

vitorhugods: The main difference are:

### Take
- Is not an terminal operator. It returns another flow.
- Once you call it, nothing happens. You still need to call something like `collect` or another terminal operator in order to start the flow.
- Is silent, if the flow is closed without emitting any value, the same behaviour will be propagated to whoever is calling take, without exceptions.

### First
- Is a terminal operator. It doesn't return another flow.
- Once you call it, the flow will start immediatelly and hang until the value is emmited.
- If the flow is closed without emitting a value, it will throw `NoSuchElementException`.


---- 
They are similar in the sense that they will make the caller wait for a value to be emitted when collecting the flow. Until the value is emitted, it hangs.

vitorhugods: In this case, I don't understand how can this code work 🧐

`take(1)` will make sure that only ONE value is collected. If you run this with two or more items, the flow will never emmit enything, no?

vitorhugods: All things considered, because we have a `take(1)` quickly followed by a `collect()`, it doesn't make much of a difference between using it `take` or `first`.

We can re-write the function to be like this:

```kt
suspend inline fun <A, B> Collection<A>.flatMapFromIterable(
    crossinline block: suspend (A) -> Flow<B>
): Flow<List<B>> = flow {
    val result = mutableListOf<B>()

    if (isEmpty()) emit(result)

    forEach { a ->
        result.add(block(a).first())
    }
    emit(result)
}
```

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Converations.sq
borichellow: that is a really cool idea!

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
```diff
@@ -17,21 +17,23 @@ import com.wire.kalium.network.api.message.MessageApi
...
 import com.wire.kalium.persistence.dao.message.MessageEntity
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.map
 
 interface MessageRepository {
-    suspend fun getMessagesForConversation(conversationId: ConversationId, limit: Int): Flow<List<Message>>
-    suspend fun persistMessage(message: Message): Either<CoreFailure, Unit>
+    suspend fun getMessagesForConversation(conversationId: ConversationId, limit: Int, offset: Int): Flow<List<Message>>
+    suspend fun persistMessage(message: Message, isMyMessage: Boolean = true): Either<CoreFailure, Unit>
```
vitorhugods, line 29: What is `myMessage`?

Is it when I am the sender? Could it be from myself but from another device?

Can't we just figure it out internally based on `userId` and `clientId`?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
```diff
@@ -0,0 +1,91 @@
...
+
+class GetNotificationsUseCaseImpl(
+    private val messageRepository: MessageRepository,
+    private val userRepository: UserRepository,
+    private val conversationRepository: ConversationRepository,
+    private val messageMapper: MessageMapper = MapperProvider.messageMapper(),
+    private val publicUserMapper: PublicUserMapper = MapperProvider.publicUserMapper()
+) : GetNotificationsUseCase {
+
+    override suspend operator fun invoke(): Flow<List<LocalNotificationConversation>> {
```
vitorhugods, line 32: Big and powerful function. No tests? 🧐

logic/src/commonMain/kotlin/com/wire/kalium/logic/functional/Flow.kt
```diff
@@ -0,0 +1,27 @@
...
```
vitorhugods, line 25:
```suggestion


suspend inline fun <A, B> Collection<A>.flatMapFromIterable(
    crossinline block: suspend (A) -> Flow<B>
): Flow<List<B>> = flow {
    val result = mutableListOf<B>()

    if (isEmpty()) emit(result)

    forEach { a ->
        result.add(block(a).first())
    }
    emit(result)
}
```

vitorhugods: Also, could add at least a test for it:

```kotlin

    @Test
    fun givenAListOfItems_whenFlatMappingFromIterable_thenTheMappedResultsShouldBeEmitted() = runTest {
        val items = listOf(1, 2)

        val result = items.flatMapFromIterable { flowOf(it * 2) }

        result.test {
            assertEquals(listOf(2, 4), awaitItem())
            awaitComplete()
        }
    }
```

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAO.kt
```diff
@@ -57,7 +57,8 @@ interface MessageDAO {
     suspend fun insertMessages(messages: List<MessageEntity>)
     suspend fun updateMessage(message: MessageEntity)
     suspend fun updateMessageStatus(status: MessageEntity.Status, id: String, conversationId: QualifiedIDEntity)
-    suspend fun getAllMessages(): Flow<List<MessageEntity>>
+    suspend fun getAllMessages(limit: Int, offset: Int): Flow<List<MessageEntity>>
```
vitorhugods, line 60:
```suggestion
    suspend fun getMessagesFromAllConversations(limit: Int, offset: Int): Flow<List<MessageEntity>>
```

vitorhugods (changes requested):
Overall I think it's a nice job. 

But this PR could have been split into smaller chunks before. Not sure if it's worth it now, but just to consider in the future:

- The renaming of `ConversationOptions` in a simple PR: `refactor: rename ConversationOptions` that could have been approved and merged fast.
- `MarkMessagesAsNotifiedUseCase` could be added on its own PR.
- `GetNotificationsUseCase` could be added on its own PR.

Also, I think the UseCases should have at least a couple of happy tests written for them.

vitorhugods (approved):
Nicely done! 👨🏼‍🍳🤌🏼

PR #378, opened 2022-04-08
fix: store jvm preferences in a more discoverable place

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Finding and deleting the data stored in `KaliumPreferences` when running the JVM target is really hard annoying.

### Solutions

Use `userNodeForPackage(EncryptedSettingsHolder::class.java)` instead of `userRoot()` which on macOS at least save the preferences in `~/Library/Preferences/com.wire.kalium.plist` instead of some obscure java plist file. 

#### How to Test

CLI app still works.

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/persistence/src/jvmMain/kotlin/com/wire/kalium/PackagePathProvider.kt
+++ b/persistence/src/jvmMain/kotlin/com/wire/kalium/PackagePathProvider.kt
@@ -1,0 +1,8 @@
+package com.wire.kalium;
+
+/**
+ * IMPORTANT: this class only exist to provide a permanent package path for
+ * the Java Preferences API and must **NOT** be moved to a different package.
+ * Doing so will result in data loss for users.
+ */
+internal class PackagePathProvider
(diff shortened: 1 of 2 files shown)


vitorhugods (commented):
I'm a bit skeptical of this.

Relying on a package in order to persist some data means that if we rename the package, or move EncryptedSettingsHolder elsewhere, the data might be lost/not-found.
A bit of a weird side-effect that's hard to track and someday may JVM apps using this when there's a new version of Kalium/persistence available.

typfel:
> I'm a bit skeptical of this.
> 
> Relying on a package in order to persist some data means that if we rename the package, or move EncryptedSettingsHolder elsewhere, the data might be lost/not-found. A bit of a weird side-effect that's hard to track and someday may JVM apps using this when there's a new version of Kalium/persistence available.

Ok, I agree that's not very nice. But having the preferences file shared with all other java applications is also not very nice. What about creating and using a dummy class at package root `com.wire.kalium`?

vitorhugods:
> > I'm a bit skeptical of this.
> > Relying on a package in order to persist some data means that if we rename the package, or move EncryptedSettingsHolder elsewhere, the data might be lost/not-found. A bit of a weird side-effect that's hard to track and someday may JVM apps using this when there's a new version of Kalium/persistence available.
> 
> Ok, I agree that's not very nice, but having the preferences file shared with all other java applications is also not very nice. What about creating and using a dummy class at package root `com.wire.kalium`?

A bit weird, but it's Java. So... sure!

A Dummy class with some documentation on top to make sure we don't move it around.

BatmanDontMoveAMuscleGIF

PR #397, opened 2022-04-12
fix: events received through websocket being re-processed durying sync

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

@borichellow noticed that we were receiving some events twice. First in real-time and then when doing sync.

Easy to reproduce:
1. Receive an event through the Websocket
2. Close the app
3. Open the app
4. Event will be received again through the notification API

### Causes

I remembered a crucial flaw when doing the receiving of events, that right now the "last processed ID" is only updated when receiving events through the API.

So *all* the events received through the Websocket are not being considered as processed, making the app request for them again when opening the app.

### Solutions

Change the current implementation, so:
- Remove the updating of the last ID from the repository, as it has two sources (REST and Websocket);
- Centralise this logic by making `ListenToEventsUseCase` responsible for updating the processed ID and always **after** processing the event.

`ListenToEventsUseCase` is completely agnostic about the source of the events, it doesn't know if the event has come through the Websocket or through the REST API. So it stays like this:
1. Event is emited
2. Event is processed
3. Last ID is updated

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ListenToEventsUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ListenToEventsUseCase.kt
@@ -32,6 +32,7 @@ class ListenToEventsUseCase(
                                 kaliumLogger.i(message = "Unhandled event id=${event.id}")
                             }
                         }
+                        eventRepository.updateLastProcessedEventId(event.id)
                     }
                 }.onFailure {
                     kaliumLogger.e(message = "Failure when receiving events: $it")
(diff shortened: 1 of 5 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ListenToEventsUseCase.kt
```diff
@@ -32,6 +32,7 @@ class ListenToEventsUseCase(
                                 kaliumLogger.i(message = "Unhandled event id=${event.id}")
                             }
                         }
+                        eventRepository.updateLastProcessedEventId(event.id)
```
alexandreferris, line 35: 👌🏻

alexandreferris (approved):

PR #413, opened 2022-04-18
fix: bug not downloading correctly asset messages from web client 

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When Web client sends asset messages, it does it with 2 messages:
1. First it sends a preview asset message, with the file metadata information such as asset width and height (in case of Images or video), file name, mimeType, and asset size. This is in case the other receiving client wanted to just show a quick preview placeholder with that metadata.
2. Secondly it sends another asset message with the same message and conversation ids than the first one, but just containing the asset RemoteData (decryption keys, asset id and asset token)

The way that the asset send and receive logic was implemented previously, was discarding any asset message that didn't include the file decryption keys. Therefore file metadata could not be displayed correctly if the original message didn't contain both.

### Solutions

Even though in AR we send now just 1 asset message with all the needed info (metadata + decryption keys), to allow retrocompatibility with web clients that implement above behavior we had to make a workaround. This consists on storing the first preview message, and then updating it accordingly once the second message arrives with the final decryption keys

#### How to Test

A PR in AR will be raised following this one containing the fix.

### Notes (Optional)

This solution is not optimal due to the fact that there is a lapse of time (between the 1st and 2nd messages being received) where the asset message received will not be able to be downloaded, as the `assetId` will be missing temporarily.

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
@@ -101,7 +101,7 @@ class MessageDataSource(
         }.onSuccess {
             Either.Right(it)
         }.onFailure {
-            Either.Left(StorageFailure.DataNotFound)
+            Either.Left(it)
         }
 
     override suspend fun markMessageAsSent(conversationId: ConversationId, messageUuid: String) =
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
@@ -1,12 +1,17 @@
 package com.wire.kalium.logic.data.message
 
 import com.wire.kalium.logic.CoreFailure
+import com.wire.kalium.logic.StorageFailure
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.id.IdMapper
 import com.wire.kalium.logic.di.MapperProvider
 import com.wire.kalium.logic.failure.SendMessageFailure
 import com.wire.kalium.logic.functional.Either
+import com.wire.kalium.logic.functional.onFailure
+import com.wire.kalium.logic.functional.onSuccess
+import com.wire.kalium.logic.kaliumLogger
 import com.wire.kalium.logic.wrapApiRequest
+import com.wire.kalium.logic.wrapStorageRequest
 import com.wire.kalium.network.api.message.MLSMessageApi
 import com.wire.kalium.network.api.message.MessageApi
 import com.wire.kalium.network.api.message.MessagePriority
@@ -15,8 +20,7 @@ import com.wire.kalium.network.utils.isSuccessful
 import com.wire.kalium.persistence.dao.message.MessageDAO
 import com.wire.kalium.persistence.dao.message.MessageEntity
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.filterNotNull
-import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.map
 
 interface MessageRepository {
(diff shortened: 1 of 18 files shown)


typfel:
Did you consider ignoring asset messages which doesn't contain `RemoteData`?

gongracr:
@typfel yes, that's what was being doine previously, and that caused that the asset messages appeared to be empty as those were the only ones including the asset metadata such as asset file name and size
Captura de Pantalla 2022-04-18 a las 18 00 22

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
```diff
@@ -88,15 +92,17 @@ class MessageDataSource(
...
-                .first()
-        )
-    }
+                .firstOrNull()?.run {
+                    messageMapper.fromEntityToMessage(this)
+                }
+        }.onSuccess {
+            Either.Right(it)
+        }.onFailure {
+            Either.Left(StorageFailure.DataNotFound)
```
saleniuk, line 104: I think that `wrapStorageRequest` can also return other failures - should we also handle them as a `DataNotFound`?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
```diff
@@ -99,9 +99,8 @@ class ProtoContentMapperImpl : ProtoContentMapper {
         val content = when (val protoContent = genericMessage.content) {
             is GenericMessage.Content.Text -> MessageContent.Text(protoContent.value.content)
             is GenericMessage.Content.Asset -> {
-                if (protoContent.value.uploaded != null) MessageContent.Asset(
-                    MapperProvider.assetMapper().fromProtoAssetMessageToAssetContent(protoContent.value)
-                ) else MessageContent.Unknown // Sometimes backend sends some preview asset messages just with img metadata and no keys or asset id, so we discard it
+                // Backend sends some preview asset messages just with img metadata and no keys or asset id, so we need to overwrite one with the other one
```
saleniuk, line 102: Maybe it would be better to distinguish them by creating a sealed class instead of passing `AssetContent.RemoteData` with empty data? 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ConversationEventReceiver.kt
```diff
@@ -77,7 +72,20 @@ class ConversationEventReceiver(
...
+                        is MessageContent.Text -> messageRepository.persistMessage(message)
+                        is MessageContent.Asset -> {
+                            messageRepository.getMessageById(message.conversationId, message.id)
+                                .onFailure {
+                                    // No asset message was received previously, so just persist the preview asset message
+                                    messageRepository.persistMessage(message)
+                                }
+                                .onSuccess { persistedMessage ->
+                                    // The asset message received contains the asset decryption keys, so update the preview message persisted previously
+                                    updateMessage(persistedMessage, message.content.value.remoteData)?.let {
```
typfel, line 84: We also need to verify that sender of the updated message is the same as the original message.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ConversationEventReceiver.kt
```diff
@@ -99,6 +107,18 @@ class ConversationEventReceiver(
         }
     }
 
+    private fun updateMessage(persistedMessage: Message, newMessageRemoteData: AssetContent.RemoteData): Message? =
```
typfel, line 110:
```suggestion
    private fun updateAssetMessage(persistedMessage: Message, newMessageRemoteData: AssetContent.RemoteData): Message? =
```

typfel (changes requested):
👍  and I would consider filtering out asset messages which only has the preview.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ConversationEventReceiver.kt
gongracr: Would it be possible to receive 2 messages with same message id but different sender ID's? If so that would be indicative of something more serious happening

typfel: It shouldn't be possible for someone else then the original sender to update his/her message. If we don't verify this we have security vulnerability since then an attacker could secretly replace the asset sent by someone else.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
gongracr: I changed it to just forward it the same `StorageFailure` failure returned by the `messageDao` instance

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
gongracr: I think that's a bit overkill, as this will probably (hopefully) be improved by the backend to accommodate to the way we do it in Android now, i.e, with just one asset message instead of 2

saleniuk (approved):

typfel (approved):

PR #421, opened 2022-04-21
chore: Rename OngoingCallsUseCase to IncomingCallsUseCase

alexandreferris:

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently there was no real use of ongoing calls and thus renaming it to be correctly used as IncomingCalls

### Solutions

Renamed the usecase to `IncomingCallsUseCase`

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
@@ -20,12 +20,10 @@ interface CallManager {
 
 expect class CallManagerImpl : CallManager
 
-val CallManager.ongoingCalls get() = allCalls.map {
+val CallManager.incomingCalls get() = allCalls.map {
     it.filter { call ->
         call.status in listOf(
-            CallStatus.INCOMING,
-            CallStatus.ANSWERED,
-            CallStatus.ESTABLISHED
+            CallStatus.INCOMING
         )
     }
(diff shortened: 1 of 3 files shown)


vitorhugods (approved):
I feel like I'm not the only one that will ask this:

What IS an incoming call?
What if it was established, but I dropped out and it's still going on. Is it classified as `incoming` and I can re-join it? Or what?


This PR is a step in the right direction, but we still have some states and edge-cases that we may want to clarify going forward. I suggest keeping and eye open as we explore, and try to cover with KDocs the different use cases and how they will affect the status of the call.

ohassine (approved):

alexandreferris:
> I feel like I'm not the only one that will ask this:
> 
> What IS an incoming call? What if it was established, but I dropped out and it's still going on. Is it classified as `incoming` and I can re-join it? Or what?
> 
> This PR is a step in the right direction, but we still have some states and edge-cases that we may want to clarify going forward. I suggest keeping and eye open as we explore, and try to cover with KDocs the different use cases and how they will affect the status of the call.

Indeed! The change was mainly because this specific usecase is now being used to listen when an incoming call is ... well, incoming to the user and having it only with `INCOMING` status will make sure that will only open the incoming call screen only once.

The way it was before, whenever the call was ANSWERED or ESTABLISHED, the incoming call UI would show multiple times and trying to keep the logic out of reloaded project I did the change directly to this usecase to solely work towards a single purpose.

PR #422, opened 2022-04-21
fix: login is kaput after logout [AR-1398]

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

login after logout, resulting in the app stuck at client register step

### Causes (Optional)

MLS and Proteus files are not cleared on hard logout

### Solutions

clear local files 

### Testing

#### How to Test

login -> logout -> login(with the same account)

### Notes (Optional)

for this to work proberly the inMemory user storage `CoreLogin.userScopeStorage`

----

----
##### References


cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/MLSClient.kt
```diff
@@ -8,6 +8,16 @@ typealias MLSKeyPackage = ByteArray
...
 
+    /**
+     * Delete local MLS DB and files
+     *
+     * @return true when delete is successful and false otherwise
+     */
+    fun clearLocalFiles(): Boolean
+
+
+    fun close()
```
vitorhugods, line 19: What is close meant to do?

Can we document this?

MohamadJaara: As I understand from the MLS docs, It's used to release any locks the C code have on the MLS resources, aka the DB.

Even though close is called, there are still some files that are not cleared, mainly the `db-wal` and `db-shm` I will ask  Mathieu about this issue.

But the logout is working correctly, but can't confirm if MLS DB state is valid after the login (login after log out)

vitorhugods: I guess I mean a bit different.

Why would I call close and not deleteFiles for example?

MohamadJaara: Good question, I guess it's only needed when deleting the files, soft logout will keep the files -> the locks can stay

MohamadJaara: I will change it and make close part of `deleteLocalFiles` function (I will do the same for `ProteusClient`)

mchenani (approved):
🔥

alexandreferris (approved):
🚀

cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/MLSClient.kt
```diff
@@ -8,6 +8,14 @@ typealias MLSKeyPackage = ByteArray
 
 interface MLSClient {
 
+    /**
+     * Release any locks the C code have on the MLS resources and
+     * delete local MLS DB and files
+     *
+     * @return true when delete is successful and false otherwise
+     */
+    fun clearLocalFiles(): Boolean
```
vitorhugods, line 17: An exception instead of true or false at least would provide more information on why it failed.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
```diff
@@ -1,27 +1,58 @@
...
         }
     }
+
+    private fun clearUserStorage() {
+        authenticatedDataSourceSet.userDatabaseProvider.nuke()
+        authenticatedDataSourceSet.kaliumPreferencesSettings.nuke()
+    }
+
+    private fun clearCrypto() {
+        authenticatedDataSourceSet.proteusClient.clearLocalFiles()
```
vitorhugods, line 41: Ignored result?

vitorhugods (commented):
What should we do if we fail to delete the local files when logging out?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
MohamadJaara: I deliberately  ignored the result here since I don't believe a logout should fail, the user should press logout the use should be logged out, and we have to handle any error internally

cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/MLSClient.kt
MohamadJaara: The `deleteRecursively()` used in JVM/Android the returns Boolean, and it throws error only if the path is null, so it will never throw.
I can't say the same about other platforms.

PR #426, opened 2022-04-22
fix: update the message time with the one received from the server after sending [AR-1437]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Messages sometimes can be wrongly ordered. The only situation that has been found that it happens is when the user has changed his time manually.
In order to maintain the correct messages order for all participants in a conversation, it must be synchronised using the time from the server, which hasn't been done before.

### Solutions

The result of the API request that is used to send messages contains the server time for such message, so now it's passed and the message datetime is updated after a successful send. To keep the order of all _pending_ messages, the time difference between the local time and the server time is calculated and such difference is applied to the datetime of all remaining _pending_ messages to keep the right order in the conversation.

#### Test Coverage (Optional)

#### How to Test

Change the timezone or even set the time manually and start sending messages.

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -78,28 +91,34 @@ class MessageSenderImpl(
...
         messageUuid: Message,
-    ): Either<CoreFailure, Unit> = suspending {
+    ): Either<CoreFailure, String> = suspending {
         messageRepository.sendEnvelope(conversationId, envelope).coFold(
             {
                 when (it) {
                     is SendMessageFailure.Unknown -> Either.Left(it)
                     is SendMessageFailure.ClientsHaveChanged -> messageSendFailureHandler.handleClientsHaveChangedFailure(it).flatMap {
-                        trySendingOutgoingMessage(conversationId, messageUuid)
+                        attemptToSend(conversationId, messageUuid)
```
vitorhugods, line 117: Does this mean that in case of a retry the message time won't be updated?

vitorhugods (changes requested):
Nice fix. Just small comments

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
saleniuk: It has been implemented so that there is a loop when `ClientsHaveChanged` (`attemptToSend` -> `attemptToSendWithProteus` -> `trySendingEnvelopeRetryingIfPossible` -> `attemptToSend` ...) and I don't know if it should be this way or not so I didn't want to modify it, just apply the "update message time" changes. It should eventually return the message time when coming out of this loop when `messageRepository.sendEnvelope` returns success.

vitorhugods: Oh yeah indeed.

We probably should have a `maxRetries` and `retryCount` for this Proteus loop in the future. But it should just work after one or MAX two tries.

Anything other than that would be a very unexpected behaviour from the backend or some internal bug in Kalium.

vitorhugods (approved):

PR #429, opened 2022-04-24
refactor: remove suspend modifier from DAOs

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

1. `suspend` modifier is used by default in all DAOs functions, which lead to unnecessary compile/run time overhead.
2. using `flow` to fetch only list from the DB with `flow.fitst()` 

### Solutions

1. remove the ´suspend´ modifier.
2. replace `query.asFlow()` with `query.executeAsList()` and using `.asFlow()` only when it's really needed (e.g. observing a list of conversations)

Needs releases with:

### Testing

### Notes (Optional)

this PR is not supposed to pass CI since the changes here requires refactoring some repositories and UseCases in ´logic´ (the logic refactor will come in a later PR)

----

----
##### References

### DIFF
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
@@ -1,6 +1,5 @@
 import com.wire.kalium.persistence.dao.QualifiedIDEntity;
 import com.wire.kalium.persistence.dao.ConversationEntity;
-import kotlin.Int;
 
 CREATE TABLE Conversation (
     qualified_id TEXT AS QualifiedIDEntity NOT NULL PRIMARY KEY,
@@ -46,3 +45,6 @@ updateConversationMutingStatus:
 UPDATE Conversation
 SET muted_status = ?, muted_time = ?
 WHERE qualified_id = ?;
+
+selectConversationByType:
+SELECT * FROM Conversation WHERE type = ?;
--- a/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAO.kt
+++ b/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAO.kt
@@ -31,25 +31,27 @@ data class Member(
 )
 
 interface ConversationDAO {
-    suspend fun getSelfConversationId(): QualifiedIDEntity
-    suspend fun insertConversation(conversationEntity: ConversationEntity)
-    suspend fun insertConversations(conversationEntities: List<ConversationEntity>)
-    suspend fun updateConversation(conversationEntity: ConversationEntity)
-    suspend fun updateConversationGroupState(groupState: ConversationEntity.GroupState, groupId: String)
-    suspend fun getAllConversations(): Flow<List<ConversationEntity>>
-    suspend fun getConversationByQualifiedID(qualifiedID: QualifiedIDEntity): Flow<ConversationEntity?>
-    suspend fun getConversationByGroupID(groupID: String): Flow<ConversationEntity?>
-    suspend fun deleteConversationByQualifiedID(qualifiedID: QualifiedIDEntity)
-    suspend fun insertMember(member: Member, conversationID: QualifiedIDEntity)
-    suspend fun insertMembers(memberList: List<Member>, conversationID: QualifiedIDEntity)
-    suspend fun deleteMemberByQualifiedID(conversationID: QualifiedIDEntity, userID: QualifiedIDEntity)
-    suspend fun getAllMembers(qualifiedID: QualifiedIDEntity): Flow<List<Member>>
-    suspend fun insertOrUpdateOneOnOneMemberWithConnectionStatus(
+    fun getSelfConversationId(): QualifiedIDEntity?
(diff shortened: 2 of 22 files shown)


typfel:
The methods were made suspend to be compatible with JS where many apis a async.

ohassine (changes requested):
But Dao methods are needed to be run asynchronously and inside a worker thread.
 If we remove this `suspend` modifier how can we do that ?

MohamadJaara:
@typfel if we are going to use `sqlDelight` it will be fine since DB operations are normal blocking functions.
Otherwise, we need to investigate if it's worth it to introduce extra overhead for android/iOS to support JS (and how much of an over head it's)

MohamadJaara:
> But Dao methods are needed to be run asynchronously and inside a worker thread. If we remove this `suspend` modifier how can we do that ?

@ohassine it's fine since regular/blocking functions can be called from Coroutines﻿.
`suspend` modifier is used by the compiler to create a state machine that optimizes using callbacks every time a function needs to suspend.
In our case, the  sqlDelight functions `asFlow(), executeAsList(), ect` are blocking functions that never suspend, so adding `suspend` is only adding compile/run time overhead with no benefits.

typfel:
> > The methods were made suspend to be compatible with JS where many apis are async.
> 
> @typfel if we are going to use `sqlDelight` it will be fine since DB operations are normal blocking functions. Otherwise, we need to investigate if it's worth it to introduce extra overhead for android/iOS to support JS (and how much of an over head it's)

Well we don't know yet if can use sql delight for JS, we definitely can't use it out of the box. Why are you in such a hurry to remove the `suspend` annotations if we haven't even measured the overhead it brings. Why is it important to do this now? Can't it wait until we've figured out the JS persistence solution?

MohamadJaara:
> > > The methods were made suspend to be compatible with JS where many apis are async.
> > 
> > 
> > @typfel if we are going to use `sqlDelight` it will be fine since DB operations are normal blocking functions. Otherwise, we need to investigate if it's worth it to introduce extra overhead for android/iOS to support JS (and how much of an over head it's)
> 
> Well we don't know yet if can use sql delight for JS, we definitely can't use it out of the box. Why are you in such a hurry to remove the `suspend` annotations if we haven't even measured the overhead it brings. Why is it important to do this now? Can't it wait until we've figured out the JS persistence solution?



> > > The methods were made suspend to be compatible with JS where many apis are async.
> > 
> > 
> > @typfel if we are going to use `sqlDelight` it will be fine since DB operations are normal blocking functions. Otherwise, we need to investigate if it's worth it to introduce extra overhead for android/iOS to support JS (and how much of an over head it's)
> 
> Well we don't know yet if can use sql delight for JS, we definitely can't use it out of the box. Why are you in such a hurry to remove the `suspend` annotations if we haven't even measured the overhead it brings. Why is it important to do this now? Can't it wait until we've figured out the JS persistence solution?

There is no hurry to change it, I just wanted to raise a quick PR after our discussion in the collective, there is another point in the PR about using `.flow()` to retrieve a list

vitorhugods:
> But I think we need it for db operations (like inserting and updating) as they may be delayed.

`suspend` isn't needed in order to run on a different thread/context/dispatcher/whatever. We can do that from the outside. Maybe the apps, maybe Kalium:Logic should do it.

Example:

```kotlin
class MyDao {
    // DAO doesn't have suspend modifiers
    fun insertStuffIntoTheDB() { /** do it **/ }
}


suspend fun main() {
    myScope.launch(Dispatchers.IO){
        // I'm now running this on IO Dispatcher
        insertStuffIntoTheDB()
    }
}
```

----

However, @MohamadJaara, trying to go a bit deeper on what @typfel is saying:

SQLite drivers for JVM, Android and iOS are synchronous, so internally they can easily work like that.



However, for JS and IndexedDB, INTERNALLY (stuff we don't see) the calls are something like this:

```kotlin
class AsyncDBDriver {
    fun insertStuff(data: Data, onInsertComplete: () -> Unit)) { /** insert and call the callback when done **/ }
}
```

And SQLDelight will have to use it somehow, and a non-suspendable version doesn't work right:
```kotlin
class NonSuspendableGeneratedSQLDelightCode {
    fun insertConversation(conversation: ConversationEntity) {
        driver.insertStuff{ 
            // WTF? How do I make this work?
        }
    }
}

class SuspendableGeneratedSQLDelightCode {
    // With suspend this will work!
    suspend fun insertConversation(conversation: ConversationEntity) = suspendCoroutine { continuation ->
        driver.insertStuff(conversation) { 
            continuation.resume()
        }
    }
}
```

Sooooooooooo, currently `SQLDelight` doesn't generate `suspendable` queries in order to support these Drivers, this means that if we don't currently need to use `suspendable`.

However, it seems kinda natural that it soon will able to generate `suspendable` queries and support async drivers, as seen here: https://github.com/cashapp/sqldelight/pull/3168


I hope this helps clarifying the issue. So, in general, I think @typfel made the right move by leaving starting with the `suspend` modifier, as we future-proof ourselves.

MohamadJaara:
After the feedback, I will close this PR and open new one only with the changes to removing ´.first()´  when reading lists from DB

PR #433, opened 2022-04-25
chore: identify lack of connection failures (AR-1312)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We currently don't identify when an API request fails because we're unable to reach the server.

### Solutions

Ktor will throw `IOException` when we can't reach the server due to any reason.
Handle `IOException` independently from generic `Exception`, mapping it to the current `NoNetworkConnection` failure.

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreFailure.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreFailure.kt
@@ -4,6 +4,7 @@ import com.wire.kalium.cryptography.exceptions.ProteusException
 import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.network.exceptions.KaliumException
 import com.wire.kalium.network.utils.NetworkResponse
+import io.ktor.utils.io.errors.IOException
 
 sealed class CoreFailure {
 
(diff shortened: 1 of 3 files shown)


MohamadJaara:
> Ktor will throw IOException when we can't reach the server due to any reason.

This was how it's implemented in `network` before, but it is not accurate since for JVM throw UnknownHostException (i think the same is true for Android)

vitorhugods:
> > Ktor will throw IOException when we can't reach the server due to any reason.
> 
> This was how it was implemented in `network` before, but it is not accurate since for JVM throw UnknownHostException (i think the same is true for Android)

This was before Ktor 2.0, which changed a lot how these exceptions are handled. `UnknownHost` is now a cause of this `IOException` that Ktor throws. Notice that it isn't  a Java/JVM/Android `IOException`, but [Ktor's IOException](https://github.com/ktorio/ktor/blob/43fa9ab942f89a7dc1aa2c2f6bb04e94f932e80a/ktor-io/common/src/io/ktor/utils/io/errors/Errors.kt#L3) now.

There's still some work to be done from Ktor side, as it seems that on JavaScript some exceptions can get away with this.
But as we currently don't have JS targets for `network` nor `logic`, it's hard to improve this or to make JS-specific code.

I suggest leaving it like this or be more brutal about it and map any uncatched exceptions to `NoNetworkAvailable` instead of `ServerMiscomunication`.

MohamadJaara (approved):

PR #442, opened 2022-04-26
feat: add filtering to the search result by only the accepted connection status

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- the search query from the DAO returned the contacts that also had the status as not accepted 

### Solution

- extended the user dao method to query by additional ConnectionState argument 
----

----

----
##### References

### DIFF
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
@@ -47,11 +47,12 @@ SELECT * FROM User WHERE connection_status = ?;
 selectByQualifiedId:
 SELECT * FROM User WHERE qualified_id = ?;
 
-selectByNameOrHandleOrEmail:
+selectByNameOrHandleOrEmailAndConnectionState:
 SELECT * FROM User
 WHERE name LIKE ('%' || :searchQuery || '%')
 OR  handle LIKE  ('%' || :searchQuery || '%')
-OR  email LIKE  ('%' || :searchQuery || '%');
+OR  email LIKE  ('%' || :searchQuery || '%')
+AND connection_status = :connectionStatus;
 
 updateUserhandle:
 UPDATE User SET handle = ? WHERE qualified_id = ?;
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
@@ -49,9 +49,9 @@ SELECT * FROM User WHERE qualified_id = ?;
 
 selectByNameOrHandleOrEmailAndConnectionState:
 SELECT * FROM User
-WHERE name LIKE ('%' || :searchQuery || '%')
+WHERE (name LIKE ('%' || :searchQuery || '%')
 OR  handle LIKE  ('%' || :searchQuery || '%')
-OR  email LIKE  ('%' || :searchQuery || '%')
+OR  email LIKE  ('%' || :searchQuery || '%'))
 AND connection_status = :connectionStatus;
 
 updateUserhandle:
--- a/persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/UserDAOTest.kt
+++ b/persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/UserDAOTest.kt
@@ -101,7 +101,7 @@ class UserDAOTest : BaseDatabaseTest() {
         val user3 = USER_ENTITY_3
         db.userDAO.insertUsers(listOf(user1, user2, user3))
         //when
-        val searchResult = db.userDAO.getUserByNameOrHandleOrEmail(user2.email!!).first()
+        val searchResult = db.userDAO.getUserByNameOrHandleOrEmailAndConnectionState(user2.email!!).first()
         //then
         assertEquals(searchResult, listOf(user2))
     }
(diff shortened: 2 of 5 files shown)


persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/UserDAOTest.kt
```diff
@@ -251,11 +251,26 @@ class UserDAOTest : BaseDatabaseTest() {
...
+
+        val mockUsers = listOf(USER_ENTITY_1, USER_ENTITY_2, USER_ENTITY_3)
+        db.userDAO.insertUsers(mockUsers)
+        //when
+        val searchResult = db.userDAO.getUserByNameOrHandleOrEmailAndConnectionState(commonNamePrefix, UserEntity.ConnectionState.ACCEPTED).first()
+        //then
+        searchResult.forEach { userEntity ->
+            assertEquals(UserEntity.ConnectionState.ACCEPTED,userEntity.connectionStatus)
+        }
+    }
```
vitorhugods, line 272: Could you insert a couple of users that are **not** connected and just one conected, making sure that is the one that will be returned?

vitorhugods (changes requested):
Nice job. Just one suggestion to test making sure the connection condition is respected when searching users.

alexandreferris (approved):
Great addition 🚀 

I do have one question but not code related, as this is to show the users we already have an accepted connection, when will we see users that we sent/received those connections but are still pending?

persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/UserDAOTest.kt
trOnk12: sure ,good idea

trOnk12:
> Great addition 🚀
> 
> I do have one question but not code related, as this is to show the users we already have an accepted connection, when will we see users that we sent/received those connections but are still pending?

good question I think it should be another ticket, but I think it will be a matter of passing a list of connection states to the DAO method instead of a single one, but perhaps that is a question for @vitorhugods  ? :P as this can be implemented pretty soon I guess

vitorhugods:
> Great addition 🚀
> 
> I do have one question but not code related, as this is to show the users we already have an accepted connection, when will we see users that we sent/received those connections but are still pending?

We should already be able to do this on Reloaded, but we don't have design for that yet I think.

```kotlin
observeConversationListDetails.collect { conversationList ->
    conversationList.filterIsInstance<ConversationDetails.OneOne>()
        .forEach { conversation ->
            conversation.otherUser.connectionStatus
    }
}
```

But right now we ignore the connection status when showing the list of conversations on Reloaded.

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
```diff
@@ -47,11 +47,12 @@ SELECT * FROM User WHERE connection_status = ?;
...
 SELECT * FROM User WHERE qualified_id = ?;
 
-selectByNameOrHandleOrEmail:
+selectByNameOrHandleOrEmailAndConnectionState:
 SELECT * FROM User
 WHERE name LIKE ('%' || :searchQuery || '%')
 OR  handle LIKE  ('%' || :searchQuery || '%')
-OR  email LIKE  ('%' || :searchQuery || '%');
+OR  email LIKE  ('%' || :searchQuery || '%')
+AND connection_status = :connectionStatus;
```
saleniuk, line 55: I think first part with all the OR has to be in brackets otherwise you won't get what you want 😄

trOnk12: hmmm unit tests are passing :P will verify it with another unit tests as Vitor suggested adding some that should give the answer to this one if it works as expected

trOnk12: good one, could anyone explain me @vitorhugods @saleniuk  why we need those brackets ? 

does it work likes this without brackets ? : 

look for a ROW with a name equal to query 
OR a ROW with a handle equal to query 
OR a ROW equal to a EMAIL AND a connection_status equal to query ?

trOnk12:
@vitorhugods @saleniuk  pushed the changes and added additional unit tests

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
saleniuk: AND has precedence over OR, so if we have something like that:
`WHERE A OR B AND C`
 without brackets it will be actually executed like
 `WHERE A OR (B AND C)`
 so if we want it to be different, we have to set the brackets ourself, in this case it should be:
 `WHERE (A OR B) AND C`

vitorhugods: > look for a ROW with a name equal to query
> OR a ROW with a handle equal to query
> OR a ROW equal to a EMAIL AND a connection_status equal to query ?

We want users that MUST be `CONNECTED`.

And match at least one of: `handle`, `email`, `name`.

Currently: `matchName || matchEmail || matchHandle && matchConnection`.
It means that if it `matchName` is true, it's enough.

It should be: `(matchName || matchEmail || matchHandle) && matchConnection`.



You can play with this example:
https://www.db-fiddle.com/f/dEjmVGeZqo5ZePLPzJukNv/0

trOnk12: thanks <3

vitorhugods (approved):
🚀🚀🚀

PR #443, opened 2022-04-26
feat(messages): retry sending after reconnecting (PART-2): Android implementation (AR-1312)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We have no implementation for the new `MessageSendingScheduler` for Android.

### Solutions

DecidingIfYouShouldTalkToYourCrushGIF

- Modify our `WorkScheduler` to implement the `MessageSendingScheduler` interface.
- The `WorkScheduler` on Android will enqueue a `ScheduledMessageWorker` for each message that needs to be retried once internet connection is restablished.
- In case the same failure happens, the cycle should repeat naturally.

### Testing

W.I.P.

----

----
##### References


logic/src/androidMain/kotlin/com/wire/kalium/logic/sync/WorkScheduler.kt
```diff
@@ -118,7 +136,33 @@ actual class WorkScheduler(private val context: Context, private val userId: Use
...
+            .build()
+
+        val request = OneTimeWorkRequest.Builder(workerClass)
+            .setConstraints(connectedConstraint)
+            .setInputData(inputData)
+            .build()
+
+        WorkManager.getInstance(context).enqueueUniqueWork(
+            WORK_NAME_PREFIX_PER_USER + userId.value,
+            ExistingWorkPolicy.APPEND,
```
trOnk12, line 155: are we sure we want to append the work and not replace it ? I assume that we are could retry the sending of the same persisted message several times before we get the the internet I think in that case it would enqueue several times of sending the same message ?

trOnk12: perhaps it is worth to add some integration tests with the WorkManager, dunno if you plan to do it or not in the next PR's :P

trOnk12: ahhhh I see it it is W.I.P, forget it

logic/src/androidMain/kotlin/com/wire/kalium/logic/sync/WorkScheduler.kt
```diff
@@ -94,21 +100,33 @@ class WrapperWorkerFactory(private val coreLogic: CoreLogic) : WorkerFactory() {
...
-actual class WorkScheduler(private val context: Context, private val userId: UserId) {
+actual class WorkScheduler(private val context: Context, private val userId: UserId) : MessageSendingScheduler {
 
-    actual fun schedule(work: KClass<out UserSessionWorker>, name: String) {
-        val inputData = Data.Builder()
-            .putString(WrapperWorkerFactory.WORKER_CLASS_KEY, work.java.canonicalName)
-            .putSerializable(WrapperWorkerFactory.USER_ID_KEY, userId)
-            .build()
-        val request = OneTimeWorkRequest.Builder(WrapperWorker::class.java)
+    private val workerClass = WrapperWorker::class.java
```
trOnk12, line 124: plx add a space between workerClass and enqueueImmediateWork <3

trOnk12 (commented):
great job as always 🥇

logic/src/androidMain/kotlin/com/wire/kalium/logic/sync/WorkScheduler.kt
vitorhugods: > are we sure we want to append the work and not replace it ? I assume that we are could retry the sending of the same persisted message several times before we get the the internet I think in that case it would enqueue several times of sending the same message ?

There's no way to try multiple times right now.
Cycle goes like this:
1. `SendWhateverMessageUseCase`
2. `MessageSender`
3. In case of failure due to bad Internet: `MessageSendingScheduler`
4. On Android, one job will be scheduled for EACH message, in order. Goes back to 2.

So it's linear. No way we spawn multiple jobs for the same message.
There's a single queue for each user: `WORK_NAME_PREFIX_PER_USER + userId.value`, and each message retry is appended in there.

If, for whatever reason we launch two workers for the same message, we can simply verify on `MessageSender` if the message is already marked as sent.

> Tests
I'm adding tests soon in this PR.

trOnk12 (approved):

vitorhugods:
I've noticed that writing tests for this is impossible right now without refactoring `UserSessionScope`.

I'll see about that _soon_. Manual tests have worked on Reloaded have worked.

PR #444, opened 2022-04-27
feat: Add current session flow use case

borichellow:
# What's new in this PR?

- Added `CurrentSessionFlowUseCase`, to have an option to get `CurrentSession` not just once but a flow that will emit every change (login, logout, etc.). 
- Added unit test for CurrentSessionFlowUseCase
- Updated unit tests for `SessionStorage` to cover a new feature


logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreFailure.kt
```diff
@@ -96,3 +99,12 @@ internal inline fun <T : Any> wrapStorageRequest(storageRequest: () -> T?): Eith
...
     }
 }
+
+internal inline fun <T : Any> wrapFlowStorageRequest(storageRequest: () -> Flow<T?>): Flow<Either<StorageFailure, T>> {
+    return try {
+        storageRequest().map { it?.let { data -> Either.Right(data) } ?: Either.Left(StorageFailure.DataNotFound) }
+    } catch (e: Exception) {
+        kaliumLogger.e(e.stackTraceToString())
+        flowOf(Either.Left(StorageFailure.Generic(e)))
+    }
```
saleniuk, line 109: There is a `catch` operator so I'm wondering if something like that would work:
```
return storageRequest()
    .map { it?.let { data -> Either.Right(data) } ?: Either.Left(StorageFailure.DataNotFound) }
    .catch { e -> 
        kaliumLogger.e(e.stackTraceToString())
        emit(Either.Left(StorageFailure.Generic(e)))
     }
```
this way we don't create a separate Flow in case of exception 😄

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/client/SessionStorage.kt
```diff
@@ -47,7 +56,9 @@ interface SessionStorage {
 class SessionStorageImpl(
     private val kaliumPreferences: KaliumPreferences
 ) : SessionStorage {
-    override fun addSession(persistenceSession: PersistenceSession) =
+    private val sessionWasChangedFlow = MutableSharedFlow<Unit>(1).also { it.tryEmit(Unit) }
```
saleniuk, line 59: I'm thinking if it wouldn't be better to create such Flow solution in `KaliumPreferences` and add for instance `getStringFlow` and `getSerializableFlow` there because if these preferences would be used and updated also outside the `SessionStorage` then we wouldn't get such update here 🤔

borichellow:
Closing after conversation with @MohamadJaara , as this Settings will be refactored soon

PR #449, opened 2022-04-27
fix: GetNotificationsUseCase null date issue

borichellow:
# What's new in this PR?

### Issues

1. Clear install app 
2. Login
3. wait until some conversations were fetched
4. Send a message to that user

Result: Notification doesn't come until user opens that conversation at least once

### Causes (Optional)

When the conversation is just fetched from the server and wasn't opened by user yet `last_notified_message_date = null` and  GetNotificationsUseCase can't get any messages from the DB, cause the "after date" in `Messages.selectMessagesByConversationIdAfterDate` is an empty string.


persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/ConversationDAOTest.kt
```diff
@@ -139,6 +139,17 @@ class ConversationDAOTest : BaseDatabaseTest() {
         assertEquals(ConversationEntity.MutedStatus.ONLY_MENTIONS_ALLOWED, result?.mutedStatus)
     }
 
+    @Test
+    fun givenAnExistingConversation_ThenConversationsForNotificationCanBeRetrieved() = runTest {
```
vitorhugods, line 143:
```suggestion
    fun givenMultipleConversations_whenGettingConversationsForNotifications_thenOnlyUnnotifiedConversationsAreReturned() = runTest {
```

vitorhugods (changes requested):
Just a tiny change to make the test name a bit easier to understand.

vitorhugods (approved):

PR #451, opened 2022-04-28
chore: add avs send http response message

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We were sending messages/notifying the other user via proteus/normal request but not also via AVS.

### Solutions

Call to `wcall_resp` on AVS to notify of any calling change.

### DIFF
--- a/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
+++ b/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
@@ -221,7 +221,7 @@ actual class CallManagerImpl(
                         is Either.Left -> {
                             wcall_resp(
                                 inst = deferredHandle.await(),
-                                status = 400,
+                                status = 400, // TODO: Handle the errorCode from CoreFailure
                                 reason = "Couldn't send Calling Message",
                                 arg = context
                             )
--- a/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
+++ b/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
@@ -103,13 +103,14 @@ actual class CallManagerImpl(
             readyHandler = { version: Int, arg: Pointer? ->
                 callingLogger.i("$TAG -> readyHandler")
             },
-            sendHandler = { _, conversationId, avsSelfUserId, avsSelfClientId, _, _, data, _, _, _ ->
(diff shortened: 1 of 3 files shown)


logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
```diff
@@ -198,7 +198,44 @@ actual class CallManagerImpl(
...
+        scope.launch {
+            messageString?.let { message ->
+                withCalling {
+                    when (sendCallingMessage(conversationId, avsSelfUserId, avsSelfClientId, message)) {
+                        is Either.Right -> {
+                            wcall_resp(
+                                inst = deferredHandle.await(),
+                                status = 200,
+                                reason = "",
+                                arg = null
```
ohassine, line 216: We need to pass something here not null, we get it from sendHandler (first parameter)

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
```diff
@@ -198,7 +198,44 @@ actual class CallManagerImpl(
...
+                            )
+                        }
+                        is Either.Left -> {
+                            wcall_resp(
+                                inst = deferredHandle.await(),
+                                status = 400,
+                                reason = "Couldn't send Calling Message",
+                                arg = null
```
ohassine, line 224: We need to pass the correct error code here, it's not always 400.
and same thing for the arg, a value is needed

ohassine, line 224: We need to pass the correct error code here, it's not always 400.
and same thing for the arg, a value is needed

ohassine (changes requested):
That's something we have missed 👍 . overall looks good ✅
I've just requested some small changes regarding some required params

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
alexandreferris: Added a TODO so we can handle the errorCode later from `CoreFailure`

alexandreferris: Added a TODO so we can handle the errorCode later from `CoreFailure`

ohassine (approved):
lgtm ✅

PR #455, opened 2022-04-29
fix(network): handle unknown device class type when getting clients of users

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When getting list of clients from users, we were expecting the `class` to be always present. Leading to crashes.
But the API doesn't always incldue that data.

### Solutions

When nothing comes up, replace with "unknown".

#### Test Coverage (Optional)

### Notes

We currently' don't use this field for anything. So just putting in a String is fine.
Once we start using it, we should probably replace it with an enum or similar.

----

----
##### References

### DIFF
--- a/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/id/IdMapperTest.kt
+++ b/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/id/IdMapperTest.kt
@@ -1,5 +1,6 @@
 package com.wire.kalium.logic.data.id
 
+import com.wire.kalium.network.api.user.client.DeviceTypeDTO
 import com.wire.kalium.network.api.user.client.SimpleClientResponse
 import kotlin.test.Test
 import kotlin.test.assertEquals
@@ -30,7 +31,7 @@ class IdMapperTest {
 
     @Test
     fun givenASimpleClientResponse_whenMappingFromSimpleClientResponse_thenTheIDShouldBeMappedCorrectly() {
-        val simpleClientResponse = SimpleClientResponse("an ID", "it doesn't matter")
+        val simpleClientResponse = SimpleClientResponse("an ID", DeviceTypeDTO.Desktop)
 
         val clientID = idMapper.fromSimpleClientResponse(simpleClientResponse)
 
--- a/network/src/commonMain/kotlin/com/wire/kalium/network/api/user/client/ClientResponse.kt
+++ b/network/src/commonMain/kotlin/com/wire/kalium/network/api/user/client/ClientResponse.kt
@@ -38,6 +38,6 @@ data class ClientsOfUsersResponse(
 @Serializable
 data class SimpleClientResponse(
     @SerialName("id") val id: String,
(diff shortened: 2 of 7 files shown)


network/src/commonMain/kotlin/com/wire/kalium/network/api/user/client/ClientResponse.kt
```diff
@@ -38,6 +38,6 @@ data class ClientsOfUsersResponse(
 @Serializable
 data class SimpleClientResponse(
     @SerialName("id") val id: String,
-    @SerialName("class") val deviceClass: String
+    @SerialName("class") val deviceClass: String? = "unknown"
```
MohamadJaara, line 41: 1. There is already an enum `DeviceTypeDTO` for client class check `ClientResponse` 
2. do we need to default it to unknown, or we can leave it as null since I assume it's not used?

vitorhugods: > 1. There is already an enum DeviceTypeDTO for client class check ClientResponse

Oooh. Nice. Replacing with `DeviceTypeDTO`.


> 2. do we need to default it to unknown, or we can leave it as null since I assume it's not used?


I think defaulting to `Unknown` makes the fact that we tried, but the backend didn't answer us (or answered with something we don't understand) a bit more clear.

If the backend returns `spaceship`, it would be a bit weird to store it as `null`, also throwing an exception for something so little could be over-reacting.

Does it make sense?

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/id/IdMapperTest.kt
```diff
@@ -30,7 +31,7 @@ class IdMapperTest {
 
     @Test
     fun givenASimpleClientResponse_whenMappingFromSimpleClientResponse_thenTheIDShouldBeMappedCorrectly() {
-        val simpleClientResponse = SimpleClientResponse("an ID", "it doesn't matter")
+        val simpleClientResponse = SimpleClientResponse("an ID", DeviceTypeDTO.Desktop)
```
vitorhugods, line 34: It matters now 😄

network/src/commonMain/kotlin/com/wire/kalium/network/api/user/client/ClientResponse.kt
```diff
@@ -38,6 +38,6 @@ data class ClientsOfUsersResponse(
 @Serializable
 data class SimpleClientResponse(
     @SerialName("id") val id: String,
-    @SerialName("class") val deviceClass: String
+    @SerialName("class") val deviceClass: DeviceTypeDTO? = DeviceTypeDTO.Unknown
```
MohamadJaara, line 41: We only need to do the same for `ClientDTO` since it can be null over there as well

MohamadJaara (approved):
✅ only one comment about making the client class nullable in `ClientDTO`

PR #457, opened 2022-05-02
feat: Added SyncPendingEventsUseCase

borichellow:
# What's new in this PR?

Added SyncPendingEventsUseCase
It's needed for 1-time syncing of all the pending events, when the app receives FCM.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncPendingEventsUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncPendingEventsUseCase.kt
@@ -16,28 +16,25 @@ class SyncPendingEventsUseCase(
 ) {
 
     /**
-     * @return Flow<Unit> that emits only once - when syncing Pending Events was done
+     * Syncing only Pending Events, to find out what did we miss
      */
-    suspend operator fun invoke(): Flow<Unit> {
+    suspend operator fun invoke() {
         syncManager.waitForSlowSyncToComplete()
 
-        return flow {
-            eventRepository.pendingEvents()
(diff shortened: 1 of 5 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncPendingEventsUseCase.kt
```diff
@@ -0,0 +1,43 @@
...
+import kotlinx.coroutines.flow.onCompletion
+
+class SyncPendingEventsUseCase(
+    private val syncManager: SyncManager,
+    private val eventRepository: EventRepository,
+    private val conversationEventReceiver: EventReceiver<Event.Conversation>
+) {
+
+    /**
+     * @return Flow<Unit> that emits only once - when syncing Pending Events was done
```
vitorhugods, line 19: It doesn't actually emit anything.

It only closes when it's done, without emitting.

```kt
syncPendingEvents().collect {
    println("Collected event") // This will never happen
}

syncPendingEvents().onCompletion {
    println("Completed flow") // This will be called
}.collect()
```

So the docs are a bit misleading.

I suggest just changing it to a `suspend` function, without returning anything.

vitorhugods (changes requested):
`SyncPendingEventsUseCase` is misleading right now. It doesn't need to be a Flow.

I guess it's the same for `StartListeningToEventsUseCase`, but that was originally done with the intention of exposing the events in the future, but I'm not sure that it will happen.

Perhaps we can improve here already, as I re-work the Sync mechanism.

vitorhugods (approved):
✅✅

PR #473, opened 2022-05-04
refactor: call Manager

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Everything related to calling was being placed in `CallManagerImpl` thus generating a very huge file.

### Solutions

We have separated almost all callbacks and handlers of calling into their own file. We also have a `CallRepository` that is now holding a `CallProfile`, where we are "storing" the calls we received or created for now.

#### Test Coverage (Optional)

Unit tests will be added to callsbacks/handlers in the future.


logic/src/jvmMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
```diff
@@ -9,6 +9,8 @@ import com.wire.kalium.logic.kaliumLogger
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 
+val calls = MutableStateFlow(listOf<Call>())
```
vitorhugods, line 12: How can these `calls` be static?

What if there are two users logged in?

ohassine: That's right, I missed that point

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/UpdateCallStatusById.kt
```diff
@@ -0,0 +1,33 @@
+package com.wire.kalium.logic.data.call
+
+import com.wire.kalium.logic.data.id.toConversationId
+import com.wire.kalium.logic.feature.call.Call
+import com.wire.kalium.logic.feature.call.CallStatus
+import com.wire.kalium.logic.feature.call.calls
+import kotlinx.coroutines.flow.update
+
+//TODO create unit test for this one in a different PR
+class UpdateCallStatusById {
```
alexandreferris, line 10: Maybe we could rename this to something like `CallStateManager` as it is only handling the call state

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
```diff
@@ -5,9 +5,12 @@ import com.wire.kalium.logic.data.call.ConversationType
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.message.Message
 import com.wire.kalium.logic.data.message.MessageContent
+import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.map
 
+val calls = MutableStateFlow(listOf<Call>())
```
alexandreferris, line 12: it can't be static here as well

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/UpdateCallStatusById.kt
ohassine: I don't really like to use some general names like `Manager`, `Handler` etc...It breaks the single responsibility  principle, I want to be more specific  : Manager could be creating, deleting or updating.
Also by using `UpdateCallStatusById` as a name, you don't have to open the file and see the what it does.

vitorhugods: I agree with @ohassine on avoiding generic names like Manager and Handler.

But when you have a instance of `UpdateCallStatusById`, how can you name it?

```kotlin
val update = UpdateCallStatusById()
update.updateById(id) { /** **/ }
```

Seems weird. 


When we name classes, we don't name them the same as a function, we name the classes and objects for what they are.

It is **not** an `Update`. It is an `Updater`. It performs updates:

```kotlin
val callUpdater = CallUpdater()
callUpdater.updateById(id) { /** **/ }
```


A silly comparison:

```kotlin
interface Cat {
    fun meow()
}
```

Instead of:
```kotlin
interface Meow {
    fun meow()
}
```

vitorhugods: The more I look at this, the more I want to put this logic inside `CallRepository`.

Because to be fair, this is what it is.
Everywhere else we are using Repositories for this.

Just because this is stored in memory instead of DB, SharedPreferences or API, doesn't mean it can't be dealt with in a Repository.

Bonus: we have plenty of tested Repositories already, so we know how to do it. And we can share for all platforms.

ohassine: Totally agree, Thanks 👌

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
```diff
@@ -26,4 +60,76 @@ internal class CallDataSource(
...
+        conversationId: ConversationId,
+        userId: UserId,
+        clientId: ClientId,
+        data: String
+    ): Either<CoreFailure, Unit> {
+        val messageContent = MessageContent.Calling(data)
+        val date = Clock.System.now().toString()
+        val message = Message(uuid4().toString(), messageContent, conversationId, date, userId, clientId, Message.Status.SENT)
+        return messageSender.sendMessage(message)
+    }
```
vitorhugods, line 88: `MessageSender` is on the domain layer, above the repositories, together with the `UseCases` and `CallManager`.

This business logic of getting the correct dates and creating a message with a generated UUID, which is only being used in `OnHttpRequest` could be left there.

Not really worth moving this logic down to the repo risking some circular dependency in the future and whatnot.

TL;DR: I suggest inlining this code inside `OnHttpRequest`, and don't depend on `MessageSender` in any repository.

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnConfigRequest.kt
```diff
@@ -0,0 +1,43 @@
...
+//TODO create unit test
+class OnConfigRequest(
+    private val calling: Calling,
+    private val callRepository: CallRepository
+) : CallConfigRequestHandler {
+    override fun onConfigRequest(inst: Handle, arg: Pointer?): Int {
+        callingLogger.i("${CallManagerImpl.TAG} - onConfigRequest")
+
+        //TODO use the same coroutine job of call manager
+        CoroutineScope(Dispatchers.IO).launch {
```
vitorhugods, line 24:
```suggestion
        callingScope.launch {
```

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnConfigRequest.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.wire.kalium.logic.feature.call.AvsCallBackError
+import com.wire.kalium.logic.feature.call.CallManagerImpl
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+
+//TODO create unit test
+class OnConfigRequest(
+    private val calling: Calling,
+    private val callRepository: CallRepository
```
vitorhugods, line 18:
```suggestion
    private val callRepository: CallRepository,
    private val callingScope: CoroutineScope
```

Maybe we can tackle this now before moving forward?

For testing, we can pass `TestScope`.

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnConfigRequest.kt
```diff
@@ -0,0 +1,43 @@
...
+            val config = callRepository.getCallConfigResponse(limit = null)
+                .fold({
+                    TODO("")
+                }, {
+                    it
+                })
+
+            calling.wcall_config_update(
+                inst = inst,
+                error = 0, // TODO: http error from internal json
```
vitorhugods, line 34: What does this `http error form interl json` mean?

vitorhugods (changes requested):
Nice refactoring in general, testable and atomic! Now we can make sure these components work :)

Just some tiny suggestions.

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnConfigRequest.kt
alexandreferris: This is the http error code we need to get from `callApi.getCallConfig(..)`, we will then need to map it from `CoreFailure` to something from `NetworkResponse.Error` and try to get the errorCode. Which is something I think we can leave for later, at least for now until we get most of the basics of calling in place.

vitorhugods (approved):
👨🏼‍🍳🤌🏻

alexandreferris:
> 👨🏼‍🍳🤌🏻

grazie bambino!

PR #480, opened 2022-05-06
fix: contacts sync feature was overriding connection status (AR-1570)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

https://wearezeta.atlassian.net/browse/AR-1570

### Issues

Currently, when we were doing a synchronization of the users ids (endpoint `/list-users`) we were overriding it's connection status.

As context, the sync process happens in the order describer here:

https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/15566946/Full+state+synchronization

### Causes (Optional)

Using `INSERT OR REPLACE INTO` was causing to override data that was not provided again, or in fact provided with the default values again, see more at https://www.sqlite.org/lang_conflict.html

### Solutions

As we in the previous steps of the sync already have the user's persisted, we can perform an update with the new data only.
As a fallback if the update detected no records, will insert the new data

----

 - [X] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----

----
##### References

### DIFF
--- a/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAOImpl.kt
+++ b/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAOImpl.kt
@@ -71,6 +71,21 @@ class UserDAOImpl(private val queries: UsersQueries) : UserDAO {
         queries.transaction {
             users.forEach { user ->
                 queries.updateUser(user.name, user.handle, user.email, user.accentId, user.previewAssetId, user.completeAssetId, user.id)
+                val recordDidNotExist = queries.selectChanges().executeAsOne() == 0L
+                if (recordDidNotExist) {
+                    queries.insertUser(
+                        user.id,
+                        user.name,
+                        user.handle,
+                        user.email,
+                        user.phone,
+                        user.accentId,
+                        user.team,
+                        user.connectionStatus,
+                        user.previewAssetId,
+                        user.completeAssetId
+                    )
+                }
             }
         }
     }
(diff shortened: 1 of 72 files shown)


persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAOImpl.kt
```diff
@@ -67,6 +67,29 @@ class UserDAOImpl(private val queries: UsersQueries) : UserDAO {
         queries.updateUser(user.name, user.handle, user.email, user.accentId, user.previewAssetId, user.completeAssetId, user.id)
     }
 
+    override suspend fun updateUsers(users: List<UserEntity>) {
+        queries.transaction {
+            users.forEach { user ->
+                queries.updateUser(user.name, user.handle, user.email, user.accentId, user.previewAssetId, user.completeAssetId, user.id)
+                val recordDidNotExist = queries.selectChanges().executeAsOne() == 0L
```
gongracr, line 74: Doesn't `selectChanges()` return the number of rows modified in the previous sql statement? what is then `executeAsOne()` statement doing?

gongracr:
Looking good. Could we add some tests to verify the expected behavior at least?

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAOImpl.kt
yamilmedina: This `selectChanges()` in effect returns the rows affected by the table operation, the `.executeAsOne()` function allow us to get the result, since the result of a query it's a raw `Query` result, for example when we return the results from a query we usually use `.asFlow()` to get a more manageable type

PS: I added the tests that you suggested =)

gongracr (approved):
Thanks for the tests 🙏  Although in the near future it would be great to have some more complex integration tests, testing the UserRepository class, and making sure `fetchUsersByIds(ids)` is not overriding the previously inserted values with the default ones.
Great job as usual 🙌 🍄

PR #485, opened 2022-05-10
feat: add log writer to the kermit config [AR-1450]

MohammedMokresh:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

without having the log writer in kermit we don't have a way to intercept the logs that been logged in kermit 

### Solutions

add the support to add a custom log writer to kermit to support writing to file 

----

----

----
##### References


logger/build.gradle.kts
```diff
@@ -43,7 +43,7 @@ kotlin {
     sourceSets {
         val commonMain by getting {
             dependencies {
-                implementation("co.touchlab:kermit:1.0.0")
+                api("co.touchlab:kermit:1.1.1")
```
alexandreferris, line 46: Why the change from `implementation` to `api`?

alexandreferris (commented):
Great work! I just have one small question tho 🤔

logger/build.gradle.kts
yamilmedina: I think it's related to having this lib automatically imported on the consumers projects. 

I'm wondering if this could cause a dependency conflict in the future, in case another lib would import this `kermit` lib transitively 🤔

yamilmedina (approved):
Looks good to me, just a small question that jumped to my head on the possibility to have deps conflicts in the future, but I think you could decide the best approach 🙌  right now I don't see an issue leaving it as api =)

logger/build.gradle.kts
MohammedMokresh: because we are using it in the reloaded as we need to build the log writer from the library , and in theory the other clients will do the same to create a custom log writer

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/EnableLoggingUseCase.kt
```diff
@@ -0,0 +1,15 @@
+package com.wire.kalium.logic.feature.user
+
+import com.wire.kalium.logic.configuration.UserConfigRepository
+
+
+class EnableLoggingUseCase(
```
yamilmedina, line 6: Could we extract an interface out of this UseCase ? It will be easy I think in the future for espresso testing in case we want to create Fakes.

You can see an example on `UploadUserAvatarUseCase`

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/IsLoggingEnabledUseCase.kt
```diff
@@ -0,0 +1,28 @@
+package com.wire.kalium.logic.feature.user
+
+import com.wire.kalium.logic.StorageFailure
+import com.wire.kalium.logic.configuration.UserConfigRepository
+import com.wire.kalium.logic.kaliumLogger
+
+class IsLoggingEnabledUseCase(
```
yamilmedina, line 7: Could we extract an interface out of this UseCase ? It will be easy I think in the future for espresso testing in case we want to create Fakes.

You can see an example on `UploadUserAvatarUseCase`

yamilmedina (commented):
Great work, Big Mo! 🪄  Just left some comments about having the Use Cases behind an interface, so We could mock/fake them later for espresso testing 🚀

yamilmedina (approved):
Good work! thanks for applying the changes 🪄 🚀

PR #487, opened 2022-05-10
feat: Improve getIncomingCallsUsecase to have more info

borichellow:
### Issues

existed `Call` data class contains not enough info for display push notifications on IncommingCalls

### Solutions

added `LocalNotificationCall` (just like `LocalNotificationConversation`) - the data class that contains all the necessary info.
and updated `GetIncomingCallsUsecase` to map generic `Call` into the `LocalNotificationCall`

### Also
added call cancelation on Rejecting the call

### Note

this change shouldn't break anything in Reloaded project, so can be merged at any time


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/GetIncomingCallsUseCase.kt
```diff
@@ -1,22 +1,52 @@
...
 internal class GetIncomingCallsUseCaseImpl(
     private val callManager: CallManager,
-    private val syncManager: SyncManager
+    private val syncManager: SyncManager,
+    private val conversationRepository: ConversationRepository
 ) : GetIncomingCallsUseCase {
 
-    override suspend operator fun invoke(): Flow<List<Call>> {
+    //TODO update UnitTests after fixing all the other TODOs
+    override suspend operator fun invoke(): Flow<List<LocalNotificationCall>> {
```
vitorhugods, line 26: Just a reminder that this change will require Reloaded to be updated so it doesn't break.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/LocalNotification.kt
```diff
@@ -1,9 +1,21 @@
...
- * Kalium local data classes that contains all the necessary data for displaying Notifications,
+ * Kalium local data classes that contains all the necessary data for displaying Call Notification, or display call itself
+ * and suppose to be mapped (in platform side) into platform-specific objects to show the notification, or call
+ */
+data class LocalNotificationCall(
+    val conversationId: ConversationId,
+    val status: CallStatus,
+    val notificationTitle: String,
+    val notificationBody: String?
+)
```
vitorhugods, line 15: Does it need to be named `LocalNotification`?

What if we don't want to display notifications but just display the items in a list?

Maybe just `DetailedCall` or `CallDetails` is good enough and represents it better.

How the notifications will be displayed, what goes in the title, what goes in the body, or whatever, is responsibility of the app. So perhaps `Kalium` can leave it to Reloaded or other apps to figure out. Maybe provide some helpers as functions, but maybe not set in stone like this.


My suggestion:

```suggestion
data class LocalNotificationCall(
    val conversationId: ConversationId,
    val status: CallStatus,
    val notificationTitle: ConversationDetails
)
```

And do the mapping to a `CallNotification` on Reloaded. This way we let the consumers of Kalium use all the `ConversationDetails` data to display a notification.

vitorhugods (changes requested):
Nice change, I just think Kalium shouldn't be so opinionated about how things are displayed on screen.

I left some suggestions here to make it more generic and flexible to whoever is calling this UseCase.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/LocalNotification.kt
ohassine: Why adding a new data class for notification ? Adding a new fields for `callerName` and `callMetaData` to `Call` data class isn't enough ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/GetIncomingCallsUseCase.kt
ohassine: exactly, this change will break the feature of incoming call

borichellow: actually no, this won't break anything in Reloaded. Cause Kotlin will replace object type from `Call` to the new one by itself and there are the same filed names (`conversationId` and `status`).

vitorhugods (approved):
🚀✅

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/GetIncomingCallsUseCase.kt
```diff
@@ -1,21 +1,59 @@
...
     }
 }
+
+data class IncomingCallData(
+    val conversationId: ConversationId,
+    val status: CallStatus,
+    val conversationDetails: ConversationDetails,
+    val callerName: String,
+    val callerTeamName: String?
+)
```
ohassine, line 59: Then we will need to create a separat data class for `outgoingCall` and `initiatedCall` ?

I would suggest to do it in different way, instead of fetching call details every time we invoke the incomingCallUseCase or whatever, we can can do that while creating the call object. This way Call object will hold of informations needed to be displayed

In the **callManagerRefactor** PR, we extracted that part of creating the call so you can do it there.

ohassine (changes requested):

ohassine (approved):
👌👌

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/Call.kt
```diff
@@ -16,6 +19,9 @@ data class Call(
     val conversationId: ConversationId,
     val status: CallStatus,
     val callerId: String,
+    val conversationDetails: ConversationDetails,
+    val caller: OtherUser?,
```
ohassine, line 23: I think we don't need to pass all of these data here, just the needed values for calling should be enough

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/call/usecase/GetIncomingCallsUseCaseTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private lateinit var getIncomingCallsUseCase: GetIncomingCallsUseCase
+
+    @BeforeTest
+    fun setUp() {
+        getIncomingCallsUseCase = GetIncomingCallsUseCaseImpl(callRepository, syncManager)
+    }
+
+    @Test
+    fun givenEmptyCallList_thenEmptyNotificationList() = runTest {
```
ohassine, line 43: This is a use case for getting incoming calls, it has nothing related to notification.
Please follow given..when..then pattern for test cases

`givenAnEmptyCallList_whenInvokingGetIncomingCallsUseCase_thenEmitsAnEmptyListOfCalls`

ohassine (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
```diff
@@ -63,7 +73,18 @@ internal class CallDataSource(
...
+            .getConversationDetailsById(conversationId)
+            .first()
+
+        val caller = userRepository.getKnownUser(callerId.toUserId()).first()
+
+        val team = caller?.team
+            ?.let { teamId -> teamRepository.getTeam(teamId).first() }
+
+        val call = Call(conversationId, status, callerId, conversation, caller, team)
+
```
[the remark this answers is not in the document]
borichellow: yeah, always forget about tests :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/Call.kt
borichellow: okay, make sense

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/call/CallRepositoryTest.kt
```diff
@@ -66,63 +93,100 @@ class CallRepositoryTest {
...
-        assertEquals(2, calls.value.size)
-        assertEquals(calls.value[0], startedCall)
-        assertEquals(calls.value[1], answeredCall)
+
+        calls.test {
+            val list = awaitItem()
+            assertEquals(2, list.size)
+            assertEquals(list[0], startedCall)
+            assertEquals(list[1], answeredCall)
+        }
```
ohassine, line 139: we should verify that `getConversationDetailsById`, `getKnownUser` and `getTeam` are being invoked once

ohassine, line 139: we should verify that `getConversationDetailsById`, `getKnownUser` and `getTeam` are being invoked once

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/call/CallRepositoryTest.kt
```diff
@@ -66,63 +93,100 @@ class CallRepositoryTest {
...
+        callRepository.createCall(sharedConversationId, CallStatus.INCOMING, "caller_id")
 
         val calls = callRepository.callsFlow()
-        assertEquals(mapOfCallProfiles.size, calls.value.size)
-        assertEquals(calls.value[0], incomingCall2)
+        calls.test {
+            val list = awaitItem()
+            assertEquals(mapOfCallProfiles.size, list.size)
+            assertEquals(list[0], incomingCall2)
+        }
```
ohassine, line 164: same here...
we should verify that getConversationDetailsById, getKnownUser and getTeam are being invoked once

ohassine, line 164: same here...
we should verify that getConversationDetailsById, getKnownUser and getTeam are being invoked once

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
```diff
@@ -63,7 +73,26 @@ internal class CallDataSource(
...
+        val call = Call(
+            conversationId = conversationId,
+            status = status,
+            callerId = callerId,
+            conversationName = conversation.conversation.name,
+            conversationType = conversation.conversation.type,
+            callerName = caller?.name,
+            callerTeamName = team?.name
+        )
+
```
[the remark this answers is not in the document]
borichellow: in that case `callerName` and `callerTeamName` are `null`, those fields are nullable in `Call`

ohassine (approved):
lgtm ✅ Great work  🚀

PR #497, opened 2022-05-13
chore: add sender id to server config

MohammedMokresh:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

add sender Id temporarily to server config ,so the staging env can receive a notifications 

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/ServerConfig.kt
```diff
@@ -15,7 +15,8 @@ data class ServerConfig(
     val blackListUrl: String,
     val teamsUrl: String,
     val websiteUrl: String,
-    val title: String
+    val title: String,
+    val androidSenderId: String?
```
vitorhugods, line 19: Why is it called `androidSenderId` ?

What happens if I register a client on JVM, for example on the `cli` module?

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/ServerConfig.kt
```diff
@@ -26,7 +27,8 @@ data class ServerConfig(
             teamsUrl = """https://teams.wire.com""",
             blackListUrl = """https://clientblacklist.wire.com/prod""",
             websiteUrl = """https://wire.com""",
-            title = "Production"
+            title = "Production",
+            androidSenderId = "782078216207"
```
vitorhugods, line 31: I feel like this shouldn't be the default, and that Reloaded should do something like this instead:

```kotlin
val prodConfig = PRODUCTION.copy(senderId = "782078216207")
val stagingConfig = STAGING.copy(senderId = "723990470614")

// and then call the usecases:


login(user, password, prodConfig)
```

Or even better, we should make `PRODUCTION` and `STAGING` private and expose a `fun createProduction(senderId: String)` and `fun createStaging(senderId: String)`.

vitorhugods (changes requested):
It kinda smells a bit too much:
- Introducing a new, temporary field, that should already be marked as `@Deprecated` from day one;
- It's a platform-specific thing being mixed with all the common code.
- It's a client-specific thing being glued into the core of a agnostic SDK.

As we discussed last week, I know it's not on you @MohammedMokresh, this is just a "temporary" fix until we sort our Firebase projects and the backend and other clients adjust.

But there's nothing more permanent than a temporary fix, so we better minimise the impact of this.


So I've left some thoughts in there on what we can do to improve this, in case it stays with us for longer than expected.

PR #498, opened 2022-05-13
feat: observe server config use case

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

1. there is no usecase to expose a list of stored server config.
2. the server API version need to be calculated even for hard-coded backends (prod, staging …)

### Solutions

1. add `ObserveServerConfigUseCase` that return a `Flow<List<ServerConfig>>` and in case of not config stored it will store any hard-coded configs (atm it will store only PROD)
2. simply `UpdateApiVersionsWorker` and `UpdateApiVersionsUseCase` there is no need to return errors since it supposed to run the update and store the new commonApiVersion locally 

Needs releases with:

### Testing

WIP
#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/UpdateApiVersionsWorker.kt
```diff
@@ -1,32 +1,17 @@
...
-            }
-            is UpdateApiVersionsResult.Failure -> {
-                kaliumLogger.e("UPDATE API VERSIONS FAILURE ${result.genericFailure}")
-                (result.genericFailure as? CoreFailure.Unknown)?.let {
-                    it.rootCause?.printStackTrace()
-                }
-                apiVersionCheckManager.changeState(ApiVersionCheckState.Failed(result.genericFailure))
-                Result.Retry // TODO should it retry or just fail?
-            }
+        return updateApiVersionsUseCase().let { Result.Success }.also {
```
saleniuk, line 13: What if we want to schedule the immediate work to be started when the app is launched just like the slow sync? Shouldn't we handle somehow the error and block the user from using the app without calculated common api version? Will such case be handled differently?

MohamadJaara: I was thinking about this case and that should be handled for each account separately because the update can success for account A put fail for account B.
we can either handle this case in the sync after (at the beginning of the sync, check the API version).
Or we can return a list of failed update attempts (this way)

PR #513, opened 2022-05-18
chore: detekt custom rules (AR-1681)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR adds custom rules set to detekt config, so we as "Wire" can create rules that make sense for us as a team.
The rule that it's added as a first step, is enforcing annotations on `@Serializable` classes, following a conversation with @MohamadJaara 

### Causes (Optional)

This will cause that the current build breaks as we need to use some annotations on fields when using `@Serializable` 

### Notes (Optional)

Later on, we could improve or add more rules for our conventions, maybe having implementations of use cases and repositories as internal, etc.

----

 - [X] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----

----
##### References

### DIFF
--- a/detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
+++ b/detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
@@ -25,8 +25,8 @@ class EnforceSerializableFields(config: Config = Config.empty) : Rule(config) {
     )
 
     override fun visitClassOrObject(kClass: KtClassOrObject) {
-        for (superEntry in kClass.annotationEntries) {
-            if (isSerializableAnnotationWithoutArgs(superEntry)) {
+        for (ktAnnotationEntry in kClass.annotationEntries) {
+            if (isSerializableAnnotationWithoutArgs(ktAnnotationEntry)) {
                 getDeclarationsOfInterestForClass(kClass).forEach { annotatedValue ->
                     val hasMatchingRequirement = annotatedValue.annotationEntries.any { annotation ->
                         val annotationName = annotation.shortName.toString()
--- a/detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
+++ b/detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
@@ -23,8 +23,9 @@ class EnforceSerializableFields(config: Config = Config.empty) : Rule(config) {
     override fun visitClassOrObject(kClass: KtClassOrObject) {
         for (superEntry in kClass.annotationEntries) {
             if (superEntry.text?.startsWith(SERIALIZABLE_CLASS_ANNOTATION) == true) {
-                kClass.getValueParameters().forEach { parameter ->
-                    val hasMatchingRequirement = parameter.annotationEntries.any { annotation ->
+                getDeclarationsOfInterestForClass(kClass).forEach { annotatedValue ->
+                    println(annotatedValue.name)
+                    val hasMatchingRequirement = annotatedValue.annotationEntries.any { annotation ->
                         val annotationName = annotation.shortName.toString()
                         annotationName == "SerialName" || annotationName == "Serializable" || annotationName == "Transient"
                     }
@@ -32,7 +33,7 @@ class EnforceSerializableFields(config: Config = Config.empty) : Rule(config) {
                         report(
                             kClass,
                             "The Serializable class '${kClass.name}' declares "
-                                    + "a field '${parameter.name}' without proper Serializable annotation "
+                                    + "a field '${annotatedValue.name}' without proper Serializable annotation "
                                     + "'@SerialName | @Serializable | @Transient'."
                         )
                     }
@@ -40,6 +41,13 @@ class EnforceSerializableFields(config: Config = Config.empty) : Rule(config) {
             }
         }
     }
+
+    private fun getDeclarationsOfInterestForClass(kClass: KtClassOrObject): List<KtAnnotated> =
+        if (kClass is KtClass && kClass.isEnum()) {
+            kClass.declarations.filterIsInstance<KtEnumEntry>()
+        } else {
+            kClass.getValueParameters()
+        }
 
     private fun report(classOrObject: KtClassOrObject, message: String) {
         report(CodeSmell(issue, Entity.atName(classOrObject), message))
--- a/detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
+++ b/detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
@@ -11,7 +11,7 @@ import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
 
 @Suppress("NestedBlockDepth")
-class EnforceSerializerFields(config: Config = Config.empty) : Rule(config) {
+class EnforceSerializableFields(config: Config = Config.empty) : Rule(config) {
 
     override val issue: Issue = Issue(
         javaClass.simpleName,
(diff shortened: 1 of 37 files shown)


detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.jetbrains.kotlin.psi.psiUtil.getValueParameters
+
+@Suppress("NestedBlockDepth")
+class EnforceSerializableFields(config: Config = Config.empty) : Rule(config) {
+
+    override val issue: Issue = Issue(
+        javaClass.simpleName,
+        Severity.Style,
+        "Use Serializable annotations on every field of the Class that use @Serializable.",
+        Debt.FIVE_MINS
```
vitorhugods, line 20:
```suggestion
        Debt(mins = DEBT_IN_MINUTES_PER_MISSING_ANNOTATION)
```

gradle/detekt.gradle
```diff
@@ -5,11 +5,18 @@ detekt {
     // allRules = false // activate all available (even unstable) rules.
     config = files("$projectDir/detekt/detekt.yml")
     input = files("$projectDir")
-    baseline = file("$projectDir/detekt/baseline.xml") // a way of suppressing issues before introducing detekt
+    baseline = file("$projectDir/detekt/baseline.xml")
```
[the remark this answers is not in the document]
yamilmedina: Yeah! I had it on the side (as the example from the detekt repo) and when applied formatting jumped to the bottom 🙃

network/src/commonMain/kotlin/com/wire/kalium/network/api/conversation/CreateConversationRequest.kt
```diff
@@ -40,6 +40,7 @@ data class CreateConversationRequest(
     val protocol: ConvProtocol?
 )
 
+@Suppress("EnforceSerializableFields")
 @Serializable(with = ReceiptMode.ReceiptModeAsIntSerializer::class)
 enum class ReceiptMode(val value: Int) {
```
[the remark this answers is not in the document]
vitorhugods: OOH.

It doesn't actually catch `enums`!

Take this for example:

```kotlin
@Serializable
enum class AssetSizeDTO {
    @SerialName("preview")
    PREVIEW,

    @SerialName("complete")
    COMPLETE;
}

AssetSizeDTO::class.getValueParameters() // empty list
```

detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
```diff
@@ -0,0 +1,52 @@
...
+                            kClass,
+                            "The Serializable class '${kClass.name}' declares "
+                                    + "a field '${parameter.name}' without proper Serializable annotation "
+                                    + "'@SerialName | @Serializable | @Transient'."
+                        )
+                    }
+                }
+            }
+        }
+    }
```
vitorhugods, line 42:
```suggestion
    override fun visitClassOrObject(kClass: KtClassOrObject) {
        for (superEntry in kClass.annotationEntries) {
            if (superEntry.text?.startsWith(SERIALIZABLE_CLASS_ANNOTATION) == true) {
                getDeclarationsOfInterestForClass(kClass).forEach { annotatedValue ->
                    println(annotatedValue.name)
                    val hasMatchingRequirement = annotatedValue.annotationEntries.any { annotation ->
                        val annotationName = annotation.shortName.toString()
                        annotationName == "SerialName" || annotationName == "Serializable" || annotationName == "Transient"
                    }
                    if (!hasMatchingRequirement) {
                        report(
                            kClass,
                            "The Serializable class '${kClass.name}' declares "
                                    + "a field '${annotatedValue.name}' without proper Serializable annotation "
                                    + "'@SerialName | @Serializable | @Transient'."
                        )
                    }
                }
            }
        }
    }

    private fun getDeclarationsOfInterestForClass(kClass: KtClassOrObject): List<KtAnnotated> =
        if (kClass is KtClass && kClass.isEnum()) {
            kClass.declarations.filterIsInstance<KtEnumEntry>()
        } else {
            kClass.getValueParameters()
        }
```

This should cover enums ;) 

After a few minutes playing around and adding logs, this seems to work!

yamilmedina: Wow! nice research 👏  I did the same, started playing with logs and debugging to get an idea where every field, annotation etc. can be grabbed :D

network/src/commonMain/kotlin/com/wire/kalium/network/api/conversation/CreateConversationRequest.kt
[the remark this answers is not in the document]
yamilmedina: I managed to get this working, too, with this line:

`kClass.annotationEntries.firstOrNull()?.getValueArgumentsInParentheses()?.isNotEmpty()`

We can have a look if the `@Serializable` annotation has arguments or not 🔬  which is the case for custom Serializers 🙌

detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
yamilmedina: Don't worry! I managed to fix it that locally ! And also added the checks in case we have Custom Serializers 👀

detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
```diff
@@ -42,12 +45,18 @@ class EnforceSerializableFields(config: Config = Config.empty) : Rule(config) {
         }
     }
 
-    private fun getDeclarationsOfInterestForClass(kClass: KtClassOrObject): List<KtAnnotated> =
-        if (kClass is KtClass && kClass.isEnum()) {
-            kClass.declarations.filterIsInstance<KtEnumEntry>()
+    private fun getDeclarationsOfInterestForClass(kClass: KtClassOrObject): List<KtAnnotated> {
+        return if (kClass is KtClass && kClass.isEnum()) {
+            if (kClass.annotationEntries.firstOrNull()?.getValueArgumentsInParentheses()?.isNotEmpty() == true) {
```
vitorhugods, line 50:
```kotlin
@Suppress("MemberVisibilityCanBePrivate")
@Serializable
enum class Whatever {
    FOO, BAR;
}
```

I think `annotationEntries.firstOrNull()` would return the `Suppress`, and not the `Serializable`.

So this wouldn't be right.


Also, I think the logic for custom serializable should be applied for classes too, not only for enums.


So, if we're going this way:

```kotlin
private fun getDeclarationsOfInterestForClass(serializableAnnotation: KtAnnotationEntry, kClass: KtClassOrObject): List<KtAnnotated> {
    if (serializableAnnotation.getValueArgumentsInParentheses().isNotEmpty()) {
        println("Ignoring internal entries for [${kClass.name}] since it has a custom [KSerializer] implementation")
        return emptyList()
    }
    ... // Resume with old logic here.
}
    

// Usage:

getDeclarationsOfInterestForClass(superEntry, kClass)

```

detekt-rules/src/main/java/com/wire/detekt/rules/EnforceSerializableFields.kt
```diff
@@ -25,8 +25,8 @@ class EnforceSerializableFields(config: Config = Config.empty) : Rule(config) {
     )
 
     override fun visitClassOrObject(kClass: KtClassOrObject) {
-        for (superEntry in kClass.annotationEntries) {
-            if (isSerializableAnnotationWithoutArgs(superEntry)) {
+        for (ktAnnotationEntry in kClass.annotationEntries) {
+            if (isSerializableAnnotationWithoutArgs(ktAnnotationEntry)) {
```
vitorhugods, line 29: Awesome 👏🏼

PR #538, opened 2022-05-25
feat: build configs [AR-1463]

MohammedMokresh:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
generate build configs from the local default.json or from remote repo if exist
----

----

----
##### References


vitorhugods:
I have some questions here. The PR template has a section called `solutions` that is quite helpful, so that could've prevented these:

## 1. In general, how to use it?

Is this a `Kalium` thing? Can we use it directly on Reloaded?
Do we need to rebuild Kalium if we want different feature flags? What 

## 2. What about multiple clients?

If I want to build Reloaded A and Reloaded B with different feature flags, what should I do?

## 3. Baseline.xml?

Why add _more_ stuff to the baseline? The idea behind this file was that we could introduce the style checker without it failing due to current issues.
After Detekt was added, we should only **remove** from the Baseline as we fix the problems, not introduce more.


----

In general, I was thinking that we could have a code / flags generator that reads from Json and exports as global stuff, however if we need to do this

vitorhugods:
In general, I am not sure how much it's Kalium's responsibility for reading JSON and turning them into flags. At least to some degree it can work (for Android and JVM for example), not sure about generating stuff that is accessible to Swift and JS.

At times, we might need config flags on each client too, as we want to hide some feature that is being developed, or some custom UI stuff that is app/platform-specific.

One thing I am certain: Kalium could and should have feature flags, so this is centralised across all platforms:

```kotlin
data class FeatureFlags(
    val isSSOEnabled: Boolean,
    val isSendingAssetsEnabled: Boolean
)

class ObserveFeatureFlagsUseCase {
    suspend operator fun invoke(): Flow<FeatureFlags>
}
```

The apps when building could read from a Json file, or whatever else they want, and when creating `CoreLogic`, could pass a `FeatureFlagsConfig`.

Maybe we could provide a Json parser, but I am not sure much we can do for all platforms.
Especially if this means re-building Kalium.

I am also not sure if iOS, JVM or even Android will consume Kalium as a git submodule or we will start deploying Kalium builds and version it once everything is stable.

MohammedMokresh:
@vitorhugods 
> ## 1. In general, how to use it?
As of this PR , it focus on generating the configs , about how to use it we can expose this submodule to the project that need to use either the logic module or the reloaded , but after reading your second comment , now I remembered that we have those features flag that comes from the endpoint , so I think this build config will only be exposed to logic and we merge it with the user flags and we have that use case which will be used in reloaded or other platforms 

> ## 2. What about multiple clients?
maybe I didn't understand properly here , but these configs are generated from the default.json , or from the remote if exist , so if you have another client that need different feature flags , you need to update those json files and build to generate the new flags 

> ## 3. Baseline.xml?
yeah this one on me , because I wanted to use the enum values as it is , I should just update to UPPER_CASE with name and use it , will update that

MohammedMokresh:
> In general, I am not sure how much it's Kalium's responsibility for reading JSON and turning them into flags. At least to some degree it can work (for Android and JVM for example), not sure about generating stuff that is accessible to Swift and JS.
> 
> At times, we might need config flags on each client too, as we want to hide some feature that is being developed, or some custom UI stuff that is app/platform-specific.


for a different client we can generate targeted build configs and maybe after that do like what you mentioned , create a use case for it and use it

PR #543, opened 2022-05-30
refactor: decuple accounts session from server metadata

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

AR-1464

### Solutions

1. decoupling `AuthSession` from server metadata.
2. using `WireDefaultRequest` ktor plugin to add API version to requests.
3. fetch and store server metadata only when needed.

Needs releases with:

### Testing

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/server/ServerConfig.kt
```diff
@@ -70,8 +69,12 @@ interface ServerConfigMapper {
     fun toDTO(links: ServerConfig.Links): ServerConfigDTO.Links
     fun toDTO(serverConfigEntity: ServerConfigEntity): ServerConfigDTO
     fun fromDTO(wireServer: ServerConfigDTO): ServerConfig
-    fun toEntity(backend: ServerConfig): ServerConfigEntity
+    fun toEntity(serverLinks: ServerConfig): ServerConfigEntity
+    fun toEntity(serverLinks: ServerConfig.Links): ServerConfigEntity.Links
+
     fun fromEntity(serverConfigEntity: ServerConfigEntity): ServerConfig
+    fun fromEntity(serverConfigEntityLinls: ServerConfigEntity.Links): ServerConfig.Links
```
saleniuk, line 76: typo "Linls" 😄

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/model/AuthSessionEntity.kt
```diff
@@ -0,0 +1,39 @@
...
+    )
+}
+
+@Serializable
+data class AuthSessionEntity(
+    @SerialName("user_id") val userId: QualifiedIDEntity,
+    @SerialName("token_type") val tokenType: String,
+    @SerialName("access_token") val accessToken: String,
+    @SerialName("refresh_token") val refreshToken: String,
+    @SerialName("wire_server") val serverLinks: ServerConfigEntity.Links
```
saleniuk, line 38: Maybe I'm missing something, but wouldn't be better to store `ServerConfigEntity.id` here instead of `ServerConfigEntity.Links`?

saleniuk: Ok, I get it, but now Links data is duplicated in both auth session and server config, wouldn't be better to keep both in respective tables, we could even keep Links and MetaData in separate tables and just operate on IDs? 
I don't know it that's possible but in case of needing to store the server config before doing any operation, we could store it already only in its proper table and keep only ID of such row in auth session data 🤔

MohamadJaara: I wanted to avoid having this connection between the ID (since it's something we added and not part of the server config) and the AuthSession

PR #544, opened 2022-05-30
feat: store system messages [AR-1490]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Conversation setting messages (system) need to be stored along with other ones in order to be able to sort and show them on conversation screen properly.

### Solutions

Added new field to messages table, added ColumnAdapter to decode and encode list of QualifiedIDEntity, added two new types of messages (MEMBER_JOIN, MEMBER_LEAVE), cleaned up models to to use proper ones on different layers (dao, api).

----

----

----
##### References


MohamadJaara (commented):
Great work, just comments about making senderClientId nullable

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ConversationEventReceiver.kt
```diff
@@ -109,17 +109,42 @@ class ConversationEventReceiverImpl(
...
-        ).onFailure { kaliumLogger.e("$TAG - failure on member join event: $it") }
+        )
+        .onSuccess {
+            val message = Message(
+                id = event.id,
+                content = MessageContent.MemberChange.Join(members = event.members),
+                conversationId = event.conversationId,
+                date = event.time,
+                senderUserId = event.addedBy,
+                senderClientId = ClientId(""), // TODO: should we keep it empty or pass some server data?
```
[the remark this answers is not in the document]
saleniuk: 👍  I changed it, in some places currently there is a need to use `!!` operator, let me know if that's fine

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq
```diff
@@ -51,8 +55,8 @@ SET visibility = 'DELETED', text_body = NULL
 WHERE id = ? AND conversation_id =? ;
 
 insertMessage:
-INSERT INTO Message(id, text_body, asset_mime_type, asset_size, asset_name, asset_image_width, asset_image_height, asset_otr_key, asset_sha256, asset_id, asset_token, asset_domain, asset_encryption_algorithm, asset_download_status, content_type, conversation_id, date, sender_user_id, sender_client_id, visibility, status)
-VALUES(?, ?, ?, ?, ?, ? ,?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+INSERT INTO Message(id, text_body, asset_mime_type, asset_size, asset_name, asset_image_width, asset_image_height, asset_otr_key, asset_sha256, asset_id, asset_token, asset_domain, asset_encryption_algorithm, asset_download_status, member_list, content_type, conversation_id, date, sender_user_id, sender_client_id, visibility, status)
```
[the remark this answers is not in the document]
saleniuk: I split this method into three, each one for every message type, I think that's better than maintaining the same `insertMessage` method for every type, but if you disagree just let me know, I'll revert it then

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq
```diff
@@ -67,18 +84,25 @@ asset_id = excluded.asset_id,
...
+INSERT INTO Message(id, member_list, content_type, conversation_id, date, sender_user_id, visibility, status)
+VALUES(?, ?, ?, ?, ?, ?, ?, ?)
+ON CONFLICT(id, conversation_id) DO UPDATE SET
+id = excluded.id,
+member_list = excluded.member_list,
+content_type = excluded.content_type,
+date = excluded.date,
+sender_user_id = excluded.sender_user_id,
+visibility = excluded.visibility,
+status = excluded.status;
```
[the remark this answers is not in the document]
MohamadJaara: ohhh my bad, this was a live demonstration of how this table is getting hard to maintain

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
```diff
@@ -43,16 +42,16 @@ sealed class Event(open val id: String) {
             override val id: String,
             override val conversationId: ConversationId,
             val addedBy: UserId,
-            val members: ConversationMembers,
-            val from: String
+            val members: List<Member>,
+            val time: String
```
gongracr, line 46: would this be a long timestamp or a `formatedTime` String? In case of the latter, I think it would be useful to be slightly more descriptive on the naming.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
```diff
@@ -1,18 +1,30 @@
...
+    data class Calling(val value: String) : Client()
+    data class Asset(val value: AssetContent) : Client()
+    data class DeleteMessage(val messageId: String) : Client()
+    data class TextEdited(val editMessageId :String, val newContent : String) : Client()
     data class DeleteForMe(
         val messageId: String,
         val conversationId: String,
         val qualifiedConversationId: QualifiedConversationId?
-    ) : MessageContent()
+    ) : Client()
```
gongracr, line 21: Shouldn't we nest these inside the `Client` sealed class in order to help compiler be exhaustive on possible `when` clauses?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
```diff
@@ -4,24 +4,46 @@ import com.wire.kalium.logic.data.conversation.ClientId
...
+    data class Client(
+        override val id: String,
+        override val content: MessageContent.Client,
+        override val conversationId: ConversationId,
+        override val date: String,
+        override val senderUserId: UserId,
+        val senderClientId: ClientId,
+        override val status: Status,
+        val editStatus : EditStatus,
+        override val visibility: Visibility = Visibility.VISIBLE
```
gongracr, line 27: Maybe move the unique `Client` parameters to the end to make it clearer they are not shared with `Server`?
```suggestion
        override val visibility: Visibility = Visibility.VISIBLE,
        override val status: Status,
        val senderClientId: ClientId,
        val editStatus : EditStatus
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
```diff
@@ -1,18 +1,30 @@
...
+
+    object Unknown : Client()
 
-    object Unknown : MessageContent()
+    // server message content types
+    sealed class MemberChange(open val members: List<Member>) : Server() {
+        data class Join(override val members: List<Member>) : MemberChange(members)
+        data class Leave(override val members: List<Member>) : MemberChange(members)
```
gongracr, line 28: Maybe rename to past tense to indicate that event already occurred?
```suggestion
        data class Joined(override val members: List<Member>) : MemberChange(members)
        data class Left(override val members: List<Member>) : MemberChange(members)
```

gongracr (approved):
It's a long PR so I probably missed some stuff, so it would be great if other peers could also review it. But so far what I saw was pretty clear and clean. Magically executed as usual Michal 🦾 🍨

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
saleniuk: I think from compiler's perspective it’s no difference, it still knows which are extensions of which class so it’s mainly a matter of how we see / format it. With having all items inside Client/Server we would need to change many lines just for it to be for instance `MessageContent.Client.Text` and we usually don’t need such detailed info in every place, so current way I think is the best of both worlds: we have hierarchy and all advantages of sealed classes without the need of changing every line of code and having overwhelming names. What do you think?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
saleniuk: Good idea, I changed it 🚀

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
saleniuk: Sure, but for event names I kept the current ones because "member-leave" and "member-join" are the exact types so it's easier to find and understand it, but here it can be changed, but I prefer "Added"/"Removed" because "Joined"/"Left" suggest that user did it himself and actually it covers both cases ("left" only if author is the only member on list)

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
gongracr: I said it because I saw multiple times, IDE warning around non exhaustive `when` clauses even though they were. And I thought it was because of that. But if you say it is not that's fine with me too. It's also true that without the extra nested class we gain in flexibility and lack of pain 😅

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
gongracr: yeah, there is no a name to rule them all, so I would leave it to your own personal preference then 🧑‍🎨

MohamadJaara (approved):
lgtm 🚀 , just small question about the date format

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
```diff
@@ -20,47 +19,47 @@ sealed class Event(open val id: String) {
...
-            val time: String,
+            val timestampIso: String,
             val content: String
         ) : Conversation(id, conversationId)
 
         data class NewConversation(
             override val id: String,
             override val conversationId: ConversationId,
-            val time: String,
+            val timestampIso: String,
```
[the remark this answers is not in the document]
saleniuk: No, I checked it and for events that have time field: `NewConversationDTO`, `NewMessageDTO`, `MemberJoinDTO`, `MemberLeaveDTO` it's actually like `2022-06-08T14:35:01.858Z` so it's a ISO 8601 timestamp 😉 
I don't know how to check `NewMLSMessageDTO` but by looking at backend code, I can assume that the same code handles events so the time format will be the same (backend json examples of "mls-message-add" also have ISO format)

PR #545, opened 2022-05-30
fix: updating user connection status (AR-1540)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently, as the app performs a sync process, we need to update/insert into User's table many times.
Right now as a last step (fetching User Info by Ids) the query was overriding every record saved, due to the harmful 🙈 `INSERT OR REPLACE`.

### Causes (Optional)

The information about the users and contacts was lost every time we performed a sync.

### Solutions

Replace the queries for the sync process to perform `upserts` instead of plain INSERTS shortcuts.
We first perform an update (by specific data available), then:
- If rows affected, then do nothing
- If no rows affected, then perform an insert with default values

#### Test Coverage (Optional)

### Notes

Also added the a GH action so we can self assign PR, I found it useful, so we can easily navigate to changes.

----

----

----
##### References


.github/workflows/pr-author-assigner.yml
```diff
@@ -0,0 +1,12 @@
+name: Auto assign author to PR
+on:
+    pull_request:
+        types: [ opened ]
+
+jobs:
+    assignAuthor:
+        runs-on: ubuntu-latest
+        steps:
```
vitorhugods, line 9: I like the fix, but can we add this tiny condition for the auto-assign action?

```suggestion
        if: github.event.pull_request.draft == false
        steps:
```

vitorhugods (approved):
Nice fix, nice automation.

Not quite atomic PR, but at least we got to see it working first-hand :D.


Just a suggestion to don't auto-assign if the PR is still a draft.

.github/workflows/pr-author-assigner.yml
```diff
@@ -0,0 +1,12 @@
+name: Auto assign author to PR
+on:
+    pull_request:
+        types: [ opened ]
```
vitorhugods, line 4:
```suggestion
        types: [ opened, ready_for_review ]
```

.github/workflows/pr-author-assigner.yml
yamilmedina: makes sense =D applied change

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAOImpl.kt
```diff
@@ -44,33 +44,43 @@ class UserDAOImpl(private val queries: UsersQueries) : UserDAO {
         )
     }
 
-    override suspend fun insertUsers(users: List<UserEntity>) {
+    override suspend fun upsertTeamMembers(users: List<UserEntity>) {
```
MohamadJaara, line 47: Can we achieve the same result by using INSERT + ON CONFLICT E.g. https://github.com/wireapp/kalium/blob/develop/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq#L25

PR #552, opened 2022-05-31
refactor: change fold so it doesn't accept nullable lambdas

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

`NullPointerException` during Runtime when returning null in `fold` nullables. The compiler doesn't complain about this:

```kotlin
val either: Either<Foo> = bar()

val result = either.fold({
    println("Left case")
    null
}, {
    println("Right case")
    it
})
```

### Causes

The definition of `fold` accepts lambdas that return null. It accepts nullable `T` types.

### Solutions

Explicitly declare `T: Any` instead of the implicit `T: Any?` when no supertype is specified.

### Testing

Compiler does it!

Can't write tests for this... as it doesn't compile anymore :) 
Type-safety FTW.

----

----
##### References


alexandreferris (approved):
🚀

PR #553, opened 2022-05-31
ci: run android emulator tests for all modules

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

CI isn't running Android tests for all modules.

### Causes

We have hardcoded modules listed when running Android emulator tests.

### Solutions

Replace hardcoded modules with project-wide task: `.connectedAndroidTest`
Still W.I.P. as I'm not sure it will work first time.

Enabling this test reviewed that some tests where not passing or actually not even building! This PR includes fixes.

### Testing

N/A

----

----
##### References

### DIFF
--- a/.github/workflows/gradle-android-tests.yml
+++ b/.github/workflows/gradle-android-tests.yml
@@ -56,7 +56,9 @@ jobs:
         with:
           api-level: ${{ matrix.api-level }}
           script: |
-            ./gradlew connectedAndroidTest
+            ./gradlew cryptography:connectedAndroidTest
+            ./gradlew persistence:connectedAndroidTest
+            ./gradlew logic:connectedAndroidTest
 
       - name: Publish Unit Test Results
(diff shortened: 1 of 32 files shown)


MohamadJaara (approved):
lgtm  🚀

vitorhugods:
22mins ONLY for the instrumented tests?

No.

No.

No.

We need something else. Either run Unit tests on the Linux job (so Instrumented run in parallel).

Or be specific about which tests run where.

Kotlin Multiplatform is quite annoying actually in this regard.
When we write tests in `common`, we can't specify if it's Android Unit or Android Instrumented.
The only solution for this that I can see is taking a similar approach to `@IgnoreJS`. Like `@AndroidUnit` or similar.

And to be fair, running is not even the biggest issue. Building is also taking a lot of time.
I'm thinking about just listing `:logic` in there too.

Still, a lot of simple tests would run both as Instrumented and Unit tests

PR #489, opened 2022-05-11
feat: (AR-1464) epic api versioning

MohamadJaara:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

AR-1464

### Solutions

this PR is a collection of already approved PRs:

1. #498 
2. #507 
3. #509
4. #514 
5. #531 
6. #527
7. #543 
8. #546 

### Dependencies (Optional)

_If there are some other pull requests related to this one (e.g. new releases of frameworks), specify them here._

Needs releases with:

- [ ] GitHub link to other pull request

#### Test Coverage (Optional)

- [ ] I have added automated test to this contribution

#### How to Test

_Briefly describe how this change was tested and if applicable the exact steps taken to verify that it works as expected._

### Notes (Optional)

_Specify here any other facts that you think are important for this issue._

### Attachments (Optional)

_Attachments like images, videos, etc. (drag and drop in the text box)_

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #557, opened 2022-06-01
feat: feature flags [AR-1463]

MohammedMokresh:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

For the build configs that we generate from the client side we needed a way to send it to kalium and read it back , so we will have SSOT of the configs (build and run time )
for now we only have the build time , but later we will combine it with the run time 

### Solutions

created a data class for the configs that we pass when we initiate the coreLogic from the client side 
and create a use case that returns a flow of this data class 
### Dependencies (Optional)

https://github.com/wireapp/wire-android-reloaded/pull/636

Needs releases with:

https://github.com/wireapp/wire-android-reloaded/pull/636

#### How to Test

after you initiate with some configs you can check if the use case returning these values 

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/featureFlags/KaliumConfigs.kt
```diff
@@ -0,0 +1,44 @@
...
+    val httpProxyUrl: String = "",
+    val websiteUrl: String = "",
+    val webSocketUrl: String = "",
+    val certificate: String = "",
+    val domain: String = "",
+    val supportEmail: String = "",
+    val teamsUrl: String = "",
+    val backendUrl: String = "",
+    val blacklistHost: String = "",
+    val accountUrl: String = "",
```
vitorhugods, line 37: I wonder about these... @MohamadJaara  might help.

I think we configure Kalium to allow or not multiple backends.
But having these Urls here is redundant. As we already pass them when needing to authenticate, etc.

logic/src/commonMain/kotlin/com/wire/kalium/logic/featureFlags/KaliumConfigs.kt
```diff
@@ -0,0 +1,44 @@
...
+    val wipeOnCookieInvalid: Boolean = false,
+    val submitCrashReports: Boolean = false,
+    val webLinkPreview: Boolean = false,
+    val developerFeaturesEnabled: Boolean = false,
+    val isLoggingEnabled: Boolean = false,
+    val isSafeLoggingEnabled: Boolean = false,
+    val enableBlacklist: Boolean = false,
+    val fileRestrictionEnabled: Boolean = false,
+    val countlyAppKey: String = "",
+    val countlyServerUrl: String = "",
```
vitorhugods, line 24: Do we still use Countly?

vitorhugods (changes requested):
Nicely done 👨🏼‍🍳🤌🏼

Just some tiny remarks, and the wish to confirm with @MohamadJaara about backend config, as #489 is on its way.

logic/src/commonMain/kotlin/com/wire/kalium/logic/featureFlags/KaliumConfigs.kt
MohamadJaara: we can delete the PROD/STAGING companion object in `ServerConfig` and take whatever is in here as the DEFAULT

MohamadJaara: and we need to add a new  feature flag for en/disable mutable accounts in different BEs (and acutely disable it)

vitorhugods: > we can delete the PROD/STAGING companion object in `ServerConfig` and take whatever is in here as the DEFAULT

I'd leave them there, but take whatever is defined here as default instead.

Leaving them there makes it easier to bootstrap Kalium into a new client/project.

logic/src/commonMain/kotlin/com/wire/kalium/logic/featureFlags/KaliumConfigs.kt
MohammedMokresh: I guess not , but I left it there as I parse the values as it is , but I think for now I can remove the things that we know that we won't use now , and can add later

logic/src/commonMain/kotlin/com/wire/kalium/logic/featureFlags/KaliumConfigs.kt
```diff
@@ -0,0 +1,44 @@
...
+    val forcePrivateKeyboard: Boolean = false,
+    val keepWebSocketOn: Boolean = false,
+    val wipeOnCookieInvalid: Boolean = false,
+    val submitCrashReports: Boolean = false,
+    val webLinkPreview: Boolean = false,
+    val developerFeaturesEnabled: Boolean = false,
+    val isLoggingEnabled: Boolean = false,
+    val isSafeLoggingEnabled: Boolean = false,
+    val enableBlacklist: Boolean = false,
+    val fileRestrictionEnabled: Boolean = false,
```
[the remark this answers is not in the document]
MohammedMokresh: my idea was to keep the values here so the reloaded or another client will be always reading from this data class , instead of either read it from the build config directly or from here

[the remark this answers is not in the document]
vitorhugods: > my idea was to keep the values here so the reloaded or another client will be always reading from this data class , instead of either read it from the build config directly or from here

It's better to add values with default parameters later, then adding them now and removing later leading to breaking changes.

If we remove `isMarketingCommunicationEnabled` because we find out that it's and old and unused feature that we only added because we had on Android, we will have to annotate with `@Deprecated`  first, and only remove it in a major version of Kalium.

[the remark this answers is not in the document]
MohammedMokresh: hmmmm true , adding will be easier than removing , so I think I will remove the things that we sure that we won't use in kalium or at all for now

logic/src/commonMain/kotlin/com/wire/kalium/logic/featureFlags/KaliumConfigs.kt
gongracr: Also, are we meant to initialise the running environment on `WireActivityViewModel.serverConfig` with these values instead?

MohamadJaara (approved):
lgtm

PR #574, opened 2022-06-07
feat: Persist self user status (AR-1741, AR-1742, AR-1743)

borichellow:
## Tasks [AR-1741](https://wearezeta.atlassian.net/browse/AR-1741),  [AR-1742](https://wearezeta.atlassian.net/browse/AR-1742),  [AR-1743](https://wearezeta.atlassian.net/browse/AR-1743)

# What's new in this PR?
- added `user_availability_status` filed into Users DB
- added `availabilityStatus` filed into Entity and Logic Users
- updated mappers for all these Users
- added `UpdateSelfAvailabilityStatusUseCase` for updating self Availability Status in the app

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
@@ -3,9 +3,12 @@ package com.wire.kalium.logic.data.publicuser
 import com.wire.kalium.logic.data.id.IdMapper
 import com.wire.kalium.logic.data.notification.LocalNotificationMessageAuthor
 import com.wire.kalium.logic.data.publicuser.model.OtherUser
+import com.wire.kalium.logic.data.user.AvailabilityStatusMapper
 import com.wire.kalium.logic.data.user.ConnectionState
+import com.wire.kalium.logic.data.user.ConnectionStateMapper
 import com.wire.kalium.logic.data.user.UserAvailabilityStatus
 import com.wire.kalium.logic.data.user.UserId
+import com.wire.kalium.logic.di.MapperProvider
 import com.wire.kalium.network.api.model.getCompleteAssetOrNull
 import com.wire.kalium.network.api.model.getPreviewAssetOrNull
 import com.wire.kalium.network.api.user.details.UserProfileDTO
@@ -18,13 +21,14 @@ interface PublicUserMapper {
     fun fromUserDetailResponse(userDetailResponse: UserProfileDTO): OtherUser
     fun fromUserApiToEntity(userDetailResponse: UserProfileDTO, connectionState: ConnectionEntity.State): UserEntity
     fun fromUserDetailResponses(userDetailResponse: List<UserProfileDTO>): List<OtherUser>
-    fun fromDaoConnectionStateToUser(connectionState: ConnectionEntity.State): ConnectionState
     fun fromPublicUserToLocalNotificationMessageAuthor(author: OtherUser?): LocalNotificationMessageAuthor
-    fun fromDaoAvailabilityStatusToUser(status: UserAvailabilityStatusEntity?): UserAvailabilityStatus?
-    fun fromUserAvailabilityStatusToDao(status: UserAvailabilityStatus?): UserAvailabilityStatusEntity?
 }
 
-class PublicUserMapperImpl(private val idMapper: IdMapper) : PublicUserMapper {
+class PublicUserMapperImpl(
+    private val idMapper: IdMapper,
+    private val availabilityStatusMapper: AvailabilityStatusMapper = MapperProvider.availabilityStatusMapper(),
+    private val connectionStateMapper: ConnectionStateMapper = MapperProvider.connectionStateMapper()
+) : PublicUserMapper {
 
     override fun fromDaoModelToPublicUser(userEntity: UserEntity) = OtherUser(
         id = idMapper.fromDaoModel(userEntity.id),
@@ -34,10 +38,10 @@ class PublicUserMapperImpl(private val idMapper: IdMapper) : PublicUserMapper {
         phone = userEntity.phone,
         accentId = userEntity.accentId,
         team = userEntity.team,
-        connectionStatus = fromDaoConnectionStateToUser(connectionState = userEntity.connectionStatus),
+        connectionStatus = connectionStateMapper.fromDaoConnectionStateToUser(connectionState = userEntity.connectionStatus),
         previewPicture = userEntity.previewAssetId,
         completePicture = userEntity.completeAssetId,
-        availabilityStatus = fromDaoAvailabilityStatusToUser(userEntity.availabilityStatus)
+        availabilityStatus = availabilityStatusMapper.fromDaoAvailabilityStatusToUser(userEntity.availabilityStatus)
     )
 
     override fun fromUserDetailResponse(userDetailResponse: UserProfileDTO) = OtherUser(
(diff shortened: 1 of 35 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
```diff
@@ -76,4 +83,22 @@ class PublicUserMapperImpl(private val idMapper: IdMapper) : PublicUserMapper {
 
     override fun fromPublicUserToLocalNotificationMessageAuthor(author: OtherUser?) =
         LocalNotificationMessageAuthor(author?.name ?: "", null)
+
+    override fun fromDaoAvailabilityStatusToUser(status: UserAvailabilityStatusEntity?): UserAvailabilityStatus? =
+        when (status) {
+            UserAvailabilityStatusEntity.AVAILABLE -> UserAvailabilityStatus.AVAILABLE
+            UserAvailabilityStatusEntity.BUSY -> UserAvailabilityStatus.BUSY
+            UserAvailabilityStatusEntity.AWAY -> UserAvailabilityStatus.AVAILABLE
```
ohassine, line 91: `AWAY` returns `AVAILABLE` here
`UserAvailabilityStatusEntity.AWAY -> UserAvailabilityStatus.AWAY`

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
```diff
@@ -194,4 +204,22 @@ internal class UserMapperImpl(private val idMapper: IdMapper = MapperProvider.id
...
+        }
+
+    override fun fromUserAvailabilityStatusToDao(status: UserAvailabilityStatus?): UserAvailabilityStatusEntity? =
+        when (status) {
+            UserAvailabilityStatus.AVAILABLE -> UserAvailabilityStatusEntity.AVAILABLE
+            UserAvailabilityStatus.BUSY -> UserAvailabilityStatusEntity.BUSY
+            UserAvailabilityStatus.AWAY -> UserAvailabilityStatusEntity.AWAY
+            UserAvailabilityStatus.NONE -> UserAvailabilityStatusEntity.NONE
+            null -> null
+        }
```
ohassine, line 224: We have already those functions in `PublicUserMapper`, can you remove this duplication ?

ohassine (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
```diff
@@ -76,4 +83,22 @@ class PublicUserMapperImpl(private val idMapper: IdMapper) : PublicUserMapper {
...
+            null -> null
+        }
+
+    override fun fromUserAvailabilityStatusToDao(status: UserAvailabilityStatus?): UserAvailabilityStatusEntity? =
+        when (status) {
+            UserAvailabilityStatus.AVAILABLE -> UserAvailabilityStatusEntity.AVAILABLE
+            UserAvailabilityStatus.BUSY -> UserAvailabilityStatusEntity.BUSY
+            UserAvailabilityStatus.AWAY -> UserAvailabilityStatusEntity.AWAY
+            UserAvailabilityStatus.NONE -> UserAvailabilityStatusEntity.NONE
+            null -> null
```
trOnk12, line 102: do we ever want to care about work with the null ? wouldn't it be better to return UserAvailibStatus.NONE?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
```diff
@@ -194,4 +204,22 @@ internal class UserMapperImpl(private val idMapper: IdMapper = MapperProvider.id
             ConnectionState.MISSING_LEGALHOLD_CONSENT -> ConnectionEntity.State.MISSING_LEGALHOLD_CONSENT
             ConnectionState.ACCEPTED -> ConnectionEntity.State.ACCEPTED
         }
+
+    override fun fromDaoAvailabilityStatusToUser(status: UserAvailabilityStatusEntity?): UserAvailabilityStatus? =
+        when (status) {
+            UserAvailabilityStatusEntity.AVAILABLE -> UserAvailabilityStatus.AVAILABLE
+            UserAvailabilityStatusEntity.BUSY -> UserAvailabilityStatus.BUSY
+            UserAvailabilityStatusEntity.AWAY -> UserAvailabilityStatus.AVAILABLE
```
gongracr, line 212: wrong parsing

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
```diff
@@ -76,4 +83,22 @@ class PublicUserMapperImpl(private val idMapper: IdMapper) : PublicUserMapper {
 
     override fun fromPublicUserToLocalNotificationMessageAuthor(author: OtherUser?) =
         LocalNotificationMessageAuthor(author?.name ?: "", null)
+
+    override fun fromDaoAvailabilityStatusToUser(status: UserAvailabilityStatusEntity?): UserAvailabilityStatus? =
```
gongracr, line 87: I think we should rename to 
```suggestion
    override fun fromDaoAvailabilityStatusToModel(status: UserAvailabilityStatusEntity?): UserAvailabilityStatus? =
```
as this is not mapping the availability status to a user object

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
borichellow: oops, copy-past issue :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
```diff
@@ -71,7 +75,8 @@ internal class UserMapperImpl(private val idMapper: IdMapper = MapperProvider.id
             accentId = userProfileDTO.accentId,
             team = userProfileDTO.teamId,
             previewAssetId = userProfileDTO.assets.getPreviewAssetOrNull()?.key,
-            completeAssetId = userProfileDTO.assets.getCompleteAssetOrNull()?.key
+            completeAssetId = userProfileDTO.assets.getCompleteAssetOrNull()?.key,
+            availabilityStatus = null
```
gongracr, line 79: What if I have a valid local value of the status? Shouldn't we make sure to update this value correctly (on the UserRepository?) with the old value to avoid nulling the user status every time we try to fetch the latest state of the user remotely?

gongracr (commented):
Looking good in general, although I wonder if the user status will be overwritten when fetching the latest state of the self user. Maybe we could add tests for that too.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
borichellow: after some rethinking i think you're right: as far as `NONE ` means "no status" it can be used as default

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
```diff
@@ -18,6 +20,8 @@ interface PublicUserMapper {
     fun fromUserDetailResponses(userDetailResponse: List<UserProfileDTO>): List<OtherUser>
     fun fromDaoConnectionStateToUser(connectionState: ConnectionEntity.State): ConnectionState
     fun fromPublicUserToLocalNotificationMessageAuthor(author: OtherUser?): LocalNotificationMessageAuthor
+    fun fromDaoAvailabilityStatusToUser(status: UserAvailabilityStatusEntity?): UserAvailabilityStatus?
```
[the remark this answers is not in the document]
borichellow: good idea, extracted it to separate mappers :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
borichellow: in DB `user_availability_status` change is ignored in user inserting and updating, it can be changed only by `updateUserAvailabilityStatus` method. So that `null`  (after update it will be `NONE`) here wouldn't break anything

gongracr (approved):
Thanks for the changes. Much fresher now 🥬 🍅 🥒 Awesome job here, now 1 step closer to have real user status in the app 🟢 !

PR #575, opened 2022-06-08
chore: add active speakers handler

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-1766](https://wearezeta.atlassian.net/browse/AR-1766)

### Issues

Handler for active speakers in a group call not implemented

### Solutions

Add implementation for active speakers in a group call now adding users that are speaking in to `activeSpeakers` attribute in `Call` data class.


calling/src/androidMain/kotlin/com/wire/kalium/calling/Calling.kt
```diff
@@ -127,6 +128,12 @@ interface Calling : Library {
         clientsJson: String
     )
 
+    @Suppress("FunctionNaming", "FunctionParameterNaming")
+    fun wcall_set_active_speaker_handler(
+        inst: Handle,
+        wcall_active_speaker_h: ActiveSpeakersHandler
```
ohassine, line 134: Could you a different name  instead of `wcall_active_speaker_h`, `activeSpeakersHandler`  ?

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
```diff
@@ -232,6 +233,7 @@ actual class CallManagerImpl(
      * onCallingReady
      * Will start the handlers for: ParticipantsChanged, NetworkQuality, ClientsRequest and ActiveSpeaker
      */
+    @Suppress("LongMethod")
```
ohassine, line 236: It would be better maybe if we move each part to a separate function and call them inside onCallingReady ?

```
private fun onCallingReady() {
    initParticipantsHandler()
    initClientsHandler()
    initNetworkHandler()
}
```

ohassine, line 236: It would be better maybe if we move each part to a separate function and call them inside onCallingReady ?

```
private fun onCallingReady() {
    initParticipantsHandler()
    initClientsHandler()
    initNetworkHandler()
}
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/ActiveSpeaker.kt
```diff
@@ -0,0 +1,8 @@
+package com.wire.kalium.logic.data.call
+
+data class ActiveSpeaker(
+    val userId: String,
+    val clientId: String,
+    val audioLevel: Int,
+    val audioLevelNow: Int
```
ohassine, line 7: what those two fields used for ?

calling/src/androidMain/kotlin/com/wire/kalium/calling/Calling.kt
alexandreferris: As this is where we are mapping the native functions I don't think we can rename them 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallActiveSpeakers.kt
```diff
@@ -0,0 +1,17 @@
...
+    @SerialName("audio_levels") val activeSpeakers: List<CallActiveSpeaker>
+)
+
+@Serializable
+data class CallActiveSpeaker(
+    @SerialName("userid") val userId: String,
+    @SerialName("clientid") val clientId: String,
+    @SerialName("audio_level") val audioLevel: Int,
+    @SerialName("audio_level_now") val audioLevelNow: Int
+)
```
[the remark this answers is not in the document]
alexandreferris: Feels right to use the same data class, but currently our data classes are a mess, and we need to fix it by having a separated module for data classes that are used project-wide. This `CallActiveSpeaker` class is specifically used only for calling logic. Where as the other data class is the one that will be exposed to AR to be used. (maybe some of this `Call..` classes could be internal or private for the module.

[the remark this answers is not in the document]
alexandreferris: Feels right to use the same data class, but currently our data classes are a mess, and we need to fix it by having a separated module for data classes that are used project-wide. This `CallActiveSpeaker` class is specifically used only for calling logic. Where as the other data class is the one that will be exposed to AR to be used. (maybe some of this `Call..` classes could be internal or private for the module.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/Call.kt
```diff
@@ -18,5 +19,6 @@ data class Call(
     val isCameraOn: Boolean,
     val callerId: String,
     val participants: List<Participant> = emptyList(),
-    val maxParticipants: Int = 0 // Was used for tracking
+    val maxParticipants: Int = 0, // Was used for tracking
+    val activeSpeakers: List<ActiveSpeaker> = emptyList()
```
ohassine, line 23: Isn't better if we put active speaker fields in Participant  object ?

alexandreferris: Great suggestion!

ohassine (approved):
lgtm ✅ Great work 👌🚀

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallMapper.kt
```diff
@@ -64,10 +65,25 @@ class CallMapper {
...
+    }
+
+    inner class ActiveSpeakerMapper {
+        fun mapParticipantsActiveSpeaker(
+            participants: List<Participant>,
+            activeSpeakers: CallActiveSpeakers
+        ) : List<Participant> = participants.map { participant ->
+            participant.copy(
+                isSpeaking = activeSpeakers.activeSpeakers.any {
+                    it.userId == participant.id.toString()
```
ohassine, line 77: we should compare with clientId also, the same user can join a group call with different clients

ohassine (changes requested):

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
```diff
@@ -267,19 +275,19 @@ actual class CallManagerImpl(
                 callingLogger.d("$TAG - wcall_set_network_quality_handler() called")
             }
         }
+    }
 
-        // Clients Request
+    private fun initNetworkHandler() {
```
ohassine, line 280: `initClientsHandler()` 🙃

ohassine (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallMapper.kt
alexandreferris: Nice! great catch

ohassine (approved):
Well done 👌✅

PR #576, opened 2022-06-09
feat: (AR-1553) Splitting search query for federated search

dmkasperski:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

This is logic needed in the android reload repo https://github.com/wireapp/wire-android-reloaded/pull/649

### Causes (Optional)

### Solutions

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References
https://wearezeta.atlassian.net/browse/AR-1553

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/id/QualifiedId.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/id/QualifiedId.kt
@@ -18,6 +18,7 @@ data class QualifiedID(
 }
 
 const val VALUE_DOMAIN_SEPARATOR = "@"
+val FEDERATION_REGEX = Regex("[^@.]+@[^@.]+\\.[^@]+")
 
 typealias ConversationId = QualifiedID
 
@@ -38,4 +39,12 @@ fun String.parseIntoQualifiedID(): QualifiedID {
     val components = split("@")
     if (components.size < 2) throw IllegalStateException("The string trying to parse is not a valid one")
     return QualifiedID(value = components.first(), domain = components.last())
+}
+
+fun String.parseSearchQueryToQualifiedID(): QualifiedID {
+    val isFederationSearch = matches(FEDERATION_REGEX)
+
+    return if (isFederationSearch) parseIntoQualifiedID() else
+        QualifiedID(this, "")
+
 }
--- a/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/id/FederationRegexTest.kt
+++ b/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/id/FederationRegexTest.kt
@@ -1,0 +1,84 @@
+package com.wire.kalium.logic.data.id
+
+import kotlin.test.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class FederationRegexTest {
+    @Test
+    fun givenUser_correctFederatedId_matchesRegex() {
+        val userId = "damian@wire.com"
+
+        assertTrue(userId.matches(FEDERATION_REGEX))
+    }
+
+    @Test
+    fun givenUser_correctFederatedId_doubleDotFederation_matchesRegex() {
+        val userId = "damian@wire.com.pl"
+
+        assertTrue(userId.matches(FEDERATION_REGEX))
+    }
+
+    @Test
+    fun givenUser_correctFederatedId_tripleDotFederation_matchesRegex() {
+        val userId = "damian@wire.com.pl.anything"
+
+        assertTrue(userId.matches(FEDERATION_REGEX))
+    }
+
+    @Test
+    fun givenUser_correctFederatedId_containsNumericCharacters_matchesRegex() {
+        val userId = "damian1@wire1.com1.pl1.anything1"
+
+        assertTrue(userId.matches(FEDERATION_REGEX))
+    }
+
+    @Test
+    fun givenUser_correctFederatedId_containsSpecialCharacters_matchesRegex() {
+        val userId = "damian1!#$%^&*()@wire1!#$%^&*().com1!#$%^&*().pl!#$%^&*()1.anything1!#$%^&*()"
+
+        assertTrue(userId.matches(FEDERATION_REGEX))
+    }
+
+    @Test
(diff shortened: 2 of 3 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/id/QualifiedId.kt
```diff
@@ -39,3 +40,11 @@ fun String.parseIntoQualifiedID(): QualifiedID {
...
     return QualifiedID(value = components.first(), domain = components.last())
 }
+
+fun String.parseSearchQueryToQualifiedID(): QualifiedID {
+    val isFederationSearch = matches(FEDERATION_REGEX)
+
+    return if (isFederationSearch) parseIntoQualifiedID() else
+        QualifiedID(this, "")
+
+}
```
trOnk12, line 50: I am sorry for being super picky, but I wonder if this is the right decision to make it as a extension function on String, the problem I see is that in the places where we are going to consume it like for example a use case : 

```
fun invoke(searchQuery : String){
  val isFederationed = searchQuery.parseSearchQueryToQualifiedID()
  
  if(isFederationed){ 
   ...do something()
  } else{
   ... do something()
  }
}

there is not a really a way to mock the behaviour expect of passing the searchQuery that would satisfy the if condition, so I think it would be better to make this logic injectable into the consutrctor of the conusmer, that way we can mock its behaviour. 

let me know what you think.

```

dmkasperski: what other type would you rather see there? 

This is strictly used for search query which is text value (String), okay maybe input directly returns TextFieldValue, but using TextFieldValue cause other consequences/problems

you are saying about mocking behavior, IMO using/passing some variable directly, as I did in my tests is a pretty good projection of what the user does i.e. if you have a test which tests functions called when the user clicks the button then in a unit test if you assume that calling function is the same as actually clicking the button there is nothing wrong about that. So in this case having a ready, desired String, which is a result of the user typing something in the input is equivalent

Also don't forget that unit tests should be as easy as possible and test a single thing, and what's the point of complicating life with mocking user inputs, they are something that determines test results so it's good to have raw values in the test, it makes the test easier to read and understand

but I'm not here to argue, I'm still learning your patterns, maybe we could jump on some call because at this point I don't think so we are on the same page how it should be implemented, maybe I'm not on your page :D so would like better explanaition

trOnk12: I think we are not on the same page indeed :P will write you on PM about what I mean here

trOnk12:
```

fun String.isSearchWithEmail(): Boolean {
    return this.contains('@')
}

class UseCase(private val someRepo: SomeRepo) {

    fun invoke(searchQuery: String) {
        if (searchQuery.isSearchWithEmail()) {
            someRepo.searchWithEmail(searchQuery)
        } else {
            someRepo.searchWIthHande(searchQuery)
        }
    }

}

class UseCaseTest() {

    @Mock
    private val someRepo: SomeRepo

    fun "if a search query is an email search repo by email"()
    {
        // here to test if we are really searching in a repo with
        // a email, we have to once again pass a valid searchQuery containing
        // an email, but we already wrote tests for that in another class
        // probably, testing the extension, we also are not able to give the requiremnt that this usecase
        // needs to validate the email first before proceeding
    }

    fun "if a search query is not an email search repo by handle"()
    {
        // here to test if we are really not searching in a repo with
        // a email, we have to once again pass a valid searchQuery containing
        // an email, but we already wrote tests for that in another class
        // probably, testing the extension, we also are not able to give the requiremnt that this usecase
        // needs to validate the email first before proceeding
    }

}

```

```
class EmailValidator{
    
    fun isEmail(query : String) : Boolean{
        return query.contains('@')
    }
}

class UseCase(
    private val someRepo: SomeRepo,
    private val emailValidator: EmailValidator
) {

    fun invoke(searchQuery: String) {
        if (emailValidator.isSearchWithEmail()) {
            someRepo.searchWithEmail(searchQuery)
        } else {
            someRepo.searchWIthHande(searchQuery)
        }
    }

}

class UseCaseTest() {

    @Mock
    private val someRepo: SomeRepo
    
    @Mock 
    private val emailValidator: EmailValidator

    fun "if a search query is an email search repo by email"()
    {
        // here we do not have to validate the query once again,
        // if that is an email, we can just test the behaviour of the use case
        // not the logic that should be validating the email
        // also we are giving this UseCase a requirement, that you have to 
        // check for the email first, before proceeding
    }

    fun "if a search query is not an email search repo by handle"()
    {
        // here we do not have to validate the query once again,
        // if that is not an email, we can just test the behaviour of the use case
        // not the logic that should be validating the email  
        // also we are giving this UseCase a requirement, that you have to 
        // check for the email first, before proceeding
    }

}

```

dmkasperski: I think both approaches make sense, but I don't know if the second one isn't an overkill 

I won't argue my solution is better is just different and if you'd like me to align with yours I can do it or maybe we need 3rd opinion here :D

trOnk12: i'll leave it up to you , it is not a blocker or anything major, just wanted to point it out and give my concerns and some arguments

MohamadJaara: I agree with @Mateusz-appunite since when the search is not federated, the domain should not be an empty string, rather the domain of self user.

```kotlin
class FederatedSearchValidator (
private val userRepo: userRespository
) {
fun invoke(): String? {
//1. when invoked extract the domain 
//2. check if it's a valid domain and the user BE is federated with
//3. return the search domain or null if it's not a valid domain
}}
```

dmkasperski:
closing in favor of https://github.com/wireapp/kalium/pull/595

PR #577, opened 2022-06-09
fix: Persist self user status

borichellow:
### Issues

Previously there was no option to provide `UpdateSelfAvailabilityStatusUseCase` (forgot to add )
Also SelfAvailabilityStatus was removed on every inserting userData into DB

### Solutions

fixed both issues :)


alexandreferris (approved):

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
```diff
@@ -24,8 +24,18 @@ deleteUser:
...
+ON CONFLICT(qualified_id) DO UPDATE SET
+name = excluded.name,
+handle = excluded.handle,
+email = excluded.email,
+phone = excluded.phone,
+accent_id = excluded.accent_id,
+team = excluded.team,
+connection_status = excluded.connection_status,
+preview_asset_id = excluded.preview_asset_id,
+complete_asset_id = excluded.complete_asset_id;
```
MohamadJaara, line 38: `INSERT OR REPLACE` have the same result as updating all the values, if there is a value that should not be updated when an already stored user is inserted you need to remove the `excluded` for that value.

MohamadJaara: Great, my bad 😅

MohamadJaara (approved):
lgtm 🚀

PR #579, opened 2022-06-10
fix: ignore or hide some message types [AR-1827]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently all message types, even ones not yet handled by app, are stored in DB as "Unknown" and visible with "content is not available".

### Solutions

Add Visibility.HIDDEN to messages that shouldn't be visible but only persisted, ignore some types that are actually more like actions than messages and shouldn't be stored.

#### Test Coverage (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
@@ -5,31 +5,36 @@ import com.wire.kalium.protobuf.messages.QualifiedConversationId
 
 sealed class MessageContent {
 
-    sealed class Client : MessageContent()
-    sealed class Server : MessageContent()
+    sealed class System : MessageContent()
+    sealed class FromProto : MessageContent()
+    sealed class Regular : FromProto()
+    sealed class Signaling : FromProto()
 
     // client message content types
-    data class Text(val value: String) : Client()
-    data class Calling(val value: String) : Client()
-    data class Asset(val value: AssetContent) : Client()
-    data class DeleteMessage(val messageId: String) : Client()
-    data class TextEdited(val editMessageId :String, val newContent : String) : Client()
+    data class Text(val value: String) : Regular()
+    data class Calling(val value: String) : Regular()
+    data class Asset(val value: AssetContent) : Regular()
+    data class DeleteMessage(val messageId: String) : Regular()
+    data class TextEdited(val editMessageId :String, val newContent : String) : Regular()
     data class DeleteForMe(
         val messageId: String,
         val conversationId: String,
         val qualifiedConversationId: QualifiedConversationId?
-    ) : Client()
+    ) : Regular()
 
     data class Unknown( // messages that aren't yet handled properly but stored in db in case
         val typeName: String? = null,
-        val encodedData: ByteArray? = null
-    ) : Client()
-    object Ignored : Client() // messages that aren't processed in any way
-    object Empty : Client()
+        val encodedData: ByteArray? = null,
+        val hidden: Boolean = false
+    ) : Regular()
+    object Empty : Regular()
 
     // server message content types
-    sealed class MemberChange(open val members: List<Member>) : Server() {
+    sealed class MemberChange(open val members: List<Member>) : System() {
         data class Added(override val members: List<Member>) : MemberChange(members)
         data class Removed(override val members: List<Member>) : MemberChange(members)
     }
+
+    // we can add other types to be processed, but signaling ones shouldn't be persisted
+    object Ignored : Signaling() // messages that aren't processed in any way
 }
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
@@ -20,7 +20,10 @@ sealed class MessageContent {
         val qualifiedConversationId: QualifiedConversationId?
     ) : Client()
 
-    data class Unknown(val encodedData: ByteArray? = null) : Client() // messages that aren't yet handled properly but stored in db in case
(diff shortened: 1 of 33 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
```diff
@@ -109,36 +110,46 @@ class ProtoContentMapperImpl : ProtoContentMapper {
...
+        val visibility = when (genericMessage.content) {
+            is GenericMessage.Content.Text,
+            is GenericMessage.Content.Asset,
+            is GenericMessage.Content.Calling,
+            is GenericMessage.Content.Composite,
+            is GenericMessage.Content.Edited,
+            is GenericMessage.Content.External,
+            is GenericMessage.Content.Location -> Message.Visibility.VISIBLE
+            is GenericMessage.Content.Deleted -> Message.Visibility.DELETED
+            else -> Message.Visibility.HIDDEN
```
mchenani, line 151: Why the default/else is hidden?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetRecentMessagesUseCase.kt
```diff
@@ -3,11 +3,15 @@ package com.wire.kalium.logic.feature.message
...
 class GetRecentMessagesUseCase(private val messageRepository: MessageRepository) {
 
-    suspend operator fun invoke(conversationId: ConversationId, limit: Int = 100, offset: Int = 0): Flow<List<Message>> {
-        return messageRepository.getMessagesForConversation(conversationId, limit, offset)
-    }
+    suspend operator fun invoke(
+        conversationId: ConversationId,
+        limit: Int = 100,
+        offset: Int = 0,
+        visibility: List<Message.Visibility> = Message.Visibility.values().toList()
```
mchenani, line 15: When we call this useCase, which visibilities are requested? because we need the deleted/hidden cases to show deleted/hidden message indicator.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
```diff
@@ -117,24 +126,28 @@ class GetNotificationsUseCaseImpl(
...
-                            content.value.contains("@$selfUsername")
-                        } ?: false
-                        val containsSelfHandle = selfUser.handle?.let { selfHandle ->
-                            content.value.contains("@$selfHandle")
-                        } ?: false
-
-                        containsSelfUserName or containsSelfHandle
+        when {
+            message is Message.Server -> false
+            message.visibility != Message.Visibility.VISIBLE -> false
```
gongracr, line 131: Maybe extract to `shouldShowMessage(message)` function?

gongracr (approved):
Problems seem to always cram up on the edges 🙈 but hopefully this will fix it right away 🤞

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
saleniuk: We only want the specified above types to be visible, all other types shouldn't be displayed on conversation screen

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
saleniuk: Sure, done 👍

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
```diff
@@ -20,7 +20,11 @@ sealed class MessageContent {
         val qualifiedConversationId: QualifiedConversationId?
     ) : Client()
 
-    data class Unknown(val encodedData: ByteArray? = null) : Client()
+    data class Unknown( // messages that aren't yet handled properly but stored in db in case
+        val typeName: String? = null,
+        val encodedData: ByteArray? = null
+    ) : Client()
+    object Ignored : Client() // messages that aren't processed in any way
```
vitorhugods, line 27: I don't think receipts (`Read` and `Delivered`), `DataTransfer`, `Reaction` (❤️ likes), and `Availability` should not be in the same group as the `MessageContent.Client` ones.

They are signaling messages. They add content and info to other messages, or to the user sending them, etc.


It's also a bit annoying when trying to get all messages to display:

```
fun getAllMessagesForDisplaying(): List<Message.Client> = TODO("")

getAllMessagesForDisplaying().forEach { message ->
    when(message) {
        is MessageContent.Ignored -> //Why is this here? This shouldn't be here.
    }
}
```

Instead of `object Ignored: Client()`, we could be:

Or, for now, as we don't handle any signaling:

```kotlin
    object Ignored: MessageContent() // instead of Client
```

In the future, we can do like this, without having many breaking changes:

```kotlin
sealed interface Signaling: MessageContent(){ // Instead of Ignored
    // different signaling types
}
```


By doing that, it's easier to handle different and handle these different types.

We can handle:

```kotlin
fun handleSignalingMessage(signalingMessage: Signaling){
    TODO("Handle signaling message")
    // receipts, availability, etc.
}
```

vitorhugods (changes requested):
Nice job, just a suggestion to avoid adding `Ignored` as `Client` messages.


I think we should also rename `Client` to something like `Regular` mesages, or `UserCreated` messages, or something like that.

So we can more easily understand what is the difference. So Signaling messages won't come up in the list, but rather modify/add content to the list of messages.

vitorhugods (approved):
Breaking changes on Reloaded maybe?

But great work! A nice step to handling all messages gracefully 👏🏼

PR #584, opened 2022-06-14
chore: Remove Calling domain before federation

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Calling is not working due to not having support for federation yet.

### Solutions

We are removing the domain from the self user Id and the conversation Id in calling only so it works for now, as soon as federation is implemented we will then rework this code to support federation.

So this code is not final.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
@@ -2,6 +2,7 @@ package com.wire.kalium.logic.data.call
 
 import com.wire.kalium.logic.CoreFailure
 import com.wire.kalium.logic.callingLogger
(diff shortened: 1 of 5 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
```diff
@@ -83,9 +84,10 @@ internal class CallDataSource(
 
     override fun updateCallStatusById(conversationId: String, status: CallStatus) {
         val callProfile = _callProfile.value
-        callProfile.calls[conversationId]?.let { call ->
+        val modifiedConversationId = conversationId.toConversationId().toString()
```
MohamadJaara, line 87: `conversationId.toConversationId()` maybe something need to be renamed here 😅

MohamadJaara (approved):
lgtm

PR #586, opened 2022-06-14
feat: Apply user status to notifications

borichellow:
[Task-1262 ](https://wearezeta.atlassian.net/browse/AR-1262)

# What's new in this PR?
Made UserAvailabilityStatus be applied for showing (or not) notifications.
And fixed tests accordingly


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
```diff
@@ -146,8 +189,31 @@ class GetNotificationsUseCaseImpl(
...
+
+    private fun allMuted(conversationMutedStatus: MutedConversationStatus, selfUser: SelfUser) =
+        conversationMutedStatus == MutedConversationStatus.AllMuted
+                || selfUser.availabilityStatus == UserAvailabilityStatus.AWAY
+
+    private fun onlyMentionsAllowed(conversationMutedStatus: MutedConversationStatus, selfUser: SelfUser) =
+        conversationMutedStatus == MutedConversationStatus.OnlyMentionsAllowed
+                || selfUser.availabilityStatus == UserAvailabilityStatus.BUSY
+
+    private fun isSupportedForNotificationsMessageType(message: Message): Boolean =
```
tmspzz, line 209: Knit pick: I suggest naming this `isMessageContentSupportedInNotifications`

tmspzz (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
```diff
@@ -118,20 +116,65 @@ class GetNotificationsUseCaseImpl(
             .distinctUntilChanged()
     }
 
+    private suspend fun observeMessagesList(conversation: Conversation) =
+        if (conversation.lastNotificationDate == null) {
+            // that is a new conversation, lets just fetch last 100 messages for it
+            messageRepository.getMessagesByConversationIdAndVisibility(
+                conversation.id,
+                100,
```
gongracr, line 124: You'll probably have to extract this to a variable to avoid the "MagicNumber" detekt warning

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCaseTest.kt
```diff
@@ -14,303 +15,321 @@ import com.wire.kalium.logic.data.message.MessageRepository
...
-            .whenInvoked()
-            .thenReturn(flowOf(TestUser.SELF))
-        given(conversationRepository)
-            .coroutine { getConversationsForNotifications() }
-            .then { flowOf(listOf()) }
-
-        val notificationsListFlow = getNotificationsUseCase()
+        val (_, getNotifications) = Arrangement()
+            .withSelfUserId()
+            .withSelfUser()
```
gongracr, line 46: Ideally it would be better to pass the expected value per parameter to these arrangement config methods to increase visibility. That way it is clearer what is set up and expected instead of "trusting" the arrangement to do the correct thing "under the hood". Something like this:
```suggestion
            .withSelfUser(TestUser.SELF)
```
It does the same thing in both cases, but on this one, the developer knows right away the type of user that will be returned by the api w/o having to look into the method

gongracr (approved):
Great work refactoring the test suite and accommodating to the Arrangement pattern. I left a tiny comment, but feel free to address it now or not. Good stuff 👷 🏗️

PR #589, opened 2022-06-14
fix: member removed not stored in DB is blocking the flow

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Right now, we get old system messages and it can create the situation where we need data of user that was removed even before we started using the app so we don't have any data about that user stored in DB.

### Solutions

Simply don't return such user, later in the app it will be possible to check if the user is on the returned list, and if not, replace it with some generic name.

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/ObserveMemberDetailsByIdsUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/ObserveMemberDetailsByIdsUseCase.kt
@@ -5,6 +5,7 @@ import com.wire.kalium.logic.data.user.UserId
 import com.wire.kalium.logic.data.user.UserRepository
 import com.wire.kalium.logic.data.user.UserTypeMapper
 import com.wire.kalium.logic.sync.SyncManager
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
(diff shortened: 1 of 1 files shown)


logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/conversation/ObserveMemberDetailsByIdsUseCaseTest.kt
```diff
@@ -118,4 +124,37 @@ class ObserveMemberDetailsByIdsUseCaseTest {
             awaitComplete()
         }
     }
+
+    @Test
+    fun givenUserIsNotKnown_whenObservingMembers_thenDoNotBlockTheFlowWaitingForItsData() = runTest {
```
saleniuk, line 129: @vitorhugods do you think this one is enough to prevent it from occurring again?

vitorhugods: I think so, at least for this UseCase.

Something to keep in mind in the future.

I guess I and other people were a bit too optimistic on data always being there, constraints in the database, etc.

In the end, due to the focus on Privacy, users will be deleted for good, and we need to accept and handle that. It's a change of mindset: missing data is a feature, really.

vitorhugods:
I'm merging, as it's an important fix.

PR #592, opened 2022-06-15
chore: create buffer for pending events

vitorhugods:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

We need a way to store pending events as we collect them through the notifications API, so later when processing live events from the WebSocket we can check if there are any duplications.

### Solutions

Create a `PendingEventsBuffer` which lets us add the collected pending events, check if events were added previously and clear it if needed.

#### Test Coverage

- [X] I have added automated test to this contribution

### Notes

Another PR is coming with the full utilisation of this bad boy.

----
#### PR Post Merge Checklist for internal contributors

 - [X] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #593, opened 2022-06-15
fix: Improve sync

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

1. We are fetching all notifications since the beginning of time when first synching.
2. We are prone to missing a few events between fetching the last page of notifications and connecting the websocket.

### Causes

1. We never implemented one step of the Sync, which is "if no event was ever processed, request the last ID from the backend".
2. We first fetch all the notifications, and then we connect to the websocket and start collecting.

#### Don't fetch pending events on first sync.

- If `lastProcessedEventId` is `null`, we fetch the last event from the server.

#### Start collecting pending events AFTER connecting to the websocket

Pretty much it. We first wait for the Websocket to be open before collecting notifications.
This means that we could collect some events twice (it comes in the websocket while we are still fetching some notification pages). In order to avoid it, we store all pending events being collected in a buffer, and everytime we receive a live event through the websocket, we check that buffer.
In order to free memory, we clear the buffer when we process the very last event in the buffer, as they are ordered. So events added previously can be safely removed from the buffer.

### Dependencies

Needs releases with:

- #592

#### Test Coverage

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
```diff
@@ -81,6 +104,23 @@ class EventDataSource(
...
+                    val lastNotificationResult = notificationApi.lastNotification(clientId.value)
+                    if (lastNotificationResult.isSuccessful()) {
+                        Either.Right(lastNotificationResult.value.id)
+                    } else {
+                        Either.Left(NetworkFailure.ServerMiscommunication(lastNotificationResult.kException))
+                    }
+                }
+            )
+        }
+
```
MohamadJaara, line 123: here as well
```suggestion
    override suspend fun lastEventId(): Either<CoreFailure, String> =
        eventInfoStorage.lastProcessedId?.let {
            Either.Right(it)
        } ?: run {
            clientRepository.currentClientId().flatMap { clientId ->
                wrapApiRequest { notificationApi.lastNotification(clientId.value) }.map { it.id }
            }
        }

```

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SlowSyncWorker.kt
```diff
@@ -14,7 +14,8 @@ class SlowSyncWorker(
 
     override suspend fun doWork(): Result {
 
-        val result = userSessionScope.users.syncSelfUser()
+        val result =
+            userSessionScope.users.syncSelfUser()
```
MohamadJaara, line 18:
```suggestion
        val result = userSessionScope.users.syncSelfUser()
```

network/src/commonMain/kotlin/com/wire/kalium/network/api/notification/NotificationApiImpl.kt
```diff
@@ -66,32 +74,46 @@ class NotificationApiImpl internal constructor(
...
+        kaliumLogger.i("Websocket open")
+
+        emit(WebSocketEvent.Open())
+
+        session
+            .incoming
+            .consumeAsFlow()
+            .onCompletion {
+                kaliumLogger.w("Websocket Closed", it)
+                emit(WebSocketEvent.Close(it))
```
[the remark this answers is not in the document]
vitorhugods: The flow will be closed anyway.

We could get it to better understand the reason, but we don't really care, as there's no agreement with the backend about the possible closing reasons.

MohamadJaara (approved):
lgtm 🚀

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncManager.kt
```diff
@@ -194,23 +238,23 @@ class SyncManagerImpl(
 
     override suspend fun waitUntilSlowSyncCompletion() {
         startSyncIfIdle()
-        syncRepository.syncState.first { it is SyncState.ProcessingPendingEvents || it is SyncState.Live }
+        syncRepository.syncState.first { it in setOf(SyncState.GatheringPendingEvents, SyncState.Live) }
```
yamilmedina, line 241: Just out of curiosity, we are guaranteed that` syncRepository.syncState.first` will always have at least one element ? I'm not sure if by any chance this could throw something like `NoSuchElementException` 🤔

yamilmedina (approved):
Great work there! I just left a comment/ question, that maybe it's a possibility that could never happen 🤔  but I prefer to ask =) 

🚀  🍍

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncManager.kt
vitorhugods: Yes. It's a `StateFlow` (requires an initial value) and is initialized with a default value of `WAITING`.

PR #596, opened 2022-06-16
feat: Apply user status to incoming calls

borichellow:
[Task-1060](https://wearezeta.atlassian.net/browse/AR-1060)
----
# What's new in this PR?
Appling `UserAvailabilityStatus` and `MutedConversationStatus` for the incoming calls. 
So the call shouldn't be shown to the user if a user is Away, or the conversation where that call comes from is muted.


gongracr (approved):
Just a small naming suggestion. AWESOME work with the tests 🦾 🤖 🦾

PR #601, opened 2022-06-17
chore: enhancements on managing ongoing and closed calls

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR aims to solve some issues we were having in calling, such as:
- Multiple Incoming Call popup (this might only be fully fixed when the AR PR is merged after this one, they need to work together 💪🏻 )
- Silent exception we were having in `OnCloseCall` handler were sometimes the `clientId` was null and we were not treating it as nullable.
- Not removing the call information from our map holder anymore when closed, which now we keep its correct status and not just removing it.

PS: This is the Kalium part of it, the AR PR will be followed soon, so maybe somethings might not seem to be very clear looking only from this side. But if you have any doubt why something was done in a specific way, please feel free to comment and we can discuss it :)


calling/src/commonMain/kotlin/com/wire/kalium/calling/CallClosedReason.kt
```diff
@@ -0,0 +1,22 @@
...
+    TIMEOUT(avsValue = 2),
+    LOST_MEDIA(avsValue = 3),
+    CANCELLED(avsValue = 4),
+    ANSWERED_ELSEWHERE(avsValue = 5),
+    IOERROR(avsValue = 6),
+    STILL_ONGOING(avsValue = 7),
+    TIMEOUT_ECONN(avsValue = 8),
+    DATA_CHANNEL(avsValue = 9),
+    REJECTED(avsValue = 10),
+    UNKNOWN(avsValue = -1);
```
vitorhugods, line 15: For future reference, I think we should map these guys to something "ours", as most of these errors are useless to the end user and clients consuming Kalium. Something like this, maybe:

```kotlin
sealed interface ReasonForClosingCall {

    /**
     * Call was finished by one of the participants
     */
    object Normal: ReasonForClosingCall

    /**
     * //I'm not sure this is correct: Call was dropped after ringing 
     * for too long without being picked-up
     */
    object Timeout: ReasonForClosingCall

    /**
     * [OneOnOne] call was picked up by this user on another device
     */
    object AnsweredByAnotherClient: ReasonForClosingCall

    /**
     * Receiver rejected the call
     */
    object Rejected: ReasonForClosingCall

    /**
     * Caller cancelled the call
     */
    object Cancelled: ReasonForClosingCall

    /**
     * Something unexpected happened, out of the control of the user
     */
    class Error(val rootCause: Throwable): ReasonForClosingCall // Can be an AvsException
}

class AvsException(message: String): RunTimeException(message)

// AVS-specific errors can be:
ReasonForClosingCall.Error(KaliumAVSException("Got DATA_CHANNEL(avsValue=9) for closing call))
```

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnCloseCall.kt
```diff
@@ -1,18 +1,38 @@
...
+
+        val avsReason = CallClosedReason.fromInt(value = reason)
+        val callStatus = if (avsReason === STILL_ONGOING) CallStatus.ONGOING else CallStatus.CLOSED
+
+        callRepository.updateCallStatusById(
+            conversationId = conversationId.toConversationId().toString(),
+            status = callStatus
+        )
+
+        callingLogger.i("OnCloseCall -> incoming call status for conversation $conversationId updated to $callStatus..")
```
vitorhugods, line 36: Can we start testing this now?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/ObserveConversationListDetailsUseCase.kt
```diff
@@ -10,17 +11,35 @@ import kotlinx.coroutines.flow.map
...
+                when (it) {
+                    is ConversationDetails.Self,
+                    is ConversationDetails.Connection,
+                    is ConversationDetails.OneOne-> it
+                    is ConversationDetails.Group -> it.copy(
+                        hasOngoingCall = (it.conversation.id in calls.map { call -> call.conversationId })
+                    )
+                }
+            }
+        }
```
vitorhugods, line 43: Could you write tests for this?

##### Given an ongoing call, when fetching conversation details, then the conversation should have ongoing call

1. Return a conversation from the mocked ConversationRepo
2. Return an ongoing call on this conversation from the mocked CallRepo
3. Assert it `hasOngoingCall == true`

##### Given a conversation without ongoing call, when fetching conversation details, then the conversation should not have ongoing call

1. Return a conversation from the mocked ConversationRepo
2. Return no ongoing calls from the mocked CallRepo
3. Assert it `hasOngoingCall != true`

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/GetEstablishedCallsUseCase.kt
```diff
@@ -5,12 +5,12 @@ import com.wire.kalium.logic.feature.call.Call
 import com.wire.kalium.logic.sync.SyncManager
 import kotlinx.coroutines.flow.Flow
 
-class GetOngoingCallUseCase(
+class GetEstablishedCallsUseCase(
```
vitorhugods, line 8:
```suggestion
class ObserveEstablishedCallsUseCase(
```

As we agreed on using the prefix `Observe` for Flows.

vitorhugods (approved):
Nicely done 🚀🚀

Just a reminder about naming flow-returning-stuff with `Observe`.

ohassine (approved):
Well done 🚀🚀

PR #610, opened 2022-06-21
refactor: cleanup deprecated Sync use cases

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- `ListenToEventsUseCase` and `SyncPendingEventsUseCase` have been deprecated for long and not needed anymore as all it does is trigger sync (something other UseCases already do if needed)

### Solutions

- Delete them. Delete the usage on Reloaded too.

### Dependencies

As soon as this PR is merged, we need to merge the one on Reloaded: https://github.com/wireapp/wire-android-reloaded/pull/690

### Testing

N/A

----

----
##### References


saleniuk (approved):

ohassine (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/WaitUntilLiveUseCase.kt
```diff
@@ -0,0 +1,17 @@
+package com.wire.kalium.logic.sync
+
+import com.wire.kalium.logic.data.sync.SyncState
+
+/**
+ * Starts sync (if not yet started) and suspends until all pending events were processed.
+ * Will resume when all pending events are processed and the client is considered "online".
+ * @see SyncState
+ */
+class WaitUntilLiveUseCase(
```
yamilmedina, line 10: Is this going to be used in reloaded?  If that's the case, maybe we can extract an interface out of this and make the implementation internal. What do you think ? 😬

yamilmedina (approved):
Looking good! Just a small not blocking comment about a nice to have for us, going in the same direction about usecases =D

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/WaitUntilLiveUseCase.kt
vitorhugods: The idea was that, to expose a way to suspend until the app is online and only then process the notifications.

But perhaps we can instead adapt the UseCases to do it, and Reloaded won't need to call this.

I kinda prefere this second way. I'll adapt this PR to only remove the deprecated UseCases, and we can figure out the `waitUntilLive` later.

PR #624, opened 2022-06-23
feat(calling): sort participants alphabetically

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We are passing non ordered participants list in a call

### Solutions

Order it alphabetically

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

On AR, start a group call to check if the titles are ordered.

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
@@ -179,9 +179,10 @@ internal class CallDataSource(
         callProfile[conversationId]?.let { call ->
             callingLogger.i("updateCallParticipants() - conversationId: $conversationId with size of: ${participants.size}")
 
+            val sortedParticipants = participants.sortedBy { it.name }
             val updatedCalls = callProfile.calls.toMutableMap().apply {
                 this[conversationId] = call.copy(
-                    participants = participants,
+                    participants = sortedParticipants,
                     maxParticipants = max(call.maxParticipants, participants.size + 1)
                 )
             }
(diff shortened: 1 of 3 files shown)


mchenani (commented):
I have a general question about your changes, isn't better to have this kind of structure in the ui? e.g what if we want to sort the base on other conditions? and have this condition as an option in the UI? I think it be better the leave this kind of decision out of the repo and the logic layer, and implement it in the UI layer.

ohassine:
> I have a general question about your changes, isn't better to have this kind of structure in the ui? e.g what if we want to sort the base on other conditions? and have this condition as an option in the UI? I think it be better the leave this kind of decision out of the repo and the logic layer, and implement it in the UI layer.

I don't really agree, this is a shared business logic that should apply to all different clients. We don't need to change on each client if the busniess requirement changes.

But yeah sorting shouldn't be done in repositories, those ones should be kept dummy classes without a business logic.
I will refactor this PR..

mchenani:
> > > I have a general question about your changes, isn't better to have this kind of structure in the ui? e.g what if we want to sort the base on other conditions? and have this condition as an option in the UI? I think it be better the leave this kind of decision out of the repo and the logic layer, and implement it in the UI layer.
> > 
> > 
> > I don't really agree, this is a shared business logic that should apply to all different clients. We don't need to change on each client if the busniess requirement changes.
> > But yeah sorting shouldn't be done in repositories, those ones should be kept dummy classes without a business logic. I will refactor this PR..
> 
> If this is an wanted behaviour to be shared by other platforms, I guess it makes sense. The UI can still re-order in a fancy way if they so desire.
> 
> Could you please add tests for this sorting? Add two participants and check if they are ordered correctly when fetching the UseCase.

Thanks for clarifying it @ohassine, now if we want to include the sort in the usecase, I suggest to pass comparable object to get the order base on that! 
Take a look on that: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-comparable/
By the way!, I don't want to block such a simple task with over engineering, what you've done already is good enough.

mchenani (approved):
LGTM🚀, also I added one general comment

alexandreferris (approved):

PR #637, opened 2022-06-28
fix: params setup for pagination support in api

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Border cases, right now we had some code and logic for calling paginated endpoints, but we were not setting this correctly on the request body, so we were stuck in a loop trying to get always more pages. 

This fixes the issue, creating a request class that can be used in paginated endpoints, right now: Conversations and Connections.

Thanks to @vitorhugods for the detailed eagle eye 🦅 

----

----

----
##### References

### DIFF
--- a/network/src/commonMain/kotlin/com/wire/kalium/network/api/pagination/PaginationRequest.kt
+++ b/network/src/commonMain/kotlin/com/wire/kalium/network/api/pagination/PaginationRequest.kt
@@ -1,0 +1,12 @@
+package com.wire.kalium.network.api.pagination
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class PaginationRequest(
+    @SerialName("paging_state")
+    val pagingState: String?,
+    @SerialName("size")
+    val size: Int? = null // Set in case you want specific number of pages, otherwise, the backend will return default per endpoint
+)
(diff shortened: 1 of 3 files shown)


network/src/commonMain/kotlin/com/wire/kalium/network/api/pagination/PaginationRequest.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class PaginationRequest(
+    @SerialName("paging_state")
+    val pagingState: String?,
+    @SerialName("size")
+    val size: Int? = null // Set in case you want specific number of pages, otherwise, the backend will return default per endpoint
```
trOnk12, line 11: Perhaps worth to mention what the default size is ? I even wonder if we want to relay on backend default size?

trOnk12 (approved):
First step to pagination yey

network/src/commonMain/kotlin/com/wire/kalium/network/api/pagination/PaginationRequest.kt
yamilmedina: It depends, from endpoint to endpoint, for example connection requests default size is 100, for conversations 1000 😅 
So the idea is you can handle on the client/endpoint if you want something specific

trOnk12: Small details but what would be the case to relay on backend default values wouldnt it make more sense to have it set on the clients?

PR #645, opened 2022-07-01
fix(calling): fix incoming call flow

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

IncomingCall flow returns the same value many times after accepting a call.
This causes some issues on the clients like displaying the same incoming call screen multiple times.

### Causes (Optional)

After accepting a call, the Participants callback will occur causing some changes on the `callsFlow`(since we are updating the participants list there). The old and the new values of the incoming call is not the same, so the new value will be emitted in the flow

### Solutions

Filter also by participants, incoming calls should not have an empty list of participants

Needs releases with:

- [x] GitHub link to other pull request

#### How to Test

On the Android app, receive a 1:1 call and accept it
You should not see incoming call screen twice

----

----

----
##### References


alexandreferris (approved):
nice catch 👌🏻

PR #648, opened 2022-07-01
fix: incoming call screen sometimes not appearing for group calls

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

On Android app, the incoming call screen is not appearing when we have an incoming call

### Causes (Optional)

It's a regression of this PR : [fix(calling): fix incoming call flow](https://github.com/wireapp/kalium/pull/645)

### Solutions

Using `distinctUntilChanged` to ensure that we are not emitting  the same incoming call many times
We need to check by id only because for group calls we can have already a list of participants that is not empty

Needs releases with:

#### How to Test

Receive 1:1  and group calls

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/GetIncomingCallsUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/GetIncomingCallsUseCase.kt
@@ -51,7 +51,9 @@ internal class GetIncomingCallsUseCaseImpl(
             .flatMapLatest {
                 //if user is AWAY we don't show any IncomingCalls
                 if (it.availabilityStatus == UserAvailabilityStatus.AWAY) flowOf(listOf())
-                else callRepository.incomingCallsFlow()
+                else callRepository.incomingCallsFlow().distinctUntilChanged {
+                        old, new -> old.firstOrNull()?.conversationId == new.firstOrNull()?.conversationId
+                }
             }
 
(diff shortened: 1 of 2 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/GetIncomingCallsUseCase.kt
```diff
@@ -51,7 +51,9 @@ internal class GetIncomingCallsUseCaseImpl(
             .flatMapLatest {
                 //if user is AWAY we don't show any IncomingCalls
                 if (it.availabilityStatus == UserAvailabilityStatus.AWAY) flowOf(listOf())
-                else callRepository.incomingCallsFlow()
+                else callRepository.incomingCallsFlow().distinctUntilChanged {
+                        old, new -> old.firstOrNull()?.conversationId == new.firstOrNull()?.conversationId
```
alexandreferris, line 55: Aren't both conversationId from the `old` and the `new` always the same if we are receiving a call from the same conversation? 🤔

ohassine: yes right, but they might have different participants..That't why `distinctUntilChanged()` will not work and we need to check by `conversationId`

alexandreferris (approved):

PR #649, opened 2022-07-04
Feat: refactor assets path [AR-1578]

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In our current asset management implementation we were storing and handling assets in their raw form as blobs of binary data / Byte arrays. This presented an immediate limitation which is that we couldn't handle large asset files. Therefore we needed to move to something more scalable. Also we couldn't use any of the Java `File` methods as these were not KMP friendly.

### Solutions

We decided to bring to the house a new library called `Okio`, provided by the Square team, which was already working under the hood (OkHttp uses it to manage all the internal networking calls). 
This library allows to handle file read & write operations, sending and consuming data streams, and also provided multiplatform methods for encryption and decryption, which would solve a lot of coding hours for other platform targets implementing these encryption algorithms. 

#### Test Coverage (Optional)

### Notes (Optional)

I made extensive code conflicts solving to the extent of my abilities, so some tiny parts might have been rewritten or removed. So please be gentle when reviewing :)

Also, resampling/pre-compression of large images is still not implemented, but will hopefully come soon too.

### Attachments (Optional)

https://user-images.githubusercontent.com/2468164/177191897-82e2549a-095c-4aa7-a850-d8e299106ef9.mp4

----

----

----
##### References


logic/src/jvmMain/kotlin/com/wire/kalium/logic/data/asset/KaliumFileSystem.kt
```diff
@@ -34,4 +43,57 @@ actual class KaliumFileSystem : FileSystem() {
 
     override fun openReadWrite(file: Path, mustCreate: Boolean, mustExist: Boolean): FileHandle =
         SYSTEM.openReadWrite(file, mustCreate, mustExist)
+
+    /**
+     * Provides the root of the cache path, used to store temporary files
+     */
+    actual val rootCachePath: Path = dataStoragePaths.cachePath.value.toPath()
```
typfel, line 50: We need to make sure this directory gets created at a suitable place otherwise the CLI app crashes when it tries to write an asset file.

logic/src/jvmMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -51,6 +55,9 @@ actual class CoreLogic(
             val rootAccountPath = "$rootPath/${userId.domain}/${userId.value}"
             val rootProteusPath = "$rootAccountPath/proteus"
             val rootStoragePath = "$rootAccountPath/storage"
+            val rootFileSystemPath = AssetsStorageFolder(rootStoragePath)
+            val rootCachePath = CacheFolder("cache/$rootAccountPath")
```
typfel, line 59:
```suggestion
            val rootCachePath = CacheFolder("$rootAccountPath/cache")
```

typfel (changes requested):
Some issues I found when trying to run the CLI app

logic/src/jvmMain/kotlin/com/wire/kalium/logic/data/asset/KaliumFileSystem.kt
typfel: You resolved this but I don't see any code which makes sure the directory exists.

gongracr: It should be pushed now

cli/src/main/kotlin/com/wire/kalium/cli/ConversationsApplication.kt
```diff
@@ -73,6 +73,9 @@ class ConversationsApplication : CliktCommand() {
     private val cliFileSystem = CliFileSystem(DataStoragePaths(assetsStoragePath, cachePath))
 
     override fun run(): Unit = runBlocking {
+        cliFileSystem.createDirectory(cachePath.value.toPath())
```
typfel, line 76: The `ConversationApplication` is some legacy app which I don't know if we even want to maintain. The CLI app is called `CLIApplication`.

Also would make sense maybe to create the directory somewhere so that we don't have to repeat this setup code in each app which uses Kalium?

typfel:
>I created a platform dependent class called KaliumFileSystem that is basically a wrapper around the Okio FileSystem class, providing an initial implementation on each platform. JS implementation will be slightly more complicated as they handle I/O operations and access to storage in a different way than JVM/Android.

`KaliumFileSystem` still exposes all of the Okio FileSystem internals and on the Okio website it says:

> FileSystem: Available on all platforms. For JavaScript this requires [Node.js](https://nodejs.org/api/fs.html).

Which means it won't run the browser.

Are we sure it's possible implement the JS version?

gongracr:
I'm not really an expert on JS nor the current implications we have re: using Node.js. By 
> Which means it won't run the browser.
you meant that using Node.js in our JS variant is not a possibility? Then I guess we will need to give it an extra spin to this proposed solution. Maybe creating an extra wrapper class around `KaliumFileSystem`

vitorhugods:
> I'm not really an expert on JS nor the current implications we have re: using Node.js. By
> 
> > Which means it won't run the browser.
> 
> you meant that using Node.js in our JS variant is not a possibility? Then I guess we will need to give it an extra spin to this proposed solution. Maybe creating an extra wrapper class around `KaliumFileSystem`

To give more context:
NodeJS is a runtime environment created to run JavaScript _outside_ browsers. So you can run JS in backends, etc.

But Kalium wants to make JS available for our web clients, which run in browsers or Electron (Browser++ 😃).



----

Random ideas, not sure if they make sense:

1. Can we take the whole `FileSystem` away? Maybe just work with streams? Then each platform passes a stream to Kalium to send attachments? Or we can define `expect/actual` so each platform can pass its own thing and Kalium transforms to a stream.


2. Or, following the line of wrapper around `KaliumFileSystem`: an "in-memory" or

vitorhugods:
Also, considering the amount of work already put here. Depending on the complexity of this JavaScript work-around, I'd definitely keep my eyes closed for JavaScript right now. As we don't even support JavaScript in `logic` module at the moment, and there probably will be other issues anyway when the time comes.

typfel:
> Can we take the whole `FileSystem` away? Maybe just work with streams? Then each platform passes a stream to Kalium to send attachments? Or we can define `expect/actual` so each platform can pass its own thing and Kalium transforms to a stream.

Like this idea, then potentially the JS version could have an in-memory implementation of the streams if the browser doesn't support file system access.

typfel:
> A completely different issue that with these changes is that we start storing assets unencrypted on disk.

Maybe a simple solution would to only keep the encrypted version of the asset on disk and decrypt it on demand when accessing the asset using the `ChipherSource`?

gongracr:
So if I understand correctly, the issue here will be that JS won't be able to resolve / make use of any of the Okio `FileSystem` methods? But wouldn't be enough if the JS platform simply had its own implementation of the `FileSystem` or as Jacob referred to, in-memory implementation of the streams conversion?

typfel:
Look good to me now! Let's get the tests working and I think someone else you also review this massive PR.

typfel (approved):
🚀

cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/utils/AESUtils.kt
```diff
@@ -1,13 +1,53 @@
...
+        } catch (e: Exception) {
+            kaliumLogger.e("There was an error while encrypting the asset:\n $e}")
+        }
+        return sizeWithPaddingAndIV(encryptedDataSize)
+    }
+
+    private fun sizeWithPaddingAndIV(size: Long): Long {
+        val ivSize = 16L
+        val ivWithPaddingSize = 32L
+        return size + (ivWithPaddingSize - (size % ivSize))
```
comawill, line 47:
```suggestion
    private fun sizeWithPaddingAndIV(size: Long): Long {
        // IV + data + pkcs7 padding
        return IV_SIZE + size + (AES_BLOCK_SIZE - (size % AES_BLOCK_SIZE))
```

cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/utils/AESUtils.kt
```diff
@@ -44,7 +114,8 @@ internal class AESDecrypt(private val secretKey: AES256Key) {
         val decryptedData = cipher.doFinal(encryptedData.data)
 
         // We ignore the first 16 bytes as they are reserved for the Initialization Vector
-        return PlainData(decryptedData.copyOfRange(16, decryptedData.size))
+        val ivSize = 16
+        return PlainData(decryptedData.copyOfRange(ivSize, decryptedData.size))
```
comawill, line 118:
```suggestion
        return PlainData(decryptedData.copyOfRange(IV_SIZE, decryptedData.size))
```

comawill (changes requested):
Please add initialization and usage of IVs for AES

cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/utils/AESUtils.kt
```diff
@@ -1,13 +1,53 @@
...
+        var encryptedDataSize = 0L
+        try {
+            // Fetch AES256 Algorithm
+            val cipher = Cipher.getInstance(KEY_ALGORITHM_CONFIGURATION)
+
+            // Parse Secret Key from our custom AES256Key model object
+            val symmetricAESKey = SecretKeySpec(key.data, 0, key.data.size, KEY_ALGORITHM)
+
+            // Init the encryption
+            cipher.init(Cipher.ENCRYPT_MODE, symmetricAESKey)
```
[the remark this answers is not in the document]
typfel: Seems on newer Android they generate a random IV if left out. But yes, safer to do it ourselves.

https://doridori.github.io/Android-Security-Beware-of-the-default-IV/#sthash.bpbcHfIN.dpbs

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/Asset.kt
```diff
@@ -48,11 +28,20 @@ enum class RetentionType {
...
-    object PNG : ImageAsset(name = "image/png")
+sealed class AssetType(open val mimeType: String)
+data class FileAsset(val fileExtension: String) : AssetType("file/$fileExtension")
+sealed class ImageAsset(override val mimeType: String) : AssetType(mimeType) {
+    object JPEG : ImageAsset(mimeType = "image/jpeg")
+    object JPG : ImageAsset(mimeType = "image/jpg")
+    object PNG : ImageAsset(mimeType = "image/png")
 }
 
+fun isValidImage(mimeType: String): Boolean =
```
Garzas, line 39: Should we support also HEIC for iOS, because I don't have full context here 😅

cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/utils/AESUtils.kt
```diff
@@ -23,16 +63,46 @@ internal class AESEncrypt {
...
+        var size = 0L
+        try {
+            // Fetch AES256 Algorithm
+            val cipher = Cipher.getInstance(KEY_ALGORITHM_CONFIGURATION)
+
+            // Parse Secret Key from our custom AES256Key model object
+            val symmetricAESKey = SecretKeySpec(secretKey.data, 0, secretKey.data.size, KEY_ALGORITHM)
+
+            // Init the decryption
+            cipher.init(Cipher.DECRYPT_MODE, symmetricAESKey, IvParameterSpec(ByteArray(16)))
```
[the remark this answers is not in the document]
gongracr: I did add already some custom IV random initialisation

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/Asset.kt
gongracr: I added the heic as valid inline images. Only problem there still is that web client doesn't seem to like these images format and doesn't know how to render them correctly.

comawill (approved):

vitorhugods (changes requested):
Good to see this in its final stretch!

I think some unwanted stuff were added to the `baseline.xml` file by accident. So I had some styling comments.

And one question about passing the `storePath` and `cacheAssetPath` to the `UserDatabaseProvider`.

Other than that, great job!

persistence/src/androidMain/kotlin/com/wire/kalium/persistence/db/UserDatabaseProvider.kt
```diff
@@ -61,6 +64,8 @@ actual class UserDatabaseProvider(
                 db.execSQL("PRAGMA foreign_keys=ON;")
             }
         }
+        storePath.mkdirs()
+        cacheAssetPath.mkdirs()
```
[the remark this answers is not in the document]
gongracr: I decided to create them here so that we synchronise all the db / storage folders creation. These folders are used to provide assets persistent and temporary storage folders, depending on their nature & purpose.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/id/PlainId.kt
```diff
@@ -5,4 +5,10 @@ import kotlin.jvm.JvmInline
 @JvmInline
 value class PlainId(val value: String)
 
+@JvmInline
+value class CacheFolder(val value: String)
+
+@JvmInline
+value class AssetsStorageFolder(val value: String)
```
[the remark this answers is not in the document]
gongracr: Moved them to `DataStoragePaths`

persistence/src/androidMain/kotlin/com/wire/kalium/persistence/db/UserDatabaseProvider.kt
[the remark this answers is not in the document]
vitorhugods: Couldn't they be created on demand?

Like, if I want to write a file `app/cache/file1.jpg`, the first action would be creating `app/cache/` directories if needed, and throw a failure if couldn't create them, instead of a silent failure here.

Putting this inside DB is kinda unexpected. They're just kinda hanging around.
DB essentially has nothing to do with these folders.

There's no guarantee that I will write files to this `storePath` and `cacheAssetPath`.

One could create a `UserDatabaseProvider` with a `cacheAssetPath = "app/cache"`. And then attempt to save files to `"app/files"` without any idea about what went wrong.


Also, this same operations seem to be done in `UserSessionScopeCommon.kt#452`:
```kt
    val kaliumFileSystem: KaliumFileSystem by lazy {
        // Create the cache and asset storage directories
        KaliumFileSystemImpl(dataStoragePaths).also {
            if (!it.exists(dataStoragePaths.cachePath.value.toPath()))
                it.createDirectory(dataStoragePaths.cachePath.value.toPath())
            if (!it.exists(dataStoragePaths.assetStoragePath.value.toPath()))
                it.createDirectory(dataStoragePaths.assetStoragePath.value.toPath())
        }
    }
```

Which I kinda prefer, as at least these are handled around the `FileSystem` itself, and not DB classes.

PR #656, opened 2022-07-06
feat: avs federation calling support (AR-1680) 

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Adds support for federated calls, passing to AVS the parsed id, this is: 
- if federated, always with domain
- if not federated, without domain

https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/555090587/Use+case+deciding+which+backend+API+version+to+use#Step-5%3A-AVS-and-API-versioning

### Solutions

Created a component `FederatedIdMapper` that will handle the parsing of the ids, depending on the config prev. stored from `/api-version` endpoint. These values are stored in kalium preferences, so we can use it without using coroutines, and reuse it in several parts of the lib

#### Test Coverage (Optional)

#### How to Test

You should be able to call in federated environments and no federated environments. Prev. to this PR. you can't start or receive calls in federated environments, ie: anta, bella

----

----

----
##### References


logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
```diff
@@ -84,24 +86,31 @@ actual class CallManagerImpl(
...
     private fun startHandleAsync(): Deferred<Handle> = scope.async(start = CoroutineStart.LAZY) {
-        val selfUserId = userId.await().value
+        val selfUserId = userId.await()
         val selfClientId = clientId.await().value
 
         val waitInitializationJob = Job()
 
         val handle = calling.wcall_create(
-            userId = selfUserId,
+            userId = federatedIdMapper.parseToFederatedId(selfUserId),
```
alexandreferris, line 95: Seems like we are only using `selfUserId` to be mapped, maybe we can move the `federatedIdMapper.parseToFederatedId(..` directly to line 89,, such as:

`val selfUserId = federatedIdMapper.parseToFederatedId(userId.await())`

what do you think?

alexandreferris (approved):
Great work @yamilmedina ! I just have one tiny suggestion

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnParticipantListChanged.kt
```diff
@@ -34,9 +41,13 @@ class OnParticipantListChanged(
 
         val participantsChange = Json.decodeFromString<CallParticipants>(data)
         callingScope.launch {
+            val memberList: List<Member> = conversationRepository
+                .observeConversationMembers(remoteConversationIdString.parseIntoQualifiedID())
+                .first()
+
```
ohassine, line 47: I am wondering why getting member list ? `participantsChange` already have list of participants with their ids

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/id/FederatedIdMapperImpl.kt
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Mapper that enables parsing [QualifiedID] into a [String] having in consideration federation.
+ *
+ * In detail, if [isFederationEnabled] is [true] then the full qualified form will be used
+ * otherwise the plain value will be used
+ */
+class FederatedIdMapperImpl(private val kaliumPreferences: KaliumPreferences) : FederatedIdMapper {
+
+    private fun isFederationEnabled() = kaliumPreferences.getBoolean(FEDERATION_ENABLED, false)
+    private fun getCurrentDomain() = kaliumPreferences.getString(CURRENT_DOMAIN) ?: "wire.com"
```
ohassine, line 23: could you move this hardcoded value into const ? `DEFAULT_DOMAIN` for example ?

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnIncomingCall.kt
```diff
@@ -36,7 +37,7 @@ class OnIncomingCall(
             callRepository.createCall(
                 conversationId = conversationId.toConversationId(),
                 status = CallStatus.INCOMING,
-                callerId = userId,
+                callerId = federatedIdMapper.parseToFederatedId(userId),
```
ohassine, line 40: Here we are parsing the id, we need to do the same I think in the other `callRepository.createCall()`  when we start a call ?

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnParticipantListChanged.kt
```diff
@@ -48,7 +59,6 @@ class OnParticipantListChanged(
             }
 
             callRepository.updateCallParticipants(
-                //TODO should be handled properly after supporting federated calls
                 conversationId = remoteConversationIdString.toConversationId().toString(),
```
ohassine, line 62: Since we have a mapper for federationId, could move all extension functions in QualifiedID to that class ?

ohassine (commented):
Good one 🚀 I just left some questions and suggestions

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnParticipantListChanged.kt
yamilmedina: The problem here is that the member list ids come without domain when non federated. So we need these ids fully qualified later on to fetch user details, ie: name, avatar, etc. Because the endpoint to fetch `userInfo` doesn't support not sending a domain :(

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnParticipantListChanged.kt
yamilmedina: I'll prefer to leave it there, since the moving the logic, will imply to touch many classes. Maybe we can address that in other PR, when needed :D

PR #667, opened 2022-07-11
fix: Sent message moves conversation to the top

borichellow:
# What's new in this PR?

Fix for [Bug-1985 ](https://wearezeta.atlassian.net/browse/AR-1985)

### Issues

when user sends message, conversation doesn't move to the top of the list

### Causes (Optional)

`updateConversationModifiedDate` was not called

### Solutions

call `updateConversationModifiedDate` on sending the message


MohamadJaara:
nice work,

I was considering implementing this in a clear way that can also provide some fine control over which messages update the conversation order 
where we add a boolean value to the `Message` model
```kotlin
sealed class Message(
    ....
) {
    abstract val updateConversationOrder: Boolean

    data class Regular(
        .....
    ) : Message(id, content, conversationId, date, senderUserId, status, visibility) {
        override val updateConversationOrder: Boolean
            get() = true
    }

    data class System(
        .....
    ) : Message(....) {
        override val updateConversationOrder: Boolean
            get() = when(content) {
                is MessageContent.MemberChange.Added -> false
                is MessageContent.MemberChange.Removed -> false
                MessageContent.MissedCall -> true
            }
    }
....

and we create a new inernal useCase to insert message to the DB that will do the check and update the order when nessery
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/SendAssetMessageUseCase.kt
```diff
@@ -124,7 +127,12 @@ internal class SendAssetMessageUseCaseImpl(
             status = Message.Status.PENDING,
             editStatus = Message.EditStatus.NotEdited
         )
-        messageRepository.persistMessage(message).map { message }
+        messageRepository.persistMessage(message)
+            .map { message }
+            .onSuccess {
+                conversationRepository.updateConversationNotificationDate(message.conversationId, message.date)
```
MohammedMokresh, line 133: maybe we can have update date function in the conversationRepository and call these 2 functions in it ? 
So we won't forget to call one and not the other in the future cases !

borichellow: i think this problem is solved by new changes that Mohamad suggested :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/PersistMessageUseCase.kt
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Internal UseCase that should be used instead of MessageRepository.persistMessage(Message)
+ * It automatically updates ConversationModifiedDate and ConversationNotificationDate if needed
+ */
+interface PersistMessageUseCase {
+    suspend operator fun invoke(message: Message): Either<CoreFailure, Unit>
+}
+
+class PersistMessageUseCaseImpl(
```
MohamadJaara, line 17:
```suggestion
internal interface PersistMessageUseCase {
    suspend operator fun invoke(message: Message): Either<CoreFailure, Unit>
}

internal class PersistMessageUseCaseImpl(
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
```diff
@@ -29,6 +29,10 @@ import kotlinx.coroutines.flow.map
 
 @Suppress("TooManyFunctions")
 interface MessageRepository {
+    @Deprecated(
+        "shouldn't be used directly",
```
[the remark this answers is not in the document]
borichellow: yeah, but i didn't find any better way to inform everybody who will want to use it directly to do not do it.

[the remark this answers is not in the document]
MohammedMokresh: just comment on it , and in future someone used that they will see it , or we can detect that while reviewing

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/PersistMessageUseCase.kt
```diff
@@ -0,0 +1,49 @@
...
+                    conversationRepository.updateConversationModifiedDate(message.conversationId, message.date)
+
+                if (userId == message.senderUserId)
+                    conversationRepository.updateConversationNotificationDate(message.conversationId, message.date)
+            }
+    }
+
+    private fun MessageContent.shouldUpdateConversationOrder(): Boolean =
+        when (this) {
+            is MessageContent.MemberChange -> false
```
MohamadJaara, line 36:
```suggestion
            is MessageContent.MemberChange -> true
```

As I remember from wolfgang the `MemberChange` messages should update the conversation list order

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
```diff
@@ -29,6 +29,10 @@ import kotlinx.coroutines.flow.map
...
+        replaceWith = ReplaceWith("com.wire.kalium.logic.data.message.PersistMessageUseCase")
+    )
```
MohamadJaara, line 35: Since it's used in the internal usecase you don't need to deprecate it

MohamadJaara (changes requested):
Great work as always 🚀, only some minor changes to when the conversation order is updated and removing deprecation from the essageRepository.persistMessage() function

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/PersistMessageUseCase.kt
borichellow: it can't be `internal` as it's used in some other packages

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -190,14 +192,17 @@ abstract class UserSessionScopeCommon(
...
+
     private val callRepository: CallRepository by lazy {
         CallDataSource(
             callApi = authenticatedDataSourceSet.authenticatedNetworkContainer.callApi,
             conversationRepository = conversationRepository,
             userRepository = userRepository,
             teamRepository = teamRepository,
             timeParser = timeParser,
-            messageRepository = messageRepository
+            persistMessage = persistMessage
```
typfel, line 205: Hmm but now we are injecting use cases into repositories, or is the only place and the CallDataSource should refactored?

borichellow:
waiting for @typfel approve :)

PR #671, opened 2022-07-11
fix: conversations not loading on missing/deleted members (AR-2000)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

https://wearezeta.atlassian.net/browse/AR-2000

### Issues

When loading big old accounts, there is a chance that one to one conversations doesn't have the other end (deleted account for example). So this PR fixes the issues, by emitting a null values, so the flow of data can end and display the information that does exists.

This is a temporary fix, we should align with product later on to see how we should handle this cases, (displaying deleted user for example, or continue in this direction, not showing this conversations)

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -202,38 +205,41 @@ class ConversationDataSource(
...
-            ConversationDetails.Group(
-                conversation, LegalHoldStatus.DISABLED // TODO(user-metadata): get actual legal hold status
-            )
-        )
-        // TODO(connection-requests): Handle requests instead of filtering them out
+        // TODO(user-metadata): get actual legal hold status
+        Conversation.Type.GROUP -> flowOf(ConversationDetails.Group(conversation, LegalHoldStatus.DISABLED))
         Conversation.Type.CONNECTION_PENDING, Conversation.Type.ONE_ON_ONE -> {
             val selfUser = userRepository.observeSelfUser().first()
+            when (val conversationMembers = getConversationMembers(conversation.id)) {
```
MohammedMokresh, line 212: why we go here with this approach to create a variable and pass it to the function 
isn't keep the fold here and pass the actual value not the Either to the success and error functions  better?

yamilmedina: I found it to be more clear in terms of readability and helped to debug better. But doesn't make any difference in terms of the bug fix. 

I'll try to find a better way to write it, if I'm not able, I'll just rollback these changes =D

MohammedMokresh (approved):

PR #682, opened 2022-07-14
chore: run instrumented test only on modules affected

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

This PR aims to reduce the amount of load used in our CI macs VMs, since we have limited resources, sometimes it doesn't makes sense to run instrumented tests, for example if no code was touched in a module that affects its dependants.

This has one limitation, as stated in `dag-command` :

> At the moment, this project doesn't detect changes in the root folder, so the configuration of the gradle modules should also be a module, it is recommended the buildSrc folder. When the pluggin detects changes in the buildSrc folder, it considers that all modules were affected by the PR.

So this is the reason why for this PR it will run everything

### Dependencies (Optional)

https://github.com/leandroBorgesFerreira/dag-command

----

----

----
##### References


buildSrc/src/main/java/OnlyAffectedConnectedTestTask.kt
```diff
@@ -0,0 +1,55 @@
...
+                    executable("./gradlew")
+                }
+            }
+        }
+    }
+
+    private companion object {
+        // makes sense to maintain dependencies in a logical order of execution
+        private val targetModules = listOf("cryptography", "persistence", "logic")
+        private const val testTaskSuffix = "connectedAndroidTest"
```
vitorhugods, line 52: It shouldn't be hard to replicate this for all tests, right?

iOS, JVM and JS for example.

I guess the issue is that not all modules have iOS and JS for example. But in general I guess we can make an `abstract` task and then just change the `targetModules` and `testTaskSuffix`.

Or, maybe we can figure out if the module has iOS and JS. 🧐 
I'll leave that to future versions of us to figure out.

vitorhugods (changes requested):
Overall I think it's a great change.

But I was wondering "What about Gradle configuration changes?"

I gave a look on the [readme of dag-command](https://github.com/leandroBorgesFerreira/dag-command#limitations), and turns out we need to handle that.

Can we check if `buildSrc` changed, and if so we test everything?

It still doesn't cover EVERYTHING. Like `build.gradle.kts` changes. But I think those are more likely to lead to compilation errors (in other platforms than Android too) than runtime errors.

buildSrc/src/main/java/OnlyAffectedConnectedTestTask.kt
yamilmedina: Makes total sense. I was wondering the same with iOS, at least I can give it a shot to apply this also for iOS tests (our other slow brother). Since we can read configuration from proj objt, I think it's possible to get if a module has this tests or not 🤔

vitorhugods: Get test by name should be really simple actually.

```kotlin
tasks
    .firstOrNull { it.name.equals("iOSTest", true }
    ?.execute()
```

yamilmedina: Ok, made some changes, now we are reusing some code, doing some "auto" config using some Kotlin DSL goodies.

I'm not positive about applying it to jvm tests since we could mess with the coverage 🤔 . We can research later, but anyway those tests are the fastest now =) Let me know what you think :D

PR #684, opened 2022-07-14
feat: refill key packages after sync

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Automatically refill key packages when they run out after the quick sync finishes. This partially implements the [use case](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/556499553/Use+case+refill+key+packages+MLS?focusedCommentId=612860005&src=mail&src.mail.action=view&src.mail.notification=com.atlassian.confluence.plugins.confluence-notifications-batch-plugin%3Abatching-notification&src.mail.recipient=8a7f80897280744b0172850c5c4d0179&src.mail.timestamp=1657706696293)

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/androidMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -100,6 +100,7 @@ actual class CoreLogic(
                 globalPreferences.value,
                 kaliumConfigs
             ).also {
+                it.keyPackageManager.startObservingKeyPackageCount()
```
typfel, line 103: Would love suggestions for a better place to start the observation. Currently this is logic duplicated in all targets.

typfel: >I thought starting these in the constructor would be nice enough. But I can see, just like for Sync start/stop, that at some point they must stop. But I question myself: When?

I had them in the constructor initially but then I can't do a `get() = KeyPackageManagerImpl()` in the `UserSessionScope` because then it would never start. So having them in the constructor forces us to initiate all dependencies immediately.

>When should we stop these jobs/async stuff? When the client applications asks us to? Because they went to the background or something? When do we stop?!

I don't we think we need necessarily need to start & stop these jobs, at least not initially. They can stop when the UserSessionScope gets destroyed. It would be future improvement to be able to stop them based certain cases like app going to the background.

typfel: I'll change it so we start the observing from the constructor with the lazy pattern.

typfel: Done

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -287,6 +289,13 @@ abstract class UserSessionScopeCommon(
             authenticatedDataSourceSet.authenticatedNetworkContainer.notificationApi, eventInfoStorage, clientRepository
         )
 
+    internal val keyPackageManager: KeyPackageManager
+        get() = KeyPackageManagerImpl(
```
vitorhugods, line 293: By keeping it in a `get() =` it will never be built unless `keyPackageManager` is called.

Maybe just:

```kotlin
internal val keyPackageManager: KeyPackageManager = KeyPackageManagerImpl(
    syncRepository,
    lazy { keyPackageRepository },
    lazy { client.refillKeyPackages }
)
```

This way it will start observing `syncRepository` right away, and internally will only call the dependencies when needed.

typfel: Yes! 🤦 missed removing that.

vitorhugods (approved):
🚀🚀🚀🚀

PR #696, opened 2022-07-20
chore: migrate verbose logging now with flowId (AR-1872)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

This adds some context to the current logging. To achieve this, we are using a TAG to correctly log the application user/flow:
I came up with this list, but we can be either more generic or more specific:

 - SYNC
 - EVENT_RECEIVER
 - CONVERSATIONS 
 - CONNECTIONS
 - MESSAGES
 - SEARCH
 - LOGIN
 - REGISTER
 - CLIENTS
 - CALLING

### Causes (Optional)

If we want to opt-in to have more detailed logs, we'll have to make some adjustments in calls to logs

### Solutions

Use the `withTag` function as a middle step to log, already provided by the Kermit lib

### Notes

Next PR's will come to adjust the rest of the logging levels (error, info, warn, debug). But first I want to validate this approach

----

----

----
##### References


alexandreferris (commented):
Great work @yamilmedina !

I just have a question, will we be needing to add `.withFlowId(APPLICATION_FLOW)` to every log?

yamilmedina:
> Great work @yamilmedina !
> 
> I just have a question, will we be needing to add `.withFlowId(APPLICATION_FLOW)` to every log?

Oh! Sadly yes. 

I was expecting to do something like an [MDC context ](https://www.slf4j.org/apidocs/org/slf4j/MDC.html#put(kotlin.String,kotlin.String)) here (we have it on `Thread` world, thanks to `ThreadLocal`), but here we can't since Coroutines are different, but most important we are doing logs outside coroutines

vitorhugods:
> > Great work @yamilmedina !
> > I just have a question, will we be needing to add `.withFlowId(APPLICATION_FLOW)` to every log?
> 
> Oh! Sadly yes.
> 
> I was expecting to do something like an [MDC context ](https://www.slf4j.org/apidocs/org/slf4j/MDC.html#put(kotlin.String,kotlin.String)) here (we have it on `Thread` world, thanks to `ThreadLocal`), but here we can't since Coroutines are different, but most important we are doing logs outside coroutines

We could, maybe create a custom element and add it to the CoroutineContext.
`coroutineContext[LoggingInfo] = SomeLogStuff(data, etc)`


BUT, these informations would only be accessible if the logging methods were suspendable. Non-suspending stuff wouldn't be able to access the information in the `CoroutineContext`.

Food for thought tho.

yamilmedina:
> > > Great work @yamilmedina !
> > > I just have a question, will we be needing to add `.withFlowId(APPLICATION_FLOW)` to every log?
> > 
> > 
> > Oh! Sadly yes.
> > I was expecting to do something like an [MDC context ](https://www.slf4j.org/apidocs/org/slf4j/MDC.html#put(kotlin.String,kotlin.String)) here (we have it on `Thread` world, thanks to `ThreadLocal`), but here we can't since Coroutines are different, but most important we are doing logs outside coroutines
> 
> We could, maybe create a custom element and add it to the CoroutineContext. `coroutineContext[LoggingInfo] = SomeLogStuff(data, etc)`
> 
> BUT, these informations would only be accessible if the logging methods were suspendable. Non-suspending stuff wouldn't be able to access the information in the `CoroutineContext`.
> 
> Food for thought tho.

Yup! We can't do it from non-suspended contexts 😓. So I went for this approach, trying to be more explicit when we are logging. 

I don't know if it makes sense to continue with this 😅 ? I think it is a step forward (verbose tough) but we can have better track. Do you, guys @alexandreferris / @vitorhugods  have more ideas there  🪄 ? Since continuing will require to modify other parts of the code

alexandreferris:
> > > > Great work @yamilmedina !
> > > > I just have a question, will we be needing to add `.withFlowId(APPLICATION_FLOW)` to every log?
> > > 
> > > 
> > > Oh! Sadly yes.
> > > I was expecting to do something like an [MDC context ](https://www.slf4j.org/apidocs/org/slf4j/MDC.html#put(kotlin.String,kotlin.String)) here (we have it on `Thread` world, thanks to `ThreadLocal`), but here we can't since Coroutines are different, but most important we are doing logs outside coroutines
> > 
> > 
> > We could, maybe create a custom element and add it to the CoroutineContext. `coroutineContext[LoggingInfo] = SomeLogStuff(data, etc)`
> > BUT, these informations would only be accessible if the logging methods were suspendable. Non-suspending stuff wouldn't be able to access the information in the `CoroutineContext`.
> > Food for thought tho.
> 
> Yup! We can't do it from non-suspended contexts 😓. So I went for this approach, trying to be more explicit when we are logging.
> 
> I don't know if it makes sense to continue with this 😅 ? I think it is a step forward (verbose tough) but we can have better track. Do you, guys @alexandreferris / @vitorhugods have more ideas there 🪄 ? Since continuing will require to modify other parts of the code

I'm all up to having a way more verbose log, even though it might cost a bit more time as we need to add new code but I think that in the long run it will only benefit us.

And if in case we find something that helps better we can always change it again and again 😄

alexandreferris (approved):

logger/src/commonMain/kotlin/com/wire/kalium/logger/KaliumLogger.kt
```diff
@@ -72,6 +72,9 @@ class KaliumLogger(config: Config, vararg logWriters: LogWriter = arrayOf()) {
 
     val severity = config.severity
 
+    @Suppress("unused")
+    fun withFlowId(flowId: ApplicationFlow) = kermitLogger.withTag(flowId.name.lowercase())
```
yamilmedina, line 76: `TODO`: change it to `FeatureId`

vitorhugods:
> > > > > Great work @yamilmedina !
> > > > > I just have a question, will we be needing to add `.withFlowId(APPLICATION_FLOW)` to every log?
> > > > 
> > > > 
> > > > Oh! Sadly yes.
> > > > I was expecting to do something like an [MDC context ](https://www.slf4j.org/apidocs/org/slf4j/MDC.html#put(kotlin.String,kotlin.String)) here (we have it on `Thread` world, thanks to `ThreadLocal`), but here we can't since Coroutines are different, but most important we are doing logs outside coroutines
> > > 
> > > 
> > > We could, maybe create a custom element and add it to the CoroutineContext. `coroutineContext[LoggingInfo] = SomeLogStuff(data, etc)`
> > > BUT, these informations would only be accessible if the logging methods were suspendable. Non-suspending stuff wouldn't be able to access the information in the `CoroutineContext`.
> > > Food for thought tho.
> > 
> > 
> > Yup! We can't do it from non-suspended contexts 😓. So I went for this approach, trying to be more explicit when we are logging.
> > I don't know if it makes sense to continue with this 😅 ? I think it is a step forward (verbose tough) but we can have better track. Do you, guys @alexandreferris / @vitorhugods have more ideas there 🪄 ? Since continuing will require to modify other parts of the code
> 
> I'm all up to having a way more verbose log, even though it might cost a bit more time as we need to add new code but I think that in the long run it will only benefit us.
> 
> And if in case we find something that helps better we can always change it again and again 😄

Can't avoid it. It's a price we gotta pay :D

MohammedMokresh (approved):

PR #705, opened 2022-07-22
feat: join existing MLS conversations on login AR-2060

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Implements: [Use case: join existing conversation from a new device (MLS)](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/576454924/Use+case+join+existing+conversation+from+a+new+device+MLS)

During the slow sync we send a request to join all existing MLS conversations.

#### Todo:
- [x] Unit tests
- [x] Execute join requests in parallel

#### Test Coverage (Optional)

#### How to Test

1. Login to an account which has previously been added to MLS conversations
2. Check the logs that we are sending request to join these conversations3. 
### Notes (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
@@ -24,6 +24,7 @@ interface ConversationMapper {
     fun fromDaoModel(daoModel: ConversationEntity): Conversation
     fun toDAOAccess(accessList: Set<ConversationAccessDTO>): List<ConversationEntity.Access>
     fun toDAOAccessRole(accessRoleList: Set<ConversationAccessRoleDTO>): List<ConversationEntity.AccessRole>
+    fun toDAOGroupState(groupState: com.wire.kalium.logic.data.conversation.ProtocolInfo.MLS.GroupState): GroupState
     fun toApiModel(access: Conversation.Access): ConversationAccessDTO
     fun toApiModel(accessRole: Conversation.AccessRole): ConversationAccessRoleDTO
     fun toApiModel(protocol: ConversationOptions.Protocol): ConvProtocol
@@ -91,6 +92,14 @@ internal class ConversationMapperImpl(
         }
     }
 
+    override fun toDAOGroupState(groupState: com.wire.kalium.logic.data.conversation.ProtocolInfo.MLS.GroupState): GroupState =
+        when (groupState) {
+            com.wire.kalium.logic.data.conversation.ProtocolInfo.MLS.GroupState.ESTABLISHED -> GroupState.ESTABLISHED
+            com.wire.kalium.logic.data.conversation.ProtocolInfo.MLS.GroupState.PENDING_JOIN -> GroupState.PENDING_JOIN
+            com.wire.kalium.logic.data.conversation.ProtocolInfo.MLS.GroupState.PENDING_WELCOME_MESSAGE -> GroupState.PENDING_WELCOME_MESSAGE
+            com.wire.kalium.logic.data.conversation.ProtocolInfo.MLS.GroupState.PENDING_CREATION -> GroupState.PENDING_CREATION
+        }
+
     override fun toApiModel(name: String?, members: List<UserId>, teamId: String?, options: ConversationOptions) =
         CreateConversationRequest(qualifiedUsers = if (options.protocol == ConversationOptions.Protocol.PROTEUS) members.map {
             idMapper.toApiModel(it)
@@ -142,7 +151,7 @@ internal class ConversationMapperImpl(
 
     private fun ConversationResponse.getProtocolInfo(mlsGroupState: GroupState?): ProtocolInfo {
         return when (protocol) {
-            ConvProtocol.MLS -> ProtocolInfo.MLS(groupId ?: "", mlsGroupState ?: GroupState.PENDING)
+            ConvProtocol.MLS -> ProtocolInfo.MLS(groupId ?: "", mlsGroupState ?: GroupState.PENDING_JOIN, epoch ?: 0UL)
             ConvProtocol.PROTEUS -> ProtocolInfo.Proteus
         }
     }
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
@@ -24,7 +24,7 @@ interface ConversationMapper {
     fun fromDaoModel(daoModel: ConversationEntity): Conversation
     fun toDAOAccess(accessList: Set<ConversationAccessDTO>): List<ConversationEntity.Access>
     fun toDAOAccessRole(accessRoleList: Set<ConversationAccessRoleDTO>): List<ConversationEntity.AccessRole>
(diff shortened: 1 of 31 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/JoinExistingMLSConversationsUseCase.kt
```diff
@@ -0,0 +1,70 @@
...
+    val conversationRepository: ConversationRepository,
+    kaliumDispatcher: KaliumDispatcher = KaliumDispatcherImpl
+) {
+    /**
+     * A dispatcher with limited parallelism of 10.
+     * This means using this dispatcher only a single coroutine will be processed at a time.
+     */
+    @Suppress("MagicNumber")
+    @OptIn(ExperimentalCoroutinesApi::class)
+    private val dispatcher = kaliumDispatcher.default.limitedParallelism(10)
```
typfel, line 35: @MohamadJaara would this also work as an alternative to the thread pool API?

vitorhugods: Not sure what was the need, but just my 2 cents here.

The dispatchers we use:
----

- `Default`: Targeted for general computation. By default is backed by 1 thread for each CPU core (but minimum = 2).
- `Main`: Currently only available on JVM (other platforms use `Default` underneath). It's the main. It has only 1 thread in its pool: the one used to draw the UI.
- `IO`: Currently only available on JVM. Targeted for Input/Output (_e.g._ networking, filesystem stuff). Defaults to 64 Threads or CPU core count (whichever is largest). When using `limitedParalellism`, it can create MORE threads if needed.

In general, in Kalium, we can use `Default` for most stuff. During networking, for example, Ktor will switch to its own dispatchers to performing requests. Even thought it's not _ideal_, we are using `Default` as `IO` for `iOS` for example. We don't use the file system so much that this would be performance costly anyway.


Back to the point
----

I'm not sure why the number `10`. Why not more, or why not less? But using `kaliumDispatcher.default.limitedParallelism(10)` will assure that no more than 10 coroutines will run in parallel.

If the `Default` dispatcher is backed by only 8 threads (8 CPU cores), the `limitedParallelism(10)` will not change anything at all, as it was already limited to 8 in the first place.

typfel: My thinking was that a user potentially needs to join hundreds of MLS groups on login so I want to perform the requests in parallel. Now I don't want to perform 100 HTTP requests in parallel but if you say that Ktor anyway dispatches its network calls on it own dispatcher with it's own internal limits, maybe it fine to remove `limitedParallelism(10)` completely.

vitorhugods: This is indeed limiting to only 10 requests (or less depending on number of CPU cores).
I think we can just use `kaliumDispatcher.io` and leave Ktor to do its magic underneath.

If we notice any issues, we can come back and customise the Dispatcher to better suit our needs.

Also, out of curiosity, are there plans to optimise this in the future? I see one of MLS's goals is reducing unnecessary encryptions and backend requests compared to Proteus, being more scalable.

Maybe grouping together all the requests or similar.

This is definitely a point it is currently failing, although it only once after registering the client, so might be a price we're willing to pay anyway.

typfel: >Maybe grouping together all the requests or similar.

Backends response to this is that it's unnecessary if we use HTTP2 multiplexing

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAOImpl.kt
```diff
@@ -79,12 +79,12 @@ class ConversationDAOImpl(
                 teamId,
                 if (protocolInfo is ConversationEntity.ProtocolInfo.MLS) protocolInfo.groupId
                 else null,
-
                 if (protocolInfo is ConversationEntity.ProtocolInfo.MLS) protocolInfo.groupState
                 else ConversationEntity.GroupState.ESTABLISHED,
+                if (protocolInfo is ConversationEntity.ProtocolInfo.MLS) protocolInfo.epoch.toLong()
+                else 0,
```
mchenani, line 85: can we have this as a const somewhere?

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
```diff
@@ -69,6 +71,9 @@ SELECT * FROM Conversation WHERE qualified_id = ?;
 selectByGroupId:
 SELECT * FROM Conversation WHERE mls_group_id = ?;
 
+selectByGroupState:
+SELECT * FROM Conversation WHERE mls_group_state = ? AND protocol = ?;
```
mchenani, line 75: since the state is only valid for mls groups, isn't good idea to limit it protocol = mls?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -91,8 +92,17 @@ internal class ConversationMapperImpl(
         }
     }
 
+    override fun toDAOGroupState(groupState: Conversation.ProtocolInfo.MLS.GroupState): GroupState =
```
mchenani, line 95: since all the states are the same, isn't good idea to map them with ordinal number in one line -> like
Conversation.Protocol.Info.MLS.values()[groupstate.ordinal] ...

mchenani (approved):
Great work 🚀 some comments to discuss!

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
typfel: That's what I do in code. Protocol is a enum though and I couldn't figure out how use that enum in the `sq` file.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAOImpl.kt
typfel: I'll move to a const

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
typfel: Seems fragile since then we rely on that the ordering is the same.

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
mchenani: I didn't notice that. great then! there is no clean solution for it as far as I know, if it's an enum we have to hard code it.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAOImpl.kt
typfel: Done

PR #714, opened 2022-07-26
chore: adjust calling logs

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Was a bit hard to understand what was happening in calling.

### Solutions

Add more logs, specially on `callRepository.createCall(..)` so we can better understand what happened if something goes wrong.


yamilmedina (approved):
lgtm! We have now merged this new function about adding the `featureId` to the logs, but I don't think is necessary for calling, since it has its own logger 🚀  so it will be easy to identify :D

PR #715, opened 2022-07-26
feat: Receive availability status from the others

borichellow:
[Task-2090](https://wearezeta.atlassian.net/browse/AR-2090)

# What's new in this PR?

- Added handling Availability status changing from the other users and saving it into DB

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
@@ -33,7 +33,6 @@ import com.wire.kalium.persistence.dao.UserDAO
 import com.wire.kalium.persistence.dao.UserEntity
 import com.wire.kalium.persistence.dao.client.ClientDAO
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.firstOrNull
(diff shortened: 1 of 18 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -54,21 +58,26 @@ interface UserRepository {
     suspend fun observeUser(userId: UserId): Flow<User?>
     suspend fun userById(userId: UserId): Either<CoreFailure, OtherUser>
     suspend fun updateSelfUserAvailabilityStatus(status: UserAvailabilityStatus)
+    suspend fun updateOtherUserAvailabilityStatus(userId: UserId, status: UserAvailabilityStatus)
     suspend fun getAllKnownUsersNotInConversation(conversationId: ConversationId): Either<StorageFailure, List<OtherUser>>
+    suspend fun getUsersFromTeam(teamId: TeamId): Either<StorageFailure, List<OtherUser>>
+    suspend fun getTeamRecipients(teamId: TeamId): Either<CoreFailure, List<Recipient>>
```
borichellow, line 64: these changes are needed for the future when we'll broadcast our own status to others. I just left it here, cause needed to cut off the work on this task for now, and it doesn't break anything

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -213,13 +222,35 @@ class UserDataSource(
...
+                .filter { it.id != selfUserId }
+                .map(publicUserMapper::fromDaoModelToPublicUser)
+        }
+    }
+
+    override suspend fun getTeamRecipients(teamId: TeamId): Either<CoreFailure, List<Recipient>> =
+        getUsersFromTeam(teamId)
+            .map { users -> users.map { idMapper.toApiModel(it.id) } }
+            .flatMap {
+                wrapApiRequest { clientApi.listClientsOfUsers(it) }
```
vitorhugods, line 250: Is this really necessary? Can't we just use what we have from the database? We have a `ClientDAO`?

borichellow: it requires some additional changes, but good point :)

PR #717, opened 2022-07-26
fix: remove self user from public self user search

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

public search result was including the self user

### Solutions

filter out the self user from the public user search api results

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
@@ -81,7 +81,7 @@ internal class UserSearchApiWrapperImpl(
                     val isConversationMember = conversationMembersId?.contains(domainId) ?: false
                     val isSelfUser = selfUser.id == domainId
 
-                    // neglate it because that is exactly what we do not want to have in filter results
+                    // negate it because that is exactly what we do not want to have in filter results
                     !(isConversationMember || isSelfUser)
                 }
             } else {
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
@@ -2,6 +2,9 @@ package com.wire.kalium.logic.data.publicuser
 
 import com.wire.kalium.logic.NetworkFailure
 import com.wire.kalium.logic.data.id.IdMapper
+import com.wire.kalium.logic.data.user.SelfUser
+import com.wire.kalium.logic.data.user.UserDataSource
+import com.wire.kalium.logic.data.user.UserMapper
 import com.wire.kalium.logic.di.MapperProvider
 import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.logic.functional.map
@@ -10,7 +13,15 @@ import com.wire.kalium.network.api.contact.search.UserSearchApi
 import com.wire.kalium.network.api.contact.search.UserSearchRequest
 import com.wire.kalium.network.api.contact.search.UserSearchResponse
 import com.wire.kalium.persistence.dao.ConversationDAO
+import com.wire.kalium.persistence.dao.MetadataDAO
+import com.wire.kalium.persistence.dao.QualifiedIDEntity
+import com.wire.kalium.persistence.dao.UserDAO
+import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.flatMapMerge
+import kotlinx.coroutines.flow.map
+import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.json.Json
 
 internal interface UserSearchApiWrapper {
     /*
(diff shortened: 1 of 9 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
```diff
@@ -36,36 +50,67 @@ internal class UserSearchApiWrapperImpl(
...
             )
+        }.map { userSearchResponse ->
+            filter(userSearchResponse, searchUsersOptions)
         }
 
-        return if (searchUsersOptions.conversationExcluded is ConversationMemberExcludedOptions.ConversationExcluded) {
-            val conversationMembersId = conversationDAO.getAllMembers(
-                qualifiedID = idMapper.toDaoModel(qualifiedID = searchUsersOptions.conversationExcluded.conversationId)
-            ).firstOrNull()?.map { idMapper.fromDaoModel(it.user) }
+    private suspend fun filter(
```
Garzas, line 66: Hmm, if we want to give other apps flexibility to decide if they want to filter own user from search this functionality should be in use cases

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
```diff
@@ -36,36 +50,67 @@ internal class UserSearchApiWrapperImpl(
...
-                val filteredContactResponse = it.documents.filter { contactDTO ->
-                    !(conversationMembersId?.contains(idMapper.fromApiModel(contactDTO.qualifiedID)) ?: false)
+                userSearchResponse.documents.filter { contactDTO ->
+                    val domainId = idMapper.fromApiModel(contactDTO.qualifiedID)
+
+                    val isConversationMember = conversationMembersId?.contains(domainId) ?: false
+                    val isSelfUser = selfUser.id == domainId
+
+                    // negate it because that is exactly what we do not want to have in filter results
+                    !(isConversationMember || isSelfUser)
```
Garzas, line 85: Maybe extract those to separate functionality and name it instead of commenting for better readability?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
trOnk12: good point

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
trOnk12: in my opinion without comment it should be clear enough :P I added the comment just to make it even more clear enough I think a private function would be a overkill here

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
Garzas: @vitorhugods what do you think? 😄

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/user/SearchUserUseCaseTest.kt
```diff
@@ -27,7 +34,10 @@ import io.mockative.classOf
 import io.mockative.eq
 import io.mockative.given
 import io.mockative.mock
+<<<<<<< HEAD
+=======
 import io.mockative.verify
+>>>>>>> develop
```
vitorhugods, line 40: 🧐

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/user/SearchUserUseCaseTest.kt
```diff
@@ -4,15 +4,22 @@ import com.wire.kalium.logic.NetworkFailure
 import com.wire.kalium.logic.data.connection.ConnectionRepository
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.id.QualifiedID
+<<<<<<< HEAD
+=======
 import com.wire.kalium.logic.data.publicuser.ConversationMemberExcludedOptions
 import com.wire.kalium.logic.data.user.UserId
+>>>>>>> develop
```
vitorhugods, line 11: 🧐

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/UserSearchApiWrapper.kt
vitorhugods: Indeed.

I think it would be just way simpler to do it in the Search UseCase.

Inside the UseCase:
- get all results
- get self user (easily accessible)
- before returning all results, filter them out if wanted.

Instead, we're manipulating the search, and passing this logic all the way down to a `APIWrapper` unnecessarily, adding extra complexity and parameters to simple mechanisms.

We need to stop adding business logic to API, DAOs and Repositories.

trOnk12: I honestly disagree with it, UserSearchApiWrapper responsibility is to extend the functionality of the API that does not support searching for the users that are not included in the conversation. If use cases or repositories need this functionality they simply can inject it, adding this logic to a use case would limit us to not being able to use it inside the repositories if needed in future, also whenever the back-end perhaps going to support searching for the users excluded from the conversation we simply can drop the filtering and just delegate the call. I do not see any benefit of moving all this filtering to a use case, but perhaps I oversee something ? Also filtering is not really a buisness logic to me ,it is more of a repository responsiblity to simply manipulate the data

vitorhugods (approved):

PR #718, opened 2022-07-27
fix: multiple incoming call

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-2079](https://wearezeta.atlassian.net/browse/AR-2079) and [AR-2026](https://wearezeta.atlassian.net/browse/AR-2026)

### Issues

We were receiving multiple incoming calls when using the app, recurring calls or even ended calls kept on re-appearing after some time.

### Causes (Optional)

When rejecting/declining the incoming call, we would get the conversationID wrongly (combining `conversationId` and `id` would never match, as it was getting each from different calls).

### Solutions

Adjust the SQL query to get the correct call for conversation to update its status.


vitorhugods (approved):
I will steal @WolfiWire's meme here:

image


Nice to have it fixed! 🎉

PR #725, opened 2022-07-28
chore: add verification for shouldRing when incoming call

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-2026 - Late Call notifications](https://wearezeta.atlassian.net/browse/AR-2026)

### Issues

If call was received and app was closed, when opening the app the incoming call screen would pop up because it doesn't know the AVS event of incoming call is a GROUP/CONFCHECK and not a real incoming call, just a signal that it is still running.

### Solutions

Add implementation of `shouldRing` param received from OnIncomingCall from AVS.

### Test
- Install App
- Open App
- Close App (background -> force close/swipe)
- Receive group call (with app closed)
- Open App
- `join` call button should be shown and NOT incoming call screen.

### DIFF
--- a/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnIncomingCall.kt
+++ b/logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnIncomingCall.kt
@@ -33,7 +33,7 @@ class OnIncomingCall(
         callingLogger.i("[OnIncomingCall] -> ConversationId: $conversationId | UserId: $userId | shouldRing: $shouldRing")
         val mappedConversationType = callMapper.fromIntToConversationType(conversationType)
         val isMuted = mappedConversationType == ConversationType.Conference
-        val status = if(shouldRing) CallStatus.INCOMING else CallStatus.STILL_ONGOING
+        val status = if (shouldRing) CallStatus.INCOMING else CallStatus.STILL_ONGOING
         scope.launch {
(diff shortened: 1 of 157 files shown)


saleniuk (approved):

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/call/scenario/OnIncomingCall.kt
```diff
@@ -39,7 +39,8 @@ class OnIncomingCall(
                 status = CallStatus.INCOMING,
                 callerId = federatedIdMapper.parseToFederatedId(userId),
                 isMuted = isMuted,
-                isCameraOn = isVideoCall
+                isCameraOn = isVideoCall,
+                shouldRing = shouldRing
```
ohassine, line 43: Do we have to pass `shouldRing` to `createCall` ?
 we could check and pass the right status here before calling `createCall()`

alexandreferris: Yes, because the logic to handle the rest of the status is being done inside the `createCall` in `callRepository`, doing the logic to send the correct status already would have to refactor the whole `createCall` function and it will take some time and its not something we will want to do right now as we have more calling features to implement before fully diving deep on optimization on this code.

ohassine (approved):
let's go 🚀

PR #727, opened 2022-07-28
ci: abort current jobs when updating PR or develop branch

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Tests currently take a ton of time to run, and when multiple pushes are sent in a row, this creates a big queue of jobs to run on Github Actions.

For example, if someone sends three pushes in a short interval, which is a possible scenario (a couple of review suggestions + merge from develop, for example), all tests will run 3 times.

Android and iOS tests run on Mac machines, which are very limited for us. Which can cause looooong waits until all jobs are through.

We can manually abort all jobs, but this is error-prone.

### Solutions

Automatically abort jobs running in a PR when a new push is made.

For now, only long-lasting jobs are affected (JVM/JS, Android and iOS tests), so code-style (which runs on Linux anyway) can be seen.

#### Implementation details

Add a concurrency group to long-lasting jobs.
The concurrency group looks like: `[gitHubWorkFlow]-[prNumber]`.
Where `gitHubWorkFlow` is the name of the job. This avoids job A cancelling job B when job A is manually retried.

For the JVM/JS tests, that also run on `develop`, they look like `[gitHubWorkFlow]-[branchName]`, so the job will be aborted if there's a new push. This job's only purpose is to add coverage to Codecov. So it doesn't add much keeping an already-outdated reference.

### Testing

Manually testing in this PR by triggering an empty commit.

### Attachments

As you can see, the three jobs were automatically cancelled once I pushed a new empty commit to this PR:
image

----

----
##### References

### DIFF
--- a/.github/workflows/gradle-android-tests.yml
+++ b/.github/workflows/gradle-android-tests.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
     types: [ opened, synchronize ] # Don't rerun on `edited` to save time
 
+concurrency:
+    group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+    cancel-in-progress: true
+
 jobs:
   gradle-run-tests:
     runs-on: macos-latest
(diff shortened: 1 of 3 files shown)


MohamadJaara (approved):
Nice catch 🚀

yamilmedina (approved):
Amazing, didn't knew about this "concurrency" config 🙌

PR #746, opened 2022-08-02
feat: add decryption failure messages [AR-1657]

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Messages that failed to decrypt on the process of receiving new messages were being lost. Now we want to persist these kind of errors in order to allow displaying a visual error message on the consuming app layer.

### Solutions

I created a new type of Message Content, aka `MessageContent.FailedDecryption` and `MessageEntityContent.FailedDecryption` as well as the necessary MessageDao persistance and retrieval related operations.

#### Test Coverage (Optional)

#### How to Test

This will need to be tested on the AR side of things. For now only unit tests were provided

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/PersistMessageUseCase.kt
```diff
@@ -45,8 +45,10 @@ internal class PersistMessageUseCaseImpl(
             is MessageContent.RestrictedAsset -> true
             is MessageContent.DeleteForMe -> false
             is MessageContent.Unknown -> false
+            is MessageContent.Availability -> false
+            is MessageContent.FailedDecryption -> true
+            MessageContent.MissedCall -> true
```
alexandreferris, line 50: Whats the difference when using `is` and not using it?

gongracr: Using `is` cast on `when` clauses is only needed for `data class`es objects. However, for plain `object`s, we can simply ignore it. Therefore is up to us to decide whether to add it as a way of keeping visual coherence, or not (as a way to imply that it is a plain `object` class.

gongracr: I just realised that the way I implemented was creating some confusion, so I will opt for adding `is` cast to all types.

alexandreferris: Oh thanks a lot for the explanation 🙇🏻

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ConversationEventReceiver.kt
```diff
@@ -141,7 +148,41 @@ class ConversationEventReceiverImpl(
...
+                date = timestampIso,
+                senderUserId = senderUserId,
+                senderClientId = senderClientId,
+                status = Message.Status.SENT,
+                editStatus = Message.EditStatus.NotEdited,
+                visibility = Message.Visibility.VISIBLE
+            )
+            processMessage(message)
+        }
+    }
```
alexandreferris, line 166: Would having this be always saving the failed decryption message as forced `editStatus = NotEdited`?

I might be wrong, but aren't then we assuming that and edit will never fail to decrypt?

alexandreferris (commented):
Great work @gongracr ! I just have a quick question regarding the code

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ConversationEventReceiver.kt
gongracr: Good question! My understanding is that the message decryption will be always be triggered on newly received messages, so these ones can't be edited messages. Once these are decrypted successfully, they will be persisted unencrypted on our SQLite database and they will never adopt this error message type anymore.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ConversationEventReceiver.kt
```diff
@@ -135,6 +139,7 @@ class ConversationEventReceiverImpl(
 
                     else -> kaliumLogger.withFeatureId(EVENT_RECEIVER).e("$TAG - Failure when processing message: $it")
                 }
+                handleFailedProteusDecryptedMessage(event)
```
saleniuk, line 142: Shouldn't this be called only when `is ProteusFailure ->`? Right now it's called for every failure type 🤔

alexandreferris (approved):
👌🏻

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ConversationEventReceiver.kt
saleniuk: To find out if it's an "edit" message, we need to decrypt the content first, so if error happens at this point, we don't know yet what type of message it is, but still we need to store this message, so I think storing it as a not edited FailedDecryption is ok because we don't want any FailedDecryption message type to appear as an edited one.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
```diff
@@ -51,4 +51,6 @@ sealed class MessageContent {
 
     // we can add other types to be processed, but signaling ones shouldn't be persisted
     object Ignored : Signaling() // messages that aren't processed in any way
+
+    data class FailedDecryption(val encodedData: ByteArray? = null) : Regular()
```
borichellow, line 55: why did you make it a child of `Regular` and not `Signal` ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/ConversationEventReceiver.kt
gongracr: After private discussion with Jacob, we decided to group all these errors into the same generic Decryption Failure message. Also the current designs for this error handling are quite vague and we are not aiming to inform the final user about all the nuances of our own internal error handling.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
gongracr: Signal messages are not persisted in the DB. In this case, we want to persist these errors may we implement in the future a different way to trigger again the decryption process on them.

typfel: Maybe because this is a visual message not a signalling message but why is not a System message?

typfel (changes requested):

gongracr:
> Maybe because this is a visual message not a signalling message but why is not a System message?

Because this is a UI wrapper around a failed message @typfel. I thought this error message could eventually be replaced visually by the correct `Regular` message once we succeed with its decryption. If I declare it as a `System` message though, we will need to remove it from the chat and add it as a new message on the bottom of the list for example, which imho it feels a bit overkill.

saleniuk:
> Maybe because this is a visual message not a signalling message but why is it not a System message?

`Regular` messages are the ones sent by users, `System` messages are sent by the system, so they don't have any author. If we want to show `FailedDecryption` message as received from the user (which means with the header containing the avatar and username but with content replaced with some error info) then it needs to be stored as a `Regular` message. 😄

typfel:
> `Regular` messages are the ones sent by users, `System` messages are sent by the system, so they don't have any author. If we want to show `FailedDecryption` message as received from the user (which means with the header containing the avatar and username but with content replaced with some error info) then it needs to be stored as a `Regular` message. 😄

Ok, that makes sense.

>Because this is a UI wrapper around a failed message @typfel. I thought this error message could eventually be replaced visually by the correct Regular message once we succeed with its decryption. If I declare it as a System message though, we will need to remove it from the chat and add it as a new message on the bottom of the list for example, which imho it feels a bit overkill.

We can't recover messages we've failed decrypt they are lost forever but I guess it makes to expose it as regular if we want to display it as coming from a specific user. Looked at the design in the ticket and I don't understand it, it seems to include the message content.

typfel (approved):

borichellow (approved):
make sense for me :) 👍

PR #750, opened 2022-08-02
fix: Include blocked users into search knowns

borichellow:
### Issues

Blocked users were displayed not as "Contacts" but as a part of "Public Wire"
While working on [Task-1593](https://wearezeta.atlassian.net/browse/AR-1593) I found out that it's wrong

### Causes (Optional)

In DB select-query user `connection_status == ACCEPTED` which exclude blocked users.

### Solutions

Add `connection_status == BLOCKED` into DB select-query


persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
```diff
@@ -85,12 +85,12 @@ SELECT * FROM User
 WHERE (name LIKE ('%' || :searchQuery || '%')
 OR  handle LIKE  ('%' || :searchQuery || '%')
 OR  email LIKE  ('%' || :searchQuery || '%'))
-AND connection_status = :connectionStatus;
+AND connection_status IN :connectionStatuses;
```
gongracr, line 88: Where is this new `connectionStatuses` being declared?

gongracr: Shouldn't this need to be named `States` instead of `Statuses` ?

borichellow: i just added `es` to the end, but good point, changing it :)

borichellow: and about declaration: there is no need to declare it, we can just pass there `?` (the same as `: searchQuery` above) and SQLDelight "parse" it as a "second param in method", but to add more context to it me name this parameter.

gongracr (approved):
Reads good to me except for those `state`, `status` and `statuses`,  but I guess that falls a bit out of scope of this PR. 
Great job in any case getting ahead of problems! 🍒 🍈

PR #766, opened 2022-08-05
chore: add sync retry mechanism

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Clients drop connection and never recover from it.

### Solutions

Add a retry mechanism to Slow and Incremental syncs.
The retry mechanism is quite simple at the moment:
- A real exception (not cancellation) is thrown
- Wait 10 seconds
- Try again

> **Note**: In the future we can optimise this. Retrying less frequently and observing connectivity changes (like mobile phones moving from Wi-Fi to Mobile Data, for example).
> However, on Android, for example, the WebSocket can be dropped when the app is put into the background or when the screen locks.
> So this delay-retry covers more ground and is more reliable for an initial implementation.

#### Test Coverage

#### How to Test

Opening the Reloaded app with these changes:
- Open a conversation
- Receive messages
- Turn Wi-Fi off and on again

##### Expectations:
- The app should keep receiving messages
- The connectivity status bar should go to Yellow on connectivity drop, green and disappear again once connection is re-established

### Attachments

https://user-images.githubusercontent.com/9389043/183217799-e46cbb42-24d6-4b64-9f41-edfb36d6d123.mp4

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/EventGatherer.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/EventGatherer.kt
@@ -96,22 +96,17 @@ internal class EventGathererImpl(
     private suspend fun FlowCollector<Event>.handleWebsocketEvent(webSocketEvent: WebSocketEvent<Event>) = when (webSocketEvent) {
         is WebSocketEvent.Open -> onWebSocketOpen()
         is WebSocketEvent.BinaryPayloadReceived -> onWebSocketEventReceived(webSocketEvent)
-        is WebSocketEvent.Close -> throwOnWebSocketClosed(webSocketEvent)
+        is WebSocketEvent.Close -> handleWebSocketClosure(webSocketEvent)
         is WebSocketEvent.NonBinaryPayloadReceived -> logger.w("Non binary event received on Websocket")
     }
 
-    private fun throwOnWebSocketClosed(webSocketEvent: WebSocketEvent.Close<Event>): Nothing =
-        throw when (val cause = webSocketEvent.cause) {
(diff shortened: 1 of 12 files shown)


gongracr (approved):
lgtm 🍎 🍓

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/EventGatherer.kt
```diff
@@ -96,22 +96,17 @@ internal class EventGathererImpl(
...
-        is WebSocketEvent.Close -> throwOnWebSocketClosed(webSocketEvent)
+        is WebSocketEvent.Close -> handleWebSocketClosure(webSocketEvent)
         is WebSocketEvent.NonBinaryPayloadReceived -> logger.w("Non binary event received on Websocket")
     }
 
-    private fun throwOnWebSocketClosed(webSocketEvent: WebSocketEvent.Close<Event>): Nothing =
-        throw when (val cause = webSocketEvent.cause) {
+    private fun handleWebSocketClosure(webSocketEvent: WebSocketEvent.Close<Event>) =
+        when (val cause = webSocketEvent.cause) {
+            null -> logger.i("Websocket closed normally")
```
mchenani, line 105: Is that really a normal condition? I mean isn't better to mention that the cause is null for further investigation?

vitorhugods: This means that either us chose to close it, by calling cancel, or the backend close to do so by sending a close websocket message.

AFAIK the server doesn't do it. And we only do it when the connection policy is `DISCONNECT_AFTER_PENDING_EVENTS`.

So indeed, it's a normal/expected case.

PR #774, opened 2022-08-10
feat: Unblock user UseCase

borichellow:
[Task-2084](https://wearezeta.atlassian.net/browse/AR-2084)

# What's new in this PR?
- `UnblockUserUseCase` for unblocking blocked users
- UnitTests for it


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/connection/UnblockUserUseCase.kt
```diff
@@ -0,0 +1,36 @@
...
+fun interface UnblockUserUseCase {
+    /**
+     * Use case [UnblockUserUseCase] operation
+     *
+     * @param userId the target user whom to unblock
+     * @return a [Either] indicating the operation result:
+     * Left - [CoreFailure] if occurred,
+     * Right - [Unit] is success
+     */
+    suspend operator fun invoke(userId: UserId): Either<CoreFailure, Unit>
```
yamilmedina, line 24: I believe that this return type was supposed to not leak to clients, if I recall correctly from Prev. Android collectives 🤔  We are wrapping the result in a new type.

yamilmedina (commented):
Great work, just a small comment about `Either` leaking into clients, but I would prefer other opinions :D

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/connection/UnblockUserUseCase.kt
gongracr: Yepp, it was already agreed that Kalium Use Cases should not expose the `Either` construction to consuming app. The correct way would be to provide a wrapping `Result` class in the same fashion than other use cases do.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/connection/UnblockUserUseCase.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.wire.kalium.logic.kaliumLogger
+
+/**
+ * Use Case that allows the current self user to unblock another previously blocked user
+ */
+fun interface UnblockUserUseCase {
+    /**
+     * Use case [UnblockUserUseCase] operation
+     *
+     * @param userId the target user whom to block
```
gongracr, line 17:
```suggestion
     * @param userId the target user whom to unblock
```

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/connection/UnblockUserUseCaseTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    fun givenUnblockingSuccessful_thenBlockResultIsSuccess() = runTest {
+        val (_, blockUser) = Arrangement()
+            .withBlockResult(Either.Right(TestConnection.CONNECTION))
+            .arrange()
+
+        val result = blockUser(TestUser.USER_ID)
+
+        assertTrue(result is UnblockUserResult.Success)
```
gongracr, line 42: Can we check at least that the `connectionRepository.updateConnectionStatus` was invoked?

gongracr (approved):
Tiny non blocking comments but looking good otherwise 🕺 🎈

PR #779, opened 2022-08-11
chore: update kotlin to 1.7.10

vitorhugods:
### What else changed?

Because of this, Compose needed to be upgraded, which caused a need to update Android Gradle Plugin, which needed a `compileVersion` 32, etc. etc. etc.

The summary:

| Dependency | Old Version | New Version |
| ---- | ---- | ---- |
| Gradle | 7.1.1 | 7.4.2 |
| Android Build Tools | 7.0.4 | 7.2.0 |
| compileSdk | 31 | 32 |
| Coroutines | 1.6.0 | 1.6.4 |
| Compose | 1.1.0-rc01 | 1.3.0 |
| Compose Material | 1.1.0-rc01 | 1.2.0 |
| OkHttp | 4.9.3 | 4.10.0 |
| ktor | 2.0.0-beta | 2.1.0 |
| ktx-datetime | 0.3.2 | 0.4.0 |
| ktxSerialization | 1.3.2 | 1.3.3 |
| multiplatformSettings | 0.8.1 | 0.9 |
| kermit | 1.0.0 | 1.1.3 |
| ksp | 1.0.2 | 1.0.6 |
| SQLDelight | 2.0.0-alpha01 | 2.0.0-alpha03 |

### Breaking changes?

- `when` blocks must be exhaustive now. So I've addressed the places where it wasn't.
- Ktor changed the `ContentNegotiation` package.
- SQLDelight changed the packakges from `com.squareup` to `app.cash`
- Protobuf for JavaScript seems broken, so I've disabled tests related to it. To be fair, we are not using it and this is delaying us from upgrading for around 3 months already. So I find it easier to fix this later when necessary than delaying upgrades all around our stack because of something we don't use yet.


buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -58,7 +58,7 @@ object Plugins {
         scope.kotlin("jvm")
 
     fun ksp(scope: PluginDependenciesSpec) =
-        scope.id("com.google.devtools.ksp").version("1.6.10-1.0.2")
+        scope.id("com.google.devtools.ksp").version("1.7.0-1.0.6")
```
MohamadJaara, line 61:
```suggestion
        scope.id("com.google.devtools.ksp").version("1.7.10-1.0.6")
```

vitorhugods: Yep. I got plenty of warnings for missing that one out :D 

Addressed.

MohamadJaara (approved):
LGTM 🚀

PR #780, opened 2022-08-11
feat: synchronise last read for the same user on different client

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Solutions

We want to synchronise, the last read state of the conversation across the user devices. 

In order to do that I created : 

- a SQL query that gets the lastUnreadMessage for a conversation, that lastUnreadMessage is used by the AR to redirect the user that messages on the screen.
- handled the LastRead message, when we receive that message from a user with a senderID equal to our it means it is another client, at that point we update the lastReadDate of a conversation to match this message date time.
- Send the message a cross the devices so that the other clients can perform the above. 

### Testing
- To be followed after the implementation idea gets accepted


borichellow (commented):
I'm not sure about the name `LastUnreadMessage`, shouldn't it be `FirstUnreadMessage`? 
Cause last unread message is always the last message in conversation :)

trOnk12:
> I'm not sure about the name `LastUnreadMessage`, shouldn't it be `FirstUnreadMessage`? Cause last unread message is always the last message in conversation :)

from my point of view lastUnreadMessage is correct because for example someone sent the following : 

1. hi - 12:30 
2. hi - 13:30
3. hi - 14:30 

last unread message in time would be 1. because 14:30 is first message in time 2. is second message in time :P But I guess it depends from waht context you look at the word "last " :P I am fine changing it to firstUnreadMessage if that is more clear for

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/message/LastReadContentHandler.kt
```diff
@@ -0,0 +1,39 @@
...
+    // TODO: for now we are just handeling the case when the self user has read the conversation
+    // on another device, in the future we could handle the case when other user have read our message
+    // here too
+    suspend fun handle(
+        message: Message,
+        messageContent: MessageContent.LastRead
+    ) {
+        val isMessageComingFromOtherClient = message.senderUserId == userRepository.getSelfUserId()
+
+        if (isMessageComingFromOtherClient) {
```
typfel, line 28: We should also only process the LastRead if it was sent in the `selfConversation`

logic/src/commonMain/kotlin/com/wire/kalium/logic/util/TimeParser.kt
```diff
@@ -31,6 +31,8 @@ interface TimeParser {
      * @return true of time1 is before time2 or false otherwise
      */
     fun isTimeBefore(time1: String, time2: String): Boolean
+
+    fun fromEpochTimeStampToDate(timeStamp: Long): String
```
typfel, line 35:
```suggestion
    fun fromEpochTimeStampToDate(timestamp: Long): String
```

typfel:
> > I'm not sure about the name `LastUnreadMessage`, shouldn't it be `FirstUnreadMessage`? Cause last unread message is always the last message in conversation :)
> 
> from my point of view lastUnreadMessage makes more sense because for example someone sent the following :
> 
> 1. hi - 12:30
> 2. hi - 13:30
> 3. hi - 14:30
> 
> last unread message in time would be 1. because 14:30 is first message in time 13:30 is second message in time and 12:30 is the last message in time, if we order it chronological in time because then you got 3,2,1 :PI think it also makes sense from SQL query perspective if you sort it descending. But I guess it depends from waht context you look at the word "last " :P I am fine changing it to firstUnreadMessage if that is more clear or something else that would make more sense here

We also refer to this as the `lastUnreadMessage` on other platforms (iOS).

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/ConversationEventReceiver.kt
```diff
@@ -358,7 +361,21 @@ internal class ConversationEventReceiverImpl(
...
-                is MessageContent.Text, is MessageContent.FailedDecryption -> persistMessage(message)
+                is MessageContent.Text, is MessageContent.FailedDecryption -> {
+                    val isMessageComingFromOtherClient = message.senderUserId == userRepository.getSelfUserId()
+
+                    if (isMessageComingFromOtherClient) {
+                        // if the message is coming from other client it means
+                        // that at that time the user has read the conversation
+                        // so we want to update it with date that the message
+                        // has been sent
+                        conversationRepository.updateConversationReadDate(
```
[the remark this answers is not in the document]
trOnk12: I am not sure about that the PersistMessageUseCase is used to send messages too or even when the call is needed I do not think we want to keep updating it in all those cases ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
```diff
@@ -26,8 +26,9 @@ sealed class MessageContent {
 
     data class DeleteForMe(
         val messageId: String,
-        val conversationId: String,
-        val qualifiedConversationId: QualifiedConversationId?
+        @Deprecated("Use qualified id instead", ReplaceWith("conversationId"))
+        val unqualifiedConversationId: String,
+        val conversationId: ConversationId?,
     ) : Regular()
```
trOnk12, line 32: @mchenani

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/message/DeleteForMeHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+
+        if (message.conversationId == conversationRepository.getSelfConversationId())
+            messageRepository.deleteMessage(
+                messageUuid = messageContent.messageId,
+                conversationId = conversationId
+            )
+        else kaliumLogger.withFeatureId(KaliumLogger.Companion.ApplicationFlow.EVENT_RECEIVER).i(message = "Delete message sender is not verified: $message")
+    }
+
+}
```
trOnk12, line 35: @mchenani

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
```diff
@@ -26,8 +26,9 @@ sealed class MessageContent {
 
     data class DeleteForMe(
         val messageId: String,
-        val conversationId: String,
-        val qualifiedConversationId: QualifiedConversationId?
+        @Deprecated("Use qualified id instead", ReplaceWith("conversationId"))
+        val unqualifiedConversationId: String,
```
typfel, line 30: I don't understand why we add the `unqualifiedConversationId` and immediately mark it as deprecated. Wouldn't it be easier to inject the self domain into the `ProtobufContentMapper` and let it convert unqualified IDs into qualified ones.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/message/DeleteForMeHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+) {
+
+    suspend fun handle(
+        message: Message,
+        messageContent: MessageContent.DeleteForMe
+    ){
+        // The conversationId comes with the hidden message[content] only carries the conversationId VALUE,
+        // we need to get the DOMAIN from the self conversationId[here is the message.conversationId]
+        val conversationId = messageContent.conversationId ?:
+        ConversationId(messageContent.unqualifiedConversationId, userRepository.getSelfUserId().domain)
```
typfel, line 25: Why not do this inside the `ProtobufContentMapper` as I mentioned in another comment.

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageContent.kt
vitorhugods: Hehehe, that was one of the solutions Mateusz and I thought about.

The main reason to go this way was to follow the same approach done for `HideMessage`.

But we can break this pattern now.

I suggest one improvement more:

We could just stop using `UserRepository` everywhere just to get self user ID. We could inject `selfUserId` directly into the constructor, because the `UserSessionScope` already has a `userId` anyway when we create it.


This way we could inject a `defaultDomain` or similar, that would help solving this Mapper thing.

typfel: Sounds like a plan! If you already discussed this among yourself I'm fine with doing this refactoring in a separate PR.

trOnk12: I think tackeling EventReceiver etc in a seperate refactor would be a good idea to include this issue as well so I am going to leave as it is for now

typfel (approved):

PR #781, opened 2022-08-11
feat: ObserveUserInfoUseCase

borichellow:
## Needed for  [Task-1593](https://wearezeta.atlassian.net/browse/AR-1593)

# What's new in this PR?

`ObserveUserInfoUseCase` that allows to observe UserInfo changes and fetch it from API if there is no such one in DB

### Issues

No such UseCase

### Causes (Optional)

It wasn't created 

### Solutions

Create it

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/ObserveUserInfoUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/ObserveUserInfoUseCase.kt
@@ -11,7 +11,6 @@ import com.wire.kalium.logic.data.user.type.UserType
 import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.logic.functional.flatMapRightWithEither
 import com.wire.kalium.logic.functional.fold
-import com.wire.kalium.logic.functional.mapLeft
 import com.wire.kalium.logic.functional.mapRight
 import com.wire.kalium.logic.functional.mapToRightOr
 import com.wire.kalium.logic.wrapStorageRequest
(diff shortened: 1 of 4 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/ObserveUserInfoUseCase.kt
```diff
@@ -0,0 +1,83 @@
...
+                if (storageFailure is StorageFailure.DataNotFound) {
+                    userRepository.fetchUsersByIds(setOf(userId))
+                        .fold({ it }) { storageFailure }
+                } else {
+                    storageFailure
+                }
+            }
+            .filter { either ->
+                // We don't want to pass DataNotFound Error forward, cause in that case we'll fetch data from API
+                either.fold({ it !is StorageFailure.DataNotFound }) { true }
```
saleniuk, line 61: This shouldn't ever happen, because `fetchUsersByIds` isn't supposed to emit `DataNotFound`, only `getKnownUser` can do so, but it's already filtered out in `mapLeft`, so maybe even this filter isn't needed? 😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/ObserveUserInfoUseCase.kt
```diff
@@ -0,0 +1,83 @@
...
+            }
+    }
+
+    /**
+     * Users are allowed to fetch team details only if they are members of the same team
+     * @see [UserType]
+     */
+    private suspend fun getOtherUserTeam(otherUser: OtherUser): Either<CoreFailure, Team?> {
+        return if (otherUser.teamId != null && otherUser.userType == UserType.INTERNAL) {
+            val localTeam = teamRepository.getTeam(otherUser.teamId).firstOrNull()
```
saleniuk, line 71: `getTeam` returns a `Flow` so can we make `getOtherUserTeam` also return a `Flow` and `flatMap` it after `getKnownUser` instead of `foldEither` (which is a basic `map` underneath)? This way it'll emit all the updates also from `getTeam`

logic/src/commonMain/kotlin/com/wire/kalium/logic/functional/EitherFlow.kt
```diff
@@ -5,3 +5,12 @@ import kotlinx.coroutines.flow.filter
...
 fun <L, R> Flow<Either<L, R>>.onlyRight(): Flow<R> = filter { it.isRight() }.map { (it as Either.Right<R>).value }
+
+fun <L, R, T> Flow<Either<L, R>>.mapLeft(block: suspend (L) -> T): Flow<Either<T, R>> =
+    map { it.fold({ l -> Either.Left(block(l)) }) { r -> Either.Right(r) } }
+
+fun <L, R, T> Flow<Either<L, R>>.mapRight(block: suspend (R) -> T): Flow<Either<L, T>> =
+    map { it.fold({ l -> Either.Left(l) }) { r -> Either.Right(block(r)) } }
+
+fun <L, R, T : Any> Flow<Either<L, R>>.foldEither(fnL: suspend (L) -> T, fnR: suspend (R) -> T): Flow<T> =
+    map { it.fold({ l -> fnL(l) }, { r -> fnR(r) }) }
```
saleniuk, line 16: It would be better to write some short explanation in the comment for others to understand how it should behave 😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/ObserveUserInfoUseCase.kt
borichellow: not really `StorageFailure.DataNotFound` can be emitted in `mapLeft` when `userRepository.fetchUsersByIds` is succeed.

logic/src/commonMain/kotlin/com/wire/kalium/logic/functional/EitherFlow.kt
borichellow: that one was removed and the others are quit self-explanatory IMO :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/ObserveUserInfoUseCase.kt
saleniuk: I mean that `fetchUsersByIds` currently doesn't emit `DataNotFound` because currently it executes `upsertTeamMembers` and `upsertUsers` - both return `Unit` and `DataNotFound` is passed when something returns `null` 🙂 
I'm just worried because this filter won't pass the `DataNotFound` so if in the future the `fetchUsersByIds` code changes and starts emitting this error then we won't get any value and be stuck on loading for instance 🤔

saleniuk (approved):

PR #784, opened 2022-08-11
feat: persistent websocket [AR-1168]

MohammedMokresh:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

we need to save a value in our storage to know if the user enable or disabled the websocket

### Solutions

create a simple SP value to save that flag 

Needs releases with:

#### How to Test

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/webSocketStatus/EnableWebSocketUseCase.kt
```diff
@@ -0,0 +1,15 @@
...
+
+interface EnableWebSocketUseCase {
+    operator fun invoke(enabled: Boolean): Either<StorageFailure, Unit>
+}
+
+class EnableWebSocketUseCaseImpl(
+    private val userConfigRepository: UserConfigRepository
+) : EnableWebSocketUseCase {
+    override operator fun invoke(enabled: Boolean) = userConfigRepository.persistWebSocketStatus(enabled)
+}
```
alexandreferris, line 15: If we have its name as `Enable` then we should not receive a boolean and just assume we are passing a boolean true to `persistWebSocketStatus`, no? 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/webSocketStatus/EnableWebSocketUseCase.kt
```diff
@@ -0,0 +1,15 @@
+package com.wire.kalium.logic.feature.user.webSocketStatus
+
+import com.wire.kalium.logic.StorageFailure
+import com.wire.kalium.logic.configuration.UserConfigRepository
+import com.wire.kalium.logic.functional.Either
+
+interface EnableWebSocketUseCase {
+    operator fun invoke(enabled: Boolean): Either<StorageFailure, Unit>
```
yamilmedina, line 8: We should not expose `Either` to clients, so I think we need to wrap this in a `Result` class

yamilmedina (commented):
Pretty good job, just some comments about class visibility and as well if we add more tests, we could have some smiley cats (😸)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/webSocketStatus/EnableWebSocketUseCase.kt
MohammedMokresh: hmmm  , you have a point there , I think it's better to go with the name persist web socket  status

alexandreferris (approved):
👌🏻

yamilmedina (approved):
😸

vitorhugods (changes requested):
The naming is a bit weird.
Websocket is enabled. There's no way to disable it. We need it for sync.

The fact that it is `false` by default is very misleading.

Now... having a PERSISTENT Websocket connection is a different thing. So I think it makes more sense to have a clear distinction in the naming.

----

I don't quite get it.
This is an Android-specific issue. More than that... it's an F-Droid specific issue. No?

Why is this being implemented in Kalium? Would we add dark theme preference settings to Kalium too, for example? Where to draw the line?

I feel like this is a client-specific setting and that this should be implemented in Reloaded directly. If a user preference is on, we launch a service to keep the app alive in the background and set the `ConnectionPolicy` to `KEEP_ALIVE`.

vitorhugods (approved):
As we talked in private, changing this to whatever different approach now or in the future will have pretty much the same cost.
We need to discuss and bring forward solutions to having platform-specific UseCases and/or exposing a multi-platform Preferences solution so the apps can use it and create their own user-scoped settings.

PR #787, opened 2022-08-11
fix: obfuscate sensitive info in network logs 

MohammedMokresh:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

after obfuscate the log messages using regex it affects the performance so we needed to find another way

### Solutions

convert the messages in the ktor response and request to JSon so we can detect the sesiteve info

#### How to Test

run using this branch and see if you still see any sensitive info in the network level that not obfuscated

----

----

----
##### References


network/src/commonMain/kotlin/com/wire/kalium/network/KaliumKtorCustomLogging.kt
```diff
@@ -122,7 +127,7 @@ public class KaliumKtorCustomLogging private constructor(
         kaliumLogger.v("BODY Content-Type: $contentType")
         kaliumLogger.v("BODY START")
         val message = content.tryReadText(contentType?.charset() ?: Charsets.UTF_8) ?: "[response body omitted]"
-        kaliumLogger.v(message)
```
mchenani, line 125: isn't better to inject the obfuscation mechanism into the kaliumLogger and switch it on/off by configuration?

mchenani: then we can easily decide to obfuscate all the logs not by replacing each line of the logs.

MohammedMokresh: the thing is the logger module is not aware about the network stuff or Json elemnt and whatsoever , so I am not sure if we want to import these in the logger module just for this case

vitorhugods: > the thing is the logger module is not aware about the network stuff or Json elemnt and whatsoever , so I am not sure if we want to import these in the logger module just for this case

This would be a circular dependency.
`network -> logger -> network ERROR`

network/src/commonMain/kotlin/com/wire/kalium/network/KaliumKtorCustomLogging.kt
```diff
@@ -122,7 +127,7 @@ public class KaliumKtorCustomLogging private constructor(
...
+        obfuscateAndLogMessage(message)
```
vitorhugods, line 130: `obfuscateAndLogMessage` returns a String, why isn't it being used?

network/src/commonMain/kotlin/com/wire/kalium/network/KaliumKtorCustomLogging.kt
```diff
@@ -240,6 +249,49 @@ public class KaliumKtorCustomLogging private constructor(
...
+        } else {
+            obfuscateJsonElement(obj)
+        }
+        obj.toString()
+    } catch (e: Exception) {
+        "error the body content while logging "
+    }
+}
+
+fun obfuscateJsonElement(obj: JsonElement) {
```
vitorhugods, line 269: This doesn't `obfuscateJsonElement`, but rather `logObfuscatedJsonElement`

```suggestion
fun logObfuscatedJsonElement(obj: JsonElement) {
```

network/src/commonMain/kotlin/com/wire/kalium/network/KaliumKtorCustomLogging.kt
```diff
@@ -240,6 +249,49 @@ public class KaliumKtorCustomLogging private constructor(
...
+                kaliumLogger.e("${it.key} : ${it.value}")
+            }
+        }
+    }
+
+}
+
+private val obfuscateJsonKeys by lazy { listOf("password", "authorization") }
+private val obfuscateJsonIdKeys by lazy { listOf("conversation", "id", "user", "team") }
+private val obfuscateObjects by lazy { listOf("qualified_id") }
```
vitorhugods, line 291: These variables have function names.

```suggestion
private val sensitiveJsonKeys by lazy { listOf("password", "authorization") }
private val sensitiveJsonIdKeys by lazy { listOf("conversation", "id", "user", "team") }
private val sensitiveJsonObjects by lazy { listOf("qualified_id") }
```

network/src/commonMain/kotlin/com/wire/kalium/network/KaliumKtorCustomLogging.kt
```diff
@@ -240,6 +249,49 @@ public class KaliumKtorCustomLogging private constructor(
...
+    return try {
+        val obj = (Json.decodeFromString(text) as JsonElement)
+        if (obj.jsonArray.size > 0) {
+            obj.jsonArray.map {
+                obfuscateJsonElement(it)
+            }
+        } else {
+            obfuscateJsonElement(obj)
+        }
+        obj.toString()
```
vitorhugods, line 263: Why does it return a String if it's ignored?

vitorhugods (changes requested):

network/src/commonMain/kotlin/com/wire/kalium/network/KaliumKtorCustomLogging.kt
mchenani: `import kotlinx.serialization.decodeFromString
import kotlinx.serialization.json.Json
import kotlinx.serialization.json.JsonElement
import kotlinx.serialization.json.jsonArray
import kotlinx.serialization.json.jsonObject`

based on these lines, we need only the serialization library? I mean if we only add it to the logger that would be ok and no circular dependancy.
@vitorhugods @MohammedMokresh

network/src/commonMain/kotlin/com/wire/kalium/network/KaliumKtorCustomLogging.kt
MohammedMokresh: I was using it at first and after that removed the log to the function and forgot to remove the return , will update

network/src/commonMain/kotlin/com/wire/kalium/network/KaliumKtorCustomLogging.kt
```diff
@@ -240,6 +249,49 @@ public class KaliumKtorCustomLogging private constructor(
...
+}
+
+fun logObfuscatedJsonElement(obj: JsonElement) {
+    obj.jsonObject.entries.toMutableSet().map {
+        when {
+            sensitiveJsonKeys.contains(it.key.lowercase()) -> {
+                kaliumLogger.v("${it.key} : ******")
+            }
+            sensitiveJsonIdKeys.contains(it.key.lowercase()) -> {
+                kaliumLogger.v("${it.key} : ${it.value.toString().substring(START_INDEX, END_INDEX)}")
```
MohamadJaara, line 276: The function String.subString() can throw `StringIndexOutOfBoundsException`, it's fine if it is guarantied that all IDs are UUID and will never be shorter than 9 chars

MohamadJaara (approved):
LGTM 🚀, only one question about the usage of `String.subString()`

network/src/commonMain/kotlin/com/wire/kalium/network/KaliumKtorCustomLogging.kt
MohammedMokresh: yeah by theory it should not be less than 9 chars, and anyway we are handling a general exception while logging these, so that should not affect the app

PR #791, opened 2022-08-13
fix: websocket not receiving messages after some time

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

After some time, the websocket dies. It completely stops receiving messages without any sings of errors, exceptions, or anything else being called.

### Causes

As we do not send messages to the backend, it seems that the backend expects at least some pings every so often to keep the connection alive.
After some time without receiving anything, the backend just leaves the connection in a limbo and doesn't send any more messages through it.

### Solutions

Add a ping every so many seconds.
It seems that @makingthematrix also went through this problem back in 2019 for the Scala app (see [this commit](https://github.com/wireapp/wire-android-sync-engine/commit/8b1866424a26b72e35e6209a7d4b091806faf8f6#diff-6a7a11f3c4ce9a86047f331475ffbf5c3cfec2b1601b5e14297c1694c8869e7fR61)), and 30s has been the magic number that carried the Scala app until now.

Unfortunately OkHttp doesn't handle dynamic pinging, so it must be configured directly into the HttpEngine, and it seems that [the configuration in WebSockets is just ignored if you're using OkHttp](https://youtrack.jetbrains.com/issue/KTOR-4752/Ping-configuration-is-silently-not-respected-when-setting-pingInterval-on-WebSockets-using-OkHttp-engine).

So, even though adding it to OkHttp fixes it for Android and JVM, I've added it to Ktor as well so it also works for other platforms.

> **Note**: During my investigation I also added a lower-level logging to the WebSocket and I guess there's no harm keeping it.

### Testing

Manually tested as we can't yet mock WebSockets using Ktor.

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/EventProcessor.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/EventProcessor.kt
@@ -36,6 +36,7 @@ internal class EventProcessorImpl(
             is Event.FeatureConfig -> featureConfigEventReceiver.onEvent(event)
             is Event.Unknown -> kaliumLogger.withFeatureId(EVENT_RECEIVER).i("Unhandled event id=${event.id}")
         }
+        kaliumLogger.withFeatureId(EVENT_RECEIVER).i("Updating lastProcessedEventId ${event.id}")
         eventRepository.updateLastProcessedEventId(event.id)
     }
 }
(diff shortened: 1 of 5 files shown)


vitorhugods:
Labeler is currently broken: https://github.com/srvaroa/labeler/issues/43.

Seems to be caused by: https://github.com/niemeyer/gopkg/issues/63

ohassine (approved):
Bravo! Nice catch 👌

MohamadJaara (commented):
LGTM 🚀

alexandreferris (approved):

PR #792, opened 2022-08-15
chore: close open calls

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-2124 - Join call button doesn't go away](https://wearezeta.atlassian.net/browse/AR-2124)

### Issues

Join call button does not go away even if there is no more active call.

### Causes (Optional)

Due to the implementation of saving calls information in the local database, if a call was wrongly closed (app was closed without the chance to update the call status) when re-opening the app the join call button would still be there even though there was no active call in place.

### Solutions

`Close` app open calls when opening the app (because if there is still an ongoing call, AVS will let the device know about it and then we can follow the normal flow as its supposed to be.

#### Test Coverage (Optional)

#### How to Test

- Open App
- Receive/start group call
- Accept -> end Call / or / Reject the call | `Just so the join call button appears`
- close the app
- wait for group call to fully finish (all participants left the call and there is no active call anymore)
- open the app and you should not see the join call button anymore.

### Notes (Optional)

When talking with @vitorhugods we decided on making a proper start of sync managers in `UserSessionScope` instead of randomly (or kinda depending on something else) to start them.


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -574,4 +577,20 @@ abstract class UserSessionScopeCommon(
...
+
+    override val coroutineContext: CoroutineContext = SupervisorJob()
+
+    fun onInit() {
+        launch {
+            // TODO: Add a public start function to the Managers
+            incrementalSyncManager
+            slowSyncManager
+
+            callRepository.updateOpenCallsToClosedStatus()
```
vitorhugods, line 589: I'm thinking if this shouldn't be done by the `CallScope` or similar.

Or at least something a bit more "higher-level" than the repository. Because this sets precedent for other repositories being called here directly and I can imagine `UserSessionScopeCommon` doing far too much stuff.

But it works for now!

vitorhugods (approved):
👨🏼‍🍳🤌🏻

ohassine (approved):
lgtm ✨

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
alexandreferris: Indeed! I thought on adding the UseCase (I did the implementation of it) but ended up just removing it and going straight with the repository for now 🤔

PR #800, opened 2022-08-16
fix: Notification do not come

borichellow:
[Task-2184](https://wearezeta.atlassian.net/browse/AR-2184)

# What's new in this PR?
fix for the issue

### Issues

Notifications were not coming after some changes

### Causes (Optional)

All the notifications comes from 3 places: 
- messages
- connection requests
- ephemeral notifications - things that has no record in DB, but user needs to be notified about (deleting the conversation e.g.)

but when we combine it we do it by the `Flow.combine` operator, which needs at least 1 emitting from each of the combined Flows. 
The problem is that `EphemeralNotificationsManager.observeEphemeralNotifications()` in fact is a Channel that emits only when needed, so there is no "1 emitting" that we needed to combine it with others

### Solutions

Changing the "getting notifications" flow: from now we don't need to get all the notifications in 1, but we can get it one-by-one and Android will update the existed if needed. 
So I changed `Flow.combine` into `merge` Flows

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -3,10 +3,10 @@ import com.github.leandroborgesferreira.dagcommand.DagCommandPlugin
 import com.github.leandroborgesferreira.dagcommand.extension.CommandExtension
 
 buildscript {
-    val kotlinVersion = "1.7.10"
-    val dokkaVersion = "1.7.10"
-    val sqlDelightVersion = "2.0.0-alpha03"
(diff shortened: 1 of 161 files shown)


alexandreferris (approved):

yamilmedina (approved):
Nice catch ! 🐛  💯

borichellow:
Actually now i see that this solution is not perfect too....will update it

alexandreferris (approved):

PR #820, opened 2022-08-19
fix: refactor assets download to avoid OOM exception (AR-2125)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently, the app is crashing by out of memory exception, when we are dealing with big files, ie: > 60 MB
This happens because we were handling `ByteArray` (download and decryption) being an expensive DS for the device to handle.

### Solutions

Change the approach to read the `ByteArray` by chunks and write to a temp file.
On the decryption part, being JVM code, we can handle `InputStream`s =)

----

----

----
##### References

### DIFF
--- a/cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/utils/AESUtils.kt
+++ b/cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/utils/AESUtils.kt
@@ -110,11 +110,8 @@ internal class AESDecrypt(private val secretKey: AES256Key) {
                 val source = bufferedSource.inputStream().source()
                 val contentBuffer = Buffer()
                 var byteCount: Long
-                while (source.read(contentBuffer, BUFFER_SIZE).also {
-                        byteCount = it
-                        size += byteCount
-                    } != -1L
-                ) {
+                while (source.read(contentBuffer, BUFFER_SIZE).also { byteCount = it } != -1L) {
+                    size += byteCount
                     outputSink.write(contentBuffer, byteCount)
                     outputSink.flush()
                 }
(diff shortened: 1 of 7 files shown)


cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/utils/AESUtils.kt
```diff
@@ -106,12 +107,15 @@ internal class AESDecrypt(private val secretKey: AES256Key) {
...
-                outputSink.buffer().use {
-                    it.write(data)
-                    it.flush()
+                val source = bufferedSource.inputStream().source()
+                val contentBuffer = Buffer()
+                var byteCount: Long
+                while (source.read(contentBuffer, BUFFER_SIZE).also { byteCount = it } != -1L) {
+                    size += byteCount
+                    outputSink.write(contentBuffer, byteCount)
+                    outputSink.flush()
```
yamilmedina, line 116: @MohamadJaara this did the trick on the decrypt part =) now we are able to download without issues, tested until 97 MB. We can meet again, so we discuss the writing of temp files and avoid possible collisions

network/src/commonMain/kotlin/com/wire/kalium/network/api/asset/AssetApiImpl.kt
```diff
@@ -150,3 +184,5 @@ class StreamAssetContent(
         channel.close()
     }
 }
+
+private const val BUFFER_SIZE = 1024 * 4L
```
gongracr, line 188: Why not 8KB?

gongracr (approved):
Clean and beautiful job!!! 🌹 🦋 
Out of curiosity, did you also try uploading big ASSets and check it goes alright?

network/src/commonMain/kotlin/com/wire/kalium/network/api/asset/AssetApiImpl.kt
yamilmedina: Just to avoid the small chance to use extra memory, but is a valid point, we could make it 8KB and tweak down if failures, would be more efficient.

Also, I'll check this uploading, I tried with a 65 MB and worked, but will try with a bigger one and more edge cases =)

yamilmedina: Just tested, with several a prev. Downloaded asset of 91 MB, and the app could not handle :(  So I believe we need to do the same for upload asset, namely for encryption, use a form of output/input stream instead of `ByteArray`. We can sync tomorrow to have a better an idea on how to do it, maybe I can raise another PR.

PR #828, opened 2022-08-23
fix: show message after removing conversation member (AR-2038)

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Message is only showed when user leaves conversation, but not where we are removing other user from conversation

### Solutions

Disable checking if selfUser left conversation


gongracr (approved):
It seems backend also doesn't forward these kind of events via the websocket.

Strange enough, seems that the same error happens when the self user is the one adding other users in a group where he is admin. If you could also add it we would kill 2 🐦 with 1 🗿 here but I also understand that falls a bit out of the scope of this PR. 

Good catch in any case 🦅

typfel:
I don't have the full context but generally the backend doesn't forward events on the websocket for the client which performs an action. The client performing the action expected to update his/her local state from the response of the action he/her performed. 

NOTE:  that the event will be returned from the /notifications on the next incremental sync though.

gongracr:
so @typfel you are saying that if we create the event id "on the fly", on next incremental sync we will observe that event duplicated as they will have different ids?

typfel:
> so @typfel you are saying that if we create the event id "on the fly", on next incremental sync we will observe that event duplicated as they will have different ids?

Yes, generally there needs to be some filter in place depending on context. For example:  insert system message for user being removed unless user has already been removed from the group.

MohamadJaara (approved):

PR #831, opened 2022-08-24
fix: group calls white screen

alexandreferris:
…ast data

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-2179 - Group Calls stuck in white screen](https://wearezeta.atlassian.net/browse/AR-2179)

### Issues

When joining group call, the user would get stuck in white screen.

### Causes (Optional)

Due to receiving a data Pointer in SFT Callback, when really working with it inside the scope, it would get toasted because the Pointer data would no more exist correctly then having an error when trying to connect to SFT servers.

### Solutions

Move logic of data Pointer to outside of calling scope when handling it.

#### How to Test

- Uninstall App
- Install App (needs to be a fresh install as it was always happening on first call)
- Login
- Receive Call and accept it
- User should not be stuck in white screen anymore.


ohassine (approved):
🚀🚀

PR #857, opened 2022-08-31
feat: unread mentions count [AR-2278]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

To make unread message badges work well with different types of muting conversations, we need to have the option to observe not only unread messages count, but also unread mentions count. The problem is that we only use very simplified way of checking if the message contains mentions (checking if the text of the message contains the username or handle) which works for notifications but in order to count messages containing mentions it's better to create a desired db query rather than take all messages and check their contents one by one.

### Solutions

Created new table to store message mentions which are taken from the proto of `Text` message type. Implemented a query to count unread messages with mentions and added as a field in `Conversation`. Changed the way such values are taken to make sure Flows are used and we can observe changes .

#### Test Coverage (Optional)

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
@@ -11,6 +11,7 @@ import com.wire.kalium.logic.data.id.QualifiedID
 import com.wire.kalium.logic.data.id.TeamId
 import com.wire.kalium.logic.data.message.Message
 import com.wire.kalium.logic.data.message.MessageMapper
+import com.wire.kalium.logic.data.user.SelfUser
 import com.wire.kalium.logic.data.user.UserId
 import com.wire.kalium.logic.data.user.UserRepository
 import com.wire.kalium.logic.di.MapperProvider
@@ -20,6 +21,7 @@ import com.wire.kalium.logic.functional.fold
 import com.wire.kalium.logic.functional.isLeft
 import com.wire.kalium.logic.functional.isRight
 import com.wire.kalium.logic.functional.map
+import com.wire.kalium.logic.functional.mapRight
 import com.wire.kalium.logic.functional.onFailure
 import com.wire.kalium.logic.functional.onSuccess
 import com.wire.kalium.logic.kaliumLogger
@@ -40,6 +42,7 @@ import com.wire.kalium.persistence.dao.QualifiedIDEntity
 import com.wire.kalium.persistence.dao.message.MessageDAO
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flatMapLatest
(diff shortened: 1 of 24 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -278,64 +280,89 @@ class ConversationDataSource(
...
+                    }
+                )
+            }
+    }
+
+    private suspend fun observeUnreadMessageCount(conversation: Conversation): Flow<Long> {
+        return if (conversation.supportsUnreadMessageCount && hasNewMessages(conversation)) {
+            messageDAO.observeUnreadMessageCount(idMapper.toDaoModel(conversation.id))
+        } else {
+            flowOf(0L)
```
typfel, line 357: If a conversation as no new messages we return `flowOf(0L)`, what happens when a new message arrives in that conversation? Will we re-start the observation then or am I missing the context of where this is used?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageMention.kt
```diff
@@ -0,0 +1,50 @@
...
+
+import com.wire.kalium.logic.data.id.IdMapper
+import com.wire.kalium.logic.data.user.UserId
+import com.wire.kalium.persistence.dao.message.MessageEntity
+import com.wire.kalium.protobuf.messages.Mention
+
+data class MessageMention(
+    val start: Int,
+    val length: Int,
+    val userId: UserId?
```
typfel, line 11: Why is the `userId` optional? Because the qualified ID on the proto mention is optional and we don't deal with the fallback to the unqualified ID?

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
```diff
@@ -126,4 +126,4 @@ isUserMember:
 SELECT user FROM Member WHERE conversation = ? AND user = ?;
 
 whoDeletedMeInConversation:
-SELECT sender_user_id FROM Message WHERE id IN (SELECT message_id FROM MessageMemberChangeContent WHERE conversation_id = :conversation_id AND member_change_type = "REMOVED" AND member_change_list LIKE :self_user_id) ORDER BY DateTime(date) DESC LIMIT 1;
+SELECT sender_user_id FROM Message WHERE id IN (SELECT message_id FROM MessageMemberChangeContent WHERE conversation_id = :conversation_id AND member_change_type = "REMOVED" AND member_change_list LIKE ('%' || :self_user_id || '%')) ORDER BY DateTime(date) DESC LIMIT 1;
```
typfel, line 129: Is this related to the mention changes?

typfel (changes requested):
💯

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
saleniuk: No, but while working on sql query for mentions I found out that this one isn't fully correct and won't work for system messages containing more than one user so I fixed that in the meantime. 😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageMention.kt
saleniuk: According to the [doc](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/157319329/Use+case+receive+and+display+mentions):
>The user_id field is optional, because in future we might support mentions to everyone. At the moment though it should contain a user ID.

So for now, we only handle direct mentions thus we ignore the ones that doesn't have userId, but we still keep them for the future.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
saleniuk: You're right, it won't work without a proper combination with `flatMap`, so I removed `hasNewMessages(conversation)`, this way it will always observe directly the DAO, the `if` with `conversation.supportsUnreadMessageCount` can stay because the type of the conversation can't be changed.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageMention.kt
typfel: Ok I see, I would suggest that we filter out null userIds at the DAO level though so that the logic level can assume that we always have a user IDs.

saleniuk: Ok 👍

typfel (approved):

PR #859, opened 2022-09-01
fix: join a call with video enabled by default

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When you end a call and you camera is on at that time, next time you join the call you will join with camera on

### Solutions

We need to set camera state to false when ending the call

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

1. In a group call turn on your camera and then end the call
2. join the same call again, you should join with camera off

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallUseCase.kt
@@ -1,5 +1,6 @@
 package com.wire.kalium.logic.feature.call.usecase
 
+import com.wire.kalium.logic.data.call.CallRepository
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.feature.call.CallManager
 import com.wire.kalium.util.KaliumDispatcher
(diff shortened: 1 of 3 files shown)


MohamadJaara (approved):
LGTM 🚀

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallUseCase.kt
```diff
@@ -8,10 +9,12 @@ import kotlinx.coroutines.withContext
 
 class EndCallUseCase(
     private val callManager: Lazy<CallManager>,
+    private val callRepository: CallRepository,
     private val dispatchers: KaliumDispatcher = KaliumDispatcherImpl
 ) {
 
     suspend operator fun invoke(conversationId: ConversationId) = withContext(dispatchers.io) {
         callManager.value.endCall(conversationId)
+        callRepository.updateIsCameraOnById(conversationId.toString(), false)
```
typfel, line 18: What if the call is not actively ended by the user but closed due to connectivity issues? Should we call `updateIsCameraOnById()` in those cases as well?

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallUseCase.kt
ohassine: No in case of timeout we don't need to call `updateIsCameraOnById()` because the call is `CLOSED`(which is no longer useful ). We just need it for calls that are still ongoing

typfel (approved):

PR #865, opened 2022-09-06
chore: Make search user result return flow (AR-2303)

borichellow:
# What's new in this PR?

Made `SearchUsersUseCase` return `Flow` in case of success, so we can observe its changes
Also removed some useless `suspend` modifications in some funs and useless `UserSearchResult`.

### Issues

During the user search users ConnectionStatus was not updating in run-time.

### Causes (Optional)

Search result was just 1-time response, not Flow that we could observe.

### Solutions

Make `SearchUserResult.Success` contains `Flow<List<OtherUsers>` so we could observe it


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
```diff
@@ -58,11 +57,11 @@ interface ConnectionRepository {
     suspend fun fetchSelfUserConnections(): Either<CoreFailure, Unit>
     suspend fun sendUserConnection(userId: UserId): Either<CoreFailure, Unit>
     suspend fun updateConnectionStatus(userId: UserId, connectionState: ConnectionState): Either<CoreFailure, Connection>
-    suspend fun getConnections(): Either<StorageFailure, Flow<List<ConversationDetails>>>
+    fun getConnections(): Either<StorageFailure, Flow<List<ConversationDetails>>>
     suspend fun insertConnectionFromEvent(event: Event.User.NewConnection): Either<CoreFailure, Unit>
-    suspend fun observeConnectionList(): Flow<List<ConversationDetails>>
```
typfel, line 63: Why are you removing the suspend annotation? It's there for compatibility with JS  since all database operations are async there.

typfel (changes requested):

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/user/SearchUserDirectoryUseCaseTest.kt
```diff
@@ -1,104 +1,104 @@
...
-                        botService = null,
-                        deleted = false
-                    )
-                }
-            }
-        )
-    }
-
-}
+// import com.wire.kalium.logic.NetworkFailure
```
typfel, line 3: Should be deleted?

PR #874, opened 2022-09-08
chore: adjust mapping of active speakers to be more optimized

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Logic for mapping active speakers is very slow.

### Causes (Optional)

We go through every participant first then check if they are active speakers or not.

### Solutions

Go through active speakers first and not participants.

#### Test Coverage (Optional)

#### How to Test

- Open App
- Create/Join group call with many participants and let them talk
- See that the app doesn't freeze and (and device)heat as much as before.
- Everything else should look the same as before.


ohassine (approved):
Good catch, well done 🚀

PR #880, opened 2022-09-09
feat: hard logout option [AR-1653]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We need to have the option to choose whether we want to do a hard or soft logout.

### Solutions

Clearing user data extracted to a separate use case, added parameter to choose the logout type.

#### Test Coverage (Optional)

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/session/SessionRepository.kt
```diff
@@ -90,7 +90,7 @@ internal class SessionDataSource(
 
     override fun allValidSessions(): Either<StorageFailure, List<AuthSession>> =
         wrapStorageRequest {
-            sessionStorage.allSessions()?.filter { it.value is AuthSessionEntity.Valid }?.values?.toList()
+            sessionStorage.allSessions()?.filter { it.value is AuthSessionEntity.Valid }?.values?.toList()?.ifEmpty { null }
```
mchenani, line 93: here is returning null, but the function return is not nullable, am I missing something here?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -572,15 +574,20 @@ abstract class UserSessionScopeCommon internal constructor(
             userId,
             userDatabaseProvider.metadataDAO
         )
+    val clearUserData: ClearUserDataUseCase
```
mchenani, line 577: can we have this usecase private? I think it would be safer!

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/ClearUserDataUseCase.kt
```diff
@@ -0,0 +1,51 @@
...
+
+class ClearUserDataUseCaseImpl @Suppress("LongParameterList") constructor(
+    private val userId: QualifiedID,
+    private val authenticatedDataSourceSet: AuthenticatedDataSourceSet,
+    private val clientRepository: ClientRepository,
+    private val mlsClientProvider: MLSClientProvider
+) : ClearUserDataUseCase {
+
+    override suspend operator fun invoke() {
+        clearCrypto()
```
mchenani, line 22: I suggest to validate the user session here, like if it's invalid then clear it! that would be safer.

saleniuk: The use case is private now so it's safer 😄 We shouldn't clear the session there because it's used in the logout so we don't want to remove the session but to to keep it as `Invalid`.

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -46,6 +46,8 @@ abstract class CoreLogicCommon(
     @Suppress("MemberVisibilityCanBePrivate") // Can be used by other targets like iOS and JS
     abstract fun getSessionScope(userId: UserId): UserSessionScope
 
+    abstract fun deleteSessionScope(userId: UserId) // TODO remove when proper use case is ready
```
MohamadJaara, line 49: i will replace it with a stand alone usecase once my PR is merged

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
```diff
@@ -1,90 +1,65 @@
...
+        sessionRepository.allValidSessions().onSuccess {
+            val updatedUserId = it.first().session.userId
+            // there is another valid session so switch to that one
+            sessionRepository.updateCurrentSession(updatedUserId)
+            return updatedUserId
         }
+        // there is no other valid session so update current one
+        sessionRepository.updateCurrentSession(userId)
+        return null
     }
```
MohamadJaara, line 61: i suggest letting consuming clients update the current account when it's possible to do so
e.g.
a CLI client can call Logout, Update and delete after each other with no issue
an Android client will execute Logout -> Update current -> navigate somewhere -> delete user stuff 
web can do what ever web does

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/GetOrRegisterClientUseCase.kt
```diff
@@ -0,0 +1,42 @@
...
+class GetOrRegisterClientUseCaseImpl(
+    private val observeCurrentClientId: ObserveCurrentClientIdUseCase,
+    private val selfClients: SelfClientsUseCase,
+    private val registerClient: RegisterClientUseCase,
+    private val clearClientData: ClearClientDataUseCase,
+    private val proteusClient: ProteusClient
+) : GetOrRegisterClientUseCase {
+
+    override suspend fun invoke(registerClientParam: RegisterClientUseCase.RegisterClientParam): RegisterClientResult {
+        val result: RegisterClientResult? = observeCurrentClientId().firstOrNull()?.let { currentClientId ->
```
MohamadJaara, line 21: can be replaced with a function that does not return a flow and remove the firstOrNull()

MohamadJaara (changes requested):
Great work as always 🚀, only small comment about replacing the current client flow

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
vitorhugods: Please make sure we call `sessionScope.cancel()` too!
This way we force some of our "lifecycle" tasks to stop too, like CallManager and some Sync stuff.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/GetOrRegisterClientUseCase.kt
saleniuk: Ok, changed 👌

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
saleniuk: Ok, changed 👌

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
saleniuk: My solution is temporary, as Mo said he works on a proper use case, but I added `.cancel()` to this temporary code

MohamadJaara (approved):

PR #881, opened 2022-09-09
fix: duplicated event handling

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Users are constantly getting "Could not decrypt message".

### Causes

On the surface it is caused by Proteus rejecting decryption of duplicated messages.

A bit deeper, the following:

Consider that the user was offline and just becoming online, and has two pages of pending events to be fetched from the server.

While fetching the first page, an event would come in through the websocket.

This event would also appear when fetching the second page of pending events.

The current mechanism stores the pending events as we proccess them, so we can prevent duplications later when handling websocket events.

HOWEVER, it compares the whole Event. `NewMessage` contains `EncryptedData`, which some time in the past stopped honoring the `equals` method. And then comparing both events would lead to a `false` result, making the `EventGatherer` publish the same event twice (from different sources).

### Solutions

Make `EncryptedData` and other "data" classes in `cryptography` correctly implement the `equals` method.

> **Note**: We need to break these classes from being value classes, as the [language specification](https://kotlinlang.org/spec/declarations.html#value-class-declaration) says that value classes can **not** override `equals`.

#### Test Coverage

----

----
##### References


vitorhugods:
> Why do we compare the whole event? Each event has an event ID isn't it enough to compare that?

Not really.
It's `1 notification - N events`.
There's the remote possibility the backend might send multiple events in a single notification because they have the same ID.

MohamadJaara (approved):

yamilmedina (commented):
🧑‍🎄

typfel:
> > Why do we compare the whole event? Each event has an event ID isn't it enough to compare that?
> 
> Not really. It's `1 notification - N events`. There's the remote possibility the backend might send multiple events in a single notification because they have the same ID.

Unfortunate model used by BE (they should change it) I'm fairly sure in practice an event never has more than one payload. But ok let's not trust the BE. I hope we comparing the ID first so that it bails out quickly and we are not comparing the bytes of each payload.

PR #885, opened 2022-09-11
feat: add a use-case to get a message by ID (AR-1882)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

While working on pagination on Reloaded, I realised that we don't have a way to fetch a message by its coordinates (ConversationID and MessageID).
At some places we are iterating through the messages. For example, to fetch Assets info when the user clicks on a message.

With pagination we don't have a simple list of messages to iterate through. And it is just more scalable to fetch from the DB instead of iterating lists of lists, or having to change this as we change how we store the data on the ViewModel.

It also allows a more decoupled ViewModel. Where one can handle the list of messages, and another one can handle the asset download solely based on message IDs.

### Solutions

Create a `GetMessageByIdUseCase` that exposes the underlying `MessageRepository$getMessageById` function.

#### Test Coverage

----

----
##### References


logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/message/GetMessageByIdUseCaseTest.kt
```diff
@@ -0,0 +1,96 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertIs
+
+class GetMessageByIdUseCaseTest {
+
+    private val testDispatchers: KaliumDispatcher = TestKaliumDispatcher
+
+    @Test
+    fun givenMessageAndConversationId_whenInvokingUseCase_thenShouldCallMessageRepository() = runTest(testDispatchers.io) {
+        val messageId = "messID"
```
Garzas, line 29: You can use TEST_MESSAGE_ID

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/message/GetMessageByIdUseCaseTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        val conversationId = TestConversation.ID
+
+        val (arrangement, getMessageByIdUseCase) = Arrangement()
+            .withRepositoryMessageByIdReturning(conversationId, messageId, Either.Left(CoreFailure.Unknown(null)))
+            .arrange()
+
+        getMessageByIdUseCase(conversationId, messageId)
+
+        verify(arrangement.messageRepository)
```
Garzas, line 38: Idk which solution is better but can this check be moved to the other two tests?

Garzas (approved):
LGTM, just small suggestions 🚀

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/message/GetMessageByIdUseCaseTest.kt
vitorhugods: I try avoiding mixing different assertions as much as I can.

Having fewer per test, trying to grouped them based on the behaviour we are usually has the benefit that when it fails we can read the test name and easily understand what is failing.

Otherwise we need to dive deeper to udnerstand which one of the assertions is failing and why it is being tested.

Imagine we don't have `givenMessageAndConversationId_whenInvokingUseCase_thenShouldCallMessageRepository`, and we move this assertion to `givenRepositoryFails_whenInvokingUseCase_thenShouldPropagateTheFailure` and then this verification fails.

This means that "should propagate failure" is failing.
But it's failing because we're not passing the proper parameter to the repository?! The test doesn't reflect the assertion.

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/message/GetMessageByIdUseCaseTest.kt
vitorhugods: Oooh yeah. Completely forgot about this 😅.

Fixed.

PR #892, opened 2022-09-13
fix: Get conversation recipients from local DB only (AR-2365)

borichellow:
# What's new in this PR?

While sending a message app was getting all the conversations clients (all client_ids for all users in the conversation) from the API. It's redundant, we can get them from the DB.
Also there was some small error in mapping of UserId in SendMessageErrorHandling

### Causes (Optional)

It was implemented like this from the beginning

### Solutions

Make kalium get client_ids from the local DB and fetch it from the API only if something is missed.
Fix UserId mapping in SendMessageErrorHandling and add test for it


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -617,9 +617,12 @@ class ConversationDataSource(
...
     override suspend fun getConversationRecipients(conversationId: ConversationId): Either<CoreFailure, List<Recipient>> =
-        getConversationMembers(conversationId).map { it.map(idMapper::toApiModel) }.flatMap {
-            wrapApiRequest { clientApi.listClientsOfUsers(it) }.map { memberMapper.fromMapOfClientsResponseToRecipients(it) }
-        }
+        getConversationMembers(conversationId)
+            .map { it.map(idMapper::toDaoModel) }
+            .flatMap {
+                wrapStorageRequest { clientDAO.getClientsOfUsersByQualifiedIDs(it) }
+                    .map { memberMapper.fromMapOfClientsEntityToRecipients(it) }
+            }
```
MohamadJaara, line 625: We can move this logic one layer down to the DB, something like inside Conversations.sq

```
SELECT user_id, GROUP_CONCAT(id) FROM Client WHERE id IN (SELECT user FROM Member WHERE conversation = ?) GROUP BY user_id;
```
This way we have one db read operation to get Recipients

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/client/ClientDAOImpl.kt
```diff
@@ -37,6 +37,15 @@ internal class ClientDAOImpl(private val clientsQueries: ClientsQueries) : Clien
...
             .map(mapper::toModel)
 
+    override suspend fun getClientsOfUsersByQualifiedIDs(ids: List<QualifiedIDEntity>): Map<QualifiedIDEntity, List<Client>> =
+        clientsQueries.transactionWithResult {
+            ids.associateWith { qualifiedID ->
+                clientsQueries.selectAllClientsByUserId(qualifiedID)
+                    .executeAsList()
+                    .map(mapper::toModel)
+            }
+        }
```
MohamadJaara, line 47: Same as the comment above

MohamadJaara (commented):
Small suggestion to read the conversation Recipients with one sql query

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/client/ClientDAOImpl.kt
```diff
@@ -37,8 +37,23 @@ internal class ClientDAOImpl(private val clientsQueries: ClientsQueries) : Clien
...
+
     override suspend fun deleteClientsOfUserByQualifiedID(qualifiedID: QualifiedIDEntity): Unit =
         clientsQueries.deleteClientsOfUser(qualifiedID)
 
     override suspend fun deleteClient(userId: QualifiedIDEntity, clientId: String) = clientsQueries.deleteClient(userId, clientId)
+
+    override suspend fun getClientsOfConversation(id: QualifiedIDEntity): Map<QualifiedIDEntity, List<Client>> =
+        clientsQueries.selectAllClientsByConversation(id.value)
```
MohamadJaara, line 55: using id.value can cause issues in federated environments

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Clients.sq
```diff
@@ -28,3 +28,6 @@ SELECT * FROM Client;
 
 selectAllClientsByUserId:
 SELECT * FROM Client WHERE user_id = ?;
+
+selectAllClientsByConversation:
+SELECT * FROM Client WHERE user_id IN (SELECT user FROM Member WHERE conversation LIKE ('%' || :conversation_id || '%'));
```
MohamadJaara, line 33: i think this will allow to use the full conversation id not just it's value
```
SELECT * FROM Client WHERE user_id IN (SELECT user FROM Member WHERE conversation = :conversationId );
```

MohamadJaara (changes requested):
small change about using only the conversation ID value without the domain

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Clients.sq
borichellow: hmm, true, i copied it from `Member.sq`

MohamadJaara (approved):
lgtm

PR #893, opened 2022-09-13
feat: Use case to send read/delivery confirmations (w/o persistence) SQPIT-1403

mythsunwind:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Kalium is missing read and delivery confirmations

### Solutions

Implementation of read/delivery confirmation sending. This PR does not include persisting of these confirmations. It just contains a use case to send confirmations.

### Testing

TODO

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendConfirmationUseCase.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.wire.kalium.logic.data.sync.SlowSyncStatus
+import com.wire.kalium.logic.data.user.UserRepository
+import com.wire.kalium.logic.functional.Either
+import com.wire.kalium.logic.functional.flatMap
+import com.wire.kalium.logic.functional.onFailure
+import com.wire.kalium.logic.kaliumLogger
+import kotlinx.coroutines.flow.first
+import kotlinx.datetime.Clock
+
+class SendConfirmationUseCase internal constructor(
```
typfel, line 19: I'd expect the use case to work differently.  I'd say this `SendConfirmationUseCase` an internal use case or maybe it's not a use case at all but an internal helper component.

The read confirmation are sent when the user "reads" messages in a conversation so it probably makes sense to call the `SendConfirmationUseCase` from the `UpdateConversationReadDateUseCase`, which is invoked when the user "reads" messages. 

Delivery confirmations is sent upon receiving messages which is tied to sync inside Kalium. Here I would expect delivery confirmations to be triggered from the `ConversationEventReceiver`.

Lastly I'd like the `SendConfirmationUseCase` to support the feature of multiple confirmations at once so that they can be batched into a single message.

typfel (changes requested):
Nice work but I think some parts are missing.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendConfirmationUseCase.kt
mythsunwind: Thanks for your review. It is touching the same points that i have about my PR. It is not a method that should be used by a client implementation but it is necessary for QA to test client behavior. We recently found a better place for such things and i moved the use case into the debug scope.

Additionally i implemented the possibility to send multiple confirmations at once. Which i forgot in my PR.

Please re-review 😃

mythsunwind:
PR was replaced by https://github.com/wireapp/kalium/pull/1354

PR #897, opened 2022-09-13
feat: kalium testservice SQPIT-1165

mythsunwind:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

QA needs to receive MLS messages of various kinds. Usually [ETS](https://github.com/wireapp/wire-web-ets/) was doing this for proteus but now we need the same for MLS. It was decided that ETS will deprecate and a new service will be implemented based on kalium. This is the kalium testservice.

----

----

----
##### References


logic/src/jvmMain/kotlin/com/wire/kalium/logic/configuration/ClientConfigImpl.kt
```diff
@@ -18,6 +18,6 @@ actual class ClientConfigImpl : ClientConfig {
     }
 
     override fun clientType(): ClientType {
-        return ClientType.Temporary
+        return ClientType.Permanent
     }
```
vitorhugods, line 22: @typfel does changing this break something for MLS and the like?

typfel: The effect is that CLI will register permanent clients. Not end of the world but we should make this configurable.

typfel: Has no impact on MLS

mythsunwind: I can make a PR for configuration of ClientType once this PR is merged

mythsunwind: I reverted the change so that this PR can be merged without affecting CLI

vitorhugods (approved):
Personally there's nothing I feel entitled to review, as this `testservice` is a project that you are leading, @mythsunwind 😅.

I just don't know about the long run. If we need to change the Kalium's structure, repository's PR format, or codestyle, for example. These projects are in the same repo, how to handle that? 🤔 

I don't mind it being together with Kalium for now, so you can keep on building and seeing if Kalium suits this project, and we can keep them growing both hand-in-hand.

Maybe in the future we start to version Kalium and publish it the artifacts, so `testservice` can be decoupled and just use `kalium-logic` as a dependency for example.

PR #899, opened 2022-09-14
chore: Remove useless clients when possible (AR-2375)

borichellow:
# What's new in this PR?

When user Persist OtherUser Clients, it also removes clients that are not in the list. 
This will increase performance a bit while sending messages.

### Issues

PersistOtherUserClientsUsecase was only saving the new Clients into DB and does nothing on Clients that do not exist anymore. So when user sends the messages to that OtherUser, app crypts message for not existed Clients too.

### Causes (Optional)

It was implemented like that before 🤷‍♂️

### Solutions

Make `PersistOtherUserClientsUsecase` not only save new Clients but remove redundant too


persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/client/ClientDAO.kt
```diff
@@ -18,4 +18,5 @@ interface ClientDAO {
     suspend fun deleteClientsOfUserByQualifiedID(qualifiedID: QualifiedIDEntity)
     suspend fun deleteClient(userId: QualifiedIDEntity, clientId: String)
     suspend fun getClientsOfConversation(id: QualifiedIDEntity): Map<QualifiedIDEntity, List<Client>>
+    suspend fun insertClientsAndRemoveRedundant(qualifiedID: QualifiedIDEntity, clients: List<Client>)
```
MohamadJaara, line 21: Missing unit test

MohamadJaara (changes requested):
great job 🚀, only missing tests for `insertClientsAndRemoveRedundant`

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/client/ClientDAO.kt
borichellow: unitTests are missed for the whole `ClientDAOImpl` ;P

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/client/ClientDAOImpl.kt
```diff
@@ -32,6 +32,14 @@ internal class ClientDAOImpl internal constructor(
         }
     }
 
+    override suspend fun insertClientsAndRemoveRedundant(qualifiedID: QualifiedIDEntity, clients: List<Client>) =
+        clientsQueries.transaction {
+            clients.forEach { client ->
+                clientsQueries.insertClient(client.userId, client.id, client.deviceType)
+            }
+            clientsQueries.deleteClientsOfUserExcept(qualifiedID, clients.map { it.id })
+        }
```
vitorhugods, line 41: Considering this is done in a loop, using [Grouped Statements](https://cashapp.github.io/sqldelight/multiplatform_sqlite/grouping_statements/) is significantly better than just a transaction, performance-wise.

vitorhugods (approved):

MohamadJaara (approved):
LGTM 🚀

PR #914, opened 2022-09-16
fix(calling): participants handling optimization

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Description

- Retrieving user info from database is taking too much time

In the old way we were getting all data of user  by making a `select *` query on db . And it took an average of 400ms to finish the full process of retrieving 

Now with just retrieving what we need, specifying the columns that we need only, we see an some improvement. It took about 30-100ms

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserModel.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserModel.kt
@@ -86,6 +86,13 @@ data class SelfUser(
     override val availabilityStatus: UserAvailabilityStatus
 ) : User()
 
+data class OtherUserMinimized(
+    val id: UserId,
+    val name: String?,
+    val completePicture: UserAssetId?,
+    val userType: UserType,
+)
+
 data class OtherUser(
     override val id: UserId,
     override val name: String?,
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
@@ -56,7 +56,7 @@ interface UserRepository {
     suspend fun updateLocalSelfUserHandle(handle: String)
     suspend fun getAllKnownUsers(): Either<StorageFailure, List<OtherUser>>
     suspend fun getKnownUser(userId: UserId): Flow<OtherUser?>
-    fun getKnownUserMinimized(userId: UserId): OtherUserMinimized?
+    suspend fun getKnownUserMinimized(userId: UserId): OtherUserMinimized?
     suspend fun observeUser(userId: UserId): Flow<User?>
     suspend fun userById(userId: UserId): Either<CoreFailure, OtherUser>
     suspend fun updateSelfUserAvailabilityStatus(status: UserAvailabilityStatus)
@@ -233,7 +233,7 @@ internal class UserDataSource(
         userDAO.getUserByQualifiedID(qualifiedID = idMapper.toDaoModel(userId))
             .map { userEntity -> userEntity?.let { publicUserMapper.fromDaoModelToPublicUser(userEntity) } }
 
-    override fun getKnownUserMinimized(userId: UserId) = userDAO.getUserMinimizedByQualifiedID(
+    override suspend fun getKnownUserMinimized(userId: UserId) = userDAO.getUserMinimizedByQualifiedID(
         qualifiedID = idMapper.toDaoModel(userId)
     )?.let {
         publicUserMapper.fromDaoModelToPublicUserMinimized(it)
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
@@ -56,6 +56,7 @@ interface UserRepository {
     suspend fun updateLocalSelfUserHandle(handle: String)
     suspend fun getAllKnownUsers(): Either<StorageFailure, List<OtherUser>>
     suspend fun getKnownUser(userId: UserId): Flow<OtherUser?>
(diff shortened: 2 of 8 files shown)


persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAOImpl.kt
```diff
@@ -179,6 +179,18 @@ class UserDAOImpl(
...
+        userQueries.selectMinimizedByQualifiedId(listOf(qualifiedID)).executeAsOneOrNull().let {
+            it?.let { user ->
+                UserEntityMinimized(
+                    user.qualified_id,
+                    user.name,
+                    user.complete_asset_id,
+                    user.user_type
+                )
+            }
+        }
```
vitorhugods, line 192: Instead of mapping after the query, it's more performant if we pass a mapper function as the last argument of `selectMinimizedByQualifiedId`:

```suggestion
        userQueries.selectMinimizedByQualifiedId(listOf(qualifiedID)) { 
            qualifiedId, name, completeAssetId, userType ->
                UserEntityMinimized(
                    qualifiedId,
                    name,
                    completeAssetId,
                    userType 
                )
        }.executeAsOneOrNull()
```

It's probably better to move this mapping function to `UserMapper` and just pass it like this:

```kotlin
        userQueries.selectMinimizedByQualifiedId(
            listOf(qualifiedID), userMapper::toMinimizedEntity
        ).executeAsOneOrNull()
```

vitorhugods (commented):
Nice solution.

Just a note:
I was also investigating the performance of fetching users by ID and ended up optimising so we only take nanoseconds to fetch a user by caching in memory. The PR is open here #915.

It contains a benchmark comparing before/after. I wrote all the conclusions there.

So I suggest you experiment and see what fits better the needs of calling and participants change.

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAOImpl.kt
ohassine: done

alexandreferris (approved):
amazing 🚀

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserModel.kt
```diff
@@ -86,6 +86,13 @@ data class SelfUser(
     override val availabilityStatus: UserAvailabilityStatus
 ) : User()
 
+data class OtherUserMinimized(
```
yamilmedina, line 89: Only one question, what's `minimized` stands for ?  I don't know if it's related to minimized in the calling view. But maybe we can improve this name for something like `OtherUserPreview` ?

Not a blocker, at all, just maybe to be changed later 💪

yamilmedina (approved):
Greeeat! I like this PR, just the needed stuff for user. I added a small naming question/suggestion. No blocker for me, can be changed in future if we decide.

🌶️ 🚀

vitorhugods:
@ohassine I suggest you run your tests with the changes from #915 and compare. Once the users are in memory, the speed will be faster than doing these optimised queries, and won't use DB connections.

So I think this might actually be less performant now.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserModel.kt
ohassine: nothing related to calling, just light version of `OtherUser`

ohassine: nothing related to calling, just light version of `OtherUser`

ohassine:
> @ohassine I suggest you run your tests with the changes from #915 and compare. Because now, once the users are in memory, the speed will be faster than doing these optimised queries, and won't use DB connections.
> 
> So I think this might actually be less performant now.

Sure I will try it

ohassine:
> @ohassine I suggest you run your tests with the changes from #915 and compare. Because now, once the users are in memory, the speed will be faster than doing these optimised queries, and won't use DB connections.
> 
> So I think this might actually be less performant now.

| users retrieved and processed    | full user data object  | minimized user object |
| ---------| ------------- | ------------- |
| 1  | 27ms | 34ms  |
| 5  | 802ms  | 51ms  |
| 5  | 1058msl  | 54ms |
| 5  | 807ms  | 49ms |
| 1  | 9ms | 9ms |
| 1  | 12ms  | 8ms |

It's still better to use a minimized object with less field

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -56,6 +56,7 @@ interface UserRepository {
     suspend fun updateLocalSelfUserHandle(handle: String)
     suspend fun getAllKnownUsers(): Either<StorageFailure, List<OtherUser>>
     suspend fun getKnownUser(userId: UserId): Flow<OtherUser?>
+    fun getKnownUserMinimized(userId: UserId): OtherUserMinimized?
```
vitorhugods, line 59:
```
    suspend fun getKnownUserMinimized(userId: UserId): OtherUserMinimized?
```

The idea of making it suspendable is that even though SQLDelight queries are not suspendable for us yet, they can be in the near future, as [Async drivers are being supported](https://github.com/cashapp/sqldelight/commit/371f5038d03090b65e10bc79dd190973849e863c) and we will need them for JS.

This way we don't need to break compatibility in the future.

ohassine: okay I will update it

PR #919, opened 2022-09-18
feat(mls): fetch and authorize public keys (AR-1635)

mchenani:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?
To allow the server to send external commits, we need to fetch the public keys of the server and add them to each conversation we create to allow the server to submit their commits.

In this PR we fetch the public sender keys from the API when we want to create a conversation.

**This PR only works on API V2!**

#### Test Coverage

----

----

----
##### References


persistence/src/commonMain/db_global/com/wire/kalium/persistence/MLSPublicKey.sq
```diff
@@ -0,0 +1,19 @@
+import com.wire.kalium.persistence.dao.ConversationEntity;
+import com.wire.kalium.persistence.model.KeyType;
+
+CREATE TABLE MLSPublicKey(
```
typfel, line 4: Are we sure we want to store the public keys in the global DB? I have a feeling this is a bit premature since we don't how we will update the keys yet and that's something we will tackle in M5. Maybe it's enough to store the public keys in memory and fetch or retrieve on demand when creating a group.

mchenani: I think fetching these keys on demand will be more costly, since we here fetching them each 24 hours that would be ok to have them in the db. I think before releasing the M5 we can patch this part to maintain the altering old keys.

typfel: I don't think fetching keys on demand will a have negligible impact on the user experience since it's only necessary when the user is creating a new conversation and the keys aren't that big.

I'm just wary of already now creating a persisting scheme and updating use cases for the public keys before we know how that will work in M5. We don't have throw away the work you've already done, move it to a separate branch.

typfel:
@mchenani updated

typfel (approved):

PR #915, opened 2022-09-17
fix: slow performance when fetching user IDs repeatedly (AR-2435)

vitorhugods:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [X] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

## Issues

Reloaded is querying a lot of User Details in a sequence, which contributes with DB connections deadlocking when multiple operations are being performed.

## Solutions

A simple LRU Cache to store the flows of `UserEntity` in the DAO.

This way:
- We have a single instance of `Flow<UserEntity>` for each UserId, as long as the cache isn't full.
- When the query is invalidated, we make sure that **only one** request to the DB will be made to fetch the updated details even if we have multiple observers for the same User.

As visible in `UserDAOImpl`, this `LRUCache` can be easily used for other entities if we want, like `Conversation` and `Team` for example.

## Benchmark

I've done a benchmark comparing before and after adding this in-memory LRU cache.
I've ran the same benchmark multiple times getting similar results. The result below was picked in one of the runs without any reason in particular.

<details>
  <summary>Description of the benchmark</summary>

Test Setup:
- Create 1000 users
- From these users, create a list of 10 random qualified IDs, called `idsToFetch`
- Insert all the users into the DB

In each test:
- Start measuring time
- Take one ID from the `idsToFetch` list
- Select 3 times the same user ID
- Take the next ID from the list, until all IDs were fetched 3 times.
- Stop measuring time

The code of the test, without including the setup, was this:

```kotlin
fun selectWithCaching() = runTest {
    measureTime {
        idsToFetch.forEach { userId ->
            repeat(TIMES_PER_USER) {
                measureTime {
                    userDAO.getUserByQualifiedID(userId).first()
                }.also {
                    println("Took $it to get $userId")
                }
            }
        }
    }.also {
        println(
            "Took $it to select $USERS_TO_FETCH random users from a table of $USER_COUNT; " +
                    "Selecting each user $TIMES_PER_USER times"
        )
    }
}
```

</details>

<details>
  <summary>Results WITHOUT cache</summary>

```
Took 12.694ms to get QualifiedIDEntity(value=userId18, domain=userDomain)
[...]

PR #923, opened 2022-09-19
fix: improve performance of fetching metadata from the database

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We query the DB too often for simple stuff like UserID, ClientID, and also for TimeStamp keys.

### Causes

- To most, lack of caching. We should keep these values in memory, especially considering that they hardly change.
- For UserId, it's something static that does NOT change after the creation of `UserSessionScope`, so this should be injected into entities that need it.

### Solutions

- Deprecate `UserRepository.getSelfUserId`, so we start injecting `selfUserId`.

Had to make `getSelfUserId` suspendable to use the cache. This also required some changes in the code that was calling it.

- Cache the Metadata using the same approach used in #915.

### Testing

N/A
Caching is already covered in tests.

----

----
##### References


mchenani (approved):
🥁

PR #929, opened 2022-09-19
chore: move event storage to db

MohammedMokresh:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

as we have few values that been saved in the shared pref, and it quite affect the performance with many operations, we decided to move those value to the metadata table in the DB

### Solutions

insert the last event id in the meta data table

#### How to Test

run the app and check if the last event id saved in the meta data table and not in the user shared pref

----

----

----
##### References


MohamadJaara (approved):
LGTM 🚀

vitorhugods:
Is it really faster?
It's definitely faster for fetching, but we don't fetch this value often. We only care about this value a handful of time when fetching the pending events.

But we update that value all the time. Once for every single event that comes through Sync. Which means we are **inserting** all the time now.
And inserting is slower than selecting. So **without benchmarking, I don't really see it as a benefit**. In fact, **I would even say that this has the possibility of slowing things down**, because now it's another DB operation on each new event.

MohammedMokresh:
> Is it really faster? It's definitely faster for fetching, but we don't fetch this value often. We only care about this value a handful of time when fetching the pending events.
> 
> But we update that value all the time. Once for every single event that comes through Sync. Which means we are **inserting** all the time now. And inserting is slower than selecting. So **without benchmarking, I don't really see it as a benefit**. In fact, **I would even say that this has the possibility of slowing things down**, because now it's another DB operation on each new event.

Hmmmm the idea here was moving the values in the SP to reduce the operations in the SP as we are getting 
Fatal Exception: java.security.KeyStoreException: the master key android-keystore://androidx_security_master_key exists but is unusable ,  for some reason and that affect the app when it goes idle

PR #943, opened 2022-09-21
fix: Update Firebase token when new comes (AR-2489)

borichellow:
# What's new in this PR?

new flow for registering Firebase token 

### Issues

Firebase token is not updated on our BE side if Firebase decides to create a new one (it's rare but possible case).

### Causes (Optional)

FB token when it comes was just saving into local DB and when user login we register that token for the user.

### Solutions

Add some flag into UserScope "shouldUpdateFirebaseToken"
On every `init` in `UserScope` check this `shouldUpdateFirebaseToken` flag if it's true - update the token for BE.
And everytime Firebase generate a new token we save it into local DB **and** set `shouldUpdateFirebaseToken` for all users to `true`.


logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/notification/NotificationTokenRepository.kt
```diff
@@ -6,22 +6,22 @@ import com.wire.kalium.logic.functional.map
 import com.wire.kalium.logic.wrapStorageRequest
 import com.wire.kalium.persistence.client.TokenStorage
 
-data class NotificationToken(val token: String, val transport: String)
+data class NotificationToken(val token: String, val transport: String, val senderId: String)
```
vitorhugods, line 9: Can we specify what is `senderId`? Is it the client ID? Or User ID?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -62,6 +62,8 @@ internal interface UserRepository {
     suspend fun getAllKnownUsersNotInConversation(conversationId: ConversationId): Either<StorageFailure, List<OtherUser>>
     suspend fun getUsersFromTeam(teamId: TeamId): Either<StorageFailure, List<OtherUser>>
     suspend fun getTeamRecipients(teamId: TeamId): Either<CoreFailure, List<Recipient>>
+    suspend fun observeUpdateFirebaseTokenFlag(): Flow<Boolean>
+    suspend fun setUpdateFirebaseTokenFlag(shouldUpdate: Boolean): Either<StorageFailure, Unit>
```
vitorhugods, line 66: This Repo already has too many responsibilities.
We could create a `PushTokenRepository` that has these two functions there.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/notificationToken/ObserveNotificationTokenUpdating.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.wire.kalium.logic.data.user.UserRepository
+import com.wire.kalium.logic.functional.flatMap
+import com.wire.kalium.logic.functional.isRight
+import com.wire.kalium.network.api.user.pushToken.PushTokenBody
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.map
+
+internal interface ObserveNotificationTokenUpdating {
```
vitorhugods, line 14: As this is not a UseCase, we can name it like a normal type and not like a function.

```suggestion
internal interface PushTokenUpdater
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/notificationToken/ObserveNotificationTokenUpdating.kt
```diff
@@ -0,0 +1,46 @@
...
+    suspend operator fun invoke()
```
vitorhugods, line 15: Perhaps

```suggestion
    suspend fun monitorTokenChanges()
```

This way, the utilisation becomes:

```kotlin
launch {
    val pushTokenUpdater = createPushTokenUpdater()
    pushTokenUpdater.monitorTokenChanges()
}
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/notificationToken/ObserveNotificationTokenUpdating.kt
```diff
@@ -0,0 +1,46 @@
...
+                                senderId = notificationToken.senderId,
+                                client = clientId.value,
+                                token = notificationToken.token,
+                                transport = notificationToken.transport
+                            )
+                        )
+                    }
+            }
+            .collect { result ->
+                if (result.isRight()) userRepository.setUpdateFirebaseTokenFlag(false)
```
vitorhugods, line 43: Really cool 😎

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -666,6 +675,8 @@ abstract class UserSessionScopeCommon internal constructor(
             slowSyncManager
 
             callRepository.updateOpenCallsToClosedStatus()
+
+            observeNotificationTokenUpdating()
         }
```
vitorhugods, line 680: Maybe we could launch this on its own job, so people don't forget that this operation is long suspending.

```suggestion
        }
        launch {
            observeNotificationTokenUpdating()
        }
```

vitorhugods (changes requested):
Really really cool 😎

Just a couple of tiny naming suggestions.

MohammedMokresh (approved):
LGTM

vitorhugods (approved):
🚀👩🏼‍🚀👨🏼‍🚀🧑🏼‍🚀

PR #956, opened 2022-09-22
chore: Add GetIncomingCallsOnceUseCase

borichellow:
# What's new in this PR?

Added `GetIncomingCallsOnceUseCase` to be able fetch calls only once and do not observe it. 
Can be useful for Push Notifications.


vitorhugods:
I'm not sure if this is needed anymore.

In any case, I'd suggest reverting the name to `GetIncomingCallsUseCase` and renaming the currently existing use-case to `ObserveIncomingCallsUseCase` so we keep a standard.

PR #963, opened 2022-09-23
fix: Send message after leaving the conversation (AR-2474)

borichellow:
# What's new in this PR?

### Issues

Leaving a conversation before a message is sent, results in the message never getting sent.

### Causes (Optional)

Sending the message is called in `ViewModelScope` which means that it will be canceled after leaving the screen, even if it wasn't finished.

### Solutions

Move sending messages into a more global scope, which lives longer than ViewModel. `UserSessionScopeCommon` extends `CoroutineScope` and can be used for that. 
Also moved running `SendTextMessageUseCase` running into `dispatchers.io` to not block UI thread.

# Note

can be merged at any time, doesn't break anything in Reloaded app


ohassine (approved):
Good catch 👍

yamilmedina (approved):
🚀 🎸

gongracr:
Would this also work for asset messages and its upload process??

tmspzz (approved):

PR #968, opened 2022-09-23
fix: doubled you left system message [AR-2422]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When a user leaves a group conversation and kills the app and opens it again, the message that he left is displayed twice.

### Causes (Optional)

When we execute the "member leave" request, we don't get the exact event in response thus we don't have the id. Since we left, server stops sending us events via socket so we have to generate this one ourselves and add to the local db, but when we kill the app and it syncs after opening again, this event is received from the REST request with a proper event id and added again to the local db.

### Solutions

Generate and save the event after the "member leave" request with time we receive in the response, and when we get "member leave" event as event via socket, check if such event was already added comparing the data that we have (conversation id, sender id, exact time and members removed list) and if so then only update its id.

#### Test Coverage (Optional)

#### How to Test

Leave the conversation, kill the app, launch the app again and open this conversation. 

### Attachments (Optional)

https://user-images.githubusercontent.com/30429749/191985100-e5629efe-4376-4c84-949f-de19b1c1a310.mp4

----
##### References


persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
```diff
@@ -68,100 +68,130 @@ class MessageDAOImpl(private val queries: MessagesQueries, private val conversat
...
-        when (val content = message.content) {
-            is MessageEntityContent.Text -> {
-                queries.insertMessageTextContent(
-                    message_id = message.id,
-                    conversation_id = message.conversationId,
-                    text_body = content.messageBody
-                )
-                content.mentions.forEach {
-                    queries.insertMessageMention(
+        if (!updateIdIfAlreadyExists(message)) {
```
MohamadJaara, line 71: Just to be a bit more efficient, we need to check if it already exists only if the sender is self

MohamadJaara (commented):
I'm not really sure if it's clients responsibly to deal with such weird edge case, in perfect world clients should receive this event only from server (receive it complete with event ID)

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
saleniuk: Sure, added it to the code 👍

saleniuk:
> I'm not really sure if it's clients responsibly to deal with such weird edge case, in perfect world clients should receive this event only from server (receive it complete with event ID)

I know and I'm also not happy about it, that would be great to rely only on what backend gives us, but we don't live in a perfect world and we have to adjust to how the backend works, so in order to provide the best possible UX we already handle it this way for some time.

MohamadJaara (approved):
Approved, but we should open a ticket for the BR team to solve it from their side

typfel:
I'm not saying should be like this but the BE works like this.

When you perform a request like removing a user from group. The response to this request is a conversation.member-left event which the client is expected to process. The event will not arrive on the websocket so it will only be processed once. However it will present in the notification stream because it must include all events. So if you haven't processed any event more events afterwards when you do an incremental sync the event will show up again.

PR #970, opened 2022-09-24
fix(performance): add conversations View to DB and improve connections requests (AR-2515)

mchenani:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?
Due to the DB performance issues, we found some places where we were querying the database nested in other queries.

Injected the selfUserId instead of querying it every time
Injected the teamId instead of querying it every time
When we fetch the connection, we need the other user info as well; we were fetching those data by querying the user info by the userId that exists in the connection; replace it with a join query, and now we return the userInfo with the connection data directly from the database.

We had many flows + queries to fetch the conversationsList! [Combine of One Observer for Connections + One for Calls + One for Conversations + one for each conversationsDetails]
To fix all of these db-connections we created a View to present a single Presenter of the ConversationsList

### Issues

Issues
Queries nested in other queries in the DAO layer instead of joins on the db level
Queries nested in other queries in the DAO layer instead of injecting the cached values like the selfUserId

Needs releases with:

- [x] GitHub link to other pull request
https://github.com/wireapp/wire-android-reloaded/pull/1022

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -96,6 +109,152 @@ internal class ConversationMapperImpl(
...
+                        name = name,
+                        accentId = 0,
+                        userType = domainUserTypeMapper.fromUserTypeEntity(userType),
+                        availabilityStatus = userAvailabilityStatusMapper.fromDaoAvailabilityStatusToModel(userAvailabilityStatus),
+                        deleted = type != ConversationEntity.Type.GROUP && otherUserId != null,
+                        botService = botService?.let { BotService(it.id, it.provider) },
+                        handle = null,
+                        completePicture = previewAssetId?.let { idMapper.fromDaoModel(it) },
+                        previewPicture = previewAssetId?.let { idMapper.fromDaoModel(it) },
+                        teamId = teamId?.let { TeamId(it) }
```
typfel, line 168: Isn't this the teamId of the self user? OtherUser could belong to a different team.

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
```diff
@@ -76,6 +76,96 @@ UPDATE Conversation
...
+    WHEN "ONE_ON_ONE" THEN  User.name
+    WHEN "CONNECTION_PENDING" THEN connection_user.name
+    ELSE Conversation.name
+    END AS name,
+Conversation.type,
+Call.status AS callStatus,
+CASE (Conversation.type)
+    WHEN "ONE_ON_ONE" THEN  User.preview_asset_id
+    WHEN "CONNECTION_PENDING" THEN connection_user.preview_asset_id
+    END AS previewAssetId,
```
typfel, line 92: I'd suggest pulling in the entire connection.user instead of extracting most fields individually. This would also simply the mapping later and make it less error prone since you can re-use the user mapper.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/ObserveConversationViewUseCaseImpl.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import com.wire.kalium.logic.data.conversation.ConversationDetails
+import com.wire.kalium.logic.data.conversation.ConversationRepository
+import kotlinx.coroutines.flow.Flow
+
+fun interface ObserveConversationViewUseCase {
+    suspend operator fun invoke(): Flow<List<ConversationDetails>>
+}
+
+internal class ObserveConversationViewUseCaseImpl(
```
typfel, line 11: I prefer to keep the old name ObserveConversationListDetailsUseCase. I don't think we should expose the "View" name at all.

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
mchenani: Good question, when I put it here, I always had in mind all the data here are related to the other user, only in the self conversation the second party is self!

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
mchenani: But then most of the fields are redundant.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/ObserveConversationViewUseCaseImpl.kt
mchenani: Yes, I missed that! thanks

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
typfel: You pull in almost all fields. I don't think the overhead is that big.

mchenani: it's not about the data, because we have all the data we need, what I'm doing there instead of having connection userId, myUserId, otherUserid -> always return userId, and based on the type of the conversation put the correct value in that!

typfel: And I say. Pull connection.user as connectedUser if type is oneToOne and deal with populating the right fields in the mapper. Less logic in the query.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/type/UserTypeMapper.kt
```diff
@@ -54,6 +54,7 @@ class DomainUserTypeMapperImpl : DomainUserTypeMapper {
             UserTypeEntity.OWNER -> owner
             UserTypeEntity.ADMIN -> admin
             UserTypeEntity.SERVICE -> service
+            else -> none
```
gongracr, line 57: I'd rather be exhaustive here and do `UserTypeEntity.NONE -> none` instead as it might be that we add new user type entities in the future that will be silently swallowed here, right?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/AvailabilityStatusMapper.kt
```diff
@@ -4,19 +4,20 @@ import com.wire.kalium.persistence.dao.UserAvailabilityStatusEntity
...
 
 internal class AvailabilityStatusMapperImpl : AvailabilityStatusMapper {
-    override fun fromDaoAvailabilityStatusToModel(status: UserAvailabilityStatusEntity): UserAvailabilityStatus =
+    override fun fromDaoAvailabilityStatusToModel(status: UserAvailabilityStatusEntity?): UserAvailabilityStatus =
         when (status) {
             UserAvailabilityStatusEntity.AVAILABLE -> UserAvailabilityStatus.AVAILABLE
             UserAvailabilityStatusEntity.BUSY -> UserAvailabilityStatus.BUSY
             UserAvailabilityStatusEntity.AWAY -> UserAvailabilityStatus.AWAY
             UserAvailabilityStatusEntity.NONE -> UserAvailabilityStatus.NONE
+            else -> UserAvailabilityStatus.NONE
```
gongracr, line 20: maybe it would be better to map both to null:
`UserAvailabilityStatusEntity.NONE, null` -> UserAvailabilityStatus.NONE

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -96,6 +109,152 @@ internal class ConversationMapperImpl(
...
+                ConversationDetails.Self(
+                    Conversation(
+                        id = idMapper.fromDaoModel(id),
+                        name = name,
+                        type = type.fromDaoModelToType(),
+                        teamId = teamId?.let { TeamId(it) },
+                        protocol = protocolInfoMapper.fromEntity(protocolInfo),
+                        mutedStatus = conversationStatusMapper.fromMutedStatusDaoModel(mutedStatus),
+                        removedBy = null, // todo: ask how to calculate?
+                        creatorId = PlainId(""), // deprecated
```
gongracr, line 126: If `creatorId` is marked as deprecated, why don't we just delete it instead of keep using it?

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAO.kt
```diff
@@ -63,6 +65,32 @@ data class ConversationEntity(
     }
 }
 
+@Suppress("FunctionParameterNaming")
+data class ConversationViewEntity(
+    val id: QualifiedIDEntity,
+    val name: String?,
```
gongracr, line 71: Should ever a conversation have an empty name?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -96,6 +109,152 @@ internal class ConversationMapperImpl(
...
+                        deleted = type != ConversationEntity.Type.GROUP && otherUserId != null,
+                        botService = botService?.let { BotService(it.id, it.provider) },
+                        handle = null,
+                        completePicture = previewAssetId?.let { idMapper.fromDaoModel(it) },
+                        previewPicture = previewAssetId?.let { idMapper.fromDaoModel(it) },
+                        teamId = teamId?.let { TeamId(it) }
+                    ),
+                    connectionState = connectionStatusMapper.fromDaoModel(connectionStatus),
+                    legalHoldStatus = LegalHoldStatus.DISABLED,
+                    userType = domainUserTypeMapper.fromUserTypeEntity(userType),
```
gongracr, line 172: Would this be the self user `userType` or the other user `userType` ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/LeaveConversationUseCase.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import com.wire.kalium.logic.data.conversation.ConversationRepository
+import com.wire.kalium.logic.data.id.ConversationId
+import com.wire.kalium.logic.data.message.PersistMessageUseCase
+import com.wire.kalium.logic.data.user.UserId
+
+interface LeaveConversationUseCase {
+
+    /**
+     * This use case will allow to leave a given group conversation while still keeping the mentioned conversation in
```
gongracr, line 11:
```suggestion
     * This use case will allow the self user to leave a given group conversation while still keeping the mentioned conversation in
```

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAO.kt
vitorhugods: 1:1 conversations have no name, actually.

From my side I see your name, from your side you see mine, for example. We should always use the `OtherUser.name` as the conversation name.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
mchenani: it will affect everywhere! I'll do it in next PR

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
tmspzz: @mchenani @typfel we will address this in another PR. For the sake of not holding other code off we'll merge as is for now.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
mchenani: it's the other thing! like federated and...

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/ConversationDAO.kt
gongracr: yes, but this field is the local entity field specific to each user's setup, so there is nothing stopping us to declare this `name` field as not nullable and map it with the appropriate name

tmspzz (approved):

PR #958, opened 2022-09-23
fix: update IncrementalSyncState state to pending after executing while policy allows (AR-2437)

vitorhugods:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [X] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

After fetching pending events, and the `ConnectionPolicy` is `DISCONNECT_AFTER_PENDING_EVENTS`, `SyncState` stays on `LIVE`.

This is unintuitive and is a problem on Reloaded, because Reloaded does:
```kotlin
upgradePolicy()
waitUntilLive()
downgradePolicy()
```

However, takes a few CPU cycles until `IncrementalSyncManager` collects the change in `ConnectionPolicy`, and during this short time the `IncrementalSyncStatus` is still `LIVE`.

Which means that `waitUntilLive` returns immediatelly.

Reloaded would need to put an arbitrary delay between `upgradePolicy()` and `waitUntilLive()` in order to properly wait until Sync finishes.

### Solutions

Make sure that **every time** that incremental sync stops, the state goes back to `Pending`.
This way, when performing `waitUntilLive`, the state won't be falsely reported as `LIVE` and return immediatelly.

#### Test Coverage

- [X] I have added automated test to this contribution

----
#### PR Post Merge Checklist for internal contributors

 - [X] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #984, opened 2022-09-27
fix: wait until live hanging endlessly (AR-2524) (AR-2533)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Some clients are calling `waitUntilLive` and are hanging forever even after Sync is `LIVE`.

### Causes

In #958 we changed IncrementalSyncState to always go back to `PENDING` as soon as it finishes, due to `ConnectionPolicy` downgrade for example.

This means that `incrementalSyncState` transitions from `Pending Events` -> `Live` -> `Pending` really fast.

`waitUntilLive` collects the values of `incrementalSyncState` until it sees `Live`. 
Because the emissions are fast and the flow isn't buffered, the `Live` state can be missed. So it looks like this:

Emitted values:
`Pending Events` -> `Live` -> `Pending`

Collector observes:
`Pending Events` -> `Pending`

And keeps waiting for a `Live` signal.

### Solutions

Drop the usage of `StateFlow`.
StateFlow is a specialised `SharedFlow` that doesn't allow repetition of values, nor buffering, nor a lot of things.
If we need to make sure that we collect all the emitted events, `StateFlow` is not the way to go.

Replace it with a buffered `SharedFlow` with replay value of `1`.
This keeps the same behaviour from `StateFlow`, that new collectors have access to the last emitted value when starting collection.
Also, being it buffered, with `SUSPEND` buffered overflow policy, we make sure that collectors won't miss events.

#### Test Coverage

New tests to make sure that even very slow collectors don't  miss emissions of `ConnectionPolicy` changes nor `IncrementalSyncStatus` changes.

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
@@ -5,6 +5,7 @@ import com.wire.kalium.logic.CoreFailure
 import com.wire.kalium.logic.NetworkFailure
 import com.wire.kalium.logic.StorageFailure
 import com.wire.kalium.logic.data.conversation.ConversationDetails
+import com.wire.kalium.logic.data.conversation.ConversationRepository
 import com.wire.kalium.logic.data.event.Event
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.id.IdMapper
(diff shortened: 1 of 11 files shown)


saleniuk (approved):
👌 
Maybe there should also be a test to make sure that it doesn't emit repeated values, as stated in the comment 😉

vitorhugods:
> 👌 Maybe there should also be a test to make sure that it doesn't emit repeated values, as stated in the comment 😉

Done :)

typfel (approved):

vitorhugods:
> Oh interesting that you can just miss state updates with `StateFlow`. Did you reproduce this issue with the unit tests?

Yes: https://github.com/wireapp/kalium/blob/2d53e5ede22d11d0937c539a2e72bbba1669c48d/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/sync/IncrementalSyncRepositoryTest.kt#L50

This test was failing before the changes.

I also observed this behaviour using logs inside `waitUntilLive`. Not 100% reproduction rate, flaky as it's probably reliant on dispatcher availability at the moment.

Also, from [StateFlow's documentation](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-state-flow/):
> State flow always has an initial value, replays one most recent value to new subscribers, does not buffer any more values
> Application of flowOn, conflate, buffer with CONFLATED or RENDEZVOUS capacity, distinctUntilChanged, or cancellable operators to a state flow has no effect.

I had the feeling you'd find it interesting so I added you as a reviewer 😅

PR #998, opened 2022-09-29
fix: Connection status filed duplication

borichellow:
# What's new in this PR?

### Issues

data class `com.wire.kalium.logic.data.conversation.ConversationDetails.OneOne` contains fileds `ConnectionState`  and `OtherUser` which also contains `ConnectionState` (basically state in OtherUser is the original one).

This duplication combined with a wrong mapping in some mappers causes problems and wrong ConnectionStatus displaying in UI

### Solutions

Remove duplication and use `com.wire.kalium.logic.data.user.OtherUser.connectionStatus` as the only source of truth.


gongracr (approved):
lgtm 🥝

PR #1004, opened 2022-10-04
feat: allow sending and receiving of multiple reactions

tmspzz:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

Currently Wire only supports one, predefined visible reaction. We would like to expand support to multiple reactions.

### Solutions

With this PR we introduce sending, storing and receiving a comma separated list of reactions. Each substring split at comma is considered a reaction. Therefore:

- `"👍,🤯,😅"` contains 3 reactions
- `"hello,ciao,goodbye"` also contains 3 reactions

----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1008, opened 2022-10-04
chore: Replace isCreator by creatorId in ConversationViewEntity

borichellow:
# What's new in this PR?

- in `ConversationViewEntity` renamed field `isSelfCreated` into `isCreator` and fixed it (previously it was calculated incorrectly). 
- in `ConversationViewEntity` fixed the calculation of `isMember` field
- in `ConversationViewEntity` fixed `JOIN`
- added all the fields from `Conversation` BD table into `ConversationDetails` DB view, so now we can use only `ConversationViewEntity` model for all `SELECT` queries. This makes more sense: only 1 source of truth, fewer mappers, and more info in model.
- updated Mappers and DB-tests accordingly 
- added DB migration, so no need to clear the app data after updating.

## Need to release with [AR -1068 PR](https://github.com/wireapp/wire-android-reloaded/pull/1068)


MohamadJaara:
the selfUser id is stored ion the metadata dao which is in the DB, and i remember @vitorhugods working on an quick way to get selfId from the DB

vitorhugods:
> the selfUser id is stored ion the metadata dao which is in the DB, and i remember @vitorhugods working on an quick way to get selfId from the DB

Indeed. It's together in #1004.

#1004 is mostly reviewed, just missing some tests.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
```diff
@@ -142,8 +141,7 @@ sealed class ConversationDetails(open val conversation: Conversation) {
         val unreadMessagesCount: Long = 0L,
         val unreadMentionsCount: Long = 0L,
         val lastUnreadMessage: Message?,
-        val isSelfUserMember: Boolean = true,
-        val isSelfCreated: Boolean = false
+        val isSelfUserMember: Boolean = true
```
gongracr, line 144: Would this be really needed if we already have access to the same field within the `conversation` object?

persistence/src/commonMain/db_user/com/wire/kalium/persistence/1.sqm
```diff
@@ -0,0 +1,92 @@
+DROP VIEW ConversatonDetails;
```
gongracr, line 1: Should we name this `1.sqm` or something more descriptive?

borichellow: `1.sqm` means "migration from verstion 1 to version 2", so it has to be `1` :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
borichellow: that's right, this is why i removed that field here :)

gongracr: But `isSelfUserMember` is still part of the `Conversation` data class in line 30 🤷

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepositoryTest.kt
```diff
@@ -437,6 +438,11 @@ class ConversationRepositoryTest {
             .whenInvokedWith(anything(), anything())
             .thenDoNothing()
 
+        given(conversationDAO)
+            .suspendFunction(conversationDAO::getConversationByQualifiedID)
+            .whenInvokedWith(anything())
+            .then { TestConversation.VIEW_ENTITY }
+
```
gongracr, line 445: I might sound like a scratched vinyl, but if you have time, adapting this test to the Arrangement pattern will save us a lot of boilerplate code and make it easier to add new one :)

borichellow: i was afraid you ask for that XD

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
borichellow: after your question i noticed that both `isSelfUserMember` and `isSelfCreated ` should be not in `Conversation` but in `ConversationDetails.OneOne` and `ConversationDetails.Group` as these fields are related only to these data classes. So I moved them :)

gongracr (approved):
Thanks for the changes 🙇 And glad that my comments could help you identify some other potential bugs 😄  Great job!

mchenani (approved):
Well done buddy 🚀 fixed many issues and cleaned up the way more than before 🥁

PR #1025, opened 2022-10-07
chore: sendMessage with recipients parameter

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-2571 - Targeted messages not implemented](https://wearezeta.atlassian.net/browse/AR-2571)

This is only `PART 1`, another part will follow where this changes are properly used.

### Issues

We were always send OTR messages to all recipients in a conversation.

### Causes (Optional)

There was no parameter for receiving target recipients.

### Solutions

Add target recipients as parameter.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
@@ -8,6 +8,7 @@ import com.wire.kalium.logic.data.conversation.Conversation
 import com.wire.kalium.logic.data.conversation.ConversationOptions
 import com.wire.kalium.logic.data.conversation.ConversationRepository
 import com.wire.kalium.logic.data.conversation.MLSConversationRepository
+import com.wire.kalium.logic.data.conversation.Recipient
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.id.GroupID
 import com.wire.kalium.logic.data.message.Message
@@ -67,7 +68,7 @@ interface MessageSender {
      * @param message that will be sent
      * @see [sendPendingMessage]
      */
-    suspend fun sendMessage(message: Message.Regular): Either<CoreFailure, Unit>
+    suspend fun sendMessage(message: Message.Regular, targetRecipients: List<Recipient>? = null): Either<CoreFailure, Unit>
 
     /**
      * Attempts to send the given Client Discovery [Message] to suitable recipients.
@@ -110,8 +111,8 @@ internal class MessageSenderImpl internal constructor(
         }
     }
 
-    override suspend fun sendMessage(message: Message.Regular): Either<CoreFailure, Unit> =
-        attemptToSend(message).map { messageRemoteTime ->
+    override suspend fun sendMessage(message: Message.Regular, targetRecipients: List<Recipient>?): Either<CoreFailure, Unit> =
+        attemptToSend(message, targetRecipients).map { messageRemoteTime ->
             updateDatesOfMessagesWithServerTime(message, messageRemoteTime)
         }
 
@@ -136,7 +137,7 @@ internal class MessageSenderImpl internal constructor(
 
     override suspend fun sendClientDiscoveryMessage(message: Message.Regular): Either<CoreFailure, String> = attemptToSend(message)
 
-    private suspend fun attemptToSend(message: Message.Regular): Either<CoreFailure, String> {
+    private suspend fun attemptToSend(message: Message.Regular, targetRecipients: List<Recipient>? = null): Either<CoreFailure, String> {
         return conversationRepository.getConversationProtocolInfo(message.conversationId).flatMap { protocolInfo ->
             when (protocolInfo) {
                 is Conversation.ProtocolInfo.MLS -> {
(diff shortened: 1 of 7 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -145,15 +146,23 @@ internal class MessageSenderImpl internal constructor(
...
                 }
             }
         }
-}
+    }
 
-    private suspend fun attemptToSendWithProteus(message: Message.Regular): Either<CoreFailure, String> {
+    private suspend fun attemptToSendWithProteus(
+        message: Message.Regular,
+        targetRecipients: List<Recipient>?
```
typfel, line 157: In this case where you pass in a subset of the possible recipients we also needs modify the `trySendingProteusEnvelope`. Currently it always send the message with the `MessageApi.QualifiedMessageOption.ReportAll` option but when a partial list is sent we need to use one of the following options:

`IgnoreAll` - If the message is sent to a single client
`ReportSome` = if the message is sent to a single user

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -145,15 +146,23 @@ internal class MessageSenderImpl internal constructor(
...
+    ): Either<CoreFailure, String> {
         val conversationId = message.conversationId
-        return conversationRepository.getConversationRecipients(conversationId).flatMap { recipients ->
+
+        return targetRecipients?.let {
+            messageEnvelopeCreator.createOutgoingEnvelope(it, message).flatMap { envelope ->
+                trySendingProteusEnvelope(envelope, message)
```
typfel, line 163: No need to call `sessionEstablisher.prepareRecipientsForNewOutgoingMessage(recipients)`?

typfel (changes requested):

vitorhugods (commented):
Pretty much what @typfel commented + it would be pretty nice to have tests where we check if:
- `prepareRecipientsForNewOutgoingMessage` is called with the right arguments when `recipients` are provided
- `getConversationRecipients` is not called when `recipients` are provided
- The envelope passed to the repo contains entries for the right recipients and the correct `Report`/`Ignore` type when `recipients` are provided.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
typfel: Might make sense to create an enum to make intention clear instead of trying to compute it.

```
sealed class MessageTarget {
  class User(val userId: UserId): MessageTarget
  class Client(val clientId: ClientId): MessageTarget
  class Conversation: MesageTarget
}
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
alexandreferris: We actually do!

typfel: Can you explain why because I don't see it.

typfel:
@alexandreferris are still working on this? I would expect a test that we are mapping to the right `QualifiedMessageOption` depending where we target the message.

typfel (approved):

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/message/MessageRepositoryTest.kt
```diff
@@ -153,6 +157,74 @@ class MessageRepositoryTest {
...
+
+        val expectedParameters = MessageApi.Parameters.QualifiedDefaultParameters(
+            sender = messageEnvelope.senderClientId.value,
+            recipients = mapOf(),
+            true,
+            MessagePriority.HIGH,
+            false,
+            messageEnvelope.dataBlob?.data,
+            MessageApi.QualifiedMessageOption.IgnoreAll
+        )
```
vitorhugods, line 220: Should we really assert things like Message Priority, data blob, etc.? These are independent from the message targets.
When we change the message priority logic, this test will fail without any reason.

I suggest to, instead, use a `matching {}` lambda to assert only the parameters we actually need:


```
        verify(arrangement.messageApi)
            .suspendFunction(arrangement.messageApi::qualifiedSendMessage)
            .with(matching {
            it.recipients.isEmpty() && it.options = MessageApi.QualifiedMessageOption.IgnoreAll
         }, anything())
```

Same thing for the test above.

vitorhugods (changes requested):
Great work.

Just some suggestions to prevent the tests breaking in the future unnecessarily and making it easier to understand what's being tested.

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/message/MessageRepositoryTest.kt
alexandreferris: Aha! Nice, I was not aware that you could work with it like that. Thanks for the heads up 🙏🏻

vitorhugods (approved):
🚀🚀🚀🚀🚀🚀

PR #1028, opened 2022-10-07
fix: implement ProteusClientProvider [AR-2577]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

"Message could not be decrypted" appearing after some time.

### Causes

This seems to be caused because we are using somehow old stored Prekeys on the AR client, that might not have been totally removed when removing the client when logging in again and creating a new session from same device.
One of the direct causes is that maybe we get error when removing Proteus client local files or probably because even if we remove the files and open the client again, we still have multiple references to the old one.

### Solutions

Provide ProteusClient via the provider, this way we can be sure that if we remove local files, the client is also nulled out and client is provided lazily so next time we need it, the new instance is created and only the new one is used. 

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/ProteusClientProvider.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.wire.kalium.cryptography.ProteusClient
+import com.wire.kalium.cryptography.ProteusClientImpl
+import com.wire.kalium.cryptography.exceptions.ProteusException
+
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlin.coroutines.cancellation.CancellationException
+
+interface ProteusClientProvider {
+    suspend fun clear()
```
typfel, line 12: Why not also use the name `clearLocalFiles()` here just `clear` is a bit ambiguous about what it does.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/ProteusClientProvider.kt
```diff
@@ -0,0 +1,38 @@
...
+    suspend fun getOrCreate(): ProteusClient
```
MohamadJaara, line 13: To avoid creating new files unintentionally, we can add another function
```kotlin
    suspend fun getOrError(): Either<CoreFailure, ProteusClient> {
          // check if the files exists or return MissingClientRegistration
}
```
because the creating new files should only happen during register client

typfel: @saleniuk you added `getOrError()` but aren't using it. As I understood @MohamadJaara you use it everywhere and only use `getOrCreate()` when registering a new client.

saleniuk: @typfel sure, my bad, now it's correct

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
```diff
@@ -33,7 +33,7 @@ class LogoutUseCaseImpl @Suppress("LongParameterList") constructor(
         deregisterTokenUseCase()
         logoutRepository.logout()
         sessionRepository.logout(userId = userId, reason)
-        if (reason == LogoutReason.SELF_HARD_LOGOUT) {
+        if (reason == LogoutReason.SELF_HARD_LOGOUT || reason == LogoutReason.REMOVED_CLIENT || reason == LogoutReason.DELETED_ACCOUNT) {
```
saleniuk, line 36: @typfel @MohamadJaara correct me if I'm wrong but I think that we should also remove client and user data when the client has been removed or account has been deleted, not only when there is a hard logout 🤔

MohamadJaara: user is only removed on hard logout
when REMOVED_CLIENT or DELETED_ACCOUNT only client data need to be removed 
on SESSION_EXPIRED or SOFT_LOGOUT no data is removed

saleniuk: Ok, thanks, so I changed it to this:
```
        if (reason == LogoutReason.REMOVED_CLIENT || reason == LogoutReason.DELETED_ACCOUNT || reason == LogoutReason.SELF_HARD_LOGOUT) {
            // we put this delay here to avoid a race condition when receiving web socket events at the exact time of logging put
            delay(CLEAR_DATA_DELAY)
            clearClientDataUseCase()
        }
        if (reason == LogoutReason.SELF_HARD_LOGOUT) {
            clearUserDataUseCase() // this clears also current client id
        } else {
            clientRepository.clearCurrentClientId()
        }
```
It should be ok now, right?

saleniuk: Ok, thanks @MohamadJaara , so I changed it to this:
```
        if (reason == LogoutReason.REMOVED_CLIENT || reason == LogoutReason.DELETED_ACCOUNT || reason == LogoutReason.SELF_HARD_LOGOUT) {
            // we put this delay here to avoid a race condition when receiving web socket events at the exact time of logging put
            delay(CLEAR_DATA_DELAY)
            clearClientDataUseCase()
        }
        if (reason == LogoutReason.SELF_HARD_LOGOUT) {
            clearUserDataUseCase() // this clears also current client id
        } else {
            clientRepository.clearCurrentClientId()
        }
```
It should be ok now, right?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
```diff
@@ -33,10 +33,13 @@ class LogoutUseCaseImpl @Suppress("LongParameterList") constructor(
...
+
+        if (reason == LogoutReason.REMOVED_CLIENT || reason == LogoutReason.DELETED_ACCOUNT || reason == LogoutReason.SELF_HARD_LOGOUT) {
             // we put this delay here to avoid a race condition when receiving web socket events at the exact time of logging put
             delay(CLEAR_DATA_DELAY)
             clearClientDataUseCase()
+        }
+        if (reason == LogoutReason.SELF_HARD_LOGOUT) {
             clearUserDataUseCase() // this clears also current client id
         } else {
             clientRepository.clearCurrentClientId()
```
MohamadJaara, line 45: maybe we can replace this with when(logoutReason ) just to make it clearer

MohamadJaara (approved):
LGTM 🚀, just small suggestion to make the code in LogoutUsecase clearer

PR #1030, opened 2022-10-10
feat: store last generated prekey

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

kalium does not store the id of the last generated prekey

### Solutions

store it

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


typfel:
Why do we need store the ID of the last generated pre key?

MohamadJaara:
> Why do we need store the ID of the last generated pre key?

the scala app is doing so and during the migration i would like to migrate this value too, because the new app will be using the same crypto box files as the old one.

typfel:
> > Why do we need store the ID of the last generated pre key?
> 
> the scala app is doing so and during the migration i would like to migrate this value too, because the new app will be using the same crypto box files as the old one.

Huh interesting but where would you actually use it. Is it some how passed as an argument to crypto box?

MohamadJaara:
> > > Why do we need store the ID of the last generated pre key?
> > 
> > 
> > the scala app is doing so and during the migration i would like to migrate this value too, because the new app will be using the same crypto box files as the old one.
> 
> Huh interesting but where would you actually use it. Is it some how passed as an argument to crypto box?

When generating new prekeys for the backend (we don't do this, atm), 
`ProteusClient.newPrekeys(from: Int, count: Int)`  I'm assuming that the parameter from is `otrLastPrekeyId + 1`

MohammedMokresh (approved):
LGTM

PR #1034, opened 2022-10-11
chore: calling target messages logic

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-2571 - Targeted messages not implemented](https://wearezeta.atlassian.net/browse/AR-2571)

This is Part 2 final which contains the logic that parses the target recipients received from AVS to be send as a target message.

### Issues

We were always send OTR messages to all recipients in a conversation.

### Causes (Optional)

There was no parameter for receiving target recipients.

### Solutions

Receive target recipients, parse it and send as parameter to `sendMessage`.

### Dependencies
Can only be merged after:

- [X] #1025

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/mapper/CallMapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/mapper/CallMapper.kt
@@ -3,12 +3,16 @@ package com.wire.kalium.logic.data.call.mapper
 import com.wire.kalium.calling.CallTypeCalling
 import com.wire.kalium.calling.ConversationTypeCalling
 import com.wire.kalium.calling.VideoStateCalling
+import com.wire.kalium.logic.data.call.CallClientList
 import com.wire.kalium.logic.data.call.CallMetadata
 import com.wire.kalium.logic.data.call.CallType
 import com.wire.kalium.logic.data.call.ConversationType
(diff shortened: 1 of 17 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/mapper/CallMapper.kt
```diff
@@ -158,4 +167,33 @@ class CallMapperImpl : CallMapper {
...
+                    )
+                )
+            } else {
+                recipientsList[recipientIndex] = recipientsList[recipientIndex].copy(
+                    clients = recipientsList[recipientIndex].clients.toMutableList().apply {
+                        add(clientId)
+                    }
+                )
+            }
+        }
```
ohassine, line 193: what about using `groupingBy` to make this mapping ?
something like this: 

```
val recipients = callClientList.clients.groupingBy { it.userId }
            .fold({ key, _ -> key to mutableListOf<ClientId>() }, { _, accumulator, element ->
                accumulator.also { (_, list) -> list.add(ClientId(element.clientId)) }
            }).run {
                values.toList().map {
                    val qualifiedUserId = qualifiedIdMapper.fromStringToQualifiedID(it.first)
                    Recipient(qualifiedUserId, it.second)
                }
            }
```

ohassine (approved):
lgtm 👌

PR #1037, opened 2022-10-12
feat: proxy support [AR-2280]

MohammedMokresh:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

As a User want to add another level of verification to the login process by using a Proxy.

### Solutions

implement proxy support with basic auth in our network layer

Needs releases with:

#### How to Test

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/autoVersioningAuth/AutoVersionAuthScopeUseCase.kt
```diff
@@ -11,11 +11,11 @@ class AutoVersionAuthScopeUseCase(
     private val serverLinks: ServerConfig.Links,
     private val coreLogic: CoreLogicCommon
 ) {
-    suspend operator fun invoke(): Result =
+    suspend operator fun invoke(proxyCredentials: (() -> Pair<String, String>)?): Result =
```
vitorhugods, line 14: I like the fact that this doesn't have a default argument.

This way you the caller needs to explicitly opt out of providing credentials.

Maybe it would be cleaner / more explicit if we had a:

```kotlin

sealed interface ProxyCredentials {

    object None: ProxyCredentials

    class UsernameAndPassword(val username: String, val password: String): ProxyCredentials
}

```

This way the caller needs to pass `ProxyCredentials.None` or `ProxyCredentials.UsernameAndPassword(username = "user@example.org", password = "123456")`

Reducing a bit the risk of calling `Pair(password, email)` instead of `Pair(email, password)` by accident.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/auth/login/ProxyCredentialsModel.kt
```diff
@@ -0,0 +1,3 @@
+package com.wire.kalium.logic.data.auth.login
+
+data class ProxyCredentialsModel(val username: String?, val password: String?)
```
typfel, line 3: How about deleting this class and move the properties to ProxyCredentials.UsernameAndPassword

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/daokaliumdb/ServerConfigurationDAO.kt
```diff
@@ -23,6 +23,9 @@ internal object ServerConfigMapper {
         domain: String?,
         commonApiVersion: Int,
         federation: Boolean,
+        apiProxy: String?,
+        needsAuthentication: Boolean?,
+        port: Int?
```
typfel, line 28: Prefix these arguments with `proxy`

persistence/src/commonMain/db_global/com/wire/kalium/persistence/ServerConfiguration.sq
```diff
@@ -14,15 +14,18 @@ CREATE TABLE ServerConfiguration (
     domain TEXT UNIQUE,
     commonApiVersion INTEGER AS Int NOT NULL,
     federation INTEGER AS Boolean NOT NULL,
+    apiProxy TEXT,
+    needsAuthentication INTEGER AS Boolean,
+    port INTEGER AS Int,
```
typfel, line 19: prefix with proxy

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
```diff
@@ -1,15 +1,50 @@
...
 import okhttp3.OkHttpClient
+import java.net.Authenticator
+import java.net.InetSocketAddress
+import java.net.PasswordAuthentication
+import java.net.Proxy
 import java.util.concurrent.TimeUnit
 
-actual fun defaultHttpEngine(): HttpClientEngine = OkHttp.create {
+actual fun defaultHttpEngine(
+    serverConfigDTOLinks: ServerConfigDTO.Links?,
```
typfel, line 15: Why pass in `ServerConfigDTO.Links` when you always access `ServerConfigDTO.Links.Proxy`?

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/model/AuthSessionEntity.kt
```diff
@@ -13,14 +13,22 @@ data class ServerConfigEntity(
...
     data class MetaData(
         val federation: Boolean,
         val apiVersion: Int,
         val domain: String?
     )
+
+    data class Proxy(
+        val needsAuthentication: Boolean?,
+        val apiProxy: String?,
+        val port: Int?
```
typfel, line 30: Are any of these properties actually optional?

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreFailure.kt
```diff
@@ -82,10 +84,14 @@ internal inline fun <T : Any> wrapApiRequest(networkCall: () -> NetworkResponse<
         is NetworkResponse.Error -> {
             kaliumLogger.e(result.kException.stackTraceToString())
             val exception = result.kException
-            if (exception is KaliumException.GenericError && exception.cause is IOException) {
-                Either.Left(NetworkFailure.NoNetworkConnection(exception))
+            if (exception.cause?.message?.lowercase()?.contains("socks") == true) {
```
typfel, line 87: Why this loose: any message containing the string "socks" is a `SocketError`? Can't we rely on a more specific exception type?

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/model/AuthSessionEntity.kt
MohammedMokresh: Whenever there is proxy object no it should be not optional, so yeah I will update it to not be nullable

persistence/src/commonMain/db_global/com/wire/kalium/persistence/ServerConfiguration.sq
tmspzz: apiProxyNeedsAuthentication
apiProxyPort

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/daokaliumdb/ServerConfigurationDAO.kt
tmspzz: apiProxyNeedsAuthentication
apiProxyPort

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreFailure.kt
MohammedMokresh: the problem here is when there is an error from the proxy it throw SocketException which is Java.net exception 
and can't detect that from here , otherwise we have to have corefailure for each target, and I didn't think it's necessary for this case
If you have better approach please let me know

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/auth/login/ProxyCredentialsModel.kt
MohammedMokresh: I was passing pair and after that pass the user name and pass to the ProxyCredentials.UsernameAndPassword
but I saw this as more organized way to pass it to the mapper and the scope , better than pass pair or pass the username and password params

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/auth/login/ProxyCredentialsModel.kt
typfel: Then let's rename this to simply `ProxyCredentials` and the sealed class can be called something like `ProxyAuthentication`

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreFailure.kt
typfel: If the exception is platform specific I guess we should wrap it in our own exception. If you don't want do this now at least add a TODO.

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
```diff
@@ -1,15 +1,50 @@
...
+): HttpClientEngine = OkHttp.create {
     // OkHttp doesn't support configuring ping intervals dynamically,
     // so they must be set when creating the Engine
     // See https://youtrack.jetbrains.com/issue/KTOR-4752
-    val client = OkHttpClient.Builder().pingInterval(WEBSOCKET_PING_INTERVAL_MILLIS, TimeUnit.MILLISECONDS).build()
-    preconfigured = client
-    webSocketFactory = KaliumWebSocketFactory(client)
+    val isProxyRequired = serverConfigDTOProxy != null
+    if (isProxyRequired) {
+        if (isProxyRequired && serverConfigDTOProxy?.isProxyNeedsAuthentication == true) {
```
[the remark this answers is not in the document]
MohammedMokresh: yeah thanks for catching this, this if should only wrap the authenticator not the whole proxy setting

android/src/main/kotlin/com/wire/kalium/presentation/MainActivity.kt
```diff
@@ -82,7 +82,7 @@ class MainActivity : ComponentActivity() {
     }
 
     private suspend fun provideAuthScope(coreLogic: CoreLogic, backendLinks: ServerConfig.Links): AuthenticationScope =
-        when (val result = coreLogic.versionedAuthenticationScope(backendLinks).invoke()) {
+        when (val result = coreLogic.versionedAuthenticationScope(backendLinks).invoke(AutoVersionAuthScopeUseCase.ProxyCredentials.None)) {
```
mchenani, line 85: Can't we set this value as default value? so we don't need to always pass it.

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/unbound/configuration/ServerConfigResponse.kt
```diff
@@ -21,3 +22,10 @@ internal data class EndPoints(
     @SerialName("accountsURL") val accountsBaseUrl: String,
     @SerialName("websiteURL") val websiteUrl: String
 )
+
+@Serializable
+data class Proxy(
+    @SerialName("needsAuthentication") val isProxyNeedsAuthentication: Boolean,
+    @SerialName("apiProxy") val apiProxy: String,
+    @SerialName("port") val proxyPort: Int
```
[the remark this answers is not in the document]
MohammedMokresh: for the serial name shouldn't be the same as how we read it from the deeplink ?

cli/src/main/kotlin/com/wire/kalium/cli/commands/LoginCommand.kt
```diff
@@ -49,7 +49,9 @@ class LoginCommand : CliktCommand(name = "login") {
     }
 
     private suspend fun provideVersionedAuthenticationScope(serverLinks: ServerConfig.Links): AuthenticationScope =
-        when (val result = coreLogic.versionedAuthenticationScope(serverLinks).invoke()) {
+        when (val result = coreLogic.versionedAuthenticationScope(serverLinks).invoke(
```
mchenani, line 52: The previous comment will solve this as well.

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreFailure.kt
```diff
@@ -82,10 +84,14 @@ internal inline fun <T : Any> wrapApiRequest(networkCall: () -> NetworkResponse<
         is NetworkResponse.Error -> {
             kaliumLogger.e(result.kException.stackTraceToString())
             val exception = result.kException
-            if (exception is KaliumException.GenericError && exception.cause is IOException) {
-                Either.Left(NetworkFailure.NoNetworkConnection(exception))
+            if (exception.cause?.message?.contains("socks", true) == true) {
```
mchenani, line 87: Could you please extract the hard-coded text and also set the variables names? and also you can remove the `== true` section from the `if` like: `if (exception.cause?.message?.contains("socks", true))`

cli/src/main/kotlin/com/wire/kalium/cli/commands/LoginCommand.kt
MohammedMokresh: we can, but I don't know if it's good to default values in use cases !

logic/src/commonMain/kotlin/com/wire/kalium/logic/CoreLogic.kt
```diff
@@ -52,7 +53,7 @@ abstract class CoreLogicCommon internal constructor (
     inline fun <T> globalScope(action: GlobalKaliumScope.() -> T): T = getGlobalScope().action()
 
     inline fun <T> authenticationScope(serverConfig: ServerConfig, action: AuthenticationScope.() -> T): T =
-        getAuthenticationScope(serverConfig).action()
+        getAuthenticationScope(serverConfig, null).action()
```
mchenani, line 56: I think we can here as well set default value for that `null` passing.

android/src/main/kotlin/com/wire/kalium/presentation/MainActivity.kt
MohammedMokresh: we can, but I don't know if it's good to default values in use cases !

logic/src/commonMain/kotlin/com/wire/kalium/logic/GlobalKaliumScope.kt
```diff
@@ -133,4 +137,12 @@ class GlobalKaliumScope internal constructor(
     val serverConfigForAccounts: ServerConfigForAccountUseCase
         get() =
             ServerConfigForAccountUseCase(serverConfigRepository)
+
+    private val proxyCredentialsRepository: ProxyCredentialsRepository
+        get() =
+            ProxyCredentialsDataSource(globalPreferences.value.proxyCredentialsStorage)
+
+    val persistProxyCredentialsUseCase: PersistProxyCredentialsUseCase
```
mchenani, line 145: as far as I remember, that should be `internal` now, am I right?

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/server/ServerConfig.kt
```diff
@@ -160,7 +181,12 @@ class ServerConfigMapperImpl(
                 teams = links.teams,
                 website = links.website,
                 title = links.title,
-                isOnPremises = links.isOnPremises
+                isOnPremises = links.isOnPremises,
+                proxy = links.proxy?.let {
+                    ServerConfigDTO.Proxy(
+                        apiProxy = it.apiProxy, isProxyNeedsAuthentication = it.isProxyNeedsAuthentication, proxyPort = it.proxyPort
+                    )
+                }
```
mchenani, line 189:
```suggestion
                isOnPremises = links.isOnPremises,
                proxy = links.proxy?.let {
                    ServerConfigDTO.Proxy(
                        apiProxy = it.apiProxy,
                        isProxyNeedsAuthentication = it.isProxyNeedsAuthentication,
                        proxyPort = it.proxyPort
                    )
                }
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/server/ServerConfig.kt
```diff
@@ -180,10 +206,15 @@ class ServerConfigMapperImpl(
             blackList = blackList,
             teams = teams,
             title = title,
-            isOnPremises = isOnPremises
+            isOnPremises = isOnPremises,
+            proxy = proxy?.let { ServerConfig.Proxy(it.isProxyNeedsAuthentication, it.apiProxy, it.proxyPort) }
```
mchenani: I see this proxy object mapped manually many times, I suggest to have a mapper for the Proxy object, then we only call it to prevent false mappings. in the end those prxyMapper functions can be tested.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
```diff
@@ -19,6 +19,7 @@ sealed class AuthenticationResult {
     ) : AuthenticationResult()
 
     sealed class Failure : AuthenticationResult() {
+        object SocketError : Failure()
```
mchenani, line 22: this socketError is kind of error that not easy to map in head of the developer! Since it's relates to the tech of the proxy itself, isn't better to rename it to ProxyError?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/autoVersioningAuth/AutoVersionAuthScopeUseCase.kt
```diff
@@ -3,19 +3,33 @@ package com.wire.kalium.logic.feature.auth.autoVersioningAuth
...
             handleError(it)
         }, { serverConfig ->
-            Result.Success(coreLogic.getAuthenticationScope(serverConfig))
+            when (proxyCredentials) {
+                is ProxyCredentials.None -> {
+                    Result.Success(coreLogic.getAuthenticationScope(serverConfig, null))
+
+                }
+                is ProxyCredentials.UsernameAndPassword -> {
+                    with(proxyCredentials.proxyCredentialsModel) {
```
mchenani: Nice, thanks 👨‍🌾

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/PersistProxyCredentialsUseCase.kt
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+}
+
+internal class PersistProxyCredentialsUseCaseImpl(
+    private val proxyCredentialsRepository: ProxyCredentialsRepository
+) : PersistProxyCredentialsUseCase {
+
+    override suspend operator fun invoke(username: String, password: String): PersistProxyCredentialsUseCase.PersistProxyCredentialsResult =
+        proxyCredentialsRepository.persistProxyCredentials(username, password).fold({
```
mchenani, line 27: tiny suggestion: since the repo name itself is explanatory why we need to have a full name function? isn't better to simply rename the `persistProxyCredentials` to `persist`?

logic/src/commonMain/kotlin/com/wire/kalium/logic/network/SessionManagerImpl.kt
```diff
@@ -60,4 +66,15 @@ class SessionManagerImpl(
     override suspend fun onClientRemoved() {
         sessionRepository.logout(userId, LogoutReason.REMOVED_CLIENT)
     }
+
+    override fun proxyCredentials(): ProxyCredentialsDTO? =
+        wrapStorageNullableRequest { proxyCredentialsStorage.getProxyCredentials() }.nullableFold({
```
mchenani, line 71: I can't get this part, we have the nullableFold then we check the nullability manually in next lines!

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
```diff
@@ -1,15 +1,49 @@
...
+            InetSocketAddress.createUnresolved(serverConfigDTOProxy?.apiProxy, serverConfigDTOProxy!!.proxyPort)
+        )
+
+        val client = OkHttpClient.Builder().pingInterval(WEBSOCKET_PING_INTERVAL_MILLIS, TimeUnit.MILLISECONDS).proxy(proxy)
+            .build()
+        preconfigured = client
+        webSocketFactory = KaliumWebSocketFactory(client)
+
+    } else {
+        val client = OkHttpClient.Builder().pingInterval(WEBSOCKET_PING_INTERVAL_MILLIS, TimeUnit.MILLISECONDS).build()
```
mchenani, line 45: I suggest to extract this settings [pingInterval and..] in a companion object so we can see them in one place and easy to config.

network/src/commonMain/kotlin/com/wire/kalium/network/networkContainer/UnboundNetworkContainer.kt
```diff
@@ -19,7 +19,7 @@ private interface UnboundNetworkClientProvider {
 
 internal class UnboundNetworkClientProviderImpl(
     val developmentApiEnabled: Boolean,
-    engine: HttpClientEngine = defaultHttpEngine()
+    engine: HttpClientEngine = defaultHttpEngine(null)
```
mchenani, line 22: default values? 🗡️

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/client/ProxyCredentialsStorage.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import com.wire.kalium.persistence.kmmSettings.KaliumPreferences
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+interface ProxyCredentialsStorage {
+    /**
+     * to save the proxy credentials after the user add and verify correct proxy config at the login
+     */
+    fun saveProxyCredentials(username: String, password: String)
```
mchenani, line 11:
```suggestion
    fun persist(username: String, password: String)
```

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/client/ProxyCredentialsStorage.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * get the proxy credentials to be used with the authenticated client
+     */
+    fun getProxyCredentials(): ProxyCredentialsEntity?
```
mchenani, line 16:
```suggestion
    fun fetch(): ProxyCredentialsEntity?
```

mchenani: Or `persistCredentials` and `fetchCredentials` ...

logic/src/commonMain/kotlin/com/wire/kalium/logic/network/SessionManagerImpl.kt
MohammedMokresh: isn't the nullable to return null if there is no value !, so here the storage can be null before we store any credentials or in the case where we don't have proxy at all , so we make it nullable and check if the value not null so we map it else return null

network/src/commonJvmAndroid/kotlin/com/wire/kalium/network/HttpEngine.kt
MohammedMokresh: it's already constants inside networkClient as we use it there as well

logic/src/commonMain/kotlin/com/wire/kalium/logic/network/SessionManagerImpl.kt
```diff
@@ -60,4 +66,15 @@ class SessionManagerImpl(
     override suspend fun onClientRemoved() {
         sessionRepository.logout(userId, LogoutReason.REMOVED_CLIENT)
     }
+
+    override fun proxyCredentials(): ProxyCredentialsDTO? =
+        wrapStorageNullableRequest { proxyCredentialsStorage.fetch() }.nullableFold({
```
mchenani, line 71: We have the nullableFold and again in the next part we have the null check, what's the reason? I think I missed something here.

mchenani (approved):
Great job🦾 Just one question.

PR #1040, opened 2022-10-12
fix: asset upload issues [AR-1671][AR-2472][AR-2585][AR-2583]

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently, the scope of the assets upload or download was bounded to the lifecycle of the screen where they were being triggered, provoking that way the failure every time we were destroying these views by navigating out of them.

### Solutions

1. With this PR, the scope of these jobs, will be bounded to the app context, as this one will be provided on the user scope creation (app's init method).
2. Since the above mentioned changes were keeping the scope only while the app was running, we needed to come up with a strategy to avoid possible asset download/upload status deadlocks when app and its scope was being destroyed/killed/swiped out. Therefore, I added a SQL method to reset these transient status on app's initialisation in a similar fashion to how we do it with hanged calls when app is being killed (kudos to @vitorhugods for nice and elegant solution tip).

#### Test Coverage (Optional)
WIP

#### How to Test

Just send an asset message, and navigate out of the conversation. Check that the asset message gets eventually delivered. If killing the app while the message is being uploaded or downloaded, check that the asset is marked as `FAILED` asset.

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
@@ -125,13 +125,10 @@ internal class ScheduleNewAssetMessageUseCaseImpl(
             )
 
             // We persist the asset message right away so that it can be displayed on the conversation screen loading
-            persistMessage(message)
-                .onSuccess {
-                    // We schedule the asset upload and return Either.Right(Unit) so later it's transformed to Success(message.id)
-                    scope.launch(dispatcher.io) { uploadAssetAndUpdateMessage(message, conversationId) }
-                }.onFailure {
-                    updateAssetMessageUploadStatus(Message.UploadStatus.FAILED_UPLOAD, conversationId, message.id)
-                }
+            persistMessage(message).onSuccess {
+                // We schedule the asset upload and return Either.Right(Unit) so later it's transformed to Success(message.id)
+                scope.launch(dispatcher.io) { uploadAssetAndUpdateMessage(message, conversationId) }
+            }
         }.fold({
             updateAssetMessageUploadStatus(Message.UploadStatus.FAILED_UPLOAD, conversationId, message.id)
             ScheduleNewAssetMessageResult.Failure(it)


vitorhugods:
For the tests to work, I guess we need to call `advanceUntilIdle` to make sure it executes whatever is inside the `scope.launch`

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetMessageAssetUseCase.kt
```diff
@@ -67,29 +72,34 @@ internal class GetMessageAssetUseCaseImpl(
...
+                }, { decodedAssetPath ->
+                    // Only update the asset download status if it wasn't downloaded before, aka the asset was indeed downloaded while
+                    // running this specific use case. Otherwise, recursive loop as described above kicks in.
+                    if (!wasDownloaded)
+                        updateAssetMessageDownloadStatus(Message.DownloadStatus.SAVED_INTERNALLY, conversationId, messageId)
+
+                    downloadResult = MessageAssetResult.Success(decodedAssetPath, assetMetadata.assetSize)
+                })
+            }
+            downloadJob.join()
```
vitorhugods, line 97: About the tests failing

## Current scenario

The test dispatchers have a limited parallelism of `1`, that is: no parallel execution.

Launching the job and calling `join` right after is locking the tests:
- Launch job
- Wait for it to complete
- How will it complete if we didn't tell the job dispatcher to run the launched jobs?
- ⚠️⚠️⚠️⚠️ Timeout

## My analysis of the issue

### _Maybe_ `UnconfinedTestDispatcher` does the trick?

Maybe one way to fix it is by using [`UnconfinedTestDispatcher`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-unconfined-test-dispatcher.html). It guarantees that `launch` and `async` blocks will be executed eagerly _at least inside the top-level of runTest_, so I'm not 100% sure it would fix it.

### Perhaps these tests show that there's room for improvement?

Even if `UnconfinedTestDispatcher` does the trick, should we need a trick in the first place? Perhaps we can rethink a bit how it's implemented:

#### Download

Do we need to keep downloading if whatever wanted the asset cancelled the execution and is no longer interested in it? If the consumers of Kalium are interested in the asset, `GetMessageAssetUseCase` will be called again. So, no need to launch using another Scope. We can perform it in the original scope of the caller.


#### Upload

Do we really need to keep waiting for the result? We can launch and forget. The item will be updated with a failure anyway 🤷🏼‍♂️ 

Suppose we have a WorkManager that would run it, we would not be able to wait for the result too.

The important thing is that the message/asset is updated later with `UPLOADED` or `UPLOAD_FAILURE`.

gongracr: I just added these jobs and waited for them to finish to be more faithful to the essence of the use cases, which is, to execute the desired actions and return an appropriate use case `Result` after it. If we take the road of fire and forget, we might end up returning result values that are not accurate. 
- For example, in the case of sending an asset while being offline, if we delegate the responsibility of uploading and sending the final asset message to a separate coroutine, without waiting for it to be finished (successfully or with an error), we will be returning a `Success` result right after launching the new coroutine, when indeed the result will be a `Failure`. 
- In the scenario of downloading an asset, the strategy of launching and forgetting is simply not possible, as the caller of the use case is expecting a valid result with a valid asset path, so we definitely need to wait for the execution of the download.

vitorhugods: ## Downloading

> In the scenario of downloading an asset, the strategy of launching and forgetting is simply not valid. The caller of the SendAssetMessageUseCase is expecting a valid result with a valid asset path, so we definitely need to wait for the execution of the download before returning anything. The alternative of not delegating the download to a different coroutine will be that we could only download the asset while being on the conversation screen... So I guess this call will need to be consensed at some point with product too...

I didn't say we should fire and forget the downloading. I said "no need to launch using another Scope. We can perform it in the original scope of the caller".
Using `scope.launch` is unnecessary. Why pass it to another scope, if we are bound to return a value anyway?

## Uploading

> For example, in the case of sending an asset while being offline, if we delegate the responsibility of uploading and sending the final asset message to a separate coroutine, without waiting for it to be finished (successfully or with an error), we will be returning a Success result right after launching the new coroutine, when indeed the result will be a Failure. Although we could agree that this might not be too bad (the message will be eventually sent and its upload status updated accordingly), it will not represent precisely the final result of it increasing the chances of a future error mapping wrongly the results by a consumer of Kalium.

I agree, we are not being true to what is expected of the `Send` UseCase. But we are free to change the UseCase and its contract. We can rename it to `ScheduleNewAssetMessageUseCase` or `ComposeNewAssetMessageUseCase` for example. In which case, returning Success with the ID of the created message is absolutely fine.

We can also change the return type of the UseCase, we could, for example, return a `Deffered<Result>`, so whoever is calling this can choose to wait for the result or not.

Personally, I prefere the first approach. We can return an ID, the caller can choose to do whatever with it, even though I think currently we have no usecase for it.

gongracr: Renamed the `SendAssetMessageUseCase` to `ScheduleNewAssetMessageUseCase`. And declared `GetMessageAssetUseCase` result type as `Deferred` so that the consuming part decides what to do with the download request, whether to launch & forget, or wait for it to finish.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
```diff
@@ -115,18 +125,19 @@ internal class SendAssetMessageUseCaseImpl(
...
-            }.onFailure {
-                kaliumLogger.e("Asset persist method failed")
-                updateAssetMessageUploadStatus(Message.UploadStatus.FAILED_UPLOAD, conversationId, message.id)
+            persistMessage(message).fold({
                 Either.Left(it)
-            }.flatMap {
-                uploadAssetAndUpdateMessage(message, conversationId)
-            }
+                                         }, {
+                // We schedule the asset upload and return the message ID
```
saleniuk, line 131: Here it returns `Unit`, not message ID so it took me a few seconds to understand it 😅 maybe change it to be more like "We schedule the asset upload and return Either.Right(Unit) so later it's transformed to Either.Right(message.id)"? 

Also, maybe subtle change but I feel like this way it would look better 😄 
```
persistMessage(message)
    .onSuccess {
        scope.launch(dispatcher.io) { uploadAssetAndUpdateMessage(message, conversationId) }
    }
    .onFailure {
        updateAssetMessageUploadStatus(Message.UploadStatus.FAILED_UPLOAD, conversationId, message.id)
    }
```

saleniuk (approved):
Just two suggestions that don't affect the code's functioning, so I'm giving a 👍  😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
gongracr: Fair enough. Thanks for making it look more readable 🤓

PR #1048, opened 2022-10-14
refactor: conversation event processing AR-2620 - part 2

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- `addMembersToCloudAndStorage` assumes the default member role instead reading it from the server response.
- `createGroupConversation` optimistically  inserts members in DB when creating a MLS group instead of waiting for the server response.
- `addMemberToMLSGroup` assumes the default member role instead reading it from the server response.
- Insertion of system messages is duplicated by `AddMemberToGroupConversation`, `RemoveMemberFromGroupConversation` `MemberJoinEventHandler` and `MemberLeaveEventHandler`.

### Solutions

- Inject `MemberJoinEventHandler` and `MemberLeaveEventHandler` everywhere where we need to handle member join & member leave events.

### Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/kalium/pull/1047

#### Test Coverage

----

----

----
##### References


network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/conversation/ConversationMemberChangeDTO.kt
```diff
@@ -18,36 +18,14 @@ sealed class ConversationMemberAddedDTO {
...
-        val fromUser: UserId,
-        @SerialName("time")
-        val time: String
-    ) : ConversationMemberAddedDTO()
+    data class Changed(val event: EventContentDTO.Conversation.MemberJoinDTO) : ConversationMemberAddedDTO()
 }
 
 @Serializable
 sealed class ConversationMemberRemovedDTO {
     // TODO: the server response with an event aka, UserAdded model is inaccurate
```
vitorhugods, line 26: What about this `TODO`?

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/conversation/ConversationMemberChangeDTO.kt
```diff
@@ -18,36 +18,14 @@ sealed class ConversationMemberAddedDTO {
...
```
vitorhugods, line 18: Same here

vitorhugods (commented):
Looks nice, I just don't quite understand the change in `ConversationMemberRemovedDTO` and `ConversationMemberAddedDTO` 🤔

typfel:
> Looks nice, I just don't quite understand the change in `ConversationMemberRemovedDTO` and `ConversationMemberAddedDTO` 🤔

The response for removing or adding a member is the same event which ends up in the notification stream. This change makes that explicit by re-using the event DTO.

This also enables me to re-use the member join / leave event handler when processing the response.

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/conversation/ConversationMemberChangeDTO.kt
typfel: Good question. I'll find out who added it. I suppose it could be removed.

vitorhugods:
> > Looks nice, I just don't quite understand the change in `ConversationMemberRemovedDTO` and `ConversationMemberAddedDTO` 🤔
> 
> The response for removing or adding a member is the same event which ends up in the notification stream. This change makes that explicit by re-using the event DTO.
> 
> This also enables me to re-use the member join / leave event handler when processing the response.

I thought it was changing the expected JSON structure.
But it isn't, unless someone tries to deserialise using these directly
Perhaps it would be more explicit of using a typealias?

Something like
```kotlin
typealias AddMemberResponse = Event.MemberChangedDTO
typealias RemoveMemberReaponse = Event.MemberChangedDTO
```
And so on

typfel:
> > > Looks nice, I just don't quite understand the change in `ConversationMemberRemovedDTO` and `ConversationMemberAddedDTO` 🤔
> > 
> > 
> > The response for removing or adding a member is the same event which ends up in the notification stream. This change makes that explicit by re-using the event DTO.
> > This also enables me to re-use the member join / leave event handler when processing the response.
> 
> I thought it was changing the expected JSON structure. But it isn't, unless someone tries to deserialise using these directly Perhaps it would be more explicit of using a typealias?
> 
> Something like
> 
> ```kotlin
> typealias AddMemberResponse = Event.MemberChangedDTO
> typealias RemoveMemberReaponse = Event.MemberChangedDTO
> ```
> 
> And so on

I don't quite follow the suggestion, there's no `Event.MemberChangedDTO`.

I can rename `ConversationMemberAddedDTO` to ConversationMemberAddedResponse` if that makes things more clear.

vitorhugods:
> > > > Looks nice, I just don't quite understand the change in `ConversationMemberRemovedDTO` and `ConversationMemberAddedDTO` 🤔
> > > 
> > > 
> > > The response for removing or adding a member is the same event which ends up in the notification stream. This change makes that explicit by re-using the event DTO.
> > > This also enables me to re-use the member join / leave event handler when processing the response.
> > 
> > 
> > I thought it was changing the expected JSON structure. But it isn't, unless someone tries to deserialise using these directly Perhaps it would be more explicit of using a typealias?
> > Something like
> > ```kotlin
> > typealias AddMemberResponse = Event.MemberChangedDTO
> > typealias RemoveMemberReaponse = Event.MemberChangedDTO
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And so on
> 
> I don't quite follow the suggestion, there's no `Event.MemberChangedDTO`.
> 
> I can rename `ConversationMemberAddedDTO` to `ConversationMemberAddedResponse` if that makes things more clear.

Before, `ConversationMemberAddedDTO.Changed` was being mapped from this:
```json
{
    "type": "conversation.member-leave",
    "qualified_conversation": {***},
    "qualified_from": {***},
    "time": "2022-10-17:11:14..."
}
```

Now it has the following structure:

```json
{
    "event": {
        "type": "conversation.member-leave",
        "qualified_conversation": {***},
        "qualified_from": {***},
        "time": "2022-10-17:11:14..."
    }
}
```

The API doesn't return this new format above ↑. It returns the old format.

So we're now making the request using `EventContentDTO.Conversation.MemberLeaveDTO` and then mapping to `ConversationMemberRemovedDTO.Changed`.

This way we are defining a DTO that does not exist in the API and could be unsafely used leading to a serialization exception. This is what I'm trying to avoid.

Looking again at this, typealias won't work, as we can't have typealias as one of the sealed class types, but I also don't see any obvious alternative. I guess leaving the way it is is fine 🤷🏼.

But we could take out the `@Serializable` annotation from `ConversationMemberRemovedDTO` and `ConversationMemberAddedDTO` as they do not represent serializable data anymore.

typfel:
> > > > > Looks nice, I just don't quite understand the change in `ConversationMemberRemovedDTO` and `ConversationMemberAddedDTO` 🤔
> > > > 
> > > > 
> > > > The response for removing or adding a member is the same event which ends up in the notification stream. This change makes that explicit by re-using the event DTO.
> > > > This also enables me to re-use the member join / leave event handler when processing the response.
> > > 
> > > 
> > > I thought it was changing the expected JSON structure. But it isn't, unless someone tries to deserialise using these directly Perhaps it would be more explicit of using a typealias?
> > > Something like
> > > ```kotlin
> > > typealias AddMemberResponse = Event.MemberChangedDTO
> > > typealias RemoveMemberReaponse = Event.MemberChangedDTO
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > And so on
> > 
> > 
> > I don't quite follow the suggestion, there's no `Event.MemberChangedDTO`.
> > I can rename `ConversationMemberAddedDTO` to `ConversationMemberAddedResponse` if that makes things more clear.
> 
> Before, `ConversationMemberAddedDTO.Changed` was being mapped from this:
> 
> ```json
> {
>     "type": "conversation.member-leave",
>     "qualified_conversation": {***},
>     "qualified_from": {***},
>     "time": "2022-10-17:11:14..."
> }
> ```
> 
> Now it has the following structure:
> 
> ```json
> {
>     "event": {
>         "type": "conversation.member-leave",
>         "qualified_conversation": {***},
>         "qualified_from": {***},
>         "time": "2022-10-17:11:14..."
>     }
> }
> ```
> 
> The API doesn't return this new format above ↑. It returns the old format.
> 
> So we're now making the request using `EventContentDTO.Conversation.MemberLeaveDTO` and then mapping to `ConversationMemberRemovedDTO.Changed`.
> 
> This way we are defining a DTO that does not exist in the API and could be unsafely used leading to a serialization exception. This is what I'm trying to avoid.
> 
> Looking again at this, typealias won't work, as we can't have typealias as one of the sealed class types, but I also don't see any obvious alternative. I guess leaving the way it is is fine 🤷🏼.
> 
> But we could take out the `@Serializable` annotation from `ConversationMemberRemovedDTO` and `ConversationMemberAddedDTO` as they do not represent serializable data anymore.

I'm not changing the JSON we are consuming, it's not wrapped in a `"event": { }` object.

But you right that `ConversationMemberRemovedDTO` is no longer represent data we that decode so I should remove the `@Serializable` annotation. They now represent the two cases which the backend respond with: 

200 = change was applied 
204 = change was already applied 

I'll rename to `ConversationMemberRemovedResponse` and remove the @Serializable` annotation. I think that makes it clear that this is no re-usable model in other contexts.

vitorhugods (approved):
🚀🚀🚀🚀🚀

PR #1050, opened 2022-10-14
test: add tools to create databases for testing

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We are always mocking the Database.
For some complex high-level tests it's way easier (and faithful to actual prod code) to use a real Database.

For example, if we want to test a scenario where we receive a Sync event, like new message, and then call a UseCase to get messages from a conversation.

Mocking all the different mechanisms would be a huge pain.

### Solutions

Create a `persistence-test` module, which can be used by consumers of `persistence`.
This module contains simple `TestUserDatabase` and `TestGlobalDatabase` classes that can be used to grab DAOs and use actual DAOs instead of mocks.

#### Example of usage

In this example (added in this PR to the `logic` module), we test the `ReactionRepository` using an actual Dao:

```kt

class ReactionRepositoryTest {

    private val userDatabase = TestUserDatabase(TestUser.ENTITY_ID)
    private val reactionRepository = ReactionRepositoryImpl(SELF_USER_ID, reactionsDao)

    @Test
    fun givenSelfUserReactionWasPersisted_whenGettingSelfUserReactions_thenShouldReturnPreviouslyStored() = runTest {
        insertInitialData()
        val emoji = "🫡"
        reactionRepository.persistReaction(TEST_MESSAGE_ID, TEST_CONVERSATION_ID, SELF_USER_ID, "Date", emoji)

        val result = reactionRepository.getSelfUserReactionsForMessage(TEST_MESSAGE_ID, TEST_CONVERSATION_ID)

        result.shouldSucceed {
            assertTrue(it.size == 1)
            assertEquals(emoji, it.first())
        }
    }

}

```

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/build.gradle.kts
+++ b/logic/build.gradle.kts
@@ -24,6 +24,9 @@ android {
     packagingOptions {
         resources.pickFirsts.add("google/protobuf/*.proto")
     }
+    // Run only Instrumented tests. No need to run Unit AND Instrumented
+    // We have JVM tests if we want to run quickly on our machines
+    sourceSets.remove(sourceSets["test"])
 }
 
 kotlin {
@@ -87,7 +90,7 @@ kotlin {
                 implementation(Dependencies.Android.work)
             }
         }
-        val androidTest by getting {
+        val androidAndroidTest by getting {
             dependencies {
                 implementation(Dependencies.AndroidInstruments.androidTestRunner)
                 implementation(Dependencies.AndroidInstruments.androidTestRules)
--- a/logic/build.gradle.kts
+++ b/logic/build.gradle.kts
@@ -68,6 +68,7 @@ kotlin {
         val commonTest by getting {
             dependencies {
                 implementation(kotlin("test"))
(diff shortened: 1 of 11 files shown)


typfel:
What are the benefits of using a real database in the tests? Less setup because you don't need to mock the database behaviour? A drawback could be that it's more expensive to create database for each test. Currently the test database is runs on disk, does it make sense to run an in memory DB for better performance?

vitorhugods:
> What are the benefits of using a real database in the tests? Less setup because you don't need to mock the database behaviour? A drawback could be that it's more expensive to create database for each test. Currently the test database is runs on disk, does it make sense to run an in memory DB for better performance?

I've edited the `Issue` in the PR tempalte to better explain the goal.

About making it in memory, I'll do some research on the topic. It would be indeed something very nice to have.

typfel (approved):

PR #1052, opened 2022-10-14
fix: MissingClientRegistration after reopening the app

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

After closing and reopening the app, some actions like sending message result in `MissingClientRegistration`.

### Causes (Optional)

ProteusClient isn't retrieved from the local files after reopening the app.

### Solutions

Split `open` method into two: `openOrCreate` and `openOrError`. First one creates Proteus local files by opening the box even when not existing, the second one checks if files already exist and only then opens the box, otherwise returns failure.

#### How to Test

Install the app, login, close the app, reopen again and try to send a message.

----
##### References

### DIFF
--- a/cryptography/src/jsMain/kotlin/com/wire/kalium/cryptography/ProteusClientImpl.kt
+++ b/cryptography/src/jsMain/kotlin/com/wire/kalium/cryptography/ProteusClientImpl.kt
@@ -31,11 +31,7 @@ actual class ProteusClientImpl actual constructor(rootDir: String) : ProteusClie
     }
 
     override suspend fun openOrError() {
-        val engine = MemoryEngine()
-        engine.init("in-memory").await()
-
-        box = Cryptobox(engine)
-        box.load().await() // TODO is the use of box.load() instead of box.create() ) correct for this method?
+        openOrCreate() // JS cryptobox is in-memory only
     }
 
     override fun getIdentity(): ByteArray {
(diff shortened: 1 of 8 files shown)


cryptography/src/jsMain/kotlin/com/wire/kalium/cryptography/ProteusClientImpl.kt
```diff
@@ -21,14 +22,22 @@ actual class ProteusClientImpl actual constructor(rootDir: String) : ProteusClie
...
         box = Cryptobox(engine)
         box.create().await()
     }
 
+    override suspend fun openOrError() {
+        val engine = MemoryEngine()
+        engine.init("in-memory").await()
+
+        box = Cryptobox(engine)
+        box.load().await() // TODO is the use of box.load() instead of box.create() ) correct for this method?
```
MohamadJaara, line 38: @typfel can you please answer this one ?

MohamadJaara (approved):
LGTM 🚀, but i would like for @typfel to take a look at the changes for the js Cryptobox

cryptography/src/jsMain/kotlin/com/wire/kalium/cryptography/ProteusClientImpl.kt
typfel: There's no nothing to load because the JS cryptobox in-memory only, so `openOrError` will always fail by definition.

saleniuk: Ok, so I think it should always do the same as `openOrCreate`, because we use this `openOrError` in the common code or we should find some other way of checking whether the box has been already created on JS

PR #1070, opened 2022-10-21
chore: add RootPathsProvider [AR-2591]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In order to perform the migration, we need to know the location of the crypto box to move the files from the old app.

### Solutions

Introduce `RootPathsProvider` which can contain paths that can be available from the `CoreLogic`, for now only account path and proteus path, so the app can retrieve the path to which the crypto files should be moved. This provider is also passed to the `UserSessionScopeProviderImpl` to create and keep paths in one place only.

----
##### References


yamilmedina (approved):
lgtm 🥇  !

PR #1072, opened 2022-10-24
chore: add PersistRegisteredClientIdUseCase [AR-2591]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In order to perform the migration, we need to be able to persist already registered client id from the old app.

### Solutions

We already have the logic of it inside `GetOrRegisterClientUseCase`, so it's extracted to the separate use case `PersistRegisteredClientIdUseCase` which checks whether the given `ClientId` is registered and persists the id in the local storage if so.

#### Test Coverage (Optional)

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/PersistRegisteredClientIdUseCase.kt
```diff
@@ -0,0 +1,48 @@
...
+interface PersistRegisteredClientIdUseCase {
+
+    /**
+     * @param clientId client id to be persisted
+     * @return success if the given id is registered, failure otherwise
+     */
+    suspend operator fun invoke(clientId: ClientId): PersistRegisteredClientIdResult
+}
+
+class PersistRegisteredClientIdUseCaseImpl(
```
yamilmedina, line 21: We still don't have a detekt rule for this, but it's a good idea to make this class internal (or its constructor), so we just communicate with clients using public interfaces :D

yamilmedina (approved):
Just a minor adjustment about the Usecase Impl visibility. Looks on point 🙌

PR #1075, opened 2022-10-25
feat: migration conversations from scala (AR-2625)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Kalium logic to expose an use case to persist the migrated conversations from the old application to the new one, this takes default values, that later might be updated by the sync process (only active conversations), deleted will only be there to retain history of messages later.

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -301,6 +303,25 @@ internal class ConversationMapperImpl(
         ConversationOptions.Protocol.MLS -> ConvProtocol.MLS
     }
 
+    override fun toDaoModel(conversation: Conversation): ConversationEntity = with(conversation) {
```
saleniuk, line 306: Shouldn't we pass during migration some more detailed object? I feel like using `Conversation` we miss some crucial data that we could and should get from the previous app, like `creatorId`, `accessRole` or `mutedStatus`, especially when we migrate it after the sync, so we should take all important data from the previous app, or am I misunderstanding this?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -566,6 +567,15 @@ internal class ConversationDataSource internal constructor(
             .map { it.map { conversationIdEntity -> idMapper.fromDaoModel(conversationIdEntity) } }
     }
 
+    override suspend fun insertConversationFromMigration(conversations: List<Conversation>): Either<CoreFailure, Unit> {
```
saleniuk, line 570: I think we don't have to mention that it's used for the migration, maybe we'll find some other cases in the future 😄 also, the name is singular "insertConversation" but the param is actually a list 😉

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
yamilmedina: I did the change and map the rest of the meaningful data we need, except for `accessRole`.
Talking with Marco, this case of migrated conversations that will not come from backend/sync, means are conversations without access anymore, so anyway we will not able to do nothing.

Nevertheless, I mapped the rest of the data to do it more generic

saleniuk (approved):

PR #1076, opened 2022-10-26
feat: Receive mentions

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-2545 - Receiving @mentions](https://wearezeta.atlassian.net/browse/AR-2545)

### Issues

We weren't selecting and mapping the results from DB View for mentions.

### Causes (Optional)

DB View wasn't accounting for mentions

### Solutions

Add mentions in DB View and map its results

### Testing

Tests will be added in another PR

----
##### References
1. https://www.sqlitetutorial.net/sqlite-group_concat/
2. https://www.sqlite.org/lang_corefunc.html

### DIFF
--- a/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageEntity.kt
+++ b/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageEntity.kt
@@ -2,6 +2,8 @@ package com.wire.kalium.persistence.dao.message
 
 import com.wire.kalium.persistence.dao.QualifiedIDEntity
 import com.wire.kalium.persistence.dao.reaction.ReactionsEntity
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
 
 @Suppress("LongParameterList")
 sealed class MessageEntity(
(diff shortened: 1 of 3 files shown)


vitorhugods:
> Tests will be added in another PR

Any reason for that? 😅

alexandreferris:
> 

Nothing in particular, was following the way my older epic PRs was like:
Part 1: logic
Part 2: tests

Can add them here if you want tho, this and the tests branch will still be pointing to the epic branch anyways 🤷🏻‍♂️

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageEntity.kt
```diff
@@ -113,10 +115,11 @@ sealed class MessageEntity(
         VISIBLE, DELETED, HIDDEN
     }
 
+    @Serializable
     data class Mention(
-        val start: Int,
-        val length: Int,
-        val userId: QualifiedIDEntity
+        @SerialName("start") val start: Int,
```
yamilmedina, line 120: I'm not sure about putting this `Serializable` annotations in our entity models, I think they belong to our network dto's more, can we do it without ? 

Also, I didn't check that deep, but if we have it on other entities, I guess should be ok 😅

yamilmedina (commented):
Small question and suggestion to try, great catch 🍞 🧀

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageEntity.kt
alexandreferris: We do have in other entities and it is done here explicitly just so it can decode from a JSON string into this data class 🤔

yamilmedina (approved):
🪁

PR #1080, opened 2022-10-26
Fix/missing notifications soft logout [AR-2612]

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Our way of handling tokens when logging out is not optimal. Not so long ago, we introduced the concept of soft logging out. This means the user wants to close his/her current session, and update all the local and remote information to reflect that his session is inactive, but without removing all the messaging history and other related session data as he could log in again eventually. 
So depending on the data that is removed/cleared on soft logout, the behavior when logging in again will be different as in the initial session.

The problem here came when we deregister the firebase token but we didn't update our internal flag `SHOULD_UPDATE_FIREBASE_TOKEN_KEY` once the user logs out. This caused that on sucessive logs in by the same client (the client data was still there as it had been a soft logout), this flag was still set to `false` causing that the firebase token will not be registered again, and then notifications will never arrive when the app was killed or in the background.

### Solutions

Added a call to `pushTokenRepository.setUpdateFirebaseTokenFlag(true)` in order to reset the flag for future sessions.

#### Test Coverage (Optional)

----

----

----
##### References


trOnk12 (approved):
well done 🧑‍🎄

PR #1085, opened 2022-10-27
feat: Receiving mentions part3

borichellow:
# What's new in this PR?

### Issues

in DB: `ConversationDetails` View Table had no `unreadMentionsCount` filed, which Is needed.

### Causes (Optional)

Nobody added it there 

### Solutions

Add it, update the DB-migration and Mappers


alexandreferris (approved):
👌🏻

PR #1069, opened 2022-10-21
refactor: break up ConversationEventReceiverTest

typfel:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

`ConversationEventReceiverTest` is huge

### Solutions

Break it up so each event is tested separately.

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1094, opened 2022-10-28
refactor: split NewMessageEventHandler

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

`NewMessageEventHandler` is BIG. It has many responsibilities and it's hard to write tests for some of the last (and very important) steps like testing how messages of specific contents are handled.

This comes from a discussion I had with @typfel on #1069

The goals of this PR are:
- Make it easier to write protocol-specific behaviour and test it without having to worry about mocking the storage of a Text Message in the end
- Make it easier to write content-specific behaviour and test it without having to worry about which protocol it came from

**Single responsibility** at its finest, really.

### Solutions

> **Note**: Name suggestions are most welcome!

#### Create the concept of `MessageUnpacker`.
A MessageUnpacker takes an event and returns a `MessageUnpackResult` in case of success. This `MessageUnpackResult` can be one of:
- `ProtocolSignalingMessage`: (only for MLS) messages that do not contain messages, the ones users are used to see. For example, MLS commits.
- `ApplicationMessage`: Classical stuff. Text, assets, calls, etc. The normal messages.

#### Create the concept of `MessageUnpacker`

Two implementations: `MLSMessageUnpacker` and `ProteusMessageUnpacker`.
These have a simple responsibility: Take in a MLS/Proteus Event, make whatever is needed in terms of Protocol, and return a `MessageUnpackResult` mentioned above.

#### Create the concept of `ApplicationMessageHandler`.

This will take care of the `ApplicationMessage`, mentioned above.
Inserts into DB, delegate stuff to the `CallManager`, etc.
Most of the meat from `NewMessageEventHandler` ends up here.

#### What happens with `NewMessageEventHandler`?
Now, its the simplest of all, and is basically a delegator:
- Receive Proteus or MLS event
- Ask specialized `MessageUnpacker` to decrypt the message and return a result.
- If it resulted in a `ApplicationMessage`, pass it to the `ApplicationMessageHandler`, which will take care of multiple.
- In case of unpacking/decryption error, pass it to `ApplicationMessageHandler` as a `DecryptionError`.

### Testing

Whilst I didn't ADD tests, no logic was added either.
I've basically reorganized everything and the test classes are matching it too.

#### Test Coverage

-------
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/MessageUnpackResult.kt
```diff
@@ -0,0 +1,20 @@
+package com.wire.kalium.logic.sync.receiver.conversation.message
+
+import com.wire.kalium.logic.data.conversation.ClientId
+import com.wire.kalium.logic.data.id.ConversationId
+import com.wire.kalium.logic.data.message.ProtoContent
+import com.wire.kalium.logic.data.user.UserId
+
+internal sealed interface MessageUnpackResult {
+
+    object ProtocolSignalingMessage : MessageUnpackResult
```
typfel, line 10: I would maybe call this `HandshakeMessage`, but we don't use this type for anything so why not remove it and return an optional  `ApplicationMessage` instead.

typfel (changes requested):
Much needed clean up! 🧹 🌟

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/MessageUnpackResult.kt
typfel: Makes sense.

typfel (approved):

PR #1104, opened 2022-11-02
fix: Make SendTextUseCase accept mentions

borichellow:
# What's new in this PR?

### Issues

sending mentions was not possible

### Causes (Optional)

- `SendTextMessageUseCase` was not accepting Mentions as a parameter 
- `ProtoContentMapper` was not mapping mentions correctly for sending messages

### Solutions

- add mentions as a parameter into `SendTextMessageUseCase`
- fix mapping in `ProtoContentMapper`


ohassine (approved):
lgtm 🚀

PR #1113, opened 2022-11-08
fix: app refreshing auth tokens more that it needs to

MohamadJaara:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

the access token has a lifetime of 15 min but it was noticed that in some cases the refresh is executed more often that it need to.

### Causes (Optional)

each user has 2 HTTP clients at a time and each have its own AuthTokens cache. 
when one of the HTTP clients update the token, the second one is still had the old token cached, which lead to the token being updated twice in a row,

### Solutions

move the caching of tokens to the `SessionManager`

Needs releases with:

- [ ] GitHub link to other pull request

#### Test Coverage (Optional)

- [ ] I have added automated test to this contribution

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1131, opened 2022-11-14
refactor: synchronizing read/write of auth session 

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

even with #1113 there is still the possibility that a HTTP client A does a request in the same time as B is updating the token. which means 2 refresh token requests right after each other.

### Solutions

use a Mutex to lock reading from the AuthToken cache while an update process is ongoing

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/network/SessionManagerImpl.kt
```diff
@@ -31,14 +33,17 @@ class SessionManagerImpl internal constructor(
...
+    private val lock: Mutex = Mutex()
+    private var session: SessionDTO? = null
     private var serverConfig: Atomic<ServerConfigDTO?> = Atomic(null)
 
-    override fun session(): SessionDTO = session.get() ?: run {
-        wrapStorageRequest { tokenStorage.getToken(idMapper.toDaoModel(userId)) }
-            .map { sessionMapper.fromEntityToSessionDTO(it) }
-            .onSuccess { session.set(it) }
-        session.get()!!
+    override suspend fun session(): SessionDTO = lock.withLock {
```
typfel, line 40: So I take it that withLock will do the following:

lock
execute lambda
unlock

network/src/commonMain/kotlin/com/wire/kalium/network/session/SessionManager.kt
```diff
@@ -72,6 +76,7 @@ fun HttpClientConfig<*>.installAuth(sessionManager: SessionManager, accessTokenA
                         null
                     }
                 }.also {
+                    sessionManager.afterTokenUpdate()
```
vitorhugods, line 79: Can we `try { } finally { }` this one?

I am a bit worried we could lock the mutex and never release it if some exception is thrown. Doesn't look like a exception could be thrown right now, but maybe the code changes and this is introduced? I am not sure. Just trying to be safe here.

network/src/commonTest/kotlin/com/wire/kalium/api/TestSessionManagerV0.kt
```diff
@@ -22,6 +22,14 @@ class TestSessionManagerV0 : SessionManager {
...
         )
 
+    override suspend fun beforeTokenUpdate() {
+        TODO("Not yet implemented")
+    }
+
+    override fun afterTokenUpdate() {
+        TODO("Not yet implemented")
+    }
+
```
gongracr, line 32: Are we planning to test this at some point?

network/src/commonMain/kotlin/com/wire/kalium/network/session/SessionManager.kt
```diff
@@ -55,6 +58,7 @@ fun HttpClientConfig<*>.installAuth(sessionManager: SessionManager, accessTokenA
             }
 
             refreshTokens {
+                sessionManager.beforeTokenUpdate()
```
typfel, line 61: Instead of doing:

```
 lock
 refresh token
 update token
 unlock 
 ```
 
 maybe we should move this whole logic to the sessionManager and only enqueue a token refresh here?

Comment (typfel) [reply network/src/commonMain/kotlin/com/wire/kalium/network/session/SessionManager.kt:61]: I think using a queue we can avoid a lock

**Review by typfel — CHANGES_REQUESTED** [review]

File: network/src/commonMain/kotlin/com/wire/kalium/network/session/SessionManager.kt (hunk 4 above)
Comment (typfel) [reply network/src/commonMain/kotlin/com/wire/kalium/network/session/SessionManager.kt:61]: A queue in Kotlin I guess is a dispatcher with parallelism of one.

File: network/src/commonTest/kotlin/com/wire/kalium/api/TestSessionManagerV0.kt (hunk 3 above)
Comment (MohamadJaara) [reply network/src/commonTest/kotlin/com/wire/kalium/api/TestSessionManagerV0.kt]: i just removed it and moved the whole logic to SessionManager

**Review by vitorhugods — APPROVED** [review]
👨‍🍳👌

File: logic/src/commonMain/kotlin/com/wire/kalium/logic/network/SessionManagerImpl.kt [hunk 5]
```diff
@@ -49,29 +69,59 @@ class SessionManagerImpl internal constructor(
...
-        }.fold({
-            TODO("IMPORTANT! Not yet implemented")
+        }.nullableFold({
+            null
         }, {
             it
         })
 
-    override suspend fun onSessionExpired() {
+    override suspend fun updateToken(accessTokenApi: AccessTokenApi, oldAccessToken: String, oldRefreshToken: String): SessionDTO? {
```
typfel, line 90: Future refactoring: inject `AccessTokenApi` into `SessionManagerImpl` and we can remove it has a parameter and we don't need to forward it all the way down to the HttpClient.

typfel (approved):

PR #1133, opened 2022-11-15
feat: receive replies (AR-2484)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Users can't see replies to messages in the chat.

### Solutions

Implement it!

#### There's one ugly thing

- When we INSERT into the database, we don't have (nor need) all the message information (with all the original sender and original message metadata). All we need is the ID of the original message.

- When we QUERY the database, a lot of data comes back.

Because of this, I added two fields to the text message: 
- `quotedMessageReference`: Used when receiving and inserting into the DB
- `quotedMessageDetails`: Result from the Database with all the Joins

The only solution I could think of is completely splitting the "insert" from the "read" model and have two models completely. This complicates stuff and I'm not sure that it's needed right now, but it's something to keep in mind going forward!

#### Test Coverage

#### How to Test

Receive a Text message with a reply.
When fetching the received message, it must contain a reference to the original message and all needed data to display it.

### Notes

Checking if the hash matches is not implemented yet.
It needs some exploration, as we need to handle the changing of content hash as the quoted message is edited.

----

----
##### References


persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/message/BaseMessageTest.kt
```diff
@@ -0,0 +1,56 @@
...
+
+import com.wire.kalium.persistence.BaseDatabaseTest
+import com.wire.kalium.persistence.dao.ConversationDAO
+import com.wire.kalium.persistence.dao.UserDAO
+import com.wire.kalium.persistence.dao.reaction.ReactionDAO
+import com.wire.kalium.persistence.utils.stubs.newConversationEntity
+import com.wire.kalium.persistence.utils.stubs.newUserEntity
+import kotlin.test.BeforeTest
+
+open class BaseMessageTest : BaseDatabaseTest() {
```
alexandreferris, line 11: This is really great 👌🏻

alexandreferris (approved):
Great work @vitorhugods 🔥

PR #1136, opened 2022-11-17
feat: Observe ConversationDetails instead of just get once

borichellow:
# Needed for the task [AR-2491](https://wearezeta.atlassian.net/browse/AR-2491?atlOrigin=eyJpIjoiZTkxZTJkNTI1OWExNDQ1YmIyMTgyOTFmY2ZmZjIzMzAiLCJwIjoiaiJ9)

# What's new in this PR?

### Issues

`GetOneToOneConversationUseCase` returns only one-time result and does not observe the changes. 

### Causes (Optional)

Was implemented like that from the beginning.

### Solutions

Make it return `Flow<Result>` to be able to observe ConversationDetails changes.


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/GetOneToOneConversationUseCase.kt
```diff
@@ -4,13 +4,21 @@ import com.wire.kalium.logic.data.conversation.Conversation
...
 
 class GetOneToOneConversationUseCase internal constructor(
-    private val conversationRepository: ConversationRepository
+    private val conversationRepository: ConversationRepository,
+    private val dispatchers: KaliumDispatcher = KaliumDispatcherImpl
 ) {
 
-    suspend operator fun invoke(otherUserId: UserId): Result = conversationRepository.getOneToOneConversationWithOtherUser(otherUserId)
-        .fold({ Result.Failure }, { Result.Success(it) })
+    suspend operator fun invoke(otherUserId: UserId): Flow<Result> =
```
yamilmedina, line 18: Let's take advantage of these changes, and we can add a small block of documentation, for the purpose of the Use case :D

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/GetOrCreateOneToOneConversationUseCase.kt
```diff
@@ -7,6 +7,7 @@ import com.wire.kalium.logic.data.conversation.ConversationGroupRepository
 import com.wire.kalium.logic.data.conversation.ConversationRepository
 import com.wire.kalium.logic.data.user.UserId
 import com.wire.kalium.logic.functional.fold
+import kotlinx.coroutines.flow.first
 
 class GetOrCreateOneToOneConversationUseCase(
```
yamilmedina, line 12: Let's take advantage of these changes, and we can add a small block of documentation, for the purpose of the Use case :D

yamilmedina (approved):
Small comment about documenting use cases, no blocker since were old classes.
Excellent catch about the root of the problem with the event bug 🪁 🌻

PR #1138, opened 2022-11-17
fix: don't parse response body on 404 errors to align with v3 API (AR-2722)

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Notification API has changed in API V3: [list of changes](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/672006169/API+changes+v2+v3)

**tl;dr** no events are returned on 404 errors any more.

### Solutions

Stop parsing the response body on errors in all API implementation and remove the `isMissingNotifications` property, consumers are now expected to handle the 404 error instead.

#### Test Coverage

----

----

----
##### References


vitorhugods (approved):

network/src/commonMain/kotlin/com/wire/kalium/network/api/v0/authenticated/NotificationApiV0.kt
```diff
@@ -63,13 +61,7 @@ internal open class NotificationApiV0 internal constructor(
         queryClient: String,
         querySince: String?
     ): NetworkResponse<NotificationResponse> {
-        return wrapKaliumResponse({
```
MohamadJaara, line 66: don't we need to do this only for v3, and keep v0-2 parser the old way?

vitorhugods: I'll write here as a way to test if I got the idea right and it makes sense, and @typfel can correct me.

The way I saw it, we can just fail (for now).
At the moment, we check that Slow Sync is performed every 7 days, and unless the device becomes offline for 21+ days or some development backend is wiped, the `404` shouldn't even happen.

### Now, let's talk about prod.

When prod comes, this 7 day thing should be disabled and we should only perform Slow Sync when logging in OR when a 404 is returned, right?

So, logic applied to V0/V1/V2 behaviour (let's call it **"Legacy Logic"**):
- Handle 404, 
- Invalidate Slow Sync and wait its completion
- Resume Incremental Sync with events returned in 404 (getting all events available in the backend).

HOWEVER, this doesn't need to be this way. We can (call it **"New Logic"**):
- Handle 404
- Invalidate Slow Sync and wait its completion
- Perform Incremental Sync without `?since=${lastEventId#}`, also getting all events available in the backend.

The **New Logic** approach should work for all API versions. But it means we also need to change the sync mechanism a bit in order to gracefully perform the slow sync and resumt Incremental sync without `lastEventId`.

This **New Logic** is not implemented in this PR, but the **Legacy Logic** was never developed in Kalium, as I was planning on doing it when lifting the 7 day logic.

typfel: You've understood It correctly @vitorhugods and since neither "Legacy Logic" or "New Logic" is currently implemented why not go for "New Logic" and avoid having to deal with API differences.

PR #1143, opened 2022-11-18
feat: add asset name to replies and support invalid replies (AR-2484)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Some unused fields being returned by the query (asset ID and domain)
- Necessary field not being returned (asset name)
- In the near future we need to store "invalid" replies (when the hash of the reply doesn't match the hash of the original message, or the original message isn't found)

### Solutions

- Change DB fields to return what's needed only
- Add `isQuoteValid` field to text content.

> **Note**: The implementation of valid/invalid quotes will be done in #1140. So the hardcoded `true` will be replaced with actual logic.

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
```diff
@@ -150,7 +150,8 @@ class ProtoContentMapperImpl(
                 protoContent.value.quote?.let {
                     MessageContent.QuoteReference(
                         it.quotedMessageId,
-                        it.quotedMessageSha256?.array
+                        it.quotedMessageSha256?.array,
+                        true // TODO: Check hash to figure out if it's valid
```
typfel, line 154: How did you plan to calculate if the hash is valid here? You would need the content of the quoted message if is not readily accessible here.

vitorhugods: Indeed, this will need to be actually calculated in the `conversation` even receiver part.

It's a bit annoying and I'm not really sure how to model it so the `ProtoContentMapper` can return "idk if it's valid". 

I can't see another way, other than having the `EventReceiver` outside actively perform a sanity check on what's returned by the `ProtoContentMapper`.

I'm open to ideas.

vitorhugods: This kind of behaviour (the `NewMessageEventHandler` actively performing sanity checks on what's returned by the `ProtoContentMapper`) isn't exactly new and unexpected, I guess.

Take Delete/Edit of messages.
`ProtoContentMapper` could say, that user A is trying to edit messages from user B.

The truth is that we don't ever trust what's returned by `ProtoContentMapper` and depending on the message type/content we should perform checks.

I would love to find a way to model this data in a way that makes it less error-prone, as right now this can easily be skipped.

Maybe adding something like "interceptors" to `ProtoContentMapper` so we can do the validation before creating the actual "Message" data? 🤔 

IDK. 

Sounds a bit too complicated, as I just think out loud here. But maybe that gives you some ideas as well.

Maybe something to experiment with. It's definitely one point in the architecture that I'm looking to improve.

typfel: Yeah, I think that's over complicating the problem. Especially for something which we intend to deprecate in the near future.

We can also re-phrase the `isValid` property to something like. `isVerified` and always return `false` from the `ProtoContentMapper`. Then `NewMessageEventHandler` can't skip the  verify content step where it updates the value to `true` on success.

vitorhugods: I like this one. I love this one. It's simple, it's great.

Promoting this fail-deadly approach by `ProtoContentMapper` sounds nice.

However, as the hashing is being implemented still and will come in a next PR. I'm thinking about leaving these changes when it comes.

typfel: Sure go ahead, maybe change the name already though unless you want to do another DB migration.

typfel (approved):

PR #1156, opened 2022-11-25
fix: session upgrade

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

1. After upgrading the session, the an upgraded access token is not used until the next time the access token expires
2. Slow sync starts before we've upgraded our session

### Solutions

1. Clear the cached token on the HttpClient after upgrading the session.
2. Persist the client ID after we've upgraded session because that's the criteria to start the slow sync.

#### Test Coverage

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/VerifyExistingClientUseCase.kt
```diff
@@ -0,0 +1,47 @@
...
+internal class VerifyExistingClientUseCaseImpl(
+    private val clientRepository: ClientRepository
+) : VerifyExistingClientUseCase {
+
+    override suspend fun invoke(clientId: ClientId): VerifyExistingClientResult {
+        return clientRepository.selfListOfClients()
+            .fold({
+                VerifyExistingClientResult.Failure.Generic(it)
+            }, { listOfClients ->
+                val client = listOfClients.firstOrNull { it.id == clientId }
```
typfel, line 30: This looks like a bug to me. We check that our local client is equal to first client in the list of self clients. Shouldn't this be a `listOfClients.contains({ it.id == clientId })`?

MohamadJaara: yes this must be replaced with `listOfClients.contains({ it.id == clientId })`

typfel: Tried to write a failing test and it was passing :) 

`firstOrNull` returns first matching element or null. So essentially same behaviour as `istOfClients.contains({ it.id == clientId })`

PR #1171, opened 2022-11-29
refactor: use proper signaling message type

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently, there's no clear separation between messages that are "shown in the conversation" and those that are just "attached" to other messages.

Example: Reactions VS Text

A Text message is inserted into the Database, it will be queried, and it will be the "main" content.
Then, users can react to that message and it will be handled internally AND displayed as a completely different thing.

This leads to things like this in multiple places: 

```kotlin
when (content) {
    is Text -> // Do something
    is Asset -> // Do something

    is Reaction, is DeleteForMe, is MessageDelete -> // Ignore. These should never be here
}
```

### Solutions

Revamp the `Message` hierarchy, by making it an interface.
There are two interfaces that a Message can extend:
- Sendable: Messages that are transmitted between clients
- Standalone: Messages like the ones mentioned before (text, system messages, attachments, etc.)

Some messages (like text) are be Sendable AND Standalone at the same time.

Some messages (like reactions) are only Sendable.

Some messages (like System messages) are only Standalone.

**See `Messages.kt` with all the brand new updated documentatino for more details.**

### Testing

Compiler did all the heavy lifting :) 

----

----
##### References


typfel (approved):
Nice refactoring 👏

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
```diff
@@ -254,16 +254,16 @@ internal class GetNotificationsUseCaseImpl internal constructor(
...
         is MessageContent.LastRead -> false
         is MessageContent.Cleared -> false
         is MessageContent.ConversationRenamed -> false
         is MessageContent.TeamMemberRemoved -> false
+        is MessageContent.Receipt -> false
     }
 
     private fun shouldMessageBeVisibleAsNotification(message: Message) =
-        message.visibility == Message.Visibility.VISIBLE
+        message is Message.Standalone && message.visibility == Message.Visibility.VISIBLE
```
typfel, line 266: Reactions aren't displayed as  in notifications in AR?

vitorhugods: At the moment? No. It's intended per design. But I'm not sure for how long.

I think this part of the notifications logic should change to query messages and reactions as different sources.

At the moment it's expected that they'd come from the same place, but I'm but sure this should be the case.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSendingInterceptor.kt
```diff
@@ -9,18 +9,19 @@ import com.wire.kalium.logic.functional.map
...
     private val messageContentEncoder: MessageContentEncoder,
     private val messageRepository: MessageRepository
 ) : MessageSendingInterceptor {
 
-    override suspend fun prepareMessage(message: Message.Regular): Either<CoreFailure, Message.Regular> {
+    override suspend fun prepareMessage(message: Message.Sendable): Either<CoreFailure, Message.Sendable> {
         val replyMessageContent = message.content
 
         if (replyMessageContent !is MessageContent.Text
+            || message !is Message.Regular
```
mchenani, line 24: Is that correct? I see we don't have Regular anymore!

vitorhugods: Regular still exists! Please look into Messages.kt for more details.


This Interceptor has the purpose of modifying text messages with quotes, calculating the hash of the quoted message, before sending.

If the message being sent isn't a text message (they must be regular with a text content), then just retur it without modifications.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/ApplicationMessageHandler.kt
```diff
@@ -111,47 +100,71 @@ internal class ApplicationMessageHandlerImpl(
...
+                callManagerImpl.value.onCallingMessageReceived(
+                    message = signaling,
+                    content = content,
+                )
             }
+
+            is MessageContent.TextEdited -> editTextHandler.handle(signaling, content)
+            is MessageContent.LastRead -> lastReadContentHandler.handle(signaling, content)
+            is MessageContent.Cleared -> clearConversationContentHandler.handle(signaling, content)
+            is MessageContent.Receipt -> TODO("Receiving of read receipts not yet implemented")
```
yamilmedina, line 167: I'm thinking if this will cause a crash if immediately merged to develop ? Or is it part of a set of PR's 🤔

yamilmedina (approved):
Great job 👏 ! I just left a question about something isn't clear 🪁

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/ApplicationMessageHandler.kt
vitorhugods: Not yet, as the mapping from Protobuf to Receipt is not done yet, so there's a total of 0 instantiations of `MessageContent.Receipt`.

This just slipped through as it was when creating this class that I noticed that we could improve some stuff :D

PR #1180, opened 2022-12-01
fix: received messages are in bad order [AR-2473]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Received messages can be in wrong order when they are sent in the same second and only differ by milliseconds.

### Causes (Optional)

In SQL we order messages, filter data and make some comparisons by timestamp using `datetime` function which doesn't take milliseconds into account, so if we have multiple messages sent within the same second, they can be ordered not truly chronologically. 

### Solutions

SQL `datetime(...)` is equal to `strftime('%Y-%m-%d %H:%M:%S', ...)` - `%S` are full seconds, but we receive timestamps that have milliseconds so we are able to order messages more precisely. To do that, we have to use `%f`, which is "fractional seconds: SS.SSS". Unfortunately, there is no elegant function to use in this case so we have to replace all `datetime(...)` with `strftime('%Y-%m-%d %H:%M:%f',...)`. 🤷‍♂️

More info here: https://www.sqlite.org/lang_datefunc.html

Visualisation of the difference here:
image

#### How to Test

Type from web very fast numbers from 1 to 9. Before this fix, some of the messages with numbers should be in the wrong order, with this fix the order should be always right.

----

----

----
##### References


ohassine (approved):
Nice catch 👌

trOnk12 (approved):
well done ! ⭕

PR #1190, opened 2022-12-02
feat: fix sync recovery when the client is deleted

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In some cases the server could return us 404 when there is no user or event found. 

This could happen for example, when two clients are logged in on the device and one of them has been deleted, the moment we switch to that account we are going to get 404 for that user. 

We had no recovery scenerio for that scenerio and any other, we were ALWAYS retrying the process. 

Therefore I added a class handling recovery both for slow sync and incremental sync. 

I also propagated the error when the server is returning a user with deleted flag as true, so that we can consume this failure inside the sync. When that happens during the slow sync we log out the user, since there is nothing we can do with it.

Also during incremental sync, when 404 happens we give the sync last spin for a slow sync by restarting that process.

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


vitorhugods (changes requested):
Pretty nice solution in general so far. Just missing tests. Sync is super critical and is very well covered to make sure we don't break it.

From the top of my head, missing coverage:
- FetchingSelfUser returns failure if deleted
- SlowSync recovery if user isn't deleted
- IncrementalSync recovery is triggered and updates the state accordingly
- The IncrementalSync recovery will reset SlowSync
- Logout is called when needed

Also, please fill the PR description explaining what's going on.

Not only for people reviewing it. But it's possible someone will look through commits in the future trying to understand something and will end up here in the PR discussion.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/IncrementalSyncRecoveryHandler.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    private fun shouldRestartSlowSyncProcess(failure: CoreFailure): Boolean {
+        return isClientOrEventNotFound(failure)
+    }
+
+    private fun isClientOrEventNotFound(failure: CoreFailure): Boolean =
+        (failure is NetworkFailure.ServerMiscommunication) &&
+                (failure.kaliumException is KaliumException.InvalidRequestError) &&
+                (failure.kaliumException.errorResponse.code == 404)
+
```
trOnk12, line 35: @mchenani

vitorhugods (approved):
Nicely done ;) 👨‍🍳🤌

PR #1194, opened 2022-12-03
chore: add convention plugins to simplify module setup

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Configuring modules is a pain in the ass at the moment, that involves a lot of copy/pasting.

### Solutions

Create a [convention plugin](https://docs.gradle.org/current/samples/sample_convention_plugins.html) to share the build logic between modules.

Currently there are just a few options, like enabling/disabling some targets (iOS, JS) and adding native code interop (for cryptography).

With this convention plugin, instead of configuring the `android` block and the `kotlin` targets, all it's needed is:

```kotlin
plugins {
    id(libs.plugins.android.library.get().pluginId)
    id(libs.plugins.kotlin.multiplatform.get().pluginId)
    id(libs.plugins.kalium.library.get().pluginId) // <-- ADD THIS
}

kaliumLibrary {
    multiplatform {
        // Default values:
        enableiOS.set(true)
        enableJs.set(true)
        enableJsTests.set(true)
        includeNativeInterop.set(false)
    }

    // Or just call multiplatform() to get all the default values
    multiplatform()
}

// No more need to add android { } with minSdk, testRunner, etc. etc.

kotlin {
    // No more need to add and configure targets manually
    sourceSets {
        val commonMain by getting {
           // Define dependencies normally
        }
    }
} 
```

If necessary, it's possible to open the specific blocks to customize further, like android desugaring, for example.

We could expand this to configure the `commonJvmAndroid` in all modules (or hide it behind one of those flags, instead of copy/pasting, etc.

### Dependencies

The PR that disable for incomplete features is needed

- https://github.com/wireapp/kalium/pull/1197

### Testing

It should build and pass all tests normally.

----

----
##### References


vitorhugods:
Instead of having to add them manually in each module, we could also do these in the near future:
- Add compiler args to suppress all the damn `OptIn` warnings.
- Add option to create `commonJvmAndroid` sourceSet (currently copy/pasted in some modules)


In general, it simplifies a lot to have these configurations in a single place.

yamilmedina (approved):
Amazing ! Finally here🪁  🧉

PR #1197, opened 2022-12-03
test: disable tests for incomplete features

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Some tests are broken since #1185, as the unread counter values are now hardcoded to empty/zero.

### Causes

These tests are using and old version of turbine, which uses the UnconfinedDispatcher and is flaky for us, as it doesn't actually perform the tests every time, sometimes the content of the `flow.test { }` block doesn't even run and doesn't trigger the AssertionErrors.

While we should upgrade Turbine to a newer version, we need to upgrade Kotlin as well. Due to our timeline, this is something we opted to do next week.

### Solutions

1. Don't use Turbine at all for these tests. Turbine isn't actually needed, as we're just asserting the first result.
2. Completely disable these tests, as there's no way they will succeed now, instead of succeeding randomly.

#### Test Coverage

----

----
##### References


vitorhugods:
Closed, as changes were added in https://github.com/wireapp/kalium/pull/1188 in order to unblock that PR

PR #1212, opened 2022-12-08
fix: do not overwrite message data (AR-2764)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- We allow message overwrite in the database, which could allow an attacker to send a message with the same ID and replace someone's message.
- Hashes of quotes between Reloaded and other clients fail when the quoted message has a LinkPreview

### Causes

For the LinkPreview thing, they are an exception.
They actually work in a way that we should allow an partial overwrite, replacing the Text content with the new one. But we should do all sorts of checks to see if the sender is the same, if the content is the same, etc. etc. and then just APPEND the LinkPreview.

We are actually allowing the overwrite and updating the message date, which modifies the message hash and breaks replies.

### Solutions

As we don't support the LinkPreview for now, completely ignore following messages. Do not allow overwrite.

Renamed the `insertMessage` to make it clear that it will ignore if the message already exists. 

> **Note**: In the future I think we should refactor this insertion and fail deadly if that happens. Callers should either catch the exception or check beforehand if there aren't messages with the same ID on that conversation, and call the appropriate DAO function to just append/edit content if necessary.

#### Test Coverage

----

----
##### References


Garzas (approved):
Nice catch 😎

PR #1214, opened 2022-12-08
chore: upgrade kotlin to 1.7.20

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We're stuck in the past with Kotlin 1.6.10 and old versions of libraries such as Ktor, Turbine, Okio, and so many others have jumped ahead of us with plenty of bugfixes that we're having to work-around.

### Causes

We have hold back for a while because of https://youtrack.jetbrains.com/issue/KT-52172 which is a bit annoying and makes the IDE go bananas, but at this point, the costs of holding are outweighing the costs of going forward.

### Solutions

Upgrade. Kotlin from `1.6.10` to `1.7.20` and all dependencies we needed to make it work as well.

Some breaking changes identified are that:
- ✅ `when` branches must be exhaustive
- ✅ SQLDelight changed the package name
- ✅ SQLDelight requires to pass a context/dispatcher when mapping flows (I added a default that mimics the old behaviour, but we should refactor to inject a dispatcher from the Database instead and use a Test Dispatcher for tests)
- ✅ Ktor had a minor package name change for `ContentNegotiation`
- ⚠️ (we can do later if needed) PBandK/Protobuf is broken for JavaScript, we need to investigate further. 

Some work-arounds for issues we added and can be removed:
- ✅ wrapping `Turbine` around a UnconfinedTestDispatcher is not necessary anymore
- ⌛ (we can do later) Remove refreshing auth token before connecting to the websocket (websocket should do it by itself)
- ⌛ (we can do later) Remove our own implementation of SqlDelight's Paging library and use the newer version which should work out of the box
- ⌛ (we can do later) - Upgrade the LibSodium wrapper to a newer version

> **Note**: After this merge, Reloaded developers won't be able to just edit Kalium files from the same IDE window and will need to open a dedicated IDE to change Kalium files, as type resolution will be broken.

### Testing

N/A ?

Compiler should compile.

I'm running Reloaded using Kalium on this branch and so far so good.

----

----
##### References
2. https://github.com/wireapp/.github#usage
3. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.


logic/src/commonTest/kotlin/com/wire/kalium/logic/data/sync/SlowSyncRepositoryTest.kt
```diff
@@ -43,7 +42,7 @@ class SlowSyncRepositoryTest {
     }
 
     // TODO: Re-enable once we can update Turbine to 0.11.0+ (Requires Kotlin 1.6.21+)
-    @Ignore
+//     @Ignore
```
MohamadJaara, line 45: we can remove the Ignore 🚀

yamilmedina:
Question, Should we merge it after code freeze, right ?

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/sync/SlowSyncRepositoryTest.kt
vitorhugods: Oooh yeah. I was commenting out just to test it and forgot to remove 😄

vitorhugods:
Adding the `DO NOT MERGE` as we need to address the fact that `Verify DB Migrations` is taking more than 1h now.

PR #1218, opened 2022-12-08
feat: group read receipts - part 2 - handle websocket event for receipt mode update

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-2680 - Notify contacts that I've read their messages](https://wearezeta.atlassian.net/browse/AR-2680)

### Issues

When the group settings for read receipts was changed remotely, it wasn't being reflected on the app.

### Causes (Optional)

It was not implemented

### Solutions

Add handling of Read Receipt Mode event.

### Dependencies (Optional)

Needs releases with:

- [X] #1217


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ReceiptModeMapper.kt
```diff
@@ -0,0 +1,41 @@
+package com.wire.kalium.logic.data.conversation
+
+import com.wire.kalium.logic.data.id.IdMapper
+import com.wire.kalium.logic.di.MapperProvider
+import com.wire.kalium.network.api.base.authenticated.conversation.ReceiptMode
+import com.wire.kalium.persistence.dao.ConversationEntity
+
+interface ReceiptModeMapper {
+    fun toDaoModel(receiptMode: Conversation.ReceiptMode?): ConversationEntity.ReceiptMode
```
MohammedMokresh, line 9: I think in the mappers it's better if we don't map nullable, and just check and handle the null before we map the values, as the mappers better to be more accurate and not have null or else branches, so it will return what is expected.

borichellow (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ReceiptModeMapper.kt
alexandreferris: It has nullable as we are handling from what we receive from a Message, normally if the message is from a 1:1 convo it will come the `expects_read_confirmation` value, but if it is from a group conversation it may or may not come, as it shouldn't be considered and we are already handling the value of messages from group conversation somewhere else 🤔

MohammedMokresh (approved):
LGTM

PR #1255, opened 2022-12-20
fix(notifications): set a default time for lastNotifiedMessageDate for convresations

mchenani:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?
Having the default null value for lastNotifiedMessageDate in conversations is causing some problem with some edge cases.
We set here the current time when we insert the conversation in the database as the lastNotifiedMessageDate.
### Issues

Some edge cases bugs on notifications.

### Solutions

Setting a default value.

### Dependencies (Optional)

Needs to update tests cases on AR to pass with new non-null enforcement.

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
```diff
@@ -226,7 +227,7 @@ internal class ConnectionDataSource(
                         teamId = null,
                         protocolInfo = ConversationEntity.ProtocolInfo.Proteus,
                         creatorId = connection.from,
-                        lastNotificationDate = null,
+                        lastNotificationDate = Clock.System.now().toString(),
```
saleniuk, line 230: Please do not use regular `.toString()`, it omits milliseconds in the string if they are equal to 0, that's why I had to do this change: https://github.com/wireapp/kalium/pull/1258

mchenani: Awesome, thanks. So I'll wait for you RP to get merged and then I'll update my PR.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
```diff
@@ -226,7 +227,7 @@ internal class ConnectionDataSource(
                         teamId = null,
                         protocolInfo = ConversationEntity.ProtocolInfo.Proteus,
                         creatorId = connection.from,
-                        lastNotificationDate = null,
+                        lastNotificationDate = DateTimeUtil.currentIsoDateTimeString(),
```
saleniuk, line 230: Now I'm not sure, won't that create similar issue as the one Vitor debugged last night? For connections maybe we can use `lastUpdate` field?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationGroupRepository.kt
```diff
@@ -68,7 +69,10 @@ internal class ConversationGroupRepositoryImpl(
             .flatMap { conversationResponse ->
                 val teamId = selfUser.teamId
                 val conversationEntity = conversationMapper.fromApiModelToDaoModel(
-                    conversationResponse, mlsGroupState = ConversationEntity.GroupState.PENDING_CREATION, teamId
+                    conversationResponse,
+                    mlsGroupState = ConversationEntity.GroupState.PENDING_CREATION,
+                    teamId,
+                    DateTimeUtil.currentIsoDateTimeString()
```
saleniuk, line 75: Same here, we should use date-time from the server, here for conversations maybe we can use `lastEventTime`?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -253,6 +254,7 @@ internal class ConversationDataSource internal constructor(
                         )
                     },
                     selfTeamIdProvider().getOrNull(),
+                    lastNotificationDate = DateTimeUtil.currentIsoDateTimeString()
```
saleniuk, line 257: Maybe we shouldn't set it here at all, because we use that `persistConversations` when syncing, right? So during every sync we will update this `lastNotificationDate` to the current device time and won't get notified about messages that happened before that time, or maybe I'm missing something 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
```diff
@@ -182,7 +183,7 @@ sealed class ConversationDetails(open val conversation: Conversation) {
             protocol = protocolInfo,
             mutedStatus = MutedConversationStatus.AllAllowed,
             removedBy = null,
-            lastNotificationDate = null,
+            lastNotificationDate = DateTimeUtil.currentIsoDateTimeString(),
```
saleniuk, line 186: We shouldn't rely on local device's time, maybe for connection it should be `lastModifiedDate` or even just don't set a default value at all 🤔

saleniuk (commented):
I think @vitorhugods should take a look as he's working on a fix for the notification timing issue, but I think we shouldn't use the device's local time, maybe for each case we can find a proper date-time field that's returned by server?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
mchenani: we don't have access to that field! this is for the conversation that we don't have anything for it and instead of null we set the current time to it.

vitorhugods:
> I think @vitorhugods should take a look as he's working on a fix for the notification timing issue, but I think we shouldn't use the device's local time, maybe for each case we can find a proper date-time field that's returned by server?

I guess we could just use `Instant.DISTANT_PAST` as well. This way we avoid unnecessary null checks?

PR #1256, opened 2022-12-20
chore: simplify makefile for faster builds and reduced downloads/checkouts

comawill:
This PR intends to simplify the Makefile (yet with a lot of duplication)
* Make sure that the actual compiled libraries are used
* don't try to install rust
* cache downloaded sources to reduce turnover when something fails
* use make dependencies to allow potential parallel builds


scripts/bootstrap.mk
```diff
@@ -15,104 +16,121 @@ ifeq ($(OS),darwin)
...
+AVS_FRAMEWORK_SRC := $(AVS_FRAMEWORK_UNZIP)/Carthage/Build/iOS/avs.framework
+AVS_FRAMEWORK_ARTIFACT := $(NATIVE_LIBS)/$(AVS_ARTIFACT_FILE)
+
+all: $(CRYPTOBOX_C_ARTIFACT) $(LIBSODIUM_ARTIFACT) $(CRYPTOBOX4J_ARTIFACT) $(AVS_FRAMEWORK_ARTIFACT)
 
 .PHONY: install-rust
 install-rust:
-	@curl https://sh.rustup.rs -sSf | sh -s -- -y
-	@source "${HOME}/.cargo/env"
+	which rustc
```
mythsunwind, line 58: I would rename the target from `install-rust` into `check-rust` or something

tmspzz (changes requested):
Nice improvement. Could we keep a way to build from source tho please?

mythsunwind (approved):

yamilmedina (approved):
Thanks for the collab, loved the changes!

tmspzz:
> > Nice improvement. Could we keep a way to build from source tho please?
> 
> I'm not sure if I understand your request wrong, it still builds from source. It just uses curl to download the released code instead of git clone/checkout

My bad. I though it was downloading the binaries. Why is this better than `git clone --depth 1` ?

scripts/bootstrap.mk
```diff
@@ -15,104 +16,115 @@ ifeq ($(OS),darwin)
...
+
+CRYPTOBOX4J_URL = https://github.com/wireapp/cryptobox4j/archive/refs/tags/$(CRYPTOBOX4J_VERSION).tar.gz
+CRYPTOBOX4J_TAR_GZ :=  $(NATIVE_TARBALLS)/cryptobox4j_$(CRYPTOBOX4J_VERSION).tar.gz
+CRYPTOBOX4J_CODE := native/cryptobox4j_$(CRYPTOBOX4J_VERSION)
+CRYPTOBOX4J_ARTIFACT= $(NATIVE_LIBS)/$(LIBCRYPTOBOX_JNI_ARTIFACT_FILE)
+
+AVS_FRAMEWORK_URL := https://github.com/wireapp/wire-avs/releases/download/$(AVS_VERSION)/avs.framework.osx.$(AVS_VERSION).zip
+AVS_FRAMEWORK_ZIP := $(NATIVE_TARBALLS)/avs.framework.osx.$(AVS_VERSION).zip
+AVS_FRAMEWORK_UNZIP := native/avs.framework_$(AVS_VERSION)
+AVS_FRAMEWORK_SRC := $(AVS_FRAMEWORK_UNZIP)/Carthage/Build/iOS/avs.framework
```
tmspzz, line 51: this is not source code. This is is a binary, probably the name should reflect that.

tmspzz (changes requested):

scripts/bootstrap.mk
comawill: It was here more meant in the sense of source=origin (`cp $src $dst`), still I see the potential for confusion so I changed it to `LOCATION` to reflect it a bit more

comawill:
> My bad. I though it was downloading the binaries. Why is this better than `git clone --depth 1` @comawill ?

It allows to benefit of the features from "make", as we can have some state, that is observable by make. 
The Makefile tracks now 3 states:
1. did I download the source
2. was the source unpacked
3. is the required library build

and all its dependencies between the different steps and the libraries.
Running `make all` twice will correctly determine that everything was successfully built, or if you delete one specific library only that library will be rebuild

Resetting a downloaded and unpacked source is much easier (`rm -r $LIB_CODE`), than trying to reset a git to a specific tag without deleting it completely.

Additionally the Makefile now no longer depends on git and could also be prefilled with the required tarballs and then allowing a full offline build.

tmspzz (approved):

PR #1258, opened 2022-12-21
chore: fix and unify parsing and getting iso date-time strings [AR-2851]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In multiple places we get the current time as iso string by using `Clock.System.now().toString()` because `Clock.System.now()` gives us `Instant` and `instant.toString()` returns the time in iso format string, but actually when we have a time with milliseconds equal to exact 0 (.000) then `.toString()` just ignores milliseconds and returns "2022-12-19T17:36:00Z", then if we try to parse it using `SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ")`, which we do for instance in `fun String.toTimeInMillis()`, we get `ParseException`.
Also, we use multiple ways of getting the current time or parsing it.

### Solutions

Implement `DateTimeUtil` which contains all required methods related to date-times and is a single source of this kind of data.

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/androidMain/kotlin/com/wire/kalium/logic/util/PlatformDateTimeUtil.kt
```diff
@@ -0,0 +1,35 @@
...
+import androidx.annotation.RequiresApi
+import kotlinx.datetime.Instant
+import kotlinx.datetime.toJavaInstant
+import java.text.SimpleDateFormat
+import java.time.format.DateTimeFormatter
+import java.util.Date
+import java.util.Locale
+import java.util.TimeZone
+
+actual open class PlatformDateTimeUtil actual constructor() {
```
MohamadJaara, line 13: maybe PlatformDateTimeUtil can be moved to the Util module just in case it is needed somewhere outside logic

saleniuk: That's why I'm trying to do right now and it's still WIP, but thanks for writing that, now I'm more sure that it should be this way 😄

util/build.gradle.kts
```diff
@@ -23,5 +24,11 @@ kotlin {
                 implementation(libs.coroutines.test)
             }
         }
+        val androidMain by getting {
```
MohamadJaara, line 27: are paging3 and WorkManager are needed in util ?

MohamadJaara (commented):
great work 🚀, just a question about adding paging3 and work manager  dependencies to Util module

util/build.gradle.kts
saleniuk: oh sorry, nice catch! I copied that, then I added `annotation` library and it's pushed already but somehow I forgot to change it also here 🙈

MohamadJaara (approved):
LGTM 🚀

PR #1274, opened 2022-12-23
fix: messages randomly not triggering notifications (AR-2867)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When receiving messages with little interval between them, not all messages trigger a notification.

### Causes

When we display a notification for a conversation, we call `markMessagesAsNotified(conversationId, DateTimeUtil.currentIsoDateTimeString())`.

What if the `currentIsoDate` is off by a few seconds in the future compared to the backend's clock? The messages that come from the server within this timespan in the next seconds will NOT trigger a notification.

### Solutions

Don't allow passing some arbitrary datetime, use the datetime of the last received message instead.

#### Test Coverage

### How To Test

I've tried the following experiments:

#### Change the phone's clock to a few minutes in the future

If you advance the device's clock on purpose, so it is 1min in "the future" compared to the server's time. After a notification is received, you should be able to receive another one quickly after. _i.e._ the device's clock should NOT interfere with notifications.

Before the fix, when doing this experiment with 5 minutes, I wasn't able to receive more notifications in the same conversation for 5 minutes.

#### Send messages in quick succession

I attempted to send around 1 message per second, with the contents: `1, 2, 3, 4, 5, 6, 7, 8, 9, 0`.

I was almost always receiving only 1 and 8 before the fix. Sometimes 1 and a 7.

Makes me believe that my emulator and the server are off by around 7 seconds.

Videos documenting it:

https://user-images.githubusercontent.com/9389043/209250161-63a0b425-38c9-4e69-aad4-38281b8be8bc.mp4

https://user-images.githubusercontent.com/9389043/209250194-e1ec1f9f-0dde-4a5c-baa1-136096329829.mp4 

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
@@ -18,6 +18,7 @@ import com.wire.kalium.logic.feature.message.MessageTarget
 import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.logic.functional.flatMap
 import com.wire.kalium.logic.functional.fold
+import com.wire.kalium.logic.functional.map
 import com.wire.kalium.logic.functional.onFailure
 import com.wire.kalium.logic.functional.onSuccess
 import com.wire.kalium.logic.wrapApiRequest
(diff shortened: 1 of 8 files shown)


mchenani:
One of the topics @MohamadJaara and I discussed couple of times, it that should be marked as notified from the Notification manager or we should mark it as soon as the message fetched from the db?
- A: Fetch the messages need to be notified, notify them then the notification manager decides to mark them as notified!
- B: Fetch the messages need to be notified, we mark them as notified then the notification manager don't need to be responsible for marking them as notified.

vitorhugods:
> One of the topics @MohamadJaara and I discussed couple of times, it that should be marked as notified from the Notification manager or we should mark it as soon as the message fetched from the db?
> 
> * A: Fetch the messages need to be notified, notify them then the notification manager decides to mark them as notified!
> * B: Fetch the messages need to be notified, we mark them as notified then the notification manager don't need to be responsible for marking them as notified.

**B** is adding unexpected side-effects and it cements the notification logic in a way that the apps are unable to customise. It would be similar as getting messages and marking them as read. If the app decides to change its notification logic, or just add some extra layer of logic on top, **B** makes it impossible.

I don't see any issues with **A**.

ohassine (approved):
Confirming the fix 👌 Nicely done ✨

mchenani:
> 
Yes, that's the result we came with, just wanted to share it here and have it on this PR as well.
Mixing logic + getting the notifications should only get them not mark them as notified, then as you said it's the app responsibility to mark them as notified as soon as the criteria is met to not show those messages again.

PR #1291, opened 2023-01-02
ci: speed up CI builds

vitorhugods:
By upgrading the JDK version to 17, enabling the parallel GC and caching dependencies between builds.

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Our CI is sloooooow.

### Causes

- We do _not_ cache gradle dependencies between builds for all workflows.
- We're using JDK 11
- We're not using parallel garbage collector

#### Caching
The [setup-java](https://github.com/actions/setup-java#caching-packages-dependencies) action is able to cache dependencies between builds. And the cache is smart enough to consider the hash of all important gradle files, so if the hash changes, the cache won't be used.

We can easily enable it by adding `cache: 'gradle'` to all missing workflows.

We had issues in the past and disabled the cache, mostly because this smart hashing thing wasn't in place and we were using transient dependencies early in development.

Also, it's easier to clear caches now, just going to [Actions/Cache](https://github.com/wireapp/kalium/actions/caches).

#### JDK upgrade

Since Gradle 7.3, JDK 17 is supported. It's not something we had in the beginning of the project, but it's a quick win now.

Adopt JDK is not a thing anymore, so I just opted for Azul's Zulu. We could maybe benchmark it and see if there's a significant difference between Zulu and Temurin and maybe be more scientific about it, but not today :)

#### Bonus

[Parallel Garbage Collector](https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/parallel.html) is now enabled.

According to [Bitrise's analysis](https://bitrise.io/blog/post/switch-to-jdk-17-parallel-gc-and-speed-up-your-android-builds-by-9-20), the combo of upgrading to JDK 17 + enabling the Parallel GC resulted in improvements from 9% to 20% in their benchmarks using three different projects.

### Testing

As this build changed the `gradle.properties` file, all tests should run.

### Notes

The JDK switch + dependency caching should only affect CI.

The parallel GC affects our local builds, but it's compatible with JDK 11 that some developers are using on their local machines.

----

----
##### References

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-org.gradle.jvmargs=-Xmx2g
+org.gradle.jvmargs=-Xmx2g -XX:+UseParallelGC
 android.useAndroidX=true
 kotlin.native.binary.memoryModel=experimental
 kotlin.js.generate.executable.default=false
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -15,8 +15,8 @@ jobs:
       - name: Set up JDK 11
         uses: actions/setup-java@v3
         with:
-          java-version: '11'
-          distribution: 'adopt'
+          java-version: '17'
+          distribution: 'zulu'
           cache: gradle
       - name: Validate Gradle wrapper
         uses: gradle/wrapper-validation-action@e6e38bacfdf1a337459f332974bb2327a31aaf4b
--- a/.github/workflows/generate-dokka.yml
+++ b/.github/workflows/generate-dokka.yml
@@ -17,8 +17,8 @@ jobs:
             - name: Set up JDK 11
               uses: actions/setup-java@v3
               with:
-                java-version: '11'
-                distribution: 'adopt'
+                java-version: '17'
+                distribution: 'zulu'
                 cache: gradle
 
             - name: Validate Gradle wrapper
--- a/.github/workflows/gradle-android-tests.yml
+++ b/.github/workflows/gradle-android-tests.yml
@@ -24,8 +24,9 @@ jobs:
             - name: Set up JDK 11
               uses: actions/setup-java@v3
               with:
-                  java-version: '11'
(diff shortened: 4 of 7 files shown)


vitorhugods:
I will wait for this current build to finish, and then trigger `update branch` to see if the cache is being used for all workflows and no dependencies are re-downloaded.

gradle.properties
```diff
@@ -1,4 +1,4 @@
-org.gradle.jvmargs=-Xmx2g
+org.gradle.jvmargs=-Xmx2g -XX:+UseParallelGC
```
gongracr, line 1: Should we replicate the same on our `local.properties` or would this one override the local one?

gongracr (approved):
LGTM 🍨 Hoping that this can save us many hours of development! 🤞

ohassine (approved):

vitorhugods:
For a comparison:

Android's `build the samples` step was taking around 10 minutes, as seen in [this build](https://github.com/wireapp/kalium/actions/runs/3823037845/jobs/6503781519) from a few minutes ago.

Now, [it takes around 3min40s](https://github.com/wireapp/kalium/actions/runs/3823488683/jobs/6504714133), plus the added 50s to grab the build cache.

Most of the improvement is cache related, no doubt. It's a bit tricky to compare the JDK change, but in any case, it's a step away from deprecation and towards modern tools with long-term-support anyway.

gradle.properties
vitorhugods: I'm not sure if `org.gradle.jvmargs` inside `local.properties` actually does anything 🤔.

`local.properties` is a Android Gradle Plugin thing, shouldn't affect "non-android-specific" things, unless added by other parties.

Like we did with the GitHub token thingy. We explicitly read from that file. But Gradle doesn't AFAIK.

In any case, this `global.properties` change should affect all developers.

PR #1294, opened 2023-01-03
chore: move detekt to external project and migrate config to plugin (AR-2904)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Migrate detekt rules to an external project https://github.com/wireapp/wire-detekt-rules for later reusage of rules on reloaded side. This PR:

- Removes custom defined `detekt-rules` module
- Migrates `Detekt` to be used as Kotlin DSL plugin instead of the old groovy config
- Configures the necessary boilerplate to use the artifact of https://github.com/wireapp/wire-detekt-rules where now `Detekt` rules exist

### Solutions

Detekt custom rules are flexible enough to be packaged as a dependency/lib, so we are using this approach to share/reuse custom rules in Kotlin codebases (at the moment Kalium and Reloaded). Also, since we may not want to mess up with artifact publishing right now, we are leveraging in Gradle system + ivy repositories flexibility to use the jar as a dependency.

Note: Also took advantage of this effort and added global ignore of files and classes, instead of ignoring rule by rule, as suggested in https://detekt.dev/docs/gettingstarted/gradle/#kotlin-dsl-5

----

----

----
##### References

[SQPIT-764]: https://wearezeta.atlassian.net/browse/SQPIT-764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


buildSrc/src/main/kotlin/scripts/detekt.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+configurations {
+    all {
+        resolutionStrategy {
+            eachDependency {
+                if (requested.group == "com.wire" && requested.name == "detekt-rules") {
+                    // enforce snapshot version check for updates everytime instead of 24 hrs.
+                    cacheChangingModulesFor(0, "SECONDS")
```
vitorhugods, line 26: Doesn't that prevent people from working offline?

vitorhugods (commented):
I really like the idea, just have a couple of concerns.

Changing the rules repository. We need a way to test locally before pushing new builds. Otherwise, someone changing stuff in the rules just to test something could be breaking stuff in other repositories, both for CI and developers.

I think this can be addressed with something like adding `mavenLocal()` and `publishToMavenLocal` task.
So we can work on the rulesets, publish to local, and then enable `mavenLocal()` on Reloaded/Kalium to test it out.

buildSrc/src/main/kotlin/scripts/detekt.gradle.kts
yamilmedina: Yup, that's correct. I removed it, so we get the default cache behavior of 24 hrs. If we need to test faster because of changes, we can use `mavenLocal` as suggested and also added

yamilmedina: I will add in the rules repo now the complementary part:
- publish plugin to publish to local
- automation to tag versions

yamilmedina: I added to https://github.com/wireapp/wire-detekt-rules : 
- the publish plugin, so we can test locally
- the gh action to update and tag an updated artifact

:D

vitorhugods (approved):
👨‍🍳👌

Super nice

PR #1297, opened 2023-01-04
fix: sender userid in preview message [AR-2895]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Added `senderUserId` and `selfUserId` to differentiate in UI when the user left the conversation by himself or was removed by another user

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
```diff
@@ -344,7 +344,9 @@ data class MessagePreview(
     val content: MessagePreviewContent,
     val date: String,
     val visibility: Message.Visibility,
-    val isSelfMessage: Boolean
+    val isSelfMessage: Boolean,
+    val senderUserId: UserId,
+    val selfUserId: UserId?
```
saleniuk, line 349: If we pass both `senderUserId` and `selfUserId` then `isSelfMessage` is just `senderUserId == selfUserId`, right? So it's redundant 😄  
I'm not sure if we should add `selfUserId` here to `MessagePreview` because it's not actually a part of message data, especially if we use that only for `MemberChange` system messages. 
For regular messages, list of user ids is passed so it's exactly like now in `MessagePreviewEntityContent.MemberChange` and the details of each user for UI are taken using `ObserveUserListByIdUseCase`, so you can get all user names that were added/removed and check whether any of them is the `selfUserId`, because from this use case we get list of `User` which can be either of type `OtherUser` or `SelfUser`. If we don't need the names and all of that, then maybe it would still be better to pass `selfUserId` later in AR since it's only to display the text properly, what do you think? 😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageMapper.kt
```diff
@@ -371,10 +371,15 @@ private fun MessagePreviewEntityContent.toMessageContent(): MessagePreviewConten
     is MessagePreviewEntityContent.ConversationNameChange -> MessagePreviewContent.WithUser.ConversationNameChange(adminName)
     is MessagePreviewEntityContent.Knock -> MessagePreviewContent.WithUser.Knock(senderName)
     is MessagePreviewEntityContent.MemberChange -> when (type) {
-        MessageEntity.MemberChangeType.ADDED -> MessagePreviewContent.WithUser.MembersAdded(adminName = adminName, count = count)
-        MessageEntity.MemberChangeType.REMOVED -> MessagePreviewContent.WithUser.MembersRemoved(adminName = adminName, count = count)
+        MessageEntity.MemberChangeType.ADDED -> MessagePreviewContent.WithUser.MembersAdded(
+            adminName = adminName,
```
saleniuk, line 375: Is `adminName` the name of the person who added/removed someone? Maybe we could use better name here as conversation can have multiple admins, also if someone who is not an admin leaves, this system message type is also created even if he's not an admin 😉

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageEntity.kt
```diff
@@ -251,7 +252,9 @@ data class MessagePreviewEntity(
     val content: MessagePreviewEntityContent,
     val date: String,
     val visibility: MessageEntity.Visibility,
-    val isSelfMessage: Boolean
+    val isSelfMessage: Boolean,
+    val senderUserId: QualifiedIDEntity,
+    val selfUserId: UserIDEntity?,
```
MohamadJaara, line 257: I'm not sure about the benefits of including self user ID in every message when it will be always the same for all messages.
The check is self removed was removed/added is needed only for the UI to display the correct copy; therefore it can be left to the UI to do this comparison

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
Garzas: I changed to have this information already in entity class

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageMapper.kt
Garzas: Renamed to sender

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageEntity.kt
Garzas: Removed 😄

saleniuk (approved):

PR #1302, opened 2023-01-04
fix(dev): ghost calls after restoring backup from other client [AR-2893]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When we create a backup while being in a call, log out, clear data, log in and restore it, while the call is still ongoing, we end up alone in a call which can’t be hanged up, just continues forever.

### Causes (Optional)

When we clear data and log in again, the app has to register a new client so calls intended for a previous client won't work anymore.

### Solutions

Check if the backup we want to restore is from another client, if so then do not restore calls table because all calls there are not meant for the client that we currently try to do it on.

Needs releases with:

- https://github.com/wireapp/kalium/pull/1301

#### Test Coverage (Optional)

#### How to Test

STR:
- be in an active call
- create a backup while being in the call without password
- logout of the app, clear data
- login again
- restore the backup
- original call shouldn't be ongoing anymore

----

----

----
##### References

[SQPIT-764]: https://wearezeta.atlassian.net/browse/SQPIT-764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SQPIT-764]: https://wearezeta.atlassian.net/browse/SQPIT-764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ohassine:
If we ignore call table then the user will not see missed calls that had on the other device. Is that fine ?

saleniuk:
> If we ignore call table then the user will not see missed calls that had on the other device. Is that fine ?

For every missed call we persist a `SystemMessage.MissedCall` and we show it this way on conversation screen, so we have info about missed calls in `Message` table. 😃

ohassine (approved):
Looks good to me 👌 ✅

PR #1304, opened 2023-01-05
fix: allow receiving of null receipt modes

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently, we allow creation of DTOs with null read receipts. Whilst it makes sense because we can receive that from the backend, it also allows us sending `null` to the backend.

We should reduce this permissiveness so it only works when coming in without allowing the creation of nullable stuff internally within Kalium.

### Solutions

Remove the nullables and make it so the custom Serializer accepts nullable integers, but we convert that to `Disabled` inside of it.

### Testing

Some tests for handling of nullability can actually be removed now.

#### Test Coverage

----

----
##### References


alexandreferris (approved):
👌🏻👌🏻

PR #1309, opened 2023-01-06
feat: Group Conversation - Receipt Mode System Message

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-2712 - Show a system message when read receipts are turned on](https://wearezeta.atlassian.net/browse/AR-2712)

This PR is only for handling receipt mode when a new group event is received or in-app user creates a group.
The handling of receipt mode from changes in group settings screen will be handled in a later PR.

### Issues

New group conversations [from event or user in-app created] were not adding a system message for receipt mode.

### Causes (Optional)

It was not implemented.

### Solutions

Add implementation to always add a system message informing the members of the conversation what is the receipt mode status if the creator/receiver of the new group event matches the criteria of:
- isSelfATeamMember
- conversation is of GROUP type
- receipt mode status is NOT NULL (for users without a team it will be null when invoking `CreateGroupConversationUseCase` thus not showing for them)

### Dependencies (Optional)

A PR in AR will be raised upon this one being merged

#### How to Test

Currently the only way to test this in Kalium is to build this branch in AR and create/receive a new group event and then check `MessageNewConversationReceiptModeContent` table values that is contains the selected receipt mode.


logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/conversation/CreateGroupConversationUseCaseTest.kt
```diff
@@ -137,6 +149,40 @@ class CreateGroupConversationUseCaseTest {
             .wasInvoked(exactly = once)
     }
 
+    @Test
+    fun givenNameMembersAndOptions_whenCreatingGroupConversation_thenPersistSystemMessageForReceiptMode() = runTest {
```
vitorhugods, line 153: Could you also test for when the user is not a team member, so it doesn't insert the system message?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/CreateGroupConversationUseCase.kt
```diff
@@ -20,11 +25,15 @@ import com.wire.kalium.util.DateTimeUtil
...
  */
+@Suppress("LongParameterList")
 class CreateGroupConversationUseCase internal constructor(
     private val conversationRepository: ConversationRepository,
     private val conversationGroupRepository: ConversationGroupRepository,
     private val syncManager: SyncManager,
-    private val currentClientIdProvider: CurrentClientIdProvider
+    private val currentClientIdProvider: CurrentClientIdProvider,
+    private val selfUserId: UserId,
+    val persistMessage: PersistMessageUseCase,
```
vitorhugods, line 35:
```suggestion
    private val persistMessage: PersistMessageUseCase,
```

If necessary, please make `PersistMessageUseCase` internal as well.

It shouldn't be public in the first place, and from what I've seen, making it `internal` won't cause any breaking changes

vitorhugods (changes requested):
Great work!

Just a couple of tiny changes.

vitorhugods (approved):
👨‍🍳🤌

ohassine (approved):
LGTM 👌 🚀 just before merging, could you please check that AR can be compiled with these changes.

alexandreferris:
> LGTM 👌 🚀 just before merging, could you please check that AR can be compiled with these changes.

AR will break but I already have a PR to be opened and reviewed so it doesn't stay broken for too long

PR #1318, opened 2023-01-10
fix(notification): limit notifications to 10 messages when fetching messages from repo

mchenani:
----

- The **PR Title**

- The **PR Description**
----
### Issues
One ques if you have lot of unread and not notified messages experience slowness and ANR in the app.

### Causes (Optional)
When we querying the not notified messages for the user, there is no size limit how many messages should be fetched.
### Solutions

Final and best Solution[not implemented here - blocked by sqldelight]
We can use window functions in sql to filter the messages based on conversations id in the query; but at the moment sqldelight has an issue with `OVER` and `PARTITION` keywords
See: https://github.com/cashapp/sqldelight/issues/2799

Solution in this PR:
Filtering the messages after fetching them from the db in the Repo.

#### Test Coverage (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
@@ -188,8 +188,7 @@ class MessageDataSource(
                     // todo: needs some clean up!
                     id = conversationId.toModel(),
                     conversationName = messages.first().conversationName ?: "",
-                    messages = messages.sortedBy { message -> message.date }
-                        .take(messageSizePerConversation)
+                    messages = messages.take(messageSizePerConversation)
                         .map { message -> messageMapper.fromMessageToLocalNotificationMessage(message) },
                     isOneToOneConversation = messages.first().conversationType?.let { type ->
                         type == ConversationEntity.Type.ONE_ON_ONE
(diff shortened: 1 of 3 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
```diff
@@ -187,7 +188,9 @@ class MessageDataSource(
                     // todo: needs some clean up!
                     id = conversationId.toModel(),
                     conversationName = messages.first().conversationName ?: "",
-                    messages = messages.map { message -> messageMapper.fromMessageToLocalNotificationMessage(message) },
+                    messages = messages.sortedBy { message -> message.date }
```
vitorhugods, line 191: Can't we sort in the SQL-level?

vitorhugods (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
mchenani: but still not sure, if we sort them in total, and then filter them based on the conversationId in kotlin, the result would be sorted or not! I'll check that.

vitorhugods: According to Kotlin's [groupBy](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/group-by.html), it should keep the order it was in:

> The returned map preserves the entry iteration order of the keys produced from the original array.

So sorting them as a whole and then grouping by conversation Id should already result in the correct result.

Just gotta be careful to fetch the newest first, and then we only care about the newest 10.

PR #1327, opened 2023-01-12
fix: incoming call keeps showing up even after rejecting it (AR-2934)

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Call keeps popping up many times even after rejecting it.

 STEPS TO REPRODUCE
- Login with the same account on two different devices ( one of them AR)
- When you receive a call, reject or answer it on the other platform, not Reloaded
- You should see the incoming call screen appearing multiple times (randomly) 
You will be able to reproduce it on old accounts, not fresh installs

### Causes (Optional)

The root cause is that call status is not changed in database and stays on INCOMING state.

#### Why?
When we receive a call we create a call object with `INCOMING` state and we save it on database. If the call got answered reject on other platform then we update the status to `CLOSED`.

On old accounts, due some performance issues, the flow is inverted:
 - The call to update is invoked first(You already have some closed calls with that person) through `OnCloseCall` callback, so it passes
 - Then create call object is called next with `INCOMING` state through `OnIncomingCall` callback

Since `OnCloseCall` already happened and there is the only place where we update the status to CLOSED
--> we have now  a call saved in database with `INCOMING` state that will never change. And at this time incoming screen will start popping up ⬆️ 

### Solutions

This PR will not fix the issue of inverted calls where `UPDATE` called before `CREATE`. This needs some refactoring.

In this PR I added a new status called `CLOSED_INTERNALLY`  that will be used when the user terminates the ongoing call.
Also, I used the REJECTED status when the user rejects an incoming call.

This way even that issue happens, we have a way to stop it.

#### Test Coverage (Optional)

#### How to Test

Steps Above 👆 

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/RejectCallUseCase.kt
```diff
@@ -1,14 +1,27 @@
...
-class RejectCallUseCase(private val callManager: Lazy<CallManager>) {
+class RejectCallUseCase(
+    private val callManager: Lazy<CallManager>,
+    private val callRepository: CallRepository,
+    private val dispatchers: KaliumDispatcher = KaliumDispatcherImpl
+) {
+
+    suspend operator fun invoke(conversationId: ConversationId) = withContext(dispatchers.io) {
+        callingLogger.d("[RejectCallUseCase] -> Updating call status to CLOSED_INTERNALLY")
+        callRepository.updateCallStatusById(conversationId.toString(), CallStatus.REJECTED)
```
alexandreferris, line 23:
```suggestion
        callingLogger.d("[RejectCallUseCase] -> Updating call status to REJECTED")
        callRepository.updateCallStatusById(conversationId.toString(), CallStatus.REJECTED)
```

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/call/EndCallUseCaseTest.kt
```diff
@@ -55,5 +64,65 @@ class EndCallUseCaseTest {
...
+
+        verify(callRepository)
+            .function(callRepository::updateIsCameraOnById)
+            .with(eq(conversationId.toString()), eq(false))
+            .wasInvoked(once)
+
+        verify(callRepository)
+            .suspendFunction(callRepository::persistMissedCall)
+            .with(any())
+            .wasNotInvoked()
```
alexandreferris, line 110: Shouldn't this be invoked per name of the test? 🤔

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/call/EndCallUseCaseTest.kt
```diff
@@ -55,5 +64,65 @@ class EndCallUseCaseTest {
...
+    @Test
+    fun givenNonEstablishedCall_whenEndCallIsInvoked_thenSaveMissedCall() = runTest {
+        given(callRepository)
+            .suspendFunction(callRepository::persistMissedCall)
+            .whenInvokedWith(eq(conversationId))
+            .thenDoNothing()
+
+        given(callRepository)
+            .suspendFunction(callRepository::establishedCallsFlow)
+            .whenInvoked().then { flowOf(listOf(call.copy(status = CallStatus.INCOMING))) }
```
alexandreferris, line 83: `establishedCallsFlow` returns only calls with `ESTABLISHED` or `ANSWERED` status, for incoming calls we should use `::incomingCallsFlow`

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/call/RejectCallUseCaseTest.kt
```diff
@@ -36,12 +41,22 @@ class RejectCallUseCaseTest {
...
         rejectCallUseCase.invoke(conversationId)
 
         verify(callManager)
             .suspendFunction(callManager::rejectCall)
             .with(eq(conversationId))
             .wasInvoked(once)
+
+        verify(callRepository)
+            .suspendFunction(callRepository::updateCallStatusById)
+            .with(eq(conversationId.toString()),  eq(CallStatus.REJECTED))
```
alexandreferris, line 58:
```suggestion
            .with(eq(conversationId.toString()), eq(CallStatus.REJECTED))
```

Just some unnecessary extra space here

alexandreferris (commented):
Great work addressing this issue, I do have some comments so let me know if something doesn't make sense

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/call/EndCallUseCaseTest.kt
ohassine: Yeah right . I will update it to return empty list

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/call/EndCallUseCaseTest.kt
alexandreferris: the name of the test is: `givenNonEstablishedCall_whenEndCallIsInvoked_thenSaveMissedCall`, so shouldn't `callRepository::persistMissedCall` be invoked at least once here?

ohassine: Oh! yes good catch

alexandreferris (approved):
Great work here!

Something we need to keep in mind now (doesn't need to be done in this PR), but to update:

`updateLastCallStatusByConversationId` and `updateOpenCallsToClosedStatus` to maybe add the status `REJECTED`.

PR #1340, opened 2023-01-16
fix: Observe notifications only when Live

borichellow:
# What's new in this PR?

### Issues

When user opens the app after long offline-period (and missed A LOT of messages), then notifications spams him/her while app syncing.

### Causes (Optional)

Observing the notifications starts when the app opened. So in case of tons of missed messages user receives 1 notification for each new message => spaming

### Solutions

Start observing the notifications only when sync is finished, so user will get 1 huge notification, instead of tons of small.


logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCaseTest.kt
```diff
@@ -34,16 +34,41 @@ import io.mockative.mock
...
+    fun givenSyncStateIsPending_thenOnlyEphemeralNotificationsObserved() = runTest {
+        val (arrange, getNotifications) = Arrangement()
+            .withEphemeralNotification()
+            .withIncrementalSyncState(IncrementalSyncStatus.FetchingPendingEvents)
+            .withConnectionList(listOf())
+            .withConversationsForNotifications(listOf())
+            .arrange()
+
+        getNotifications().test {
+            awaitComplete()
```
vitorhugods, line 56: Will it ever finish?

`GetNotificationsUseCase` returns an endless flow that never completes.
I think we should await the first result is emitted, and then cancel the collection (to cancel the flow).

vitorhugods (changes requested):
Not sure about the test, but overall nice solution.

vitorhugods (approved):
👨‍🍳👌

PR #1342, opened 2023-01-17
refactor: use timestamp to store conversation and message dates (AR-2082)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently we’re using both ISO date/time String and timestamp to store date/time information in the database.

Performance-wise, the way we're operating dates is also not optimal, as are not creating function-based indexes either.

### Causes

Bad architecture choice some months ago. Something I deeply regret.

### Solutions

Trying to avoid creating a gigantic PR, these changes reduce the usage of String in the `persistence` module, by replacing date strings and epoch seconds in the most performance-critical columns with the unix timestamp in millis.

#### Avoiding Breaking changes 

There are **no** breaking changes.
I've created a secondary constructor and deprecated-getters for affected classes to keep compatibility with `logic`.

### Testing

I've tested migration and the app manually too after these changes.

#### Test Coverage (Optional)

Not applicable

----

----
##### References


persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
```diff
@@ -69,23 +75,23 @@ WHERE mls_group_id = ?;
...
 WHERE qualified_id = ?;
 
 updateAllUnNotifiedConversationsNotificationsDate:
 UPDATE Conversation
-SET last_notified_message_date = ?
-WHERE last_notified_message_date ISNULL
-OR strftime('%Y-%m-%dT%H:%M:%fZ', last_modified_date) > strftime('%Y-%m-%dT%H:%M:%fZ', last_notified_message_date);
+SET last_notified_instant = ?
+WHERE last_modified_instant ISNULL
+OR last_modified_instant > last_notified_instant;
```
mchenani, line 85: getting rid of extra conversions 🌪️

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
```diff
@@ -140,21 +146,21 @@ CASE
...
+Conversation.mls_last_keying_material_update_instant,
 Conversation.mls_group_state,
 Conversation.access_list,
 Conversation.access_role_list,
 Conversation.team_id,
 Conversation.mls_proposal_timer,
 Conversation.muted_time,
 Conversation.creator_id,
-Conversation.last_modified_date,
+Conversation.last_modified_instant,
```
mchenani, line 163: tbh, I found the `_date` more indicates the purpose of the field and not the type! So do we need to rename all of them with `_instant` ending? I thin `_date` would be OK!

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionMapper.kt
```diff
@@ -30,7 +33,7 @@ internal class ConnectionMapperImpl(
     override fun fromApiToDao(state: ConnectionDTO): ConnectionEntity = ConnectionEntity(
         conversationId = state.conversationId,
         from = state.from,
-        lastUpdate = state.lastUpdate,
+        lastUpdateInstant = state.lastUpdate.takeIf { it.isNotBlank() }?.toInstant() ?: Instant.UNIX_FIRST_DATE,
```
mchenani, line 36: maybe as an extension fun to all instants?

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
typfel: I tend to agree with @mchenani here, with keeping `_date`. Using a consistent suffix is nice though.

typfel (commented):
Badly needed refactoring!  🏎️  🏁

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
vitorhugods: Although it works in the DB layer, what about the Entity models? 

`ConversationEntity.lastModifiedDate` still exists (but it's deprecated now), same for other date fields 🤔 

And then it becomes a bit weird, mapping the `last_modified_date` column to something else than the `lastModifiedDate` field...

I'm trying to come up with a solution here but I can't see it yet.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionMapper.kt
vitorhugods: I did it for the `ConnectionMapper`, but this is not currently used elsewhere and I'd prefer to remove the passing of date strings around and remove the need for this parsing completely instead of making this function public

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
typfel: This is a temporary problem though I assume, and the usage of deprecated functions will be removed in a follow up PR.

Maybe something like this:
PR 1. introduce `lastModifiedInstant` and deprecate `lastModifiedDate` (this PR). 
PR 2. remove usage of deprecate functions
PR 3. rename `lastModifiedInstant` to `lastModifiedDate`

vitorhugods: I took another look at the usages of the deprecated stuff. And there aren't many.

Replacing them with the new ones actually simplifies some stuff and reduces modifications 😅 

So I just did all the 3 steps in a single commit.

typfel (approved):
🚀

PR #1351, opened 2023-01-18
fix: Force update (AR-2960)

borichellow:
# What's new in this PR?

Force user to update was not working properly 

### Issues

Found 2 small possible problems, need to check if it was it.

### Causes (Optional)

wrong `true/false` checking 

### Solutions

fix it


logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/appVersioning/AppVersionRepository.kt
```diff
@@ -17,7 +17,7 @@ internal class AppVersionRepositoryImpl(
         wrapApiRequest { api.fetchAppVersionBlackList(blackListUrl) }
             .fold({
                 kaliumLogger.e("$TAG: error while fetching VersionBlacklist: $it")
-                true
+                false
```
saleniuk, line 20: so if there's an error, for instance no network, then if we return false, we allow the user to use the version that can be blacklisted but we don't yet know about it - is that ok?

saleniuk (approved):

PR #1371, opened 2023-01-20
fix: emit only non empty notifications (RC)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We're emitting non-empty notifications unnecessarily.
Clients need to manually check if there's any notification.

Relodaded, for instance, is "marking as notified" without checking if there actually are notifications, causing an infinite loop of notification stream.

### Solutions

As it's a pain in the ass to write tests on Reloaded's `WireNotificationManager`, and as there are no purposes in emitting empty notifications, we might as well not do it from Kalium.

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
@@ -13,6 +13,7 @@ import kotlinx.coroutines.flow.buffer
 import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.merge
@@ -61,6 +62,7 @@ internal class GetNotificationsUseCaseImpl internal constructor(
                     .map { list -> list.filter { it.messages.isNotEmpty() } }
             }
             .distinctUntilChanged()
+            .filter { it.isNotEmpty() }
             .buffer(capacity = 3) // to cover a case when all 3 flows emits at the same time
     }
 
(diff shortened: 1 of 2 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
```diff
@@ -61,6 +62,7 @@ internal class GetNotificationsUseCaseImpl internal constructor(
                     .map { list -> list.filter { it.messages.isNotEmpty() } }
             }
             .distinctUntilChanged()
+            .filter { it.isNotEmpty() }
```
mchenani, line 65: we had that before, and we removed it due to the user profile update issue! that issue is not valid anymore?

vitorhugods: The user profile issue was that by flatMapping the `User` -> `Notifications`, when the `User` changed, we would cause the creation of a new flow that would emit the same values as before.

The fact that we filter empty notifications or not doesn't really matter to that problem.

vitorhugods:
In any case, as @MohamadJaara pointed out, this shouldn't fix the actual CAUSE of the problem that Reloaded is facing.

As the `distinctUntilChanged()` should already prevent that.

However, it corrects the symptom, at least for now.

I don't think there are arguments AGAINST this change. I think we should merge and I'll keep investigating what's actually causing multiple collections of empty notifications.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
mchenani: thanks for the details, don't know but it would be nice to have such a test to make sure that won't happen.

mchenani (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetNotificationsUseCase.kt
vitorhugods: Yeaaah. I want to bring a NotificationManager to inside Kalium actually, make it work on `UserSessionScope`, to make sure there's a app-lifecycle-independent worker doing this kind of things, instead of `ViewModel`, `PersistentWebSocketService`, and push notification receiver fighting over notifications.

The current `WireNotificationManager` in Reloaded is messy and used by too many entities and requires the caller to be very careful.

PR #1391, opened 2023-01-24
fix: repeating notification (RC) (AR-2978)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

I've noticed some notifications showing up again after I dismiss them.

### Causes

1. The `ConversationDAO` function `updateAllConversationsNotificationDate` is calling the `updateAllUnNotifiedConversationsNotificationsDate`; The DAO function doesn't do what it claims to do.
2. This function only updates if the modification date of the function is newer than the notification date.
3. We consider the `conversation.creationDate` as the value to update the modification date of a converstion. However, this is a value our local device clock can alter, so there's the possibility we mark as notified and our own client modifies the conversation to a `modifiedDate` that is smaller than the `notifiedDate`.

### Solutions

Make sure that `ConversationDAO.updateAllConversationsNotificationDate` actually does what its name suggest, and update the notification date for all conversations.

The only usage of this function is when `MarkMessagesAsNotifiedUseCase` is called for `AllConversations`. I don't see any benefit in not updating that value for all conversations. WYSIWYG

#### Test Coverage

----

----
##### References

### DIFF
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
@@ -78,11 +78,9 @@ UPDATE Conversation
 SET last_notified_date = ?
 WHERE qualified_id = ?;
 
-updateAllUnNotifiedConversationsNotificationsDate:
+updateAllNotifiedConversationsNotificationsDate:
 UPDATE Conversation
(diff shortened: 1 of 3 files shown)


borichellow (approved):
I'm not sure if this solution is better from a performance perspective, but I like it more :)

PR #1404, opened 2023-01-25
fix: ongoing call when importing backup

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When importing backup data with ongoing calls, when joining an ongoing call from backup it would crash the app, there are two reasons:

- Open Calls [Ongoing / Started / Answering or other] would differ from what would be received when receiving CONFCHECK or other AVS Signals.
- Multiple usages of current `keepReference` list resulting into `ArrayIndexOutOfBoundsException` from `add` into the `arrayList`.

### Solutions

- Added query to close all open calls from the backup as we are doing when user opens the app.
- Change current `mutableListOf<Any>` for `strongKeepReference` to `Collections.synchronizedList(mutableListOf<Any>())` so it doesn't throws `ArrayIndexOutOfBoundsException` anymore.

### DIFF
--- a/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/backup/DatabaseImporter.kt
+++ b/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/backup/DatabaseImporter.kt
@@ -47,7 +47,7 @@ class DatabaseImporterImpl(private val sqlDriver: SqlDriver) : DatabaseImporter 
         restoreTable("User")
         restoreConversations()
         restoreTable("Message")
-        if (!fromOtherClient) restoreTable("Call")
+        if (!fromOtherClient) restoreCalls()
         restoreAssets()
         restoreTable("MessageConversationChangedContent")
         restoreTable("MessageFailedToDecryptContent")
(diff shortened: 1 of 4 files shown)


vitorhugods (approved):
👨‍🍳👌 Noice

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/backup/DatabaseImporter.kt
```diff
@@ -47,7 +47,7 @@ class DatabaseImporterImpl(private val sqlDriver: SqlDriver) : DatabaseImporter
         restoreTable("User")
         restoreConversations()
         restoreTable("Message")
-        if (!fromOtherClient) restoreTable("Call")
+        if (!fromOtherClient) restoreCalls()
```
ohassine, line 50: I think we can remove restoring this table, missed calls are saved in `message` table

alexandreferris: Missed Calls yes, but in the future based on design we will have a new tab beside `Conversations` named `Calls` to show the history of calls, thus the main reason we have a table for storing calls.

ohassine: makes sense

PR #1407, opened 2023-01-25
chore: add intellij copyright notice template for project

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Given that we want to add copyright notice to our project files, we can have it automatically 

- Added using IntelliJ copyright profiles.
- Added githook pre-push to install using `make setup/pre-push-hook` so we run detekt locally (saving ci time)

### Causes (Optional)

Every new file will have the copyright notice as a header, including the current year as the creation of the file.

### Dependencies (Optional) 

We need, after merging, close the project in the IDE and re-open it again, so the configurations are loaded.

----

----

----
##### References


MohamadJaara (approved):

.githooks/pre-push
```diff
@@ -0,0 +1,26 @@
...
+# along with this program. If not, see http://www.gnu.org/licenses/.
+#
+
+set -e
+
+echo "-----------------------------------------------------------"
+echo "> Running detekt checks locally before push..."
+echo "-----------------------------------------------------------"
+
+./gradlew clean detekt
```
vitorhugods, line 26: Really needed?

Annoying when working offline, and Detekt is already running before tests anyway, so CI time is safe.

We are grown up and we should be responsible for our workflow.
If people aren't proactive enough to run Detekt locally before pushing, and prefer going back and forward seeing CI output instead of checking first.
I don't think we should babysit.

Sometimes I do forget about it, and I pay the price.
To me, having to run `detekt` on commits where I just changed some SQL or some other stuff that won't be detected would be more detrimental.

Personally, I'll disable the hook on my IDE.

yamilmedina: Yep yep! Totally agree, we are grown-ups to take care of this, as a hook running locally of course, we have total control to disable it or just ignore it with `--no-verify` that's the beauty of git, everything is possible 🧽 🙌 🌈 

I think is more of a misunderstanding on this particular hook itself, this will enforce no one to use it, it's more of an _opt-in_ feature if we want to look at it, being, if no one wants to install or ever run the `make setup/pre-push-hook` the hook will be living there for free on the repo, with 0 usages, so nothing to disable =)

vitorhugods (approved):
👌👌👌

PR #1423, opened 2023-01-27
feat: Move get asset size logic to use case in Kalium

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
Currently, the logic regarding what was the maximum size allowed for assets to be sent was only used on the `MessageComposerViewModel`. Since now we are introducing the import external media assets feature, we also needed to access this logic, and therefore the obvious choice was to move this logic into Kalium to be consumed as a use case.

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetAssetSizeLimitUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+fun interface GetAssetSizeLimitUseCase {
+    /**
+     * Returns the maximum size of an asset that can be uploaded to the Wire backend.
+     * @param isImage whether the asset to upload is an image or not
+     */
+    suspend operator fun invoke(isImage: Boolean): Long
+}
+
+internal class GetAssetSizeLimitUseCaseImpl internal constructor(
+    private val isSelfATeamMember: IsSelfATeamMemberUseCase
```
yamilmedina, line 14: I think we can inject here the `Dispatcher` and switching the execution here, so we don't have to do it on clients :D

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/IsSelfATeamMemberUseCase.kt
```diff
@@ -1,33 +1,19 @@
...
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program. If not, see http://www.gnu.org/licenses/.
- */
-
```
yamilmedina, line 18: I think we can maintain the copyright notice

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/IsSelfATeamMemberUseCase.kt
```diff
@@ -1,33 +1,19 @@
...
+    suspend operator fun invoke(): Boolean
+}
+
 /**
- * Return if self user is part of a team or noz
+ * Return if self user is part of a team or not
  */
-class IsSelfATeamMemberUseCase internal constructor(
+internal class IsSelfATeamMemberUseCaseImpl internal constructor(
     private val selfTeamIdProvider: SelfTeamIdProvider
```
yamilmedina, line 14: Same here, we can inject the dispatchers and gain some standardization on our Use Cases

yamilmedina (commented):
Great idea with the refactor needed also for the import feature.
Just some small comments 🪁

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetAssetSizeLimitUseCase.kt
gongracr: Not sure if we should switch the context on the use case level or rather on the repository, data layer level. Maybe something to discuss on next collective?

PR #1424, opened 2023-01-27
fix: clear firebase token only on soft logout [AR-3012]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When user tries to logout with clearing all data, application sometimes crashes. 

The reason about it was that updating firebase token flag in past was moved from SharedPreferences to SqlDelight DB. When user do the hard logout we don't need to clear token flag, because db is already cleared.

### Causes (Optional)

After opening app user has empty conversation list and he can't do nothing before clearing app data.

### Solutions

Update firebase token flag only when we do not do hard logout.

#### Test Coverage (Optional)

#### How to Test

1. Logout with clearing data when you are receiving messages
2. Logs should not contain any errors for MetadataDAO queries

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
@@ -59,6 +59,7 @@ internal class LogoutUseCaseImpl @Suppress("LongParameterList") constructor(
         deregisterTokenUseCase()
         logoutRepository.logout()
         sessionRepository.logout(userId = userId, reason)
+        logoutRepository.onLogout(reason)
 
         when (reason) {
             LogoutReason.SELF_HARD_LOGOUT -> {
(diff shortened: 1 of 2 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
```diff
@@ -74,17 +75,17 @@ internal class LogoutUseCaseImpl @Suppress("LongParameterList") constructor(
                 clearClientDataUseCase()
                 clientRepository.clearCurrentClientId()
                 clientRepository.clearHasRegisteredMLSClient()
+                // After logout we need to mark the Firebase token as invalid locally so that we can register a new one on the next login.
```
Garzas, line 78: @MohamadJaara Looking at the reasons should we also remove it from here? I assume that in that case we also clearing entire DB?

yamilmedina (approved):
Tricky debugging session, but the insect was found 🪁  great job!

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
MohamadJaara: no need to update it here since all of the local data will be wiped anyway

PR #1430, opened 2023-01-31
fix: removed client resync [AR-2989]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Simplified handler for logout use case, because after some previous changes we don't need to divide behaviour between `SELF_HARD_LOGOUT` and `REMOVED_CLIENT`.  

### Issues

After receiving `Removed Client` event user cannot sync after logging again.

### Causes (Optional)

User cannot use app after login again.

### Solutions

Simplify logout handler to use the same functionality for `SELF_HARD_LOGOUT` and `REMOVED_CLIENT`.


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
```diff
@@ -62,23 +62,13 @@ internal class LogoutUseCaseImpl @Suppress("LongParameterList") constructor(
         logoutRepository.onLogout(reason)
 
         when (reason) {
-            LogoutReason.SELF_HARD_LOGOUT -> {
+            LogoutReason.SELF_HARD_LOGOUT, LogoutReason.REMOVED_CLIENT, LogoutReason.DELETED_ACCOUNT -> {
```
MohamadJaara, line 65: this will delete user data when a client get remove, like database message history

MohamadJaara (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutUseCase.kt
Garzas: Yup, is that wrong?

MohamadJaara: 1. other clients do not have the same behavior.
2. in some cases, clients can miss this event if they are offline for a while.
3. we need to look into what caused the issue mentioned above because this logic was working just fine.

MohamadJaara: lets take it 1:1 and dive deep to get what exactly is going causing the issue

Garzas: Added changed discussed with @MohamadJaara

MohamadJaara (approved):

PR #1441, opened 2023-02-03
fix: OnClose call crashing

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[AR-3029 - App Crashing when receiving OnClose call from AVS when there is no call in db](https://wearezeta.atlassian.net/browse/AR-3029)

### Issues

From a fresh install build that doesn't have any (or at least the expected Call) saved in the database, when receiving a `OnCloseCall` signal from AVS it would trigger to save a `MissedCall` system message, but it requires a `CallerId` which it doesn't have and thus its crashing.

### Solutions

- Update DB to return `executeAsOneOrNull`, in case it returns null the `MissedCall` system message won't be saved as the user did not have any info about that call anyway.
- In case it returns a correct `CallerId` (from a Call that exists in DB) it will persist the message as normal.

#### Test Coverage (Optional)

#### How to Test

- Fresh install
- Receive a `OnCloseCall` signal from AVS without having any call prior
- App should not crash anymore


ohassine (approved):
Good catch!

PR #1456, opened 2023-02-07
fix: issues on `ScheduleNewAssetMessageUseCase` and dispatchers [AR-2198]

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The `ScheduleNewAssetMessageUseCase` class didn't allow the concurrent execution by multiple threads as it kept the current state on a lateinit variable shared among the rest of threads that would invoke the same created instance.  Therefore, the path of the asset could be silently changed by each one of the threads causing unexpected behaviors.

Also I identified other parts of the code where data sources should be closed, or operations were running in main thread, so also fixed those.

### Solutions

Transform the `currentAssetMessageContent` to a value that gets passed across functions and doesn't interfere with the concurrent logic of multiple threads.

----
##### References


MohammedMokresh (approved):
lgtm

PR #1461, opened 2023-02-08
fix: [RC] return replies to notification

borichellow:
# What's new in this PR?

### Issues

Notifications about replies to my message are displayed as regular text messages.

### Causes (Optional)

isSelfReply field was not fetched while getting data from DB

### Solutions

fix it


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/LocalNotification.kt
```diff
@@ -42,8 +42,7 @@ sealed class LocalNotificationMessage(
         override val author: LocalNotificationMessageAuthor,
         override val time: Instant,
         val text: String,
-        val isQuotingSelfUser: Boolean = false,
-        val isMentionedSelf: Boolean = false
+        val isQuotingSelfUser: Boolean = false
```
MohamadJaara, line 45: can we cover this case with tests to avoid breaking it in the future ?

borichellow: good point :)

PR #1477, opened 2023-02-10
fix(rc): sync exponential retry and wait until reconnected [AR-3076]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When the application cannot connect to the API host the number of retries is unbounded.

### Causes (Optional)

Syncs (both slow and incremental) are retried every 10 seconds if there's an failure.

### Solutions

Implemented an exponential delay and provided a way to observe not only network changes but also the internet connection availability. When failure, delay by given amount (calculated exponentially) and wait until the internet connection is restored, otherwise it makes no sense to retry as it will fail again.

#### How to Test

Try syncing without the internet connection, especially when there is a network but without the internet.

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/network/NetworkStateObserver.kt
```diff
@@ -0,0 +1,54 @@
...
+            // Drop the current value, so it will complete only if the connection changed again to connected during that time.
+            observeNetworkState()
+                .distinctUntilChanged()
+                .drop(1)
+                .filter { it is NetworkState.ConnectedWithInternet }
+                .firstOrNull()
+        }
+        // At this point, if the block above timed out, it means that it haven't reconnected so wait until reconnected.
+        // If the block above completed before the timeout, just make sure it's still connected.
+        observeNetworkState().firstOrNull { it is NetworkState.ConnectedWithInternet }
```
vitorhugods, line 44: If network state is not `ConnectedWithInternet`, this will make it wait indefinitely, possibly breaking the 10min max interval.

Not sure if intended or not, if buggy or not.

logic/src/androidMain/kotlin/com/wire/kalium/logic/network/NetworkStateObserverImpl.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }.shareIn(
+        scope = scope,
+        started = SharingStarted.WhileSubscribed(),
+        replay = 1
+    )
+
+    private fun NetworkCapabilities?.toState(): NetworkState {
+        val hasInternet = this?.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) == true
+        val isValidated = this?.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED) == true
```
vitorhugods, line 81: I wonder if there are some edge cases where on-premise environments could be considered "not validated" and should still be able to make requests.

Based on [the docs](https://developer.android.com/training/basics/network-ops/reading-network-state#introducing-net-capabilities):

> [NET_CAPABILITY_VALIDATED](https://developer.android.com/reference/android/net/NetworkCapabilities#NET_CAPABILITY_VALIDATED): This capability indicates the network has been found to provide actual access to the public internet last time it was probed.

Some customers MIGHT be affected, but I believe this is a great first iteration anyway and this can continue the way it is for now. We can add a `TODO` or just regular comment for now and check with customers.

logic/src/androidMain/kotlin/com/wire/kalium/logic/sync/WrapperWorker.kt
```diff
@@ -50,9 +50,9 @@ class WrapperWorker(
     CoroutineWorker(appContext, params) {
 
     override suspend fun doWork(): AndroidResult = when (innerWorker.doWork()) {
-        is KaliumResult.Failure -> AndroidResult.success()
-        is KaliumResult.Retry -> AndroidResult.failure()
-        is KaliumResult.Success -> AndroidResult.retry()
+        is KaliumResult.Failure -> AndroidResult.failure()
+        is KaliumResult.Retry -> AndroidResult.retry()
+        is KaliumResult.Success -> AndroidResult.success()
```
vitorhugods, line 55: image

logic/src/jvmMain/kotlin/com/wire/kalium/logic/network/NetworkStateObserverImpl.kt
```diff
@@ -0,0 +1,27 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.kalium.logic.network
+
+import kotlinx.coroutines.flow.Flow
+
+actual class NetworkStateObserverImpl : NetworkStateObserver {
+
+    override fun observeNetworkState(): Flow<NetworkState> {
+        TODO("Not yet implemented")
```
vitorhugods, line 25: This will most likely completely brake JVM clients, no?

On any network error, `CLI Client` and `TestService` will *poof*.

Better to just treat as always connected with internet then leaving the TODO here.

vitorhugods (changes requested):
Solid work. Just a few comments.

----

**On a side note**, I think we could extract some of these small mechanisms outside of `logic`, aiming to make the module smaller, aiding with compilation time and parallelisation of work.

In this case, `NetworkStateObserver` and its implementations don't necessarily need to be in `logic`. They could be in `network` or on a module of its own actually 🤔

I'm pretty sure it also applies to other stuff, like the WorkScheduler, and so on.

logic/src/commonMain/kotlin/com/wire/kalium/logic/network/NetworkStateObserver.kt
saleniuk: @vitorhugods that's right, that can take more than 10min, but the sync won't success anyway until we have the internet connection. It will retry when the connection state changes to `ConnectedWithInternet` so it won't try to make any requests unnecessarily. 
Would you prefer to retry sync again after the delay even if it's still not connected?

logic/src/androidMain/kotlin/com/wire/kalium/logic/network/NetworkStateObserverImpl.kt
saleniuk: Ok, I added a comment to keep that in mind 😉

logic/src/androidMain/kotlin/com/wire/kalium/logic/sync/WrapperWorker.kt
saleniuk: 😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/network/NetworkStateObserver.kt
saleniuk: I removed the lat "observe" - it's safer to retry every 10min even if there's no Internet than to be stuck.

vitorhugods: Not that I think we should retry even without internet, it's just for bug hunting purposes.

If someone ever asks themselves:
"Why isn't it respecting the 10min limit?"

vitorhugods (approved):

PR #1479, opened 2023-02-10
chore: fix dokka issues execution oom

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

If we run Dokka with stacktrace flag, we found out that is failing because of something similiar to https://github.com/Kotlin/dokka/issues/1405
So this PR tries to solve by suggested actions mentioned there.

- Disabled samples
- Increase memory metaspace and heap for gradle

----

----

----
##### References


vitorhugods:
It seems it fails mainly due to `Unresolved reference: otr`:

```
> Task :network:compileCommonMainKotlinMetadata
e: /Users/runner/work/kalium/kalium/network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/message/EnvelopeProtoMapper.kt: (22, 33): Unresolved reference: otr
```

🤔 Curious about this one, considering it usually builds nicely

vitorhugods:
One thing is quite curious:

This problem started EXACTLY when we added `darwin` source sets for `network`, `calling` and `logic` IIRC

yamilmedina:
> One thing is quite curious:
> 
> This problem started EXACTLY when we added `darwin` source sets for `network`, `calling` and `logic` IIRC

Yees ! exactly around that date started, the thing is Dokka seems to have failed in the past with memory management 😓  -like now, having a `StackOverFlowError` so I opened also an issue there, just in case they can help us. https://github.com/Kotlin/dokka/issues/2853

The other thing is to create many partials docs (this seems fine) but would be a bit overkill to merge it later, I think we still can try a couple of things more 🤞

yamilmedina:
I think I'll close this one, as I tried we many approaches, some insights:

- dokka is well known for having bad memory management
- network module is causing a big pain
- executing each task htmlPartial by itself in separate runs, doesn't brake it
- because logic module depends on network, logic never gets to finish running or generating docs.
- excluding cli, logic, and network make it "pass" but loses the point
- maybe is worth taking a shot at https://github.com/adamko-dev/dokkatoo

PR #1490, opened 2023-02-14
feat(calling): leave sub-conversation when a MLS conference ends

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

One of many PRs adding support for MLS conference calling

---

Leave the conference sub-conversation when you've leave an MLS conference or it ends, as described in [Use case: end a call (ongoing/outgoing)](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/616956023/Use+case+end+a+call+ongoing+outgoing)

#### Test Coverage

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
```diff
@@ -134,6 +135,7 @@ internal class CallDataSource(
     private val teamRepository: TeamRepository,
     private val mlsClientProvider: MLSClientProvider,
     private val joinSubconversationUseCase: JoinSubconversationUseCase,
+    private val leaveSubconversationUseCase: LeaveSubconversationUseCase,
```
alexandreferris, line 138: We can remove the suffix `UseCase`

alexandreferris, line 138: I was thinking here if maybe it would be possible to not have the repository calling a usecase and rather the other way around?

Just so we can continue the flows we already have, such as:
`UI -> ViewModel -> UseCase -> Repository -> DAO/API`

alexandreferris (commented):
Looking nice! I just have two small comments 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
typfel: It would not make sense to have the view model call this use case since it's an internal use case. Maybe it shouldn't be a use case? Join and leave a subconversation could be actions on a repository.

alexandreferris: Hmmmm, maybe then having it as actions on a repository does make more sense but it is not something causing a problem. all good :)

typfel: Yes, I'll think about how this can re-factored. But prefer not do it in this PR.

alexandreferris (approved):
👌🏻

PR #1521, opened 2023-02-21
fix(dev): ConnectivityManager$TooManyRequestsException [AR-3134]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We receive `android.net.ConnectivityManager$TooManyRequestsException`, because `registerDefaultNetworkCallback` has a limit of 100 registered callbacks per app and for some reason our app exceeds that limit.

### Causes (Optional)

The app doesn't call `unregisterNetworkCallback`. 
I checked multiple scenarios to find possible explanations and the issue probably is that `callbackFlow` which is suggested way of registering such callbacks, doesn't work well when used with `.first()` or `withTimeout { }`. Basically it won't unregister if the flow closes before the execution comes to the `awaitClose` line, so the flow doesn't even know that it has something to unregister.
Here are two examples:

image
In this case it didn't unregister because timeout happened and flow was closed on long register operation (simulated by `delay`).

image
Here it didn't unregister because it emits first value initially but we wait only for this first value, so it closed the flow right after receiving the first value, again before `awaitClose`.

### Solutions

Get rid of `callbackFlow` as it's problematic and register a single callback for the whole lifecycle of the app, every element can subscribe and start observing this flow and replay also works the way it used to because it's a `StateFlow`.

----

----

----
##### References

### DIFF
--- a/logic/src/androidMain/kotlin/com/wire/kalium/logic/network/NetworkStateObserverImpl.kt
+++ b/logic/src/androidMain/kotlin/com/wire/kalium/logic/network/NetworkStateObserverImpl.kt
@@ -22,59 +22,41 @@ import android.content.Context
 import android.net.ConnectivityManager
 import android.net.Network
 import android.net.NetworkCapabilities
-import com.wire.kalium.util.KaliumDispatcher
-import com.wire.kalium.util.KaliumDispatcherImpl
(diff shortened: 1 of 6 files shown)


MohamadJaara (approved):
LGTM 🚀

PR #1523, opened 2023-02-22
feat: inform about new devices (AR-2964)

borichellow:
# What's new in this PR?

- `NewClientManagerImpl` singleton object used for passing forward NewClientEvents.
- `ObserveNewClientsUseCase` is used on the platform side to observe new clients for any user on a device.
- improved `ObserveValidAccountsUseCase` to get all the necessary info in 1 DB query, instead of 2 
- added `Users.selectWithTeamByQualifiedId` DB-query to get the user and his team in 1 query


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/ObserveNewClientsUseCase.kt
```diff
@@ -0,0 +1,72 @@
...
+                .map { currentAccInfo ->
+                    if (currentAccInfo.userId == userId) flowOf(NewClientResult.InCurrentAccount(newClient))
+                    else userSessionScopeProvider.get(userId)?.let {
+                        it.users
+                            .getSelfUser()
+                            .map { selfUser -> NewClientResult.InOtherAccount(newClient, userId, selfUser.name, selfUser.handle) }
+                    } ?: flowOf()
+                }.getOrElse(flowOf(NewClientResult.Error))
+        }
+}
```
ohassine, line 61: Can you please cover this useCase with unit test ?

ohassine (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/ObserveNewClientsUseCase.kt
borichellow: I tried, but i can't mock `UserSessionScope` and `UserScope` in tests

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/ObserveNewClientsUseCase.kt
```diff
@@ -0,0 +1,82 @@
...
+                            .map { selfUser -> NewClientResult.InOtherAccount(newClient, userId, selfUser.name, selfUser.handle) }
+                    } ?: flowOf(NewClientResult.Error)
+                }.getOrElse(flowOf(NewClientResult.Error))
+        }
+}
+
+/** this interface needed to be able to test [ObserveNewClientsUseCaseImpl] */
+interface GetSelfUserUseCaseProvider {
+    fun get(userId: UserId): GetSelfUserUseCase?
+}
```
gongracr, line 66: I'm still not sure why you needed to add this extra interface since they both do the same thing. Can't you just mock `GetSelfUserUseCase` inside the `ObserveNewClientsUseCase` constructor?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/ObserveNewClientsUseCase.kt
```diff
@@ -0,0 +1,82 @@
...
+
+class ObserveNewClientsUseCaseImpl(
+    private val sessionRepository: SessionRepository,
+    private val getSelfProvider: GetSelfUserUseCaseProvider,
+    private val newClientManager: NewClientManager
+) : ObserveNewClientsUseCase {
+
+    override suspend operator fun invoke(): Flow<NewClientResult> = newClientManager.observeNewClients()
+        .flatMapLatest { (newClient, userId) ->
```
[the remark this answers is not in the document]
borichellow: well, there is no any cashing algorithm and we can display only 1 "New device" dialog at a time on the android side. 
So i'd say: if we would like to handle every single event in the future - we'll need to change this usecase anyway :)

gongracr (approved):
Greatee stuuuff. Thanks for the constant changes. But now it definitely looks much cleaner and performant! 🤌 👨‍🍳 🧀

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAO.kt
```diff
@@ -54,6 +54,25 @@ data class UserEntity(
...
+    val connectionStatus: ConnectionEntity.State = ConnectionEntity.State.NOT_CONNECTED,
+    val previewAssetId: UserAssetIdEntity?,
+    val completeAssetId: UserAssetIdEntity?,
+    val availabilityStatus: UserAvailabilityStatusEntity,
+    val userType: UserTypeEntity,
+    val botService: BotEntity?,
+    val deleted: Boolean,
+    val teamName: String?,
+    val teamIcon: String?
+)
```
vitorhugods, line 74: Why `UserEntityWithTeam`?

What about something like `Pair<UserEntity, TeamEntity>` or similar ?

Then we can more easily map `UserEntity` to `User` and `TeamEntity` to `Team` for example, reusing the current mapping functions if needed.

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/user/ObserveSelfUserWithTeamUseCaseTest.kt
```diff
@@ -0,0 +1,79 @@
...
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class ObserveSelfUserWithTeamUseCaseTest {
+
+    private val arrangement = Arrangement()
+
+    @Test
+    fun testIfValidDataIsPassedForward_0() = runTest {
```
vitorhugods, line 39: What's the difference between `testIfValidDataIsPassedForward_0` and `testIfValidDataIsPassedForward_1`?

The name could be clearer and explain what each test is actually testing.

Also, no need to add `testIf` before each test.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserMapper.kt
```diff
@@ -151,6 +154,27 @@ internal class UserMapperImpl(
...
+            previewAssetId?.toModel(),
+            completeAssetId?.toModel(),
+            availabilityStatusMapper.fromDaoAvailabilityStatusToModel(availabilityStatus)
+        )
+
+        val team = if (team != null && teamName != null && teamIcon != null) {
+            Team(team!!, teamName!!, teamIcon!!)
+        } else null
+        user to team
+    }
```
vitorhugods, line 176: This could be simplified if the DB already returned a pair, then we could just call a mapper for each element (`first` and `second`)

vitorhugods (changes requested):
Great and clean job overall.

Just noticed the opportunity to simplify the mapping of `User/Team` and improve the name of a couple of tests :)

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/user/ObserveSelfUserWithTeamUseCaseTest.kt
borichellow: well, the UseCase is quit simple: just get data from repository and pass it forward, so I wasn't even sure about the names for tests :) 
But I'll try to make something up

vitorhugods (approved):
🪄✨

PR #1493, opened 2023-02-14
feat: pt1. send messages offline backends - api handle federation error type (AR-3084)

yamilmedina:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

This work is part of a series of PR's that will be covering "sending messages" while some backends or recipients are not available due to federation, connectivity issues, so this will allow a user to:

- Sent even if it some of the participants are owned by a backend that is offline.
- See why some of the users were not able to receive the message and why.

This part 1, focus on handling the new error detail `federation` and persisting that failure to cover own backend unavailability and others backend unavailability.

### Solutions

Added a new type in `Message.Status` called `FAILED_REMOTELY` so we can later on -displaying in ui- show the specific error message in the context of a federated conversation.

#### Test Coverage (Optional)

- [x] I have added automated test to this contribution

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1466, opened 2023-02-08
feat: federated offline send messages test, mapping new field and coverage for all cases (AR-3082)

yamilmedina:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

Include new _optional_ field for mapping `failed_to_send` response field on send message endpoint.
Before starting that I realized no tests were covered for `qualifiedSendMessages`, so this also adds the coverage for the previous cases.

#### Test Coverage (Optional)

- [x] I have added automated test to this contribution

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [x] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1525, opened 2023-02-22
feat: federation msg failed to send handling - part 1:  base persistence and failed to send case (AR-3015)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This is the first part of the groundwork for handling the sent message response with a partial success.

- What is a partial success? 
  - A partial success is a message that was sent, but due to some backend connectivity issues, some recipients did not get the message, _eg: Federated conversations context._

Since we have 2 possibilities to have partial errors, while sending a message, eg:
- **(I will open another PR for this case)** We couldn't get an encrypted session with a user (eg. backend to get prekeys is currently offline) 
- **(This PR)** We have all sessions established, the message was sent, but for example some users in a federated backend are currently down.

### Solutions

Adds a new table in DB with relationship to a message, mapping its values, and having the possibility to persist the list of users IDs that did not get the message.

### Dependencies (Optional)

Previous related work (already merged)

#1493 
#1466 

#### Test Coverage (Optional)

** Will be addressed in part 2 of this PR. **

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -229,12 +235,25 @@ internal class MessageSenderImpl internal constructor(
         messageRepository.sendEnvelope(message.conversationId, envelope, messageTarget).fold({
             when (it) {
                 is ProteusSendMessageFailure -> messageSendFailureHandler.handleClientsHaveChangedFailure(it).flatMap {
-                    attemptToSend(message, messageTarget)
+                    attemptToSendWithProteus(message, messageTarget)
```
yamilmedina, line 238: Note: Changed this line, as at this point we know for sure the protocol, saving an extra query.

alexandreferris (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
```diff
@@ -387,3 +391,12 @@ enum class AssetType {
 
 typealias ReactionsCount = Map<String, Int>
 typealias UserReactions = Set<String>
+
+sealed class RecipientFailure {
```
typfel, line 395: I find it bit strange to have a `RecipientFailure` type which also have "no error" case. If it should also include the success case I'd suggest something along:

```
sealed class DeliveryStatus {
  data class PartialDelivery(...) : DeliveryStatus()
  object CompleteDelivery : DeliveryStatus()
}

File: logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt [hunk 3]
```diff
@@ -334,8 +343,9 @@ class MessageDataSource(
                 else -> networkFailure
             }
             Either.Left(failure)
-        }, {
-            Either.Right(it.time)
+        }, { response: QualifiedSendMessageResponse ->
+            // not sure about this forced cast, but it seems should be the case here to receive a QualifiedSendMessageResponse.MessageSent
+            Either.Right(SendMessagePartialFailureMapperImpl.fromDTO(response as QualifiedSendMessageResponse.MessageSent))
```
typfel, line 348: Couldn't the two folds with `sendMessageFailureMapper` and `SendMessagePartialFailureMapperImpl` go away and be replaced with:

```
when(response) {
  is QualifiedSendMessageResponse.MessageSent -> ...
  is QualifiedSendMessageResponse.MissingDevicesResponse -> ..
}

File: logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt [hunk 4]
```diff
@@ -186,11 +187,16 @@ internal class MessageSenderImpl internal constructor(
 
         return target.flatMap { recipients ->
             sessionEstablisher.prepareRecipientsForNewOutgoingMessage(recipients).map { recipients }
-        }.flatMap { recipients ->
+            // map a filtered recipients in case there is an error for x,y,z clients of users
+        }.fold({
+            // if (it is NetworkFailure.FederatedBackendError)
+            // handle federated failure to filter clients and add to QualifiedMessageOption.IgnoreSome
```
typfel, line 193: Is this a TODO or should it be implemented?

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
yamilmedina: Makes total sense, this `DeliveryStatus` name suits better, thanks

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
yamilmedina: This was a `TODO`, added the corresponding `TODO(federation)` tag to the comments.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
yamilmedina: Delegated this `when` and mapping down to the mapper =)

typfel: Mm I didn't quite understand the this code, also not so cool to have case which never will happen. 

I think we should consider mapping the 412 to be a success so that both response cases can be handled in a single `when` statement.

But should probably happen in a separate PR.

PR #1536, opened 2023-02-24
feat: SCIM or provided users, readonly profiles handling (AR-2186)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In order to handle checks for read only profiles, namely, users that are provided.
In simple words, users not managed by wire, their profiles are readonly.

We already had the network mapping for the SelfUser, now we add the persistence layer to it and mapping to model, in order to check on Reloaded.

#### Test Coverage (Optional)

_*Will add later, before merge*_

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/publicuser/PublicUserMapper.kt
@@ -106,7 +106,8 @@ class PublicUserMapperImpl(
             availabilityStatus = availabilityStatusMapper.fromModelAvailabilityStatusToDao(availabilityStatus),
             userType = userEntityTypeMapper.fromUserType(userType),
             botService = botService?.let { BotEntity(it.id, it.provider) },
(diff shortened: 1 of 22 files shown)


MohamadJaara:
@yamilmedina 
if you need it only for self user it does not make much sense to add it as DB column, you can use `sessionRepository.ssoId(selfUserId)`to get self userSSOID, and it already has the managed by value

yamilmedina:
> @yamilmedina if you need it only for self user it does not make much sense to add it as DB column, you can use `sessionRepository.ssoId(selfUserId)`to get self userSSOID, and it already has the managed by value
> 
> you can also use `IsPasswordRequiredUseCase` as example

Makes sense to me, the only thing about it is that docs mentioned that this read-only profiles are provided by that `managed_by` field on the self user response. So if every` SSO/managed_by` user is also consequently an SCIM user, I'm happy to revert the changes, since is less for us to do =D

yamilmedina:
Will close this in favour of moving this field to `Accounts` table

PR #1552, opened 2023-03-02
feat: enqueue self deleting messages (part 1 - text)

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

-Enqueue message self deletion

### Solutions

- I added another column in Message Table : 

 - expire_after_millis
 - self_deletion_start_date
 
 Self deletion start date is used to request pending deletions messages after the restart, we somehow need to know the timestamp it was started or paused I went for when it was started.

- Another methods to MessageRepository allowing me to query all ephemeral messages in order to restart the deletion process, this will happen when the user exists the app or the system kills it, on restart we enqueue the pending deletion messages again. 

Additionaly exposed 2 use cases - 

public to client :
-EnqueueMessageSelfDeletion - enqueue the message self deletion 

internal to kalium: 
- RestardPendingSelfDeletionUsecase - restartd self deletion messages that were not finished

### Dependencies (Optional)

Needs releases with:

### Testing

- Coming in a second PR

#### Test Coverage (Optional)

- Coming in a second PR

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
```diff
@@ -60,6 +61,42 @@ sealed interface Message {
...
+            get() = ephemeralMessage.conversationId
+        override val date: String
+            get() = ephemeralMessage.date
+        override val senderUserId: UserId
+            get() = ephemeralMessage.senderUserId
+        override val status: Status
+            get() = ephemeralMessage.status
+        override val visibility: Visibility
+            get() = ephemeralMessage.visibility
+    }
```
vitorhugods: Why would Ephemeral be its own type?

It makes it so we need to FIRST know if it's ephemeral and THEN check what's the type/content.

Initially, I don't see any need for having a new "top-level" message type at all.


What if instead a nullable `EphemeralData` is added to `Regular` messages?
This could contain `expireAfterMillis` and `selfDeletionStartDate`.


So we can: check if it's a text/asset/whatever, and THEN check if the `ephemeralData` is null or not.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageMapper.kt
```diff
@@ -96,6 +94,32 @@ class MessageMapperImpl(
...
+                                is Message.EditStatus.Edited -> MessageEntity.EditStatus.Edited(editStatus.lastTimeStamp.toInstant())
+                            },
+                            visibility = visibility,
+                            senderName = senderUserName,
+                            isSelfMessage = isSelfMessage,
+                            expectsReadConfirmation = expectsReadConfirmation
+                        )
+                    )
+                }
+            }
```
vitorhugods, line 122: This whole mapping could be removed and we could just the `ephemeralData` as part of the `regular` message.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageMapper.kt
```diff
@@ -136,9 +159,34 @@ class MessageMapperImpl(
...
+                            is MessageEntity.EditStatus.Edited -> Message.EditStatus.Edited(editStatus.lastDate.toIsoDateTimeString())
+                        },
+                        visibility = visibility.toModel(),
+                        reactions = Message.Reactions(reactions.totalReactions, reactions.selfUserReactions),
+                        senderUserName = senderName,
+                        isSelfMessage = isSelfMessage,
+                        expectsReadConfirmation = expectsReadConfirmation
+                    )
+                )
+            }
```
vitorhugods, line 189: Same here

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+        MutableStateFlow(emptyMap())
+
+    override fun startSelfDeletion(conversationId: ConversationId, messageId: String) {
+        userSessionCoroutineScope.launch {
+            // in case client start deletion couple of times for the message, we secure it with mutex
+            // that will update the map with a corresponding key and default value, so that coroutine
+            // launched after the first winning the race simply returns
+            ephemeralMessageDeletionTimeLeftMapMutex.withLock {
+                val isSelfDeletionOutgoing = ephemeralMessageDeletionTimeLeftMap.value[conversationId to messageId] != null
+                if (isSelfDeletionOutgoing) return@launch
```
vitorhugods, line 50: `ongoing` instead of `outgoing`?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+
+                ephemeralMessageDeletionTimeLeftMap.update { currentState ->
+                    currentState + ((conversationId to messageId) to SelfDeletionTimeLeft(0))
+                }
+            }
+
+            messageRepository.getMessageById(conversationId, messageId).map { message ->
+                require(message is Message.Ephemeral)
```
vitorhugods, line 58: If some app tries to wrongly call a `SelfDeletingTimer` we could fail silently and just log it instead of crashing.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EnqueueMessageSelfDeletionUseCase.kt
```diff
@@ -0,0 +1,15 @@
+package com.wire.kalium.logic.feature.message
+
+import com.wire.kalium.logic.data.id.ConversationId
+
+class EnqueueMessageSelfDeletionUseCase(
+    private val ephemeralMessageDeletionHandler: EphemeralMessageDeletionHandler,
+) {
```
vitorhugods, line 7: Missing docs for the UseCase

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+
+                enqueueMessageDeletion(message)
+            }
+        }
+    }
+
+    private fun enqueueMessageDeletion(message: Message.Ephemeral) {
+        userSessionCoroutineScope.launch {
```
vitorhugods, line 66:
```suggestion
        launch {
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+interface EphemeralMessageDeletionHandler {
+    fun startSelfDeletion(conversationId: ConversationId, messageId: String)
+
+    fun observePendingMessageDeletionState(): Flow<Map<Pair<ConversationId, String>, SelfDeletionTimeLeft>>
+
+    fun enqueuePendingSelfDeletionMessages()
+}
+
+internal class EphemeralMessageDeletionHandlerImpl(
+    private val userSessionCoroutineScope: CoroutineScope,
```
vitorhugods, line 32:
```suggestion
    userSessionCoroutineScope: CoroutineScope,
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/RestartPendingSelfDeletionMessageUseCase.kt
```diff
@@ -0,0 +1,5 @@
+package com.wire.kalium.logic.feature.message
+
+internal class RestartPendingSelfDeletionMessagesUseCase {
+
+}
```
vitorhugods, line 5: Is this needed?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/ObserveOngoingSelfDeletionMessagesUseCase.kt
```diff
@@ -0,0 +1,12 @@
+package com.wire.kalium.logic.feature.message
+
+import com.wire.kalium.logic.data.id.ConversationId
+import kotlinx.coroutines.flow.Flow
+
+class ObserveOngoingSelfDeletionMessagesUseCase(
+    private val ephemeralMessageDeletionHandler: EphemeralMessageDeletionHandler
+) {
```
vitorhugods, line 8: Missing docs

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    data class Started(val deletionStartDateInMs: Long) : SelfDeletionTimerState()
+    data class OnGoing(val timeLeft: SelfDeletionTimeLeft) : SelfDeletionTimerState()
+
+    object Finished : SelfDeletionTimerState()
+}
+
+class SelfDeletionTimeLeft(val timeLeftInMs: Long){
+
+}
```
vitorhugods, line 158:
```suggestion
value class SelfDeletionTimeLeft(val timeLeftInMs: Long)
```

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/message/EnqueueMessageSelfDeletionUseCase.kt
```diff
@@ -0,0 +1,19 @@
+package com.wire.kalium.logic.feature.message
+
+import com.wire.kalium.logic.data.id.ConversationId
+import kotlinx.coroutines.test.TestScope
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+
+class EnqueueMessageSelfDeletionUseCase {
```
vitorhugods, line 8:
```suggestion
class EnqueueMessageSelfDeletionUseCaseTest {
```

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/message/EnqueueMessageSelfDeletionUseCase.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    val testScope = TestScope()
+
+    private val selfDeletingMessageEnqueuer = EphemeralMessageDeletionHandler(testScope)
+
+    @Test
+    fun test() = runTest {
+        selfDeletingMessageEnqueuer.enqueue(ConversationId("test", "test"), "test")
+    }
```
vitorhugods, line 17: Missing tests

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+            messageRepository.getEphemeralMessages()
+                .onSuccess { ephemeralMessages ->
+                    ephemeralMessages.forEach { ephemeralMessage ->
+                        require(ephemeralMessage is Message.Ephemeral)
+
+                        enqueueMessageDeletion(message = ephemeralMessage)
+                    }
+                }
+        }
+    }
```
vitorhugods, line 119: Does this means that it doesn't wait for the message to be read before it starts deleting stuff?

vitorhugods (changes requested):
Awesome job in general, and I loved the timer implementation.

But here are some suggestions to simplify the `Message` hierarchy, some missing docs and tests.

I'd love to see the Timer tested too, as coroutines-test allows us to test stuff with a virtual clock using `advanceTimeBy(millis)` to test more easily :)

Also noticed missing docs, which Detekt should warn us about.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
trOnk12: Good point I need the ephemeral messages here that are already been marked as started then yes we enqueue those because they were already read and started

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
trOnk12: Agreed good point with well explained reason why u think this way I am convinced :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
```diff
@@ -164,6 +164,9 @@ interface MessageRepository {
         millis: Long
     ): Either<CoreFailure, Unit>
 
+    suspend fun getEphemeralMessages(): Either<CoreFailure, List<Message>>
+    suspend fun markSelfDeletionStartDate(conversationId: ConversationId, messageUuid: String, deletionDate: Long): Either<CoreFailure, Unit>
```
typfel, line 168:
```suggestion
    suspend fun markSelfDeletionStartDate(conversationId: ConversationId, messageUuid: String, deletionDate: Instant): Either<CoreFailure, Unit>
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+        const val TIMER_UPDATE_INTERVAL_IN_MILLIS = 1000L
+    }
+
+    fun startTimer(expireAfterMillis: Long) = produce {
+        send(SelfDeletionTimerState.Started(Clock.System.now().toEpochMilliseconds()))
+
+        var elapsedTime = 0L
+
+        while (elapsedTime < expireAfterMillis) {
+            delay(TIMER_UPDATE_INTERVAL_IN_MILLIS)
```
typfel, line 136: Why do we need to update every second? Why not delay the remaining time until message should be deleted? 

Is the intention that the UI should also observe these timers? Because I think UI should manager their own timers.

I think this can simplified and we can get rid of the mutex locks.

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
```diff
@@ -60,6 +61,42 @@ sealed interface Message {
         val visibility: Visibility
     }
 
+    data class Ephemeral(
+        val expireAfterMillis: Long,
+        val selfDeletionStartDate: Long?,
+        val ephemeralMessage: Regular
+    ) : Standalone {
+
+        fun timeLeftForDeletion(): Long {
```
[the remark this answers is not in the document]
saleniuk: Wouldn't it be better, instead of returning the remaining time, to return the time until which it should be displayed? Usually it shouldn't make any difference, but what if something is lagged/delayed, then it can be shown for longer that it should actually

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+                            messageRepository.markSelfDeletionStartDate(
+                                message.conversationId,
+                                message.id,
+                                Clock.System.now().toEpochMilliseconds()
+                            )
+                        }
+                    }
+
+                    is SelfDeletionTimerState.OnGoing -> {
+                        println("time left ${timerEvent.timeLeft.timeLeftInMs}")
```
saleniuk, line 84: Use logger or remove if not needed 😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+        userSessionCoroutineScope.launch {
+            val selfDeletingMessageTimer = SelfDeletingMessageTimer(
+                coroutineScope = userSessionCoroutineScope
+            )
+
+            for (timerEvent in selfDeletingMessageTimer.startTimer(message.timeLeftForDeletion())) {
+                when (timerEvent) {
+                    is SelfDeletionTimerState.Started -> {
+                        if (!message.isDeletionScheduledInThePast()) {
+                            messageRepository.markSelfDeletionStartDate(
```
saleniuk, line 75: It can return Either.Left so what it it fails?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
trOnk12: Also mutexes will still be needed, because someone can request the deletion of the message couple of times for the same message, we would need to hold the map somewhere with the ones that are ongoing to prevent that, and we will also need a mutex or some concurency protection mechanism when multiple coroutines are trying to create a timer, it will be I think one less place of locking the access with mutex

typfel: But what does the UI do with these timers? It displays a countdown of some sort? Such a timer would only need to run for a message which actually visible on screen.  What if the UI wants to update the count down every millisecond?

Managing such display timer in Kalium means we'll need to run timers for all ephemeral messages regardless if they displayed or not.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,158 @@
...
+        send(SelfDeletionTimerState.Started(Clock.System.now().toEpochMilliseconds()))
+
+        var elapsedTime = 0L
+
+        while (elapsedTime < expireAfterMillis) {
+            delay(TIMER_UPDATE_INTERVAL_IN_MILLIS)
+
+            send(SelfDeletionTimerState.OnGoing(SelfDeletionTimeLeft(expireAfterMillis - elapsedTime)))
+
+            elapsedTime += TIMER_UPDATE_INTERVAL_IN_MILLIS
```
saleniuk, line 140: Shouldn't we add the time that really passed between iterations? Every time it actually waits for 1s and executes a send so it can take more than 1s and after some time it can differ 😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
trOnk12: I mean if that would be a real requirement we could add a variable that defines how often we want to see the update, and yes it display as countdown, the process of self-deletion starts when the message is displayed on the screen, there is a use case that allows to do so, and yes we indeed run a timer if the messages is displayed or not but I do not think it is a huge problem to be honest, it simply updates a value that is held inside a map

trOnk12: I cannot see the benefit of doing the work to move this all to the UI and add another mechanism in the UI that duplicates basicly what is done here, just for the sake of not updating a value when it is not being displayed

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAO.kt
```diff
@@ -122,5 +121,9 @@ interface MessageDAO {
         millis: Long
     )
 
+    suspend fun getEphemeralMessages(): List<MessageEntity>
+
+    suspend fun updateSelfDeletionStartDate(conversationId: QualifiedIDEntity, messageId: String, selfDeletionStartDate: Long)
```
saleniuk, line 126: Can it be updated or should it be set only once?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
trOnk12: However now you are saying it would not make much sense to have a timer running each second for a message deleted after 2 weeks

trOnk12: I am going to take another look on this with what u are saying in mind I think you got a point here

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
trOnk12: all this is doing on consumer side is updating a value, if there is a time difference i think it would be some miliseconds, however with mutexes it could take longer indeed, good point I am going to take a look at it with this in mind.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
typfel: > I mean if that would be a real requirement we could add a variable that defines how often we want to see the update, and yes it display as countdown, the process of self-deletion starts when the message is displayed on the screen, there is a use case that allows to do so, and yes we indeed run a timer if the messages is displayed or not but I do not think it is a huge problem to be honest, it simply updates a value that is held inside a map

In Android you want to display a countdown, maybe another client wants to display a continuous progress bar. I agree it might seem like are implementing something twice but I'd argue visual timers and a timer for deleting a message have different requirements and we shouldn't try to merge them.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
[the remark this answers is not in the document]
typfel: Yes, returning a date makes more sense.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
trOnk12: What they want to display does not mean we should remove it from the Kalium I think, they are interested in progress and the time that is left, forcing them to implement something on the client side if we have all the information on Kalium is unncessary I think if we can have this built in, I am going to find a solution so we are both happy :P

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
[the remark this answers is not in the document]
typfel: That also ties into my arguments about timers. We don't need a `flow<timeLeftInMiliseconds>, the UI can render a `CountDown(Clock.System.now() -  message.expirationDate)`, which it refreshes that at whatever interval it deems appropriate.

typfel:
@trOnk12 should we review this again?

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/ApplicationMessageHandler.kt
```diff
@@ -196,47 +198,46 @@ internal class ApplicationMessageHandlerImpl(
...
+    private suspend fun persistRegularMessage(message: Message.Regular, expireAfterMillis: Long? = null) {
+        if (expireAfterMillis != null) {
+            persistMessage(Message.Ephemeral(expireAfterMillis, null, message))
+        } else {
+            persistMessage(message)
+        }
     }
 
-    private suspend fun handleTextMessage(
+    private suspend fun adjustTextMessageWithQoutedReference(
```
typfel, line 227:
```suggestion
    private suspend fun adjustTextMessageWithQuotedReference(
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/ApplicationMessageHandler.kt
```diff
@@ -120,9 +120,11 @@ internal class ApplicationMessageHandlerImpl(
                     status = Message.Status.SENT,
                     editStatus = Message.EditStatus.NotEdited,
                     visibility = visibility,
-                    expectsReadConfirmation = content.expectsReadConfirmation
+                    expectsReadConfirmation = content.expectsReadConfirmation,
+                    expirationStatus = content.expiresAfterMillis?.let { Message.ExpirationStatus.Expiratable(it) }
```
typfel, line 124: `Message.ExpirationStatus.Expiratable` where is this defined? I can't find it when searching this PR or the branch?

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/ApplicationMessageHandler.kt
```diff
@@ -196,47 +198,46 @@ internal class ApplicationMessageHandlerImpl(
...
 
-            is MessageContent.RestrictedAsset -> TODO()
         }
     }
 
-    private suspend fun handleKnock(message: Message.Regular) {
-        persistMessage(message)
+    private suspend fun persistRegularMessage(message: Message.Regular, expireAfterMillis: Long? = null) {
+        if (expireAfterMillis != null) {
+            persistMessage(Message.Ephemeral(expireAfterMillis, null, message))
```
typfel, line 221: I thought `Message.Ephemeral` was removed flattened. Is something not pushed?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,90 @@
...
+
+interface EphemeralMessageDeletionHandler {
+
+    fun startSelfDeletion(conversationId: ConversationId, messageId: String)
+    fun enqueuePendingSelfDeletionMessages()
+}
+
+
+// TODO:Mateusz: implement failure logic
+internal class EphemeralMessageDeletionHandlerImpl(
```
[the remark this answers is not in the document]
trOnk12: for the sake the time I spend on it I going to keep it the way I implemented it, nex time pls mention it beforehand

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/ApplicationMessageHandler.kt
trOnk12: oops I forgot about that part

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/ApplicationMessageHandler.kt
trOnk12: oops I forgot about that part

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
```diff
@@ -289,6 +297,24 @@ sealed interface Message {
...
+        // time left for deletion it can be a negative value if the time difference between the self deletion start date and
+        // now is greater then expire after millis
+        fun timeLeftForDeletion(): Long {
+            return if (!isDeletionStartedInThePast()) {
+                expireAfterMillis
+            } else {
+                val timeElapsedSinceSelfDeletionStartDate = Clock.System.now() - selfDeletionStartDate!!
+
+                expireAfterMillis - timeElapsedSinceSelfDeletionStartDate.inWholeMilliseconds
+            }
```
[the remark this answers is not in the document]
trOnk12: how is this better ? 🤔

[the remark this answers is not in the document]
typfel: Updated suggestion, now without `!!`

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,90 @@
...
+
+            messageRepository.getMessageById(conversationId, messageId).map { message ->
+                if (message is Message.Regular) enqueueSelfDeletion(message = message)
+            }
+        }
+    }
+
+    private suspend fun enqueueSelfDeletion(message: Message.Regular) {
+        if (message.isEphemeralMessage) {
+            val expirationData = message.expirationData!!
```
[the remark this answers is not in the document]
trOnk12: same here, how is that better or requires this to change ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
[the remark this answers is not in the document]
trOnk12: I think 

  return if (!hasDeletionStartedInThePast()) {
                expireAfterMillis
            } else {
                val timeElapsedSinceSelfDeletionStartDate = Clock.System.now() - selfDeletionStartDate!!

                expireAfterMillis - timeElapsedSinceSelfDeletionStartDate.inWholeMilliseconds
            }
            
            is more self explaining

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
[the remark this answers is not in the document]
typfel: I generally try to avoid `!!` since it means instant crash if the value is not there. I guess in this case it can not happen since the `isEphemeralMessage = expirationData != null`.

But I actually have to go and look that up.

Why not replace `if (message.isEphemeralMessage) {}` message.expirationData?.let {} and let the compiler guarantee that the value is always there.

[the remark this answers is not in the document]
trOnk12: in this case I agree that it does not change anything, but in some cases I like to make the language of what is going on more explicit, but in this case it indeed does not change any readibility

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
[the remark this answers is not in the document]
typfel: Yes, I agree it more reads better but it has a `!!`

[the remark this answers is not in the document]
trOnk12: I guess it is the price we have to pay if we want to keep it, personally I would keep it

typfel (approved):
Looking good now but there's quite a few TODOs and missing tests. Maybe this should be merge into a  self deletion epic branch?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
```diff
@@ -313,6 +320,27 @@ sealed interface Message {
             }
     }
 
+    data class ExpirationData(val expireAfter: Duration, val selfDeletionStatus: SelfDeletionStatus = SelfDeletionStatus.NotStarted) {
+
+        sealed class SelfDeletionStatus {
```
typfel, line 325: Nice solution!

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
```diff
@@ -361,6 +358,32 @@ class ProtoContentMapperImpl(
...
+                        content = ephemeralContent.value.content,
+                        mentions = ephemeralContent.value.mentions,
+                        quote = ephemeralContent.value.quote,
+                        expectsReadConfirmation = ephemeralContent.value.expectsReadConfirmation
+                    )
+                )
+                unpackText(genericMessageTextContent)
+            }
+
+            // TODO: Mateusz : for now we ignore anything else then Text
```
vitorhugods, line 378: From an unpacking point of view, we could handle at least `Asset` for now, no?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EnqueueMessageSelfDeletionUseCase.kt
vitorhugods: Still

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
```diff
@@ -0,0 +1,89 @@
...
+    override fun enqueuePendingSelfDeletionMessages() {
+        launch {
+            messageRepository.getEphemeralMessagesMarkedForDeletion()
+                .onSuccess { ephemeralMessages ->
+                    ephemeralMessages.forEach { ephemeralMessage ->
+                        launch {
+                            if (ephemeralMessage is Message.Regular) {
+                                enqueueSelfDeletion(message = ephemeralMessage)
+                            }
+                        }
```
vitorhugods, line 84: I would expect that the function `enqueueSelfDeletion` would have a `launch` or similar inside of it anyway. _i.e._ no need to wrap it around a launch.

Can we move this `launch` block to inside the `enqueueSelfDeletion` function and keep the same behaviour similar between these?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/PersistMigratedMessagesUseCase.kt
```diff
@@ -51,7 +51,7 @@ interface PersistMigratedMessagesUseCase {
     ): Either<CoreFailure, Unit>
 }
 
-internal class PersistMigratedMessagesUseCaseImpl @OptIn(ExperimentalCoroutinesApi::class) constructor(
+internal class PersistMigratedMessagesUseCaseImpl constructor(
```
vitorhugods, line 54: Nitpick:
```suggestion
internal class PersistMigratedMessagesUseCaseImpl(
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/ApplicationMessageHandler.kt
vitorhugods: Typo remains

vitorhugods (commented):
Looks quite good to me, just as @typfel said there are some TODOs and missing tests.

Also, I couldn't help but noticed that there are quite a few of comments marked as resolved but they don't seem to be addressed, so I flipped them back here.

Make sure the changes are already in GitHub before marking them as resolved and requesting new reviews. Maybe missing a push?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
trOnk12: I plan to handle it next, after covering it with Unit tests

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
trOnk12: I do not know why I missed that :S yes thank u

trOnk12:
> Looks quite good to me, just as @typfel said there are some TODOs and missing tests.
> 
> Also, I couldn't help but noticed that there are quite a few of comments marked as resolved but they don't seem to be addressed, so I flipped them back here.
> 
> Make sure the changes are already in GitHub before marking them as resolved and requesting new reviews. Maybe missing a push?

Do not worry I will not forget about the issues mentioned here :P it is just that it was a simple POC that did not require stuff mentioned

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/EphemeralMessageDeletionHandler.kt
vitorhugods, line 1: I've just noticed this is not covered by tests.

It should be doable with
```kotlin
@Test 
fun givenMessageIsScheduledForDeletion_whenTimePasses_thenItShouldBeDeleted() = runTest(testDispatcher) {
    // Prepare message for deletion
    // Make sure the EphemeralMessageDeletionManager gets the same dispatcher used by the testDispatcher

    advanceTimeBy(timeForDeletion + 1.second)

    // Assert it was deleted
}

@Test 
fun givenMessageIsScheduledForDeletion_whenNotTimeHasPassed_thenItShouldNotBeDeletedYet() = runTest(testDispatcher) {
    // Prepare message for deletion
    // Make sure the EphemeralMessageDeletionManager gets the same dispatcher used by the testDispatcher

    advanceTimeBy(timeForDeletion - 1.second)

    // Assert it was NOT deleted yet
}
```

trOnk12: hey Vitor, thanks fo the input, but I will cover it with test with a second PR

PR #1554, opened 2023-03-02
chore: improve logging

tmspzz:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

`toString()` is abused to sometimes be an actual string representation, and sometimes to be a more complex representation for logging. Before disaster strikes and one is used in place of another, we should avoid confusion.

### Solutions

leave `toString()` alone and use `toLogString()` when logging

----

----
##### References

### DIFF
--- a/logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
+++ b/logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
@@ -26,8 +26,6 @@ import com.wire.kalium.calling.callbacks.MetricsHandler
 import com.wire.kalium.calling.callbacks.ReadyHandler
 import com.wire.kalium.calling.types.Handle
 import com.wire.kalium.calling.types.Uint32_t
-import com.wire.kalium.logger.obfuscateDomain
-import com.wire.kalium.logger.obfuscateId
 import com.wire.kalium.logic.cache.SelfConversationIdProvider
 import com.wire.kalium.logic.callingLogger
 import com.wire.kalium.logic.data.call.CallClient
@@ -117,7 +115,7 @@ class CallManagerImpl internal constructor(
     }
     private val userId: Deferred<UserId> = scope.async(start = CoroutineStart.LAZY) {
         userRepository.observeSelfUser().first().id.also {
-            callingLogger.d("$TAG - userId ${it.value.obfuscateId()}@${it.domain.obfuscateDomain()}")
+            callingLogger.d("$TAG - userId ${it.toLogString()}")
         }
     }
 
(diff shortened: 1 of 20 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/Message.kt
```diff
@@ -287,6 +288,24 @@ sealed interface Message {
     sealed class EditStatus {
         object NotEdited : EditStatus()
         data class Edited(val lastTimeStamp: String) : EditStatus()
+
+        override fun toString(): String = when (this) {
```
tmspzz, line 292: I added this toString function. IDK if this breaks something as before it would just used a default string rep

logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
```diff
@@ -188,7 +192,7 @@ class CallManagerImpl internal constructor(
         message: Message.Signaling,
         content: MessageContent.Calling,
     ) = withCalling {
-        callingLogger.i("$TAG - onCallingMessageReceived called")
+        callingLogger.i("$TAG: { \"message\" : ${message.toLogString()}}")
```
saleniuk, line 195: Maybe keep the same information as before to know where this message log came from by looking at the logs 😄 

```suggestion
        callingLogger.i("$TAG - onCallingMessageReceived called: { \"message\" : ${message.toLogString()}}")
```

typfel: We will the message twice now, once in `ApplicationMessageHandler` and again here.  I don't think we need to log it here.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
```diff
@@ -524,11 +566,12 @@ sealed class Event(open val id: String, open val transient: Boolean) {
...
             val timestampIso: String = DateTimeUtil.currentIsoDateTimeString() // TODO we are not receiving it from API
         ) : User(id, transient) {
-            override fun toString(): String {
+            override fun toLogString(): String {
                 val properties = mapOf(
-                    "id" to id.obfuscateId(),
-                    "userId" to "${userId.value.obfuscateId()}@${userId.domain.obfuscateDomain()}",
-                    "timestampIso" to timestampIso
+                    typeKey to this::class.qualifiedName,
+                    idKey to id.obfuscateId(),
```
typfel, line 572:
```suggestion
                    idKey to id.toLogString(),
```

logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
tmspzz: is it logged on ApplicationMessageHandler tho? doesn't hat just to Message.Regular?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
tmspzz: `id` is just `String` here. That wouldn't work

typfel: One potential benefit about the introduction of `toLogString` is that we could add a parameter `fun toLogString(safeLoggingEnabled: Boolean = Logging.safeLoggingEnabled)`. which would allow us to turn off obfuscation when running builds locally.

This only works though if we always use `toLogString()` consistently at all levels.

typfel (changes requested):
Great change! With more potential. 

Some small comments on duplicated logs and about consistently using `toLogString`

logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
typfel: Right, then all is good

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
typfel: Mm, I guess it's fine if the id is one off.

If we ever want do the "safeLoggingEnabled" feature we could introduce a value class in cases like this.

typfel (approved):

PR #1570, opened 2023-03-08
feat: automatically reuse 2fa codes when registering client (AR-3088)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When logging in, the user might need to input a 2FA code, and this 2FA code should be re-used when registering the client.

### Solutions

Add a in-memory storage of 2FA code using emails as keys.
The storage is unique per authentication scope, meaning that Kalium will hold a storage per backend.

When successfully logging in: persist the 2FA code in the storage.
When registering the client, check for existing 2FA code if there's no input from client.

In case of invalid 2FA code (user mistyped, or 2FA code is too old) clear whatever is in the storage and return the failure so the client can tell the user that a new 2FA code is being sent to their email address.

> **Note**: The more I look into `UserSessionScope`, `AuthenticationScope`, etc. the more I want to refactor and clean it up. One year after starting the project, we know more clearly what's needed, and there's plenty of space to simplify and make it cleaner now. Providing a `AuthenticationScope` to the `UserSessionScope` seems kind of a no-brainer, as the `User` should be able to access the scope it used to Authenticate in the past, but I was met with surprising friction in doing it.

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/auth/verification/SecondFactorVerificationRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/auth/verification/SecondFactorVerificationRepository.kt
@@ -21,6 +21,7 @@ import com.wire.kalium.logic.NetworkFailure
 import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.logic.wrapApiRequest
 import com.wire.kalium.network.api.base.unauthenticated.VerificationCodeApi
+import io.ktor.util.collections.ConcurrentMap
 
 interface SecondFactorVerificationRepository {
 
@@ -29,11 +30,39 @@ interface SecondFactorVerificationRepository {
         verifiableAction: VerifiableAction,
     ): Either<NetworkFailure, Unit>
 
+    /**
+     * Stores the verification code in the local storage for the given email address.
+     */
+    fun storeVerificationCode(
+        email: String,
+        verificationCode: String
+    )
+
+    /**
+     * @return the stored verification code for the given email address, or null if there is no stored code.
+     * @see storeVerificationCode
+     */
+    fun getStoredVerificationCode(
+        email: String
+    ): String?
+
+    /**
+     * Clears the stored verification code for the given email address.
+     * @see storeVerificationCode
+     * @see getStoredVerificationCode
+     */
+    suspend fun clearStoredVerificationCode(
+        email: String
+    )
 }
+
+private typealias Email = String
 
 internal class SecondFactorVerificationRepositoryImpl(
     private val verificationCodeApi: VerificationCodeApi
 ) : SecondFactorVerificationRepository {
+
+    private val verificationCodeStorage = ConcurrentMap<Email, String>()
 
     override suspend fun requestVerificationCode(
         email: String,
@@ -47,4 +76,15 @@ internal class SecondFactorVerificationRepositoryImpl(
         verificationCodeApi.sendVerificationCode(email, action)
     }
 
+    override fun storeVerificationCode(email: String, verificationCode: String) {
+        verificationCodeStorage[email] = verificationCode
+    }
+
+    override fun getStoredVerificationCode(email: String): String? {
+        return verificationCodeStorage[email]
+    }
+
+    override suspend fun clearStoredVerificationCode(email: String) {
+        verificationCodeStorage.remove(email)
+    }
 }
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/SessionManagerExt.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/SessionManagerExt.kt
@@ -15,17 +15,14 @@
  * You should have received a copy of the GNU General Public License
  * along with this program. If not, see http://www.gnu.org/licenses/.
  */
-
 package com.wire.kalium.logic.feature
 
-import com.wire.kalium.logic.feature.auth.AuthenticationScope
-import com.wire.kalium.logic.sync.UserSessionWorkScheduler
-import com.wire.kalium.network.networkContainer.AuthenticatedNetworkContainer
+import com.wire.kalium.logic.data.auth.login.ProxyCredentials
+import com.wire.kalium.logic.di.MapperProvider
+import com.wire.kalium.network.session.SessionManager
 
-class AuthenticatedDataSourceSet(
-    val authenticatedRootDir: String,
-    val authenticatedNetworkContainer: AuthenticatedNetworkContainer,
-    val authenticationScope: AuthenticationScope,
-    val proteusClientProvider: ProteusClientProvider,
-    val userSessionWorkScheduler: UserSessionWorkScheduler
-)
+internal fun SessionManager.getProxyCredentials(): ProxyCredentials? =
+    MapperProvider.sessionMapper().fromDTOToProxyCredentialsModel(proxyCredentials())
+
+internal fun SessionManager.getServerConfig() =
+    MapperProvider.serverConfigMapper().fromDTO(serverConfig())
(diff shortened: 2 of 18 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
```diff
@@ -133,16 +136,23 @@ internal class LoginUseCaseImpl internal constructor(
...
                 when (it) {
                     is NetworkFailure.ProxyError -> AuthenticationResult.Failure.SocketError
-                    is NetworkFailure.ServerMiscommunication -> handleServerMiscommunication(it)
+                    is NetworkFailure.ServerMiscommunication -> handleServerMiscommunication(it, isEmail, cleanUserIdentifier)
                     is NetworkFailure.NoNetworkConnection, NetworkFailure.FederatedBackendFailure ->
                         AuthenticationResult.Failure.Generic(it)
                 }
             }, {
+                if (isEmail && secondFactorVerificationCode != null) {
+                    secondFactorVerificationRepository.storeVerificationCode(cleanUserIdentifier, secondFactorVerificationCode)
```
typfel, line 145: Why is the 2fa only stored when logging in with e-email? Shouldn't we re-use the 2fa code also when logging in with the handle?

The get back the self user when logging in so we should have all the data, no?

typfel (changes requested):

typfel (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
typfel: Ok, strange. micro services something something

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/auth/verification/SecondFactorVerificationRepository.kt
```diff
@@ -29,12 +30,40 @@ interface SecondFactorVerificationRepository {
         verifiableAction: VerifiableAction,
     ): Either<NetworkFailure, Unit>
 
+    /**
+     * Stores the verification code in the local storage for the given email address.
+     */
+    fun storeVerificationCode(
+        email: String,
```
mchenani, line 37: can we use the userId here instead of email? is it always the email here?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/SessionManagerExt.kt
```diff
@@ -0,0 +1,28 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.kalium.logic.feature
+
+import com.wire.kalium.logic.data.auth.login.ProxyCredentials
+import com.wire.kalium.logic.di.MapperProvider
+import com.wire.kalium.network.session.SessionManager
+
+internal fun SessionManager.getProxyCredentials(): ProxyCredentials? =
+    MapperProvider.sessionMapper().fromDTOToProxyCredentialsModel(proxyCredentials())
```
mchenani, line 25: since its an extension function the Proxy Credentials is suggest to have the naming only like `fromDTO`; what do you think?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
```diff
@@ -133,16 +136,23 @@ internal class LoginUseCaseImpl internal constructor(
             .fold({
                 when (it) {
                     is NetworkFailure.ProxyError -> AuthenticationResult.Failure.SocketError
-                    is NetworkFailure.ServerMiscommunication -> handleServerMiscommunication(it)
+                    is NetworkFailure.ServerMiscommunication -> handleServerMiscommunication(it, isEmail, cleanUserIdentifier)
                     is NetworkFailure.NoNetworkConnection, NetworkFailure.FederatedBackendFailure ->
                         AuthenticationResult.Failure.Generic(it)
                 }
             }, {
+                if (isEmail && secondFactorVerificationCode != null) {
```
trOnk12, line 144: maybe an overkill but maybe we can move it to a temp variable with more clear indication when we are storing the verification code ? like what is the reason we are storing the verification code here ? I assume it is because the state allows us to do so, but definetly a minor think

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/auth/verification/SecondFactorVerificationRepository.kt
mchenani: found the answer, 2FA only works with email!

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
```diff
@@ -106,9 +108,10 @@ internal class LoginUseCaseImpl internal constructor(
     ): AuthenticationResult {
         // remove White Spaces around userIdentifier
         val cleanUserIdentifier = userIdentifier.trim()
-
+        var isEmail = false
```
trOnk12, line 111: hmmm.. this is a bit weird ? I know that here is not much going on, but I think mutability could bite us when we are doing more stuff here, it makes stuff less predictiable and solving bugs harder, if there is a way to go around this mutability I would say lets try to remove it

mchenani (approved):
Looks good for me, but I have the same points as @trOnk12 pointed out.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LoginUseCase.kt
```diff
@@ -133,16 +136,23 @@ internal class LoginUseCaseImpl internal constructor(
...
+                }
                 it
             })
     }
 
-    private fun handleServerMiscommunication(error: NetworkFailure.ServerMiscommunication): AuthenticationResult.Failure {
+    private suspend fun handleServerMiscommunication(
+        error: NetworkFailure.ServerMiscommunication,
+        isEmail: Boolean,
+        userIdentifier: String,
```
trOnk12, line 154: this is also abit weird and unclear to me what is happening, maybe perhaps we can leave the function as it was and call this somewhere in the place where we are handelingServerMiscommuication, beecause before it was simply mapping stuff, but now it has a side effect, which I am not directly aware of.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/RegisterClientUseCase.kt
```diff
@@ -166,18 +176,24 @@ class RegisterClientUseCaseImpl @OptIn(DelicateKaliumApi::class) constructor(
...
+            RegisterClientResult.Failure.InvalidCredentials.Missing2FA
+
+        AuthenticationCodeFailure.INVALID_OR_EXPIRED_AUTHENTICATION_CODE -> {
+            userRepository.getSelfUser()?.email?.let {
+                secondFactorVerificationRepository.clearStoredVerificationCode(it)
+            }
+            RegisterClientResult.Failure.InvalidCredentials.Invalid2FA
+        }
+    }
+
```
trOnk12, line 196: same here, it has a return type so I am expecting it to simply returns something and do not have a side effect I think it would make more sense to split it in two steps and leave the handleAuthCodeFailure() side-effect free

trOnk12 (approved):
LGTM, but perhaps worth to take a look and give a thought about the places of some function calls and the mutability going on

PR #1583, opened 2023-03-16
feat: add use case for sending message editions [AR-2900]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We don't have a way to send message editions.

### Solutions

Implement use case for sending message editions which is responsible for updating the message in the local database and making a request, which if fails, then it also marks this message as failed.
Retrying the sending will be implemented in the future.

#### Test Coverage (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendEditTextMessageUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendEditTextMessageUseCase.kt
@@ -32,6 +32,7 @@ import com.wire.kalium.logic.feature.CurrentClientIdProvider
 import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.logic.functional.flatMap
 import com.wire.kalium.logic.functional.onFailure
+import com.wire.kalium.logic.functional.onSuccess
 import com.wire.kalium.logic.kaliumLogger
 import com.wire.kalium.persistence.dao.message.MessageEntity
 import com.wire.kalium.util.DateTimeUtil
@@ -107,18 +108,17 @@ class SendEditTextMessageUseCase internal constructor(
                 }
                 .flatMap {
                     messageSender.sendMessage(message)
-                        .flatMap {
+                        .onSuccess {
                             // when succeeded, we modify the message id locally
                             messageRepository.updateTextMessage(
                                 conversationId = message.conversationId,
                                 messageContent = content,
                                 newMessageId = editedMessageId,
                                 editTimeStamp = message.date
-                            )
-                        }
-                        .flatMap {
-                            // and change the status of the message
-                            messageRepository.updateMessageStatus(MessageEntity.Status.SENT, conversationId, editedMessageId)
+                            ).flatMap {
+                                // and change the status of the message
+                                messageRepository.updateMessageStatus(MessageEntity.Status.SENT, conversationId, editedMessageId)
+                            }
                         }
                 }
         }.onFailure {


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
```diff
@@ -275,6 +275,21 @@ class ProtoContentMapperImpl(
         }
     }
 
+    private fun packEdited(readableContent: MessageContent.TextEdited): GenericMessage.Content.Edited {
+        val mentions = readableContent.newMentions.map { messageMentionMapper.fromModelToProto(it) }
```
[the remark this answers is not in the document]
trOnk12: nahhh forget it i am telling bullshit :P it is good the way it is now I see all the code

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendEditTextMessageUseCase.kt
```diff
@@ -0,0 +1,121 @@
...
+                editTimeStamp = message.date
+            ).flatMap {
+                    messageRepository.updateMessageStatus(
+                        messageStatus = MessageEntity.Status.PENDING,
+                        conversationId = message.conversationId,
+                        messageUuid = generatedMessageUuid
+                    )
+                }.map { message }
+        }.flatMap { message ->
+            messageSender.sendMessage(message)
```
trOnk12, line 108: this part is a bit weird to me, why are we exactly doing map on updateMessageStatus() ? would it be enough to do not do map{ message } and just keep the rest as it is ?

saleniuk: `updateMessageStatus` returns `Unit` when succeeds, but we need to pass a `message` to send it, that's why `Unit` is mapped to `message`, but you're right and `sendMessage` can be moved up a bit to have a direct access to `message` - I changed that.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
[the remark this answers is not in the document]
saleniuk: I didn't want to mess with the already adopted naming convention for that 😄

alexandreferris (approved):

trOnk12 (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
```diff
@@ -275,6 +275,21 @@ class ProtoContentMapperImpl(
...
+        return GenericMessage.Content.Edited(
+            MessageEdit(
+                replacingMessageId = readableContent.editMessageId,
```
gongracr, line 282: Probably a dummy question, but shouldn't we keep the original message id?

gongracr (approved):
Beautifully executed 🦋

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/ProtoContentMapper.kt
saleniuk: Every edit changes the message id, I think the dilemma would be in case when you edit the same message on two clients and one of them fails - which one to choose as a final one then, for now we agreed that if the message edit fails we should already show the updated content locally on that one client, for now it will work because we don't have the "retry" mechanism so we won't be able to send an update with outdated id, but I think it's something worth discussing - how other platforms do that, should we keep all previous ids of such message in that case or maybe the backend handles that for us. 🤔

typfel: This is problematic already now, consider this:

1. AR client1 edits message M1 but it fails
2. AR client2 also edits message M1 and succeds
3. Upon processing the incoming edit on client1 it can't apply the edit but the message ID has already been updated with the failed edit.

How can we solve this?

### Idea 1
When sending an edit, don't update the message ID until message edit signal message has been accepted by the backend.

saleniuk: Yep, I checked that by editing message at the same time on web and scala app and it created two different messages in the end so it's not handled on other platforms as well.

> When sending an edit, don't update the message ID until message edit signal message has been accepted by the backend.

@typfel but then it will behave like this:
- we have a message with text "A"
- on device 1 we update it to "B" and it fails
- on device 2 we update it to "C" and it succeeds
- we get internet back on device 1, it receives the update from device 2 and changes the text of this message to "B"
- we lose the edit that we did on device 1

saleniuk: I think the question is: what should be more important in this scenario - the pending edit that user created earlier on device 1 or newer edit from device 2 already accepted by backend and it should update the previous failed update?

typfel: Yes, we'll lose the edit but at least everything is consistent.

If we want improve this we would need to introduce something like an edit draft which is stored separately from the original message we want to edit. This is way we could cancel a failed edit and revert back to the original message. But this probably out scope for the release.

But if even if have an edit draft it's still important to only update original message ID after an edit has been accepted by the. BE otherwise we run into consistency issues when messages are edited multiple devices.

typfel: > I think the question is: what should be more important in this scenario - the pending edit that user created earlier on device 1 or newer edit from device 2 already accepted by backend and it should update the previous failed update?

Yep, this might need input from product. 

My hunch is that it's fine to overwrite failed edits since it's the same person doing the edits, so he/she actively decides edit again on another device and would expect that to have the same state synchronised across all devices after that.

saleniuk: Ok, you're right.
I think we could improve it for now by changing if the incoming edit timestamp is older or newer than our local pending one - if newer then we use the incoming one, if older then we update the message id only and keep the content of local pending one, but maybe that's an overkill for now. 😅

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendEditTextMessageUseCase.kt
```diff
@@ -83,32 +83,46 @@ class SendEditTextMessageUseCase internal constructor(
...
+                        messageStatus = MessageEntity.Status.PENDING,
+                        conversationId = message.conversationId,
+                        messageUuid = originalMessageId
+                    )
+                }
+                .flatMap {
+                    messageSender.sendMessage(message)
+                        .flatMap {
+                            // when succeeded, we modify the message id locally
+                            messageRepository.updateTextMessage(
```
typfel, line 112: There's  a race condition here:

1. Client1 sends edit (id: 2) for message1 (inflight)
2. Receives. & process  incoming edit (id: 3) for message1 from client2
3. Client1 edit is accepted by backend
4. We update the id of message1 to 2 

Step 4 is wrong, all other clients will reject client1's edit since there' no matching message with that id.

We should only update the message id to `editedMessageId` here if the previous message id still `originalMessageId`

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendEditTextMessageUseCase.kt
```diff
@@ -83,32 +83,46 @@ class SendEditTextMessageUseCase internal constructor(
...
+                                conversationId = message.conversationId,
+                                messageContent = content,
+                                newMessageId = editedMessageId,
+                                editTimeStamp = message.date
+                            )
+                        }
+                        .flatMap {
+                            // and change the status of the message
+                            messageRepository.updateMessageStatus(MessageEntity.Status.SENT, conversationId, editedMessageId)
```
typfel, line 121: I think we also need to do this in the `MessageTextEditHandler` since it might be pending

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendEditTextMessageUseCase.kt
saleniuk: I already added that:
https://github.com/wireapp/kalium/pull/1583/files#diff-dad9f1e934e93a119a9e21d403005607074ed87f1f1507e2ffe2ebc8fdfae896R82

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendEditTextMessageUseCase.kt
saleniuk: I believe it won't be the case, if the message id has been changed in the meantime then this `updateTextMessage` won't update it since the message with `originalMessageId` doesn't exist anymore in the DB (current message id which needs to be updated is passed in the `messageContent`), so I would just change the `flatMap` to `onSuccess` to not fail the whole process if the `updateTextMessage` doesn't return `Either.Right`. What do you think?

saleniuk: I believe it won't be the case, if the message id has been changed in the meantime then this `updateTextMessage` won't update it since the message with `originalMessageId` doesn't exist anymore in the DB (current message id which needs to be updated is passed in the `messageContent`), so I would just change the `flatMap` to `onSuccess` to not fail the whole process if the `updateTextMessage` doesn't return `Either.Right`. What do you think?

typfel: You're right! Sounds good

PR #1609, opened 2023-03-30
fix(rc): federated user data is not refreshed (AR-3249)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In federated environments data for users is not refreshed

### Causes (Optional)

We are not getting updates for names, and in case a backend is down, that info is never refreshed.

### Solutions

When requesting the stored user info —only in case of a federated user— we fetch and store the data. 
This also adds an in-memory cache to avoid excessive requests.

#### Test Coverage (Optional) 

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
@@ -18,35 +18,21 @@
 
 package com.wire.kalium.logic.data.user
 
-import com.wire.kalium.logic.CoreFailure
-import com.wire.kalium.logic.NetworkFailure
-import com.wire.kalium.logic.StorageFailure
+import com.wire.kalium.logic.*
 import com.wire.kalium.logic.data.conversation.MemberMapper
 import com.wire.kalium.logic.data.conversation.Recipient
 import com.wire.kalium.logic.data.event.Event
-import com.wire.kalium.logic.data.id.ConversationId
-import com.wire.kalium.logic.data.id.IdMapper
-import com.wire.kalium.logic.data.id.QualifiedIdMapper
-import com.wire.kalium.logic.data.id.TeamId
-import com.wire.kalium.logic.data.id.toApi
-import com.wire.kalium.logic.data.id.toDao
-import com.wire.kalium.logic.data.id.toModel
+import com.wire.kalium.logic.data.id.*
 import com.wire.kalium.logic.data.publicuser.PublicUserMapper
 import com.wire.kalium.logic.data.session.SessionRepository
 import com.wire.kalium.logic.data.team.Team
 import com.wire.kalium.logic.data.team.TeamMapper
 import com.wire.kalium.logic.data.user.type.DomainUserTypeMapper
 import com.wire.kalium.logic.data.user.type.UserEntityTypeMapper
+import com.wire.kalium.logic.data.user.type.isFederated
 import com.wire.kalium.logic.di.MapperProvider
 import com.wire.kalium.logic.failure.SelfUserDeleted
-import com.wire.kalium.logic.functional.Either
-import com.wire.kalium.logic.functional.flatMap
-import com.wire.kalium.logic.functional.fold
-import com.wire.kalium.logic.functional.map
-import com.wire.kalium.logic.functional.mapRight
-import com.wire.kalium.logic.kaliumLogger
-import com.wire.kalium.logic.wrapApiRequest
-import com.wire.kalium.logic.wrapStorageRequest
+import com.wire.kalium.logic.functional.*
 import com.wire.kalium.network.api.base.authenticated.self.ChangeHandleRequest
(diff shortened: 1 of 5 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -139,70 +153,91 @@ internal class UserDataSource internal constructor(
     private suspend fun updateSelfUserProviderAccountInfo(userDTO: UserDTO): Either<StorageFailure, Unit> =
         sessionRepository.updateSsoIdAndScimInfo(userDTO.id.toModel(), idMapper.toSsoId(userDTO.ssoID), userDTO.managedByDTO)
 
+    override suspend fun getKnownUser(userId: UserId): Flow<OtherUser?> =
+        userDAO.getUserByQualifiedID(qualifiedID = userId.toDao())
+            .map { userEntity ->
+                userEntity?.let {
+                    publicUserMapper.fromDaoModelToPublicUser(userEntity)
+                }
+            }.onEach { otherUser ->
```
Garzas, line 162: Maybe extract it to separate private function to make it more readable 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -139,70 +153,91 @@ internal class UserDataSource internal constructor(
...
+                // only in case of federated users and if it's expired from cache, we fetch and refresh the user info.
+                if (otherUser != null && otherUser.userType.isFederated()
+                    && federatedUsersExpirationCache[userId]?.let { DateTimeUtil.currentInstant() > it } != false
+                ) {
+                    fetchUserInfo(userId).also { kaliumLogger.d("Federated user, refreshing user info from API after 5 minutes") }
+                    federatedUsersExpirationCache[userId] = DateTimeUtil.currentInstant().plus(5.minutes)
```
Garzas, line 168: Extract time to static variable to companion and also use it in log message

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -139,70 +153,91 @@ internal class UserDataSource internal constructor(
...
+        )
+    }.flatMap { listUserProfileDTO -> persistUsers(listUserProfileDTO) }
+
+    override suspend fun fetchUserInfo(userId: UserId) =
+        wrapApiRequest { userDetailsApi.getUserInfo(userId.toApi()) }
+            .flatMap { userProfileDTO -> persistUsers(listOf(userProfileDTO)) }
+
+    private suspend fun persistUsers(listUserProfileDTO: List<UserProfileDTO>) = wrapStorageRequest {
+        val selfUser = getSelfUser()
+        val selfUserTeamId = selfUser?.teamId?.value
```
Garzas, line 212: Can we get this information by `SelfTeamIdProvider`?

Garzas (approved):
Nice one 🥇 Just comments about some code refactor

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -139,78 +156,105 @@ internal class UserDataSource internal constructor(
...
+            }.onEach { otherUser ->
+                processFederatedUserRefresh(userId, otherUser)
+            }
+
+    /**
+     * Only in case of federated users and if it's expired or not cached, we fetch and refresh the user info.
+     */
+    private suspend fun processFederatedUserRefresh(userId: UserId, otherUser: OtherUser?) {
+        if (otherUser != null && otherUser.userType.isFederated()
+            && federatedUsersExpirationCache[userId]?.let { DateTimeUtil.currentInstant() > it } != false
```
saleniuk, line 172: I wonder if we should use `ConcurrentMap.compute()` and put fetching user inside for the action to be performed atomically, because now reading and updating this map is in two separate actions so in theory two different threads can get in between, right?

yamilmedina: Hey good point! I think in that case, it would make sense to change the implementation and have it cached only once 🤔  Because we don't have a `compute` method in `ConcurrentMap` only a `computeIfAbsent`, and that not serves this purpose, because I would not be able to expire it based on time. And as Mr. T requested, I will not implement a mechanism to keep this up to date.

So I think we have 3 options:
- Not worrying about this case, I'm not seeing a real bad effect in case it happens, please help me here to see the downside.
- Change the implementation and cached in memory only, could be a, so it will be refreshed only once when the app is opened, could be a `ConcurrentSet` instead.
- Not caching at all, and that I would prefer not to do it, could be painful for the performance for someone working with federated users.

What do you think ?

alexandreferris (approved):

PR #1610, opened 2023-03-30
feat: Broadcast self availability status to others (AR-1752)

borichellow:
# What's new in this PR?

Added `fun qualifiedBroadcastMessage` into `MessageApi` which is quit similar to `qualifiedSendMessage` but does not require `ConversationId` to be provided. 
It supposes to be user for broadcasting messages that are not related to any specific conversation but have an exact list of recipients that should receive the message. 
And used it for broadcasting self-availability status changes in `UpdateSelfAvailabilityStatusUseCase`.


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -93,8 +94,7 @@ internal interface UserRepository {
     suspend fun updateSelfUserAvailabilityStatus(status: UserAvailabilityStatus)
     suspend fun updateOtherUserAvailabilityStatus(userId: UserId, status: UserAvailabilityStatus)
     fun observeAllKnownUsersNotInConversation(conversationId: ConversationId): Flow<Either<StorageFailure, List<OtherUser>>>
-    suspend fun getUsersFromTeam(teamId: TeamId): Either<StorageFailure, List<OtherUser>>
-    suspend fun getTeamRecipients(teamId: TeamId): Either<CoreFailure, List<Recipient>>
+    suspend fun getAllRecipients(): Either<CoreFailure, Pair<List<Recipient>, List<Recipient>>>
```
borichellow, line 97: Add some doc here, to clear up what's the `Pair` is

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -207,6 +238,29 @@ internal class MessageSenderImpl internal constructor(
             }
     }
 
+    private suspend fun attemptToBroadcastWithProteus(
+        message: BroadcastMessage,
+        target: BroadcastMessageTarget
+    ): Either<CoreFailure, String> {
+        return userRepository.getAllRecipients().flatMap { (teamRecipients, otherRecipients) ->
```
MohamadJaara, line 245: according to [swagger](https://staging-nginz-https.zinfra.io/api/swagger-ui/#/default/post_broadcast_proteus_messages) 
```
report_only: Takes a list of qualified UserIDs. If any clients of the listed users are missing, the message is not sent. The missing clients are reported in the response.
```
meaning as long as we have the list of user IDs we can rely on the message sender error handling to add any missing clients and there is no need to fetch the full list of clients remotely

MohamadJaara (changes requested):
one last question about letting the message sender error handler arr any missing clients

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
borichellow: `getAllRecipients()` gets clients from the DB, not remotely. 
So the error scenario would be: 
1. we get all "known clients" 
2. try sending a message to them 
3. API returns the list of the missed clients 
4. we fetch all missed clients and save them in DB
5. repeat from #1

MohamadJaara: My bad, I thought it fetches the list of clients remotely.

MohamadJaara (approved):
LGTM

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -243,31 +297,89 @@ internal class MessageSenderImpl internal constructor(
...
+        failure: CoreFailure,
+        action: String, // Send or Broadcast
+        messageLogString: String,
+        retry: suspend () -> Either<CoreFailure, String>
+    ) =
+        when (failure) {
+            is ProteusSendMessageFailure -> {
+                logger.w("Proteus $action Failure: { \"message\" : \"${messageLogString}\", \"errorInfo\" : \"${failure}\" }")
+                messageSendFailureHandler
+                    .handleClientsHaveChangedFailure(failure)
```
gongracr, line 335: What if the failure is not due to missing clients? do we want to execute this always on any kind of failure?

gongracr (approved):
LGTM too 🍇 I just left a comment about handling the missing clients always on generic errors but nothing blocker.

PR #1651, opened 2023-04-18
feat(fed): offline backends - support v4 api version of list-prekeys endpoint pt.2 (AR-3171)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Follow up of part 2 of  #1647 
Handling the prekeys failed to list case, persistence, and mapping message option for `client_mismatch_strategy`.

For more context, you can have a look at the PR linked above but in a "tweet":
We are implementing offline backend support, so this adds the partial delivery handling when some clients are not possible to get. Making use of new fields in the backend and using `client_mismatch_strategy` while sending the message.

#### Test Coverage (Optional)

--> Will add

----

----

----
##### References


network/src/commonMain/kotlin/com/wire/kalium/network/api/v4/authenticated/MessageApiV4.kt
```diff
@@ -20,9 +20,70 @@ package com.wire.kalium.network.api.v4.authenticated
...
+import io.ktor.http.contentType
 
 internal open class MessageApiV4 internal constructor(
     authenticatedNetworkClient: AuthenticatedNetworkClient,
-    envelopeProtoMapper: EnvelopeProtoMapper
-) : MessageApiV3(authenticatedNetworkClient, envelopeProtoMapper)
+    private val envelopeProtoMapper: EnvelopeProtoMapper
+) : MessageApiV3(authenticatedNetworkClient, envelopeProtoMapper) {
+
+    override suspend fun qualifiedSendMessage(
```
typfel, line 43: So this was not implemented on V0? What's does it implementation on V0 do? Shouldn't we return not supported error like we do on other endpoints?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -300,10 +322,11 @@ internal class MessageSenderImpl internal constructor(
     private suspend fun trySendingProteusEnvelope(
         envelope: MessageEnvelope,
         message: Message.Sendable,
+        messageOption: MessageApi.QualifiedMessageOption,
         messageTarget: MessageTarget
```
typfel, line 326: Not ideal to have both `messageOption` and  `messageTarget` as parameters. So the reason we have both is because `messageTarget` is needed when retrying due missing clients for example to keep original intention of sending to a `.conversation` / `.client`.

One way out of this could be to:

1. Only keep `messageTarget` parameter here
2. Modify `MessageTarget` to capture ignore semantics Message.Conversation(val ignoreSome: List<UserId>?)
3. Map MessageOption inside this function
4. Drop IgnoreSome when re-trying.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -229,21 +232,40 @@ internal class MessageSenderImpl internal constructor(
...
+            .flatMap { Either.Right(usersWithoutSessions) }
+    } else {
+        Either.Right(usersWithoutSessions)
+    }
+
+    private fun getMessageOption(
+        usersWithoutSessions: UsersWithoutSessions,
+        messageTarget: MessageTarget
+    ): MessageApi.QualifiedMessageOption = when (messageTarget) {
+        is MessageTarget.Client -> MessageApi.QualifiedMessageOption.IgnoreAll
```
typfel, line 261: Should we silently not send a targeted message to specific client if we can't establish a session?

typfel (changes requested):

yamilmedina:
Closing in favor of #1662 
Addressed comments there, and changed base branch, because of API versioning support.

PR #1653, opened 2023-04-19
fix(dev): logout properly after expired session [AR-3307]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

After the session is expired, when combined with removal of the client, we get failures when trying to decode messages.

### Causes (Optional)

When the session expires, the account is only marked as logged out, but nothing else related to logging out is being done, client's data is not being cleared and the current client id is still kept which causes troubles.
After error saying that the client used is invalid or wrong event was used to fetch notifications (404), the slow sync is reset and restarted but incremental sync doesn't start which makes the app being stuck.

### Solutions

Refactor `UserSessionScopeProvider` and move `SessionManager` creation and all other code that's common for every platform directly to the `UserSessionScope`, to be also able to use `LogoutUseCase` in the `SessionManager` and logout the user properly when the `SESSION_EXPIRED` occurs.
Retry the incremental sync callback (listen for the completion of slow sync and retry incremental sync after) when the slow sync is restarted for recovery after the app gets a response with 404 meaning that wrong event or client id was used.
Clear the outdated client data when the app discovers that the retained client is no longer registered so that we don't keep the invalid "current client id" and "last processed event id" for that client.

#### How to Test

Expire the session by for instance changing tokens in shared prefs, remove the current client from another device and open the app to see how it behaves.

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/GetOrRegisterClientUseCase.kt
```diff
@@ -53,8 +58,7 @@ class GetOrRegisterClientUseCaseImpl(
                         is VerifyExistingClientResult.Success -> RegisterClientResult.Success(result.client)
                         is VerifyExistingClientResult.Failure.Generic -> RegisterClientResult.Failure.Generic(result.genericFailure)
                         is VerifyExistingClientResult.Failure.ClientNotRegistered -> {
-                            clearClientData()
-                            clientRepository.clearRetainedClientId()
+                            clearOldClientRelatedData()
```
typfel, line 61: Are we guaranteed that the currentClientId hasn't already been cached in the user session at this point?

saleniuk: It shouldn't happen because this use case should be used only when there's no registered client id yet, but I added clearing it just in case. @typfel do you think that's enough, we could collect `observeCurrentClientId` in `init` and update `_clientId` dynamically 🤔

typfel: 👍 should be good then!

typfel (approved):

PR #1654, opened 2023-04-20
fix: not all events are emitted [AR-3310]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

`combine` function is skipping some events from websocket due to Batch-receive optimization inside this function

### Causes (Optional)

When revoking guest access in conversation some member leave events are skipped which leaves some guest in conversation.

### Solutions

Added buffer to not skip any emitted events

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/EventGatherer.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/EventGatherer.kt
@@ -95,10 +95,8 @@ internal class EventGathererImpl(
         offlineEventBuffer.clear()
         _currentSource.value = EventSource.PENDING
         eventRepository.lastEventId().flatMap {
-            kaliumLogger.d("KBX flatMap")
             eventRepository.liveEvents()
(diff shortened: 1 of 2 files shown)


saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/EventGatherer.kt
```diff
@@ -73,7 +77,8 @@ internal interface EventGatherer {
 internal class EventGathererImpl(
     private val eventRepository: EventRepository,
     private val incrementalSyncRepository: IncrementalSyncRepository,
-    private val slowSyncRepository: SlowSyncRepository
+    private val slowSyncRepository: SlowSyncRepository,
+    private val dispatchers: KaliumDispatcher = KaliumDispatcherImpl,
```
saleniuk, line 81: Added `dispatchers` are not used anywhere, also the same with `slowSyncRepository` which is removed in similar PR for develop https://github.com/wireapp/kalium/pull/1656 so here it could also be removed, right?

PR #1662, opened 2023-04-24
feat: offline backends - handling list-prekeys failed to list (AR-3171)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Follow up of  #1651 
Since the base changed, implemented the PR comments in this new PR.

- GOAL: Handling the prekeys failed to list case, persistence, and mapping message option for `client_mismatch_strategy`.

For more context, you can have a look at the PR linked above but in a "tweet":
We are implementing offline backend support, so this adds the partial delivery handling when some clients are not possible to get. Making use of new fields in the backend and using `client_mismatch_strategy` while sending the message.

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
```diff
@@ -372,6 +375,20 @@ class MessageDataSource(
         })
     }
 
+    private fun MessageTarget.toOption(ignoredUsers: List<UserId>) = when (this) {
+        is MessageTarget.Client -> if (ignoredUsers.isNotEmpty()) {
+            MessageApi.QualifiedMessageOption.IgnoreSome(ignoredUsers.map { it.toApi() })
```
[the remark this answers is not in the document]
yamilmedina: I think I was lost a bit with this, but now it makes sense to me to go with option 3. Since at this point the info needed to be displayed to the user we have it already, so for `MessageTarget.Client` we can just log and go with `IgnoreAll`.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -229,19 +233,26 @@ internal class MessageSenderImpl internal constructor(
...
-            })
+            }
+    }
+
+    private suspend fun handleUsersWithNoClientsToDeliver(
+        conversationId: ConversationId,
+        messageId: String,
+        usersWithoutSessions: UsersWithoutSessions
+    ): Either<CoreFailure, UsersWithoutSessions> = if (usersWithoutSessions.hasMissingSessions()) {
+        messageRepository.persistNoClientsToDeliverFailure(conversationId, messageId, usersWithoutSessions.users)
```
typfel, line 252: What happens if we do this for a message which isn't persisted? Calling messages for example aren't persisted into the database.

The foreign key is not found and the operation is ignored?

yamilmedina: Good point ! I think given that this list serves the purpose of being shown on the UI for regular messages (In which case we can assume it will be persisted) we can ignore it for the rest.

alexandreferris (approved):
Just checked on calling and it are apparently only sending `Message.Signaling` so it shouldn't be an issue from the comments above 👌🏻

vitorhugods:
> Just checked on calling and it are apparently only sending `Message.Signaling` so it shouldn't be an issue from the comments above 👌🏻

Maybe message delete and edits for example? We try to send them only when we're online tho.

PR #1721, opened 2023-05-10
fix: scala migration can cause duplicated messages

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

crash when running migration 33.sqm 

### Causes (Optional)

from the stacktrace @vitorhugods found that it is related to this sql statment 

```sql
WITH self_domain_cte AS (SELECT substr(id, instr(id, \'@\') + 1) AS self_domain
                         FROM SelfUser)
UPDATE Message
SET sender_user_id  = CASE
                          WHEN sender_user_id LIKE \'%@\'
                              THEN (sender_user_id || (SELECT self_domain FROM self_domain_cte))
                          ELSE sender_user_id END,
    conversation_id = CASE
                          WHEN conversation_id LIKE \'%@\'
                              THEN (conversation_id || (SELECT self_domain FROM self_domain_cte))
                          ELSE conversation_id END
                          ```
the theory is for some users there can be duplicated messages (one with the correct conversation id and other with corrupt one)
we have no idea how this can happen and it never occurred when testing

### Solutions

1. it is 2 messages somehow have the same id and conv id
2. it is happening during migration meaning effected users are still on the db 32.sqm
3. the fix should be to change the 33.sqm

UPDATE OR IGNORE messages and then delete any corrupt messages left
the ignored ones will be the duplicated and at the end we delete them

----

----

----
##### References

### DIFF
--- a/persistence/src/commonMain/db_user/migrations/33.sqm
+++ b/persistence/src/commonMain/db_user/migrations/33.sqm
@@ -21,7 +21,7 @@ WHERE qualified_id LIKE '%@';
 WITH self_domain_cte AS (
 SELECT substr(id, instr(id, '@') + 1) AS self_domain FROM SelfUser
 )
-UPDATE Message
+UPDATE OR IGNORE Message
 SET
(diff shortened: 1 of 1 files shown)


vitorhugods (approved):
❤️

PR #1737, opened 2023-05-15
feat: Add listing / searching services (AR-3318)

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There was no way of searching services by name

### Causes (Optional)
Not implemented.

### Solutions

Add usecase to search services by name

### DIFF
--- a/network/src/commonMain/kotlin/com/wire/kalium/network/api/v0/authenticated/SelfApiV0.kt
+++ b/network/src/commonMain/kotlin/com/wire/kalium/network/api/v0/authenticated/SelfApiV0.kt
@@ -28,6 +28,7 @@ import com.wire.kalium.network.api.base.model.UserDTO
 import com.wire.kalium.network.exceptions.KaliumException
 import com.wire.kalium.network.session.SessionManager
 import com.wire.kalium.network.utils.NetworkResponse
+import com.wire.kalium.network.utils.flatMap
 import com.wire.kalium.network.utils.wrapKaliumResponse
 import io.ktor.client.request.get
 import io.ktor.client.request.header
@@ -35,7 +36,6 @@ import io.ktor.client.request.put
 import io.ktor.client.request.setBody
 import io.ktor.http.HttpHeaders
 import io.ktor.http.HttpStatusCode
-import io.ktor.http.isSuccess
 
 internal open class SelfApiV0 internal constructor(
     private val authenticatedNetworkClient: AuthenticatedNetworkClient,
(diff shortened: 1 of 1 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/service/ServiceRepository.kt
```diff
@@ -58,4 +89,11 @@ internal class ServiceDataSource internal constructor(
             id = serviceMapper.fromModelToDao(serviceId = serviceId),
             conversationId = conversationId.toDao()
         ).map { it?.toModel() }
+
+    private suspend fun getAllServicesAndMapFlow(): Flow<List<ServiceDetails>> =
+        serviceDAO.getAllServices().map {
+            it.map { serviceEntity ->
+                serviceMapper.fromDaoToModel(service = serviceEntity)
+            }
+        }
```
MohamadJaara, line 98:
```suggestion
    private suspend fun getAllServicesAndMapFlow(): Flow<Either<StorageFailure, List<ServiceDetails>>> =
        serviceDAO.getAllServices().map {
            it.map { serviceEntity ->
                serviceMapper.fromDaoToModel(service = serviceEntity)
            }
        }.wrapStorageRequest()
```

can we handle errors in this case

MohamadJaara (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/service/ObserveAllServicesUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+    suspend operator fun invoke(scope: CoroutineScope): Flow<List<ServiceDetails>>
+}
+
+class ObserveAllServicesUseCaseImpl internal constructor(
+    private val serviceRepository: ServiceRepository,
+    private val teamRepository: TeamRepository,
+    private val selfTeamIdProvider: SelfTeamIdProvider
+) : ObserveAllServicesUseCase {
+
+    override suspend fun invoke(scope: CoroutineScope): Flow<List<ServiceDetails>> {
```
trOnk12, line 48: why do we need to provide a scope from outside ?

trOnk12: but I think  you want to run sync in parallel, in that case I think coroutineScope{ } would be more fitting to this ?

vitorhugods: Why not just having it like we do for other use cases/viewmodel, as suggested by @trOnk12 ? 

```kotlin
viewModelScope.launch {
    observeAllServices()
}
```

And, as @trOnk12 suggested:

```kotlin
override suspend fun invoke(): Flow<List<ServiceDetails>> {
    return coroutineScope {
        launch(Job()) { // sync stuff }
        
        serviceRepository.observeAllServices().map { either ->
            either.fold(
                { emptyList() },
                { it }
            )
        }
    }
}
```

Or maybe using the UserSessionScope as parameter in the UseCase constructor, and using that one to launch the sync in parallel, and return the flow early.

Kalium is pretty self-contained regarding coroutine scopes and it would be nice to keep it like that, especially considering multiplatform development.

typfel: @alexandreferris this got merged without @vitorhugods suggestion. Will this happen in a follow up PR?

PR #1740, opened 2023-05-16
feat(offline-backends): users and conversations without metadata refetch pt2. (AR-3123)

yamilmedina:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We need to refetch users and conversations without metadata.

### Causes (Optional)

In case we don't have data due to a remote backend being offline, the clients will be able to refetch this data on demand.

### Solutions

This PR adds:

- Logic and use case to refetch conversations without metadata on demand.
- Mechanism to be called to sync both, conversations and users without metadata in windows of 3 hours.

#### Test Coverage (Optional)

### Notes (Optional)

Will raise a new PR to add a new mechanism for refetch every 3 hours if the app comes to foreground.

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/MissingMetadataUpdateManager.kt
```diff
@@ -0,0 +1,75 @@
...
+                metadataDAO.insertValue(LAST_MISSING_METADATA_SYNC_KEY, currentInstant.toIsoDateTimeString())
+            }
+        }.onFailure {
+            kaliumLogger.e("Error while syncing users and conversations without metadata $it")
+        }.onSuccess {
+            kaliumLogger.d("Finished syncing users and conversations without metadata")
+        }
+    }
+
+    private suspend fun needsSync(currentInstant: Instant): Boolean {
```
typfel, line 64: Checkout the `TimestampKeyRepository` it solves exactly this.

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
```diff
@@ -289,3 +289,9 @@ WHERE qualified_id = ?;
 
 getGuestRoomLinkByConversationId:
 SELECT guest_room_link FROM Conversation WHERE qualified_id = ?;
+
+selectConversationIdsWithoutMetadata:
+SELECT qualified_id
+FROM Conversation
+WHERE type IS NOT 'SELF'
+AND (name IS NULL OR creator_id = '');
```
typfel, line 297: So we rely on the `name` being null as indicator for missing metadata? What if we have conversation without names? Have you considered adding a dedicated boolean field on the user and conversation entity which signals that metadata is missing?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -660,6 +662,14 @@ class UserSessionScope internal constructor(
         )
     }
 
+    val missingMetadataUpdateManager: MissingMetadataUpdateManager by lazy {
```
typfel, line 665: This is currently never instantiated so it will never run?

typfel (changes requested):

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
yamilmedina: I know that iOS is currently doing something similar, and I wanted to avoid this, since I saw a nearly impossible way to have an empty named conversation 🤔 

nvm, I can add the field

typfel: > I know that iOS is currently doing something similar, and I wanted to avoid this, since I saw a nearly impossible way to have an empty named conversation 🤔
> 
> nvm, I can add the field

We never now what happens in future.

Btw 1:1 team conversations are group conversations without a name.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
typfel: Why not run it after the incremental finishes like many other period tasks. Then we don't need any platform  specific code.

yamilmedina: Makes sense to me, and sounds better to reuse the mechanism we already have in place.

typfel (approved):
🚀 🌔

PR #1743, opened 2023-05-16
refactor: store assets locally before the upload and when it fails [AR-2408]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently, the asset data is stored in DB only after the asset file is uploaded, so we can't show such asset with error label if sent when offline for instance.

### Solutions

Persist the asset that's  being uploaded in the "file" directory already instead of temporary "cache" directory and add it with a temporary id/key to the DB. After uploading, the temporary data is being replaced with the data from the remote.
Do not try to download the asset when fetching if it's not yet uploaded and instead just take it from the local storage.
Move scheduling sending of pending messages to the `MessageSendFailureHandler` to keep handling failures in one place.

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendKnockUseCase.kt
```diff
@@ -78,16 +74,11 @@ class SendKnockUseCase internal constructor(
                 isSelfMessage = true
             )
             persistMessage(message)
-                .flatMap {
-                    messageSender.sendMessage(message)
-                }
-                .onSuccess {
-                    messageRepository.updateMessageStatus(MessageEntity.Status.SENT, conversationId, generatedMessageUuid)
```
saleniuk, line 85: This is already done in `MessageSender` (`promoteMessageToSentUpdatingServerTime`)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendTextMessageUseCase.kt
```diff
@@ -103,16 +99,11 @@ class SendTextMessageUseCase internal constructor(
                 isSelfMessage = true
             )
             persistMessage(message)
-                .flatMap {
-                    messageSender.sendMessage(message)
-                }
-                .onSuccess {
-                    messageRepository.updateMessageStatus(MessageEntity.Status.SENT, conversationId, generatedMessageUuid)
```
saleniuk, line 110: This is already done in `MessageSender` (`promoteMessageToSentUpdatingServerTime`)

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetRepository.kt
```diff
@@ -164,30 +183,39 @@ internal class AssetDataSource(
...
+
+                    // we should also consider for avatar images, the compression for preview vs complete picture
+                    assetApi.uploadAsset(metaData, { dataSource }, uploadAssetData.dataSize)
+                        .also { dataSource.close() }
+                }
+            }
+            .flatMap { assetResponse ->
+                // After successful upload, we persist the asset to a persistent path
+                persistAsset(assetResponse.key, assetResponse.domain, decodedDataPath, uploadAssetData.dataSize, extension)
+                    .also { kaliumFileSystem.delete(uploadAssetData.tempEncryptedDataPath) }
```
Garzas, line 199: Should we delete path when persistAsset() fails?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetMessageAssetUseCase.kt
```diff
@@ -72,7 +74,10 @@ internal class GetMessageAssetUseCaseImpl(
             CompletableDeferred(MessageAssetResult.Failure(it))
         }, { message ->
             val assetDownloadStatus = (message.content as MessageContent.Asset).value.downloadStatus
+            val assetUploadStatus = (message.content as MessageContent.Asset).value.uploadStatus
```
Garzas, line 77: maybe create variable to not cast twice

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
```diff
@@ -109,66 +113,108 @@ internal class ScheduleNewAssetMessageUseCaseImpl(
...
+        assetDataSize: Long,
+        assetName: String,
+        assetMimeType: String,
+        assetWidth: Int?,
+        assetHeight: Int?,
+        expireAfter: Duration?,
+        expectsReadConfirmation: Boolean
+    ): Either<CoreFailure, Pair<AssetMessageMetadata, Message.Regular>> = currentClientIdProvider().flatMap { currentClientId ->
+        // Create a temporary asset key and domain
+        val (generatedAssetUuid, tempAssetDomain) = uuid4().toString() to ""
```
Garzas, line 163: why we are passing empty domain?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSendFailureHandler.kt
```diff
@@ -80,13 +83,18 @@ class MessageSendFailureHandlerImpl internal constructor(
...
         conversationId: ConversationId,
         messageId: String,
-        messageType: String
+        messageType: String,
+        scheduleResendIfNoNetwork: Boolean
     ) {
-        when (failure) {
-            is NetworkFailure.FederatedBackendFailure -> {
+        when {
+            failure is NetworkFailure.FederatedBackendFailure -> {
```
Garzas, line 90: Why failure was moved from when clause?

Garzas (approved):
🚀  Just small questions

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetMessageAssetUseCase.kt
```diff
@@ -72,7 +74,10 @@ internal class GetMessageAssetUseCaseImpl(
...
             val wasDownloaded: Boolean = assetDownloadStatus == SAVED_INTERNALLY || assetDownloadStatus == SAVED_EXTERNALLY
+            // assets uploaded by other clients have upload status NOT_UPLOADED
+            val alreadyUploaded: Boolean = assetUploadStatus == NOT_UPLOADED || assetUploadStatus == UPLOADED
```
gongracr, line 80: hmmm, this might be a bit dangerous. How can we differentiate between an asset not uploaded from a different client and my current one? shouldn't we add an extra check here?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetRepository.kt
saleniuk: For uploading private assets, this is a temporary file located in cache directory so it will be removed anyway sooner or later but when we try to do it again, we will execute `uploadAndPersistPrivateAsset` which will store it in a cache again before sending so it's safer to clear any files before retrying. For public assets we probably won't have the retry mechanism so after it fails, we will need to do that all over again which also means create another temporary file so it's probably ok to delete it also when it fails.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetMessageAssetUseCase.kt
saleniuk: Right, I even changed the order and started with `when`, because we don't allow any content other than `MessageContent.Asset` here

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
saleniuk: It's just temporary to store asset file locally before it's uploaded so the domain shouldn't matter and generated unique asset id is enough because we will be fetching such asset only locally until it's uploaded.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSendFailureHandler.kt
saleniuk: Because the second case is not only checking the `failure` type but also `scheduleResendIfNoNetwork`:
```
failure is NetworkFailure.NoNetworkConnection && scheduleResendIfNoNetwork -> {
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/GetMessageAssetUseCase.kt
saleniuk: As I described in the comment, for all assets from proto we always set `Message.UploadStatus.NOT_UPLOADED` and when the current client wants to send an asset, the first status is `UPLOAD_IN_PROGRESS` so it looks like we can differentiate:
- NOT_UPLOADED - state set only for assets received from backend which means they are already on the backend, sometimes they may be missing remote data (if it's a preview asset) so in that case I added `&& content.value.shouldBeDisplayed` to be sure, but even without that it would still work exactly the same way it used to work before this PR
- UPLOAD_IN_PROGRESS, UPLOADED, FAILED_UPLOAD - states set only for assets sent by the current client

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendTextMessageUseCase.kt
gongracr: In case we wanted to for example, log that a self deleting message has been successfully sent (and trigger the timer for self user self deleting messages) this would be the place right?

If so maybe we could leave a comment or unify all these use cases "send successfully message event" with a common function? cc @trOnk12

gongracr (approved):
Great stuff and nice refactoring 🙌 I left a small comment that could help us in the following PRs about adding structuring logs for self deleting messages.

saleniuk:
>In case we wanted to for example, log that a self deleting message has been successfully sent (and trigger the timer for self user self deleting messages) this would be the place right?
>If so maybe we could leave a comment or unify all these use cases "send successfully message event" with a common function?

I don't know why but I can't answer directly under this comment.
The thing is now we have it in one place for every use case that sends messages, it's in `messageSender.sendMessage` which uses `messageRepository.promoteMessageToSentUpdatingServerTime`. 🙂 
Mateusz said that timer for self deleting messages is started and running on the UI so there's nothing do to in this regard in kalium and we would need to pass the information whether the message has been sent to the `UIMessage` which I did in this PR: https://github.com/wireapp/wire-android-reloaded/pull/1803

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetRepository.kt
```diff
@@ -164,30 +183,39 @@ internal class AssetDataSource(
...
         extension: String?
     ): Either<CoreFailure, UploadedAssetId> =
-        assetMapper.toMetadataApiModel(uploadAssetData, kaliumFileSystem).let { metaData ->
-            wrapApiRequest {
-                val dataSource = kaliumFileSystem.source(uploadAssetData.tempEncryptedDataPath)
-
-                // we should also consider for avatar images, the compression for preview vs complete picture
-                assetApi.uploadAsset(metaData, { dataSource }, uploadAssetData.dataSize).also { dataSource.close() }
+        assetMapper.toMetadataApiModel(uploadAssetData, kaliumFileSystem)
+            .let { metaData ->
```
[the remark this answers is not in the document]
saleniuk: To be honest I don't remember, probably one of the _happy little accidents_ 👨‍🎨 😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendTextMessageUseCase.kt
saleniuk: The thing is now we have it in one place for every use case that sends messages, it's in messageSender.sendMessage which uses messageRepository.promoteMessageToSentUpdatingServerTime. 🙂
Mateusz said that timer for self deleting messages is started and running on the UI so there's nothing do to in this regard in kalium and we would need to pass the information whether the message has been sent to the UIMessage which I did in this PR: https://github.com/wireapp/wire-android-reloaded/pull/1803

PR #1750, opened 2023-05-22
fix: Duplicate Message error when receiving a mention that does not contain user id (AR-3006)

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When we receive a mention having this proto representation the app is throwing an exception DUPLICATE MESSAGE and the message is not displayed in the client

```
{
  "messageId": "$msgId",
  "text": {
    "content": "@hallo",
    "mentions": [
      {
        "start": 0,
        "length": 2,
      }
    ]
  }
}
```

### Causes (Optional)

The proto content does not contain user id at all.
When we parse that content, we assume that the id is present in and it's not optional 

### Solutions

Return empty id if there is no user Id in the mention received

Edit: 
Do not store in database any mention that is received without userId and make it displayed as simple text

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/mention/MessageMentionMapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/mention/MessageMentionMapper.kt
@@ -41,8 +41,8 @@ class MessageMentionMapperImpl(
         return MessageMention(
             start = mention.start,
             length = mention.length,
-            userId = mention.userId.toModel(),
-            isSelfMention = mention.userId.toModel() == selfUserId
+            userId = mention.userId?.toModel(),
+            isSelfMention = mention.userId?.toModel() == selfUserId
         )
     }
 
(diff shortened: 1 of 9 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/mention/MessageMentionMapper.kt
```diff
@@ -57,10 +57,10 @@ class MessageMentionMapperImpl(
     override fun fromProtoToModel(mention: Mention): MessageMention {
         val userId = mention.qualifiedUserId?.let {
             idMapper.fromProtoUserId(it)
-        } ?: UserId(
-            mention.mentionType?.value as String,
-            selfUserId.domain
-        )
+        } ?: run {
+            val id = if (mention.mentionType != null) mention.mentionType?.value as String else ""
```
MohamadJaara, line 61: can we maybe ignore the message in case it have invalid proto.
adding empty IDs can lead to unknown behaviors in the future that can be hard to detect.

MohamadJaara (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/mention/MessageMentionMapper.kt
ohassine: No It's not invalid, it can be received in this way as mention_type and qualified_user_id are optional
```
message Mention {
  required int32 start = 1; // offset from beginning of the message counting in utf16 characters
  required int32 length = 2;
  oneof mention_type {
    // deprecated. Should be set such that old clients always fail when looking
    // up the user. Ideally, this should not be a problem, as a non-federation
    // aware user should never be part of a federated conversation.
    string user_id = 3;
  }
  // only optional to maintain backwards compatibility.
  optional QualifiedUserId qualified_user_id = 4;
}
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/mention/MessageMentionMapper.kt
```diff
@@ -41,38 +41,50 @@ class MessageMentionMapperImpl(
...
 
-    override fun fromProtoToModel(mention: Mention): MessageMention {
-        val userId = mention.qualifiedUserId?.let {
-            idMapper.fromProtoUserId(it)
-        } ?: UserId(
-            mention.mentionType?.value as String,
-            selfUserId.domain
-        )
-        return MessageMention(
+    override fun fromProtoToModel(mention: Mention): MessageMention = mention.qualifiedUserId?.let {
```
typfel, line 57: How about making this return an optional `MessageMention` instead and do a `mapNotNull` when unpacking a text message.

typfel (changes requested):

MohamadJaara (approved):
🚀

typfel (approved):

PR #1756, opened 2023-05-25
fix: block UPDATE of deleted messages

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

deleted asset messages are marked as visible if they were deleted before the upload if done

### Causes (Optional)

if goes as follows 
1. send assert
2. before if finish uploads delete the message
3. the message is marked as deleted by the upload in the background is not canceled
4. once it is done in the background it will mark a deleted message as not deleted but the content is already deleted
5. KaBoom crash the app is trying to map content that does not exists

### Solutions

prevent update of already deleted asset messages

----

----
##### References


saleniuk (approved):

vitorhugods (commented):
We also need to perform a DB migration to fix the affected messages, no?

Looks good to me tho.

vitorhugods (approved):
I see the DB migration was added [here](https://github.com/wireapp/kalium/pull/1757).

PR #1759, opened 2023-05-25
fix: add extension to assets without it [AR-3261, AR-2929, AR-3482]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Some assets uploaded from iOS or web doesn't have extension in file name.

### Causes (Optional)

Sharing those assets by share intent makes them files with unknown type

### Solutions

Get mime type from asset which doesn't have extension and map it to extension suffix to the file 

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/Asset.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/Asset.kt
@@ -71,25 +71,3 @@ fun isVideoMimeType(mimeType: String): Boolean = mimeType in setOf(
     "video/mp4", "video/webm", "video/3gpp", "video/mkv"
 )
 
-fun getExtensionFromMimeType(mimeType: String?): String? {
-    return mimeType?.let { mimeTypeToExtensionMap[it.lowercase()] }
-}
-
-private val mimeTypeToExtensionMap = mapOf(
-    "image/jpg" to "jpg",
-    "image/jpeg" to "jpeg",
-    "image/png" to "png",
-    "image/heic" to "heic",
-    "image/gif" to "gif",
-    "image/webp" to "webp",
-    "audio/mp3" to "mp3",
-    "audio/mpeg" to "mpeg",
-    "audio/ogg" to "ogg",
-    "audio/wav" to "wav",
-    "audio/x-wav" to "wav",
-    "audio/x-pn-wav" to "wav",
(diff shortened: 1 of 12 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/Asset.kt
```diff
@@ -57,3 +70,26 @@ fun isAudioMimeType(mimeType: String): Boolean = mimeType in setOf(
 fun isVideoMimeType(mimeType: String): Boolean = mimeType in setOf(
     "video/mp4", "video/webm", "video/3gpp", "video/mkv"
 )
+
+fun getExtensionFromMimeType(mimeType: String?): String? {
+    return mimeType?.let { mimeTypeToExtensionMap[it.lowercase()] }
```
gongracr, line 75: I'm wondering whether it wouldn't be more effective to simply keep the last part of the mimeType, i.e. the second part after the `/jpg` since they match 99.9% of the times. Otherwise we are limiting ourselves to this small list. Zip or pdf files will not be able to be mapped. 
In general I would say we should rely on obtaining the extension file via `Uri.getMimeType()` method in the app utils folder as it's pretty reliable and supported by the Android system.

If that doesn't work because we want to centralise this logic in Kalium or we can't have access to the app `Context` and we still want to go for a list of known mimeTypes I'd suggest to expand it with more known formats also for `application/xxx` types. The official IANA list can be found [here](https://www.iana.org/assignments/media-types/media-types.xhtml).

gongracr (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/Asset.kt
saleniuk: You're right, I also talked with Kubaz about it. I created separate implementations for different platforms - on Android we use `android.webkit.MimeTypeMap`, on jvm I added `org.apache.tika.mime.MimeTypes` which is recommended, other platforms still use previous implementation, we can update the implementation for iOS or JS if needed.

gongracr (approved):
👏

saleniuk (approved):

PR #1764, opened 2023-05-29
fix: edited message pending state and creation time [AR-3495] [AR-3496]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Edited message has still a pending state even if already sent and its creation time is being updated.

### Causes (Optional)

When we send a message, we "promote" it locally to "sent" state. The problem is that edited messages have IDs updated and this update of the ID was executed after the state was changed, but the state change method was called already with the new ID, so in practise it wasn't doing anything as the message with the given ID was not yet in the local DB.
Similar case with updating time - this message is still "pending" so it adds the `delivery_duration` to its time, but for edited messages "creation time" shouldn't be changed at all.

### Solutions

Change the order and first update the ID for the edited messages and then promote its status to "sent", but don't modify the "creation time" for edited messages, only set "server time" for newly created messages.

#### Test Coverage (Optional)

----

----

----
##### References


borichellow (approved):
Terrific work and nice catch :)

PR #1765, opened 2023-05-29
fix: Self deletion logic delegated to kalium Use Cases

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
Currently the app layer when sending a message, specifies the expiration data value that will be used to determine when a message should be deleted. Therefore Kalium, just takes this values and assumes it's the correct one, allowing for misusages or human errors when deciding the value to package the message on the protobuf envelope.

### Solutions

Add an extra db query that will update and select the `Message` with the latest self deletion data.

#### Test Coverage (Optional)

----
##### References


trOnk12 (approved):

vitorhugods (commented):
Why delegate this to the Repository?

Also, I know we don't support this at the moment, but what about offline sending or scheduling of messages?

_i.e._ User tries to send while offline with timer `X`, then changes the timer to `Y`, and then tries to send another message.

When retrying, the timer of both messages will be `Y`, violating the user's will because the timer is figured out on-the-fly.

I would expect the UseCase to contain this logic and whatever the UseCase persists should be the law, and not to be altered by the sending mechanism.

I believe this kind of coupling and doubled responsibility (_i.e._ the interceptor being _too_ smart) could hunt our dreams in the future when we re-enable sending retrying.

MohamadJaara:
@vitorhugods 
The reason for delegating the logic to kalium, so it is global and will be applied even when sending from notifications.

We can move the logic up to the use case, but then the question is if a conversation has a timer of 1m and a user schedules a message for  then the team changes the timer to 4 weeks, should the scheduled message then be 1m or 4w ?

vitorhugods:
> @vitorhugods The reason for delegating the logic to kalium, so it is global and will be applied even when sending from notifications.
> 
> We can move the logic up to the use case, but then the question is if a conversation has a timer of 1m and a user schedules a message for then the team changes the timer to 4 weeks, should the scheduled message then be 1m or 4w ?

Considering the user was offline when it happened, it's understandable something like this could happen, and the design is probably debatable.
In any case, it would be easier to handle such case by updating the pending messages to follow the new team settings once a update event happens, as sending will only happen _after_ all events are processed.

Having it on-the-fly doesn't allow for such a gracious handling of such cases.

gongracr:
> Why delegate this to the Repository?
> 
> Also, I know we don't support this at the moment, but what about offline sending or scheduling of messages?
> 
> _i.e._ User tries to send while offline with timer `X`, then changes the timer to `Y`, and then tries to send another message.
> 
What's the problem there :) ? The message will always use the most recent timer set on the conversation or user settings. Otherwise, as Mo pointed out, we might be hacking our own code/enforced settings.


> When retrying, the timer of both messages will be `Y`, violating the user's will because the timer is figured out on-the-fly.
> 
I don't think this should be violating user's will, the message will be just following the current enforced settings with a little magic under the hood which is not evil per se imho. 


> I would expect the UseCase to contain this logic and whatever the UseCase persists should be the law, and not to be altered by the sending mechanism.
> 
I'm not necessarily against this approach, but following this will imply that we need to replicate this messageDao query for each use case we have now and any future ones that we add in the future (i.e. we decide at some point that reactions or pings should also be self deleted)


> I believe this kind of coupling and doubled responsibility (_i.e._ the interceptor being _too_ smart) could hunt our dreams in the future when we re-enable sending retrying.
Wouldn't this also be acceptable? when implementing retrying the message delivery we will not need to worry about the self deletion logic as this is already implemented?

vitorhugods:
> 



> > Why delegate this to the Repository?
> > Also, I know we don't support this at the moment, but what about offline sending or scheduling of messages?
> > _i.e._ User tries to send while offline with timer `X`, then changes the timer to `Y`, and then tries to send another message.
> 
> What's the problem there :) ? The message will always use the most recent timer set on the conversation or user settings. Otherwise, as Mo pointed out, we might be hacking our own code/enforced settings.
> 
> > When retrying, the timer of both messages will be `Y`, violating the user's will because the timer is figured out on-the-fly.
> 
> I don't think this should be violating user's will, the message will be just following the current enforced settings with a little magic under the hood which is not evil per se imho.


This is the problem: `The message will always use the most recent timer set on the conversation or user settings` 😄. When the user explicitly tries to do otherwise.

> 
> > I would expect the UseCase to contain this logic and whatever the UseCase persists should be the law, and not to be altered by the sending mechanism.
> 
> I'm not necessarily against this approach, but following this will imply that we need to replicate this messageDao query for each use case we have now and any future ones that we add in the future (i.e. we decide at some point that reactions or pings should also be self deleted)

The duplication would be "getting the timer setting before persisting the message". 
So... two or three lines of code? Which we can have full control of.

> > I believe this kind of coupling and doubled responsibility (_i.e._ the interceptor being _too_ smart) could hunt our dreams in the future when we re-enable sending retrying.
> 
> Wouldn't this also be acceptable? when implementing retrying the message delivery we will not need to worry about the self deletion logic as this is already implemented?

It's a piece of business logic that we're delegating to the message sending mechanism instead of the one actually responsible for creating the message.

If I was asked where is this timer coming from, I'd check the UseCase first. As see it's not there, then have no clue where it comes from and have to dissect the sending mechanism to find a place that overwrites the original message.

----

As I said before, the UX design for this is debatable. 

_Maybe_ team - or group - enforced timer updates should overwrite pending messages composed while offline, _maybe_ not? The user hit the send button having one idea in mind.
But there's no benefit in not allowing the user to compose messages while offline with different timer settings. We've seen how sensitive users are when self-deleting messages don't work how they want recently.

By following what's persisted instead of making it on-the-fly, we have full flexibility on programming this behavior however we want.

When there's a group or team configuration change event, we can – if the UX design says so – update all the values in the DB for pending messages, before we come online and send them. Or the design might say to leave how it was, and we don't need to do anything.

But doing it on-the-fly brings no flexibility, and delegates the responsibility to an unexpected actor.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
```diff
@@ -48,7 +48,8 @@ data class Conversation(
     val accessRole: List<AccessRole>,
     val creatorId: String?,
     val receiptMode: ReceiptMode,
-    val messageTimer: Duration? // enforced time by group conversation for self deleting messages, always NULL for 1on1 conversation
+    val messageTimer: Duration?, // enforced time by group conversation for self deleting messages, always NULL for 1on1 conversation
+    val userMessageTimer: Duration? // timer duration picked by the user in absence of team or conversation enforced timers
```
vitorhugods, line 52: Kdoc would be great here. So whoever presses F1 or Ctrl+Q on the IDE can see what's this about.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSendingInterceptor.kt
```diff
@@ -27,7 +27,7 @@ import com.wire.kalium.logic.functional.map
 import com.wire.kalium.logic.util.MessageContentEncoder
 
 interface MessageSendingInterceptor {
-    suspend fun prepareMessage(message: Message.Sendable): Either<CoreFailure, Message.Sendable>
+    suspend fun prepareMessage(originalMsg: Message.Sendable): Either<CoreFailure, Message.Sendable>
```
vitorhugods, line 30:
```suggestion
    suspend fun prepareMessage(originalMessage: Message.Sendable): Either<CoreFailure, Message.Sendable>
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSendingInterceptor.kt
```diff
@@ -36,6 +36,7 @@ class MessageSendingInterceptorImpl(
 ) : MessageSendingInterceptor {
 
     override suspend fun prepareMessage(message: Message.Sendable): Either<CoreFailure, Message.Sendable> {
```
vitorhugods, line 38:
```suggestion
    override suspend fun prepareMessage(originalMessage: Message.Sendable): Either<CoreFailure, Message.Sendable> {
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSendingInterceptor.kt
vitorhugods, line 1: Or, alternatively, just revert these changes completely.

vitorhugods (approved):
A couple suggestions.

But LGTM in general.

PR #1785, opened 2023-06-07
fix: NewClient dialog flow (WPD-284)

borichellow:
# What's new in this PR?

### Issues

A "this account was used on ..." dialog is shown in the media share extension

Aslo issue that I found while working on it: 
If user didn't open the app for some time and during this time more than 1 NewClient came -> User will see a dialog only about 1 NewClient (the last one).

### Causes (Optional)

NewClients event is not saved anywhere, there was only Singleton `NewClientManager` which we were using for emitting such an event into observers (`ObserveNewClientsUseCase`) and it emits new clients one by one

### Solutions

Add new DB-table `NewClient` and save NewClients there. Now we can observe and not observe NewClients depending on screen, without losing any NewClient.

+ Instead of emitting newClients one by one `ObserveNewClientsUseCase` emits a list of newClients.

# Important: 
Breaking changes! Merge only when [android PR](https://github.com/wireapp/wire-android-reloaded/pull/1862) ready for merge


MohamadJaara:
storing those events in the global DB add a layer of complication that is unnecessary, for example on login we need to remember to clear those events.

we can keep those events in the user self DB and if you want to globally observe the list you can observe active accounts -> get events for each account separately

vitorhugods (changes requested):
My main concern is that by making `Client` serializable and storing it as JSON we have to be super careful.

Our data classes in Logic weren't made to be persisted. We have Entities and all the tooling to detect breaking changes there.

Any change in the data class will break the stored JSON in a way won't notice, unlike with DB schema changes.

So, if it is possible to use the actual entities, that would be preferable.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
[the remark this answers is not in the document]
borichellow: maybe i'll just add removing that new client from the DB, when we receive `Event.User.ClientRemove` ???

vitorhugods (approved):

vitorhugods:
Will this also be needed on `release/candidate` ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
```diff
@@ -83,6 +84,19 @@ class ClientMapper(
...
 
+    fun fromNewClientEntity(clientEntity: NewClientEntity): Client = with(clientEntity) {
+        Client(
+            id = ClientId(id),
+            type = null,
+            registrationTime = registrationDate,
+            deviceType = deviceType?.let { fromDeviceTypeEntity(deviceType) },
+            label = null,
+            model = model,
+            isVerified = true,
```
typfel, line 95: I wouldn't expect new clients to be verified by default, but I guess doesn't really matter for what you are using this for.

typfel (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientMapper.kt
borichellow: true, it's not used anywhere in this flow, but I'll change it to `false` just in case :)

borichellow:
> Will this also be needed on `release/candidate` ?

good question 🤔

PR #1786, opened 2023-06-08
feat: supported protocols for users

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Implement updating & parsing & persisting supported protocols for users.

Supported protocols is a new field on the user entity which indicates which protocols a user is able to speak (proteus, mls).

See: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/815693828/Use+case+updating+your+supported+protocols
https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/797737097/Use+case+select+protocol+for+1+1+conversation

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -199,7 +199,8 @@ internal class ConversationMapperImpl(
                             completePicture = previewAssetId?.toModel(),
                             previewPicture = previewAssetId?.toModel(),
                             teamId = teamId?.let { TeamId(it) },
-                            connectionStatus = connectionStatusMapper.fromDaoModel(connectionStatus)
+                            connectionStatus = connectionStatusMapper.fromDaoModel(connectionStatus),
+                            supportedProtocols = null // TODO jacob include in query?
```
vitorhugods, line 203: I guess so :D 

Should this be done in this PR, or later?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -233,7 +234,8 @@ internal class ConversationMapperImpl(
                         handle = null,
                         completePicture = previewAssetId?.toModel(),
                         previewPicture = previewAssetId?.toModel(),
-                        teamId = teamId?.let { TeamId(it) }
+                        teamId = teamId?.let { TeamId(it) },
+                        supportedProtocols = null // TODO jacob include in query?
```
vitorhugods, line 238: Same here

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/self/SelfApi.kt
```diff
@@ -27,7 +29,13 @@ data class ChangeHandleRequest(
     @SerialName("handle") val handle: String
 )
 
-interface SelfApi {
+interface BaseApi {
+    fun getApiNotSupportError(apiName: String, apiVersion: Int) = NetworkResponse.Error(
+        APINotSupported("${this::class.simpleName}: $apiName api is only available on API V$apiVersion")
+    )
+}
```
vitorhugods, line 36: Is this only for `SelfApi`?

It's in `SelfApi.kt`, but its name and behaviour look quite generic.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserModel.kt
```diff
@@ -42,6 +42,7 @@ sealed class User {
     abstract val previewPicture: UserAssetId?
     abstract val completePicture: UserAssetId?
     abstract val availabilityStatus: UserAvailabilityStatus
+    abstract val supportedProtocols: Set<SupportedProtocol>?
```
vitorhugods, line 45: I wonder how to separate this once it comes from the database.

I assume it should come from the database at some point.

How to differentiate `null` from an empty set, or... if there should be any distinction. Can't we assume that an empty set is the same as "unknown" and make this not-optional ?

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/model/UserDTO.kt
```diff
@@ -39,7 +40,8 @@ data class UserDTO(
     @SerialName("qualified_id") val id: UserId,
     @SerialName("service") val service: ServiceDTO?,
     @SerialName("sso_id") val ssoID: UserSsoIdDTO?,
-    @SerialName("team") val teamId: TeamId?
+    @SerialName("team") val teamId: TeamId?,
+    @SerialName("supported_protocols") val supportedProtocols: List<SupportedProtocolDTO> = listOf(SupportedProtocolDTO.PROTEUS)
```
vitorhugods, line 44: I _hope_ we don't forget to remove this hardcoded `PROTEUS` here.

Ideally we should bubble up this to the upper layers to actually have the logic to decide what should be the supported protocols.

I can see the case for leaving it as a default argument, as it would be a lot of repetition and could actually make it harder for us to replace it in multiple places.

Can we, instead of the default parameter, add a secondary constructor that will put the default `PROTEUS` argument, and annotate it with `@Deprecated`?

Other than the warning, at least we will be able to find usages of the deprecated constructor more easily when needed in the future.

``` kotlin
    @SerialName("supported_protocols") val supportedProtocols: List<SupportedProtocolDTO>
) {

    @Deprecated("supportedProtocols will become a mandatory field. Specify the supported protocols.")
    constructor(
        accentId: Int,
        assets: List<UserAssetDTO>,
        deleted: Boolean?,
        email: String?,
        expiresAt: String?,
        handle: String?,
        nonQualifiedId: NonQualifiedUserId,
        name: String,
        locale: String,
        managedByDTO: ManagedByDTO?,
        phone: String?,
        id: UserId,
        service: ServiceDTO?,
        ssoID: UserSsoIdDTO?,
        teamId: TeamId?
    ) : this(
        accentId,
        assets,
        deleted,
        email,
        expiresAt,
        handle,
        nonQualifiedId,
        name,
        locale,
        managedByDTO,
        phone,
        id,
        service,
        ssoID,
        teamId,
        listOf(SupportedProtocolDTO.PROTEUS)
    )
}
```

vitorhugods (changes requested):
Looks good.

Just a couple of observations.

The main one, is that I wonder if the nullability of the supported protocols field brings any benefits.

In practical terms, we will probably do something like:

```kotlin
user.supportedProtocols.contains(MLS)

user.supportedProtocols.contains(PROTEUS)
```

Having to do something like below, is just annoying:

```kotlin
user.supportedProtocols?.contains(MLS) ?: false
```

We can still test for "unknown" by checking if the set is empty (something that could happen even if nullable):

```kotlin
val supportedProtocols: Set<SupportedProtocol>? = emptySet()

// It's not null, but it's empty? What is this?
// Could be made mandatory instead, no?
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserModel.kt
yamilmedina: Is this field intended to some sort of backwards compatibility ? (ie: proteus) 
I'm asking if in this case then makes sense to use Proteus as default value 🤔

network/src/commonTest/kotlin/com/wire/kalium/api/v4/UserDetailsApiV4Test.kt
```diff
@@ -68,6 +68,8 @@ internal class UserDetailsApiV4Test : ApiTest() {
         val params = ListUserRequest.qualifiedIds(
             listOf(QualifiedIDSamples.one, QualifiedIDSamples.two, QualifiedIDSamples.three)
         )
+
+        println(SUCCESS_RESPONSE.rawJson)
```
yamilmedina, line 72: I guess this is a left over for debugging purposes

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/self/SelfApi.kt
typfel: Plan was to also do for other APIs since this function currently duplicated in many places. I'll clean this up.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -407,6 +409,10 @@ internal class UserDataSource internal constructor(
         userDAO.updateUserAvailabilityStatus(userId.toDao(), availabilityStatusMapper.fromModelAvailabilityStatusToDao(status))
     }
 
+    override suspend fun updateSupportedProtocols(protocols: Set<SupportedProtocol>): Either<CoreFailure, Unit> {
```
yamilmedina, line 412: Shouldn't this operation update the local value in the database as well ? 🤔

yamilmedina (commented):
Looking nice, just added some comments and questions 🔥

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserModel.kt
typfel: Yes, we can probably make it non-optional. Was initially thinking I wanted to keep it nullable to know if we had fetched it all or not. But `supportedProtocols` can always change so we anyway need to refresh regularly so it doesn't matter.

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/model/UserDTO.kt
typfel: As long as this field exists it will be optional because it's something that a user declare. If a user hasn't set it yet we assume only proteus is supported.

I can make it optional and introduce this assumption at the mapping layer in logic instead. I suspect I need to do it in many places though since we map directly from api to entity and api to model, but I can live it.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
typfel: it should! 👍

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserModel.kt
typfel: > Is this field intended to some sort of backwards compatibility ? (ie: proteus)
> I'm asking if in this case then makes sense to use Proteus as default value 🤔

Going forward we will in a world were some users will support proteus and some MLS and some will support both. This field is used to announce to other users which protocols you support with your clients so that we can pick the right protocol.

We are defaulting proteus if a user has never set which protocols he/her support but we can't always default proteus before we have even attempted to fetch the supported protocols (the user might not support proteus).

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
typfel: Done

network/src/commonTest/kotlin/com/wire/kalium/api/v4/UserDetailsApiV4Test.kt
typfel: Deleted.

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/self/SelfApi.kt
typfel: Moved to a separate file.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
typfel: Done

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
typfel: Done

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/model/UserDTO.kt
typfel: Moved the default handling to the logic layer. I'll do refactoring PR on develop to clean up code duplication in  UserMapper and PublicUserMapper

yamilmedina (approved):
🔥

vitorhugods (approved):

PR #1790, opened 2023-06-09
fix: message timer not started when message is sent

trOnk12:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

 - Once the message is sent, we were passing old instance of the message that has not updated status.
 - Also in some cases even though the functions are executed sequentially, before we do a check again after invoking enqueue method to enqueue self deletion, the database returns still old status of the message, which is weird but perhaps invalidating database takes some time ?

### Solutions

- Pass conversation id and message id instead so that the self deletion handler, retrieves the state of the message as it is inside the deletion handler.
- Instead of relaying on exact correct timing if the status is set to SENT, we observe the status instead to make it more reliable. 

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/ephemeral/EphemeralMessageDeletionHandler.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/ephemeral/EphemeralMessageDeletionHandler.kt
@@ -11,7 +11,9 @@ import com.wire.kalium.persistence.dao.message.MessageEntity
 import com.wire.kalium.util.KaliumDispatcher
 import com.wire.kalium.util.KaliumDispatcherImpl
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.takeWhile
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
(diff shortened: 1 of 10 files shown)


vitorhugods:
> * Also in some cases even though the functions are executed sequentially, before we do a check again after invoking enqueue method to enqueue self deletion, the database returns still old status of the message, which is weird but perhaps invalidating database takes some time ?

This goes against all tests and everything we assume a database does.

There might be a simpler explanation and implementation, tho:

`startSelfDeletionIfNeeded(message)` is being called _before_ the message is updated to `SENT`.

The function `sendMessage` does `attemptToSend` and then `promoteMessageToSentUpdatingServerTime`.

But we're calling `startSelfDeletionIfNeeded` _inside_ `attemptToSend`.

To me, moving the call is simpler to maintain and less error prone than launching coroutines and observing stuff in this already overly-coupled mechanism

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/ephemeral/EphemeralMessageDeletionHandler.kt
```diff
@@ -87,16 +72,45 @@ internal class EphemeralMessageDeletionHandlerImpl(
...
+                SelfDeletionEventLogger.log(
+                    LoggingSelfDeletionEvent.WaitingForMessageToBeSent(
+                        message,
+                        expirationData
+                    )
+                )
+
+                messageRepository.observeMessageStatus(message.id, message.conversationId)
+                    .takeWhile { it != MessageEntity.Status.SENT }
+                    .collect {
```
MohamadJaara, line 86: adding .first() here will guaranties that the delete is executed only once 

```kotlin
                messageRepository.observeMessageStatus(message.id, message.conversationId)
                    .takeWhile { it != MessageEntity.Status.SENT }
                    .first()
                    .let {
```

PR #1791, opened 2023-06-09
fix: immediately retry sync on connection policy upgrade

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Sometimes users on Reloaded are "stuck on connecting"

### Causes

1. The OS can block the app from connecting to the network when the app isn't visible after a while. And we do not handle the "unblocking" on Kalium's `NetworkStateObserver`.
2. Given the current backoff policy can make it grow to up to 10 minutes while on the background, even if the app is reopened, the retry counter will stay counting.

### Solutions

The first "cause" isn't handled on this PR. As Android's ConnectivityManager calls need some testing to make sure it works well and it would be safer to improve after the next code freeze.

About the second one. It's two-fold:
1. Whenever there's a policy upgrade: reset the backoff timer.
2. Whenever there's a policy upgrade: cancel the timer that is currently running.

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/IncrementalSyncManager.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/IncrementalSyncManager.kt
@@ -35,10 +35,19 @@ import com.wire.kalium.util.KaliumDispatcherImpl
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
+import kotlinx.coroutines.cancelChildren
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.cancellable
 import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.flow.dropWhile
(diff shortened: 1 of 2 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/IncrementalSyncManager.kt
```diff
@@ -101,14 +112,35 @@ internal class IncrementalSyncManager(
...
         }
     )
 
+    private suspend fun delayUntilConnectedOrPolicyUpgrade(delay: Duration): Unit = coroutineScope {
+        select {
+            async {
+                incrementalSyncRepository
+                    .connectionPolicyState
+                    .onEach {
+                        println(it)
```
typfel, line 129: Delete or add debug log?

typfel (changes requested):

saleniuk (approved):
just take care of this `println` 😄

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/incremental/IncrementalSyncManager.kt
vitorhugods: OOF. Forgot about it :D

tmspzz (approved):

PR #1817, opened 2023-06-22
feat(mls): handle wrong epoch error [WPB-1803]

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

If there's any race condition on the backend, or the device becomes offline for too long, we might lose some MLS messages, resulting in `WRONG_EPOCH`.

### Solutions

Handle the case, by creating a `MLSWrongEpochHandler`, that will be called when there's a failure to unpack MLS conversations.
This will:
1. Fetch the conversation from the backend
2. Compare the `epoch` and see if it is different from the local version
3. In case of a difference, rejoin the existing MLS conversation to match the epoch
4. In case of a rejoin, insert a MLS warning system message for the affected conversation

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/MLSWrongEpochHandler.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/MLSWrongEpochHandler.kt
@@ -55,8 +55,7 @@ internal class MLSWrongEpochHandlerImpl(
         dateIso: String,
     ) {
         logger.i("Handling MLS WrongEpoch result")
-        conversationRepository.baseInfoById(conversationId).flatMap {
-            val protocol = it.protocol
+        conversationRepository.getConversationProtocolInfo(conversationId).flatMap { protocol ->
             if (protocol is Conversation.ProtocolInfo.MLS) {
                 Either.Right(protocol)
             } else {
@@ -77,9 +76,8 @@ internal class MLSWrongEpochHandlerImpl(
 
     private suspend fun getUpdatedConversationEpoch(conversationId: ConversationId): Either<CoreFailure, ULong?> {
         return conversationRepository.fetchConversation(conversationId).flatMap {
-            conversationRepository.baseInfoById(conversationId)
-        }.map { updatedConversation ->
-            val updatedProtocol = updatedConversation.protocol
+            conversationRepository.getConversationProtocolInfo(conversationId)
+        }.map { updatedProtocol ->
             (updatedProtocol as? Conversation.ProtocolInfo.MLS)?.epoch
         }
     }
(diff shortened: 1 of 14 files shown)


typfel (changes requested):
Looking great! Just a test case which I think is missing a negation.

typfel (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/MLSWrongEpochHandler.kt
```diff
@@ -0,0 +1,100 @@
...
+) : MLSWrongEpochHandler {
+
+    private val logger by lazy { kaliumLogger.withFeatureId(KaliumLogger.Companion.ApplicationFlow.EVENT_RECEIVER) }
+
+    override suspend fun onMLSWrongEpoch(
+        conversationId: ConversationId,
+        dateIso: String,
+    ) {
+        logger.i("Handling MLS WrongEpoch result")
+        conversationRepository.baseInfoById(conversationId).flatMap {
```
mchenani, line 58: AFAIK we have a method to get the protocol info only, isn't better to use that since we need the protocol only here?

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/MLSWrongEpochHandler.kt
```diff
@@ -0,0 +1,100 @@
...
+            if (isRejoinNeeded) {
+                joinExistingMLSConversation(conversationId)
+            } else Either.Right(Unit)
+        }.flatMap {
+            insertInfoMessage(conversationId, dateIso)
+        }
+    }
+
+    private suspend fun getUpdatedConversationEpoch(conversationId: ConversationId): Either<CoreFailure, ULong?> {
+        return conversationRepository.fetchConversation(conversationId).flatMap {
```
mchenani, line 79: can we use smaller query and data flow here to fetch what we need?

mchenani (approved):
LGTM, added some change suggestions.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/MLSWrongEpochHandler.kt
vitorhugods: Indeed, it seems we do. I'll change it.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/message/MLSWrongEpochHandler.kt
vitorhugods: I didn't notice we already had a function for it. I'll change it to use it.

Otherwise, considering that it's fetching a single conversation (and not a list), I wouldn't go through the trouble to create a new query, add tests, etc. as the performance improvement is pretty much non-existent.

yamilmedina (approved):

PR #1855, opened 2023-07-03
refactor: handle event processing failure

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently, there's no possibility of gracefuly handling errors during event processing.

For MLS - for example - there are delicate event handling where network requests need to be made and in case of failure or network instability, we should abort sync and retry later again.

Currently when an exception is thrown: Sync will stop.

This is _in parts_ what we want. However, most of our code wraps exceptions and returns as `Either.Left` without throwing exceptions.

### Solutions

Streamline the processing of events, by making handlers return `Either` as well.
In case of failures, the `lastProcessedEventId` shouldn't be updated, and `IncrementalSyncWorker` should throw a `KaliumSyncException`, making Sync stop.

I have _not_ addressed 100% of the event handlers in this PR. Only a few that were easy to tackle as this refactoring would be huge.

I've opted to keep `Either.Right(Unit)` on other cases, as it keeps the same behaviour.

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
@@ -79,7 +79,7 @@ interface ClientRepository {
         verified: Boolean
     ): Either<StorageFailure, Unit>
 
-    suspend fun saveNewClientEvent(newClientEvent: Event.User.NewClient)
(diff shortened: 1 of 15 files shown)


MohamadJaara (approved):

borichellow (approved):
Good catch and useful refactoring 💪

typfel (approved):

PR #1861, opened 2023-07-04
fix: restore assets from web

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There was an issue with parsing assets from web backup

### Causes (Optional)

Only text messages were restored from web backup

### Solutions

Parse asset json models in proper way


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/web/WebMappers.kt
```diff
@@ -76,8 +76,8 @@ fun WebEventContent.toMigratedMessage(selfUserDomain: String): MigratedMessage?
                             name = data.info?.name,
                             mimeType = mimeType,
                             remoteData = AssetContent.RemoteData(
-                                otrKey = data.otrKey?.toString()?.toByteArray() ?: ByteArray(0),
-                                sha256 = data.sha256?.toString()?.toByteArray() ?: ByteArray(0),
+                                otrKey = data.otrKey?.values?.map { it.toByte() }?.toByteArray() ?: ByteArray(0),
```
MohamadJaara, line 79: maybe ignore the message is it does not have those 2 values 

```kotlin
val otrKey = data.otrKey?.values?.map { it.toByte() }?.
val sha256 = data.sha256?.values?.map { it.toByte() }?.toByteArray()
if(otrKey == null or sha256 == null) return
// do stuff and insert the message

yamilmedina (approved):

alexandreferris (approved):

PR #1862, opened 2023-07-04
chore: nullable sender [WPB-1815]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Some federation events will not contain senderId so we need to make this variable nullable in our database

Event will look like

`{
  "type": "federation.delete",
  "domains": ["backendB.com", backendC.com"]
} `

And system message will not contain senderUserId because we need to create system message locally for each of affected domain users

### Solutions

Migrate messages to new table with nullable senderId field


MohamadJaara:
this is dangers change, can you please explain a bit more about those event and that are the data provided 
maybe we can provide a work around by having self user as the sender or even a fixed is like 
UserId("server", ${serverDomain})

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/ephemeral/DeleteEphemeralMessageForSelfUserAsReceiverUseCase.kt
```diff
@@ -74,7 +74,7 @@ internal class DeleteEphemeralMessageForSelfUserAsReceiverUseCaseImpl(
                         sendDeleteMessageToOriginalSender(
                             message.id,
                             message.conversationId,
-                            message.senderUserId,
+                            message.senderUserId!!, // TODO Mohamad can we somehow secure this case from getMessageById
```
MohamadJaara, line 77: you can do the check before calling sendDeleteMessageToOriginalSender

```kotlin
message.senderUserId?.also {
sendDeleteMessageToOriginalSender()

}
```

typfel:
> this is dangers change, can you please explain a bit more about those event and that are the data provided
> maybe we can provide a work around by having self user as the sender or even a fixed is like
> UserId("server", ${serverDomain})

I agree, for system messages I think it's very reasonable to assign the sender to the self user.

Garzas:
Closing because we want to have non nullable senderUserId

PR #1897, opened 2023-07-19
fix: duplicate conversation timer change message [WPB-286]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We were not fetching event timestamp after changing conversation message timer.

### Causes (Optional)

When someone changed timer twice and then send an asset we were unable to proper filter duplicated events..

### Solutions

Fetch event timestamp when changing conversation message timer in order to easier filter it when same event comes from web-socket. 

Also added some more tests


yamilmedina (approved):
Great job, nice improvement moving the logic out of the use case 😎

alexandreferris (approved):

PR #1905, opened 2023-07-20
fix: updating isMLSCapable

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

It should be possible to update `isMLSCapable` from false to true but not the other way around.

#### Test Coverage

----

----

----
##### References

### DIFF
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Clients.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Clients.sq
@@ -39,7 +39,7 @@ label = coalesce(excluded.label, label),
 model = coalesce(excluded.model, model),
 is_valid = is_valid,
 last_active = coalesce(excluded.last_active, last_active),
-is_mls_capable = is_mls_capable;
+is_mls_capable = excluded.is_mls_capable OR is_mls_capable;
 
 selectAllClients:
 SELECT * FROM Client;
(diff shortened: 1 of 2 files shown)


borichellow:
I'm thinking if the DB is a good place for such a thing 🤔 
I don't know the reason why it can't be `true->false`, but DB-layer is only about storing the data IMO. 
Maybe the `repository` level is a better place for such "filtering" and you could add comment why it's like that

typfel:
> I'm thinking if the DB is a good place for such a thing 🤔 
> 
> I don't know the reason why it can't be `true->false`, but DB-layer is only about storing the data IMO. 
> 
> Maybe the `repository` level is a better place for such "filtering" and you could add comment why it's like that

Yes I agree. I'll see if I can move logic up without having to duplicate it. An issue we generally have is that in the logic layer we partially duplicate the mapping logic.

vitorhugods (approved):

MohamadJaara (approved):

typfel:
> > I'm thinking if the DB is a good place for such a thing 🤔
> > I don't know the reason why it can't be `true->false`, but DB-layer is only about storing the data IMO.
> > Maybe the `repository` level is a better place for such "filtering" and you could add comment why it's like that
> 
> Yes I agree. I'll see if I can move logic up without having to duplicate it. An issue we generally have is that in the logic layer we partially duplicate the mapping logic.

@borichellow I don't think it worth the complexity increased complexity to move this up to the logic layer. I would then need to fetch any existing client before inserting or updating a client in order to do the necessary comparison of the `isMLSCapable` value.

I propose to instead add a comment on in the insert sql statement explaining this logic.

PR #1930, opened 2023-07-31
feat: local state cleanup after federation delete [WPB-1816]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

When 
- self user backend (A) stops federating with other backend (B) - federation.delete event
- two other backends stops federating (B & C) - federation.connectionRemoved event

Our backend will not send tons of events about removing users from conversations, clearing connection requests etc. Instead it will send one of two payloads depending on if our backend was involved in defederation (when case above).

Because of that our client need to do entire cleanup process in local DB, which is
- remove all connection requests with defederated backend users
- remove all defederated users from conversations which are not from their backend
- remove our users from defederated backend domains 
- remove our users and defederated users from common conversations which are not from our or defederated backend

More info about use case: 
[https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/809238539/Use+case+Stopping+to+federate+with+a+domain#Receiving-a-federation.delete-event](federation.delete)
[https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/809238539/Use+case+Stopping+to+federate+with+a+domain#Receiving-a-federation.connectionRemoved-event](federation.connectionRemoved)

### Solutions

- clean up state for connection requests
- mark users as defederated to mark 1on1 conversations as defederated to hold conversation history and disable typing
- remove users from conversations that contain defederated users or we stopped federating
- send federation system message
- send conversation member removed events


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventMapper.kt
```diff
@@ -87,6 +87,22 @@ class EventMapper(
             is EventContentDTO.UserProperty.PropertiesDeleteDTO -> deleteUserProperties(id, eventContentDTO, transient)
             is EventContentDTO.Conversation.ReceiptModeUpdate -> conversationReceiptModeUpdate(id, eventContentDTO, transient)
             is EventContentDTO.Conversation.MessageTimerUpdate -> conversationMessageTimerUpdate(id, eventContentDTO, transient)
+            is EventContentDTO.Federation -> federation(id, eventContentDTO)
+        }
+
+    private fun federation(id: String, eventContentDTO: EventContentDTO.Federation): Event =
```
yamilmedina, line 93: Maybe we could use an explicit name here, something like `federationStopped` or `federationTerminated`

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/FederationEventReceiver.kt
```diff
@@ -0,0 +1,230 @@
...
+                        event,
+                        Pair("errorInfo", "$it")
+                    )
+            }
+    }
+
+    private suspend fun handleConnectionRemovedEvent(event: Event.Federation.ConnectionRemoved) =
+        withContext(dispatchers.io) {
+            val firstDomain = event.domains.first()
+            val secondDomain = event.domains.last()
```
yamilmedina, line 131: Are we sure here we are always going to receive 2 domains 😅 ? I don't think it is a problem, but maybe we can put here some logs, to monitor in case we receive an odd `> 2` case . What do you think ?

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq
```diff
@@ -114,6 +115,17 @@ CREATE TABLE MessageMemberChangeContent (
       PRIMARY KEY (message_id, conversation_id)
 );
 
+CREATE TABLE MessageFederationContent (
```
yamilmedina, line 118: Same as the event handler, we might use a name related that's related to stopped to federate ? 🤔 
Like, `MessageFederationTerminatedContent`, just an idea in case will serve only this purpose.

yamilmedina (commented):
Added some comments, looking wonderful as an overall solution.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventMapper.kt
Garzas: I was thinking that some other events may come for federation but for now we can rename it 👍

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/FederationEventReceiver.kt
Garzas: Good point, I will log error and don't trigger any action when size will be different than 2

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/FederationEventReceiver.kt
```diff
@@ -0,0 +1,241 @@
...
+                    handleFederationDeleteEvent(conversationId, event.domain)
+                    userIds.filter { it.domain == event.domain }.forEach { userId ->
+                        userRepository.defederateUser(userId)
+                    }
+                }
+
+                // remove
+                conversationsWithMembers.group.forEach { (conversationId, userIds) ->
+                    handleFederationDeleteEvent(conversationId, event.domain)
+                    when (conversationId.domain) {
```
borichellow, line 97: just a suggestion:
```
val userIdsToRemove = when (conversationId.domain) {
     // remove defederated users from self domain conversations
    selfUserId.domain -> userIds.filterNot { it.domain == selfUserId.domain }

    // remove self domain users from defederated domain conversations
    event.domain -> userIds.filter { it.domain == selfUserId.domain }

    // remove self domain and defederated domain users from not defederated and self domain conversations
    else -> userIds
}
removeMembersFromConversation(conversationId, userIdsToRemove)
```

trOnk12 (commented):
I have no comments regards the code itself, looks good to me ! Well seperated responsibilites and I think it all looks good, I am just like Boris mentions not sure about if we are doing things here most efficently, I am also not in the position to give u any advice on SQL :S but perhaps we can move some parts of the code into the SQL itself, I assume that the SQL is the best places to gain performance on querying data, but again it is just my  gut feeling and speculations.

borichellow:
After some discussion with Kubaz, we noticed that it's gonna be quit rare case, so my comments about performance might be ignored ;)

yamilmedina (approved):
I agree with @borichellow point, But I think that because of the rare case this happens (i.e., trigger by sysadmins ops.) we can measure and later create a tech debt ticket in case necessary :D

borichellow (approved):

PR #1932, opened 2023-08-01
feat: Get and save clients mlsPublicKeys (WPB-2308)

borichellow:
# What's new in this PR?

Added `mls_public_keys` field into `Client` models and DB. It's needed to display MLS info in settings.

Needs releases with:

Could be merged without any problems for the AR.


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
```diff
@@ -159,7 +159,10 @@ class ClientDataSource(
         return wrapApiRequest { clientApi.fetchSelfUserClient() }
             .onSuccess { clientList ->
                 val selfUserIdDTO = selfUserID.toApi()
-                val list = clientList.map { clientMapper.toInsertClientParam(it, selfUserIdDTO) }
+                val list = clientList.map {
+                    println("cyka ${it.model} ${it.mlsPublicKeys}")
```
typfel, line 163: Delete?

```suggestion
```

typfel (changes requested):
Some minor comments otherwise looks fine but can you also add a description to the PR which explains why we need to save the public keys?

borichellow:
> Some minor comments otherwise looks fine but can you also add a description to the PR which explains why we need to save the public keys?

it was kinda draft, now everything is fixed :)

typfel (approved):

alexandreferris (approved):

mchenani (approved):
🥳

PR #1941, opened 2023-08-02
fix: email of non-team users is cleared during sync

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

during sync email for non-team accounts is cleared 

### Causes (Optional)

the first step in sync the email is stored correctly 
but when reaching the step for fetch info for all users id stored in the db we also include the user self id and since the fetch is from the end point `POST:/list-users` and emails of non team users will be returned as null for non team accounts which leads to the delete of the user email 

### Solutions

exclude user self id from the step to get all users info durinc sync

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -81,7 +83,11 @@ import kotlin.time.Duration.Companion.minutes
 @Suppress("TooManyFunctions")
 internal interface UserRepository {
     suspend fun fetchSelfUser(): Either<CoreFailure, Unit>
-    suspend fun fetchKnownUsers(): Either<CoreFailure, Unit>
+
+    /**
+     * fetches user information for all of users id stored in the DB
+     */
+    suspend fun syncAllOtherUsers(): Either<CoreFailure, Unit>
```
typfel, line 90: Why prefix this method with `sync` instead of `fetch` like the other methods?

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
```diff
@@ -40,7 +40,7 @@ VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
 ON CONFLICT(qualified_id) DO UPDATE SET
 name = excluded.name,
 handle = excluded.handle,
-email = excluded.email,
+email = coalesce(excluded.email, email),
```
typfel, line 43: This isn't necessary now that you exclude self when syncing users profiles? Also if the self user removes his email for some reason this would prevent from being removed? Intentional?

typfel (changes requested):
Looking good just some nitpicking and questions.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
MohamadJaara: in my mind what this function does is fetches users data form remove and store it locally which is closer to a sync operation, i fetchUsersInfo

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Users.sq
MohamadJaara: i was experimenting with the reason and forgot to revert some changes.

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Sync.sq
```diff
@@ -0,0 +1,2 @@
+userIdsWithOutSelf:
```
vitorhugods, line 1:
```suggestion
userIdsWithoutSelf:
```

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Sync.sq
```diff
@@ -0,0 +1,2 @@
...
+SELECT qualified_id FROM User WHERE qualified_id != (SELECT * FROM SelfUser);
```
vitorhugods, line 2:
```suggestion
SELECT qualified_id FROM User WHERE qualified_id != (SELECT id FROM SelfUser);
```

Just so it doesn't break if we ever add another column to `SelfUser`

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/SyncDAO.kt
```diff
@@ -0,0 +1,40 @@
...
+import kotlin.coroutines.CoroutineContext
+
+interface SyncDAO {
+
+    /**
+     * @return [List] of [UserIDEntity] of all other users.
+     * the list does not contain self user ID
+     */
+    suspend fun allOtherUsersId(): List<UserIDEntity>
+}
```
vitorhugods, line 31: This might be a sync-related query. But there's no such thing as a `SyncEntity` or a `Sync` table.

Also, the database itself has no concept of what Sync is. 

It also feels a bit weird to me, that if we ever have another legit use-case where we need all users except the self-user, and it would be in some "sync" umbrella.

It looks like `UserDAO` to me.

I'd also question the need to make a query just for this.

Couldn't we just filter it after the query? We're talking about something that happens every full moon, and would take a millisecond to iterate through 1000 users and eliminate self user from it.

Sure, now that it's done... we can stick with it. But it being outside `UserDAO` is weird to me.

persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/SyncDAOTest.kt
```diff
@@ -0,0 +1,62 @@
...
+        userDAO.insertUser(user2)
+        userDAO.insertUser(user3)
+
+        syncDAO.allOtherUsersId().also { result ->
+            assertFalse {
+                result.contains(selfUser.id)
+            }
+            assertEquals(result, listOf(user1.id, user2.id, user3.id))
+        }
+
```
vitorhugods, line 60: We don't use `syncDAO` to insert stuff. It's just to read and filter from another DAO. That's the part that to me makes it extra-obvious that something is out of place.

- "How can I get all known users, except self user?"
- "Oh, just check SyncDAO"
- "?"

`UserDAO` would make it obvious and easy to discover.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/slow/SlowSyncManager.kt
```diff
@@ -158,4 +158,4 @@ internal class SlowSyncManager(
     }
 }
 
-const val CURRENT_VERSION = 3 // bump this version to perform slow sync when some new feature flag was added
+const val CURRENT_VERSION = 4 // bump this version to perform slow sync when some new feature flag was added
```
vitorhugods, line 161: This is still being useful after all :D

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -81,7 +83,11 @@ import kotlin.time.Duration.Companion.minutes
 @Suppress("TooManyFunctions")
 internal interface UserRepository {
     suspend fun fetchSelfUser(): Either<CoreFailure, Unit>
-    suspend fun fetchKnownUsers(): Either<CoreFailure, Unit>
+
+    /**
+     * fetches user information for all of users id stored in the DB
```
vitorhugods, line 88:
```suggestion
     * Fetches user information for all of users id stored in the DB
```

vitorhugods (changes requested):
Looks good to me. I just find the existence of `SyncDAO` weird.

A couple of other nitpicks too.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
typfel: Right makes some sense but the other methods named `fetch` also doesn't return anything. Why not open separate refactoring PR if you want to change how we name these methods.

persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/SyncDAOTest.kt
MohamadJaara: UserDAO already have `@Suppress("TooManyFunctions")` and i would like to offload functionality from it not add new one, that is where the SyncDAO naming comes from

vitorhugods: It's easier to just move some of the functions there to `MemberDAO` instead, like:

`getUsersNotInConversationByHandle`
`getUsersNotInConversationByNameOrHandleOrEmail`
`observeUsersNotInConversation`

`TooManyFunctions` is one indication that a class _might_ have far too many responsibilities. But it is not necessarily a deal breaker. This function suits `UserDAO` and `Sync` is just a weird thing to have in the persistence level.

vitorhugods (approved):

typfel (approved):

PR #1977, opened 2023-08-11
fix: Remove Lazy from Global DB/Preferences (WPB-3147)

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We are having this issue:  `Exception java.security.InvalidKeyException: Keystore operation failed`
from : `com.wire.kalium.persistence.kmmSettings.EncryptedSettingsBuilder.build`

### Solutions

This is more a try out instead of a real solution as we were not able to properly reproduce this crash locally, so what we did:

- Remove Lazy from Global Database and Preferences

We will continue to monitor this crash on next release, and if it continues the same we will then implement a try/catch with a retry once as it seems to also be fixing the issue.

Google has moved this issue ticket to `New`: https://issuetracker.google.com/issues/176215143 so we are also looking forward for a response from them.

#### How to Test

Unfortunately we have found no way to force this crash

### DIFF
--- a/persistence/src/androidInstrumentedTest/kotlin/com/wire/kalium/persistence/kmmSettings/EncryptedSettingsBuilderTest.kt
+++ b/persistence/src/androidInstrumentedTest/kotlin/com/wire/kalium/persistence/kmmSettings/EncryptedSettingsBuilderTest.kt
@@ -34,6 +34,7 @@ import kotlinx.coroutines.test.runTest
 import kotlinx.coroutines.test.setMain
 import kotlin.test.AfterTest
 import kotlin.test.BeforeTest
+import kotlin.test.Ignore
 import kotlin.test.Test
 
 @OptIn(ExperimentalCoroutinesApi::class)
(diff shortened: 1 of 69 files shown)


persistence/src/androidInstrumentedTest/kotlin/com/wire/kalium/persistence/kmmSettings/EncryptedSettingsBuilderTest.kt
```diff
@@ -0,0 +1,76 @@
...
+                }
+            }
+        }.joinAll()
+        testScope.advanceUntilIdle()
+    }
+
+    @AfterTest
+    fun after() {
+        Dispatchers.resetMain()
+    }
```
MohamadJaara, line 75: suggestion: this test can take some time to run maybe ignore it and we can run it when we need to ?

MohamadJaara (approved):

borichellow (approved):

PR #1988, opened 2023-08-17
fix: multiple defederation system message [WPB-3912]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When backends stops federating clients receives more than one event about it

### Causes (Optional)

Because all events have different id and doesn't contain any timestamp we can ignore generating same system messages in two ways:
- check if last message is stop federation message
- check if user in oneOnOne conversation is defederated []

### Solutions

I chosen second options because it's more performant and defederated field can be used in other places.

Also fixed:
- hide defederated users in search
- don't remove oneOnOne conversation members after slow sync, because it may remove our conversation history

Chore:
- split fetching oneOnOne and group conversation members to be more readable and more independent logic


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -783,15 +786,21 @@ internal class ConversationDataSource internal constructor(
...
-        )
+    ): Either<CoreFailure, Map<QualifiedID, List<QualifiedID>>> = wrapStorageRequest {
+        memberDAO.getGroupConversationWithUserIdsWithBothDomains(firstDomain, secondDomain)
+            .mapKeys { it.key.toModel() }
+            .mapValues { it.value.map { userId -> userId.toModel() } }
+    }
+
+    override suspend fun getOneOnOneConversationsWithFederatedMembers(
+        domain: String
+    ): Either<CoreFailure, Map<QualifiedID, QualifiedID>> = wrapStorageRequest {
```
yamilmedina, line 800: What do you think of using here an alias:

`typealias OneToOneMembers =  Map<QualifiedID, QualifiedID>` or something like that ? to be explicit in the Result intent

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -783,15 +786,21 @@ internal class ConversationDataSource internal constructor(
...
+    override suspend fun getGroupConversationsWithMembersWithBothDomains(
         firstDomain: String,
         secondDomain: String
-    ): Either<CoreFailure, ConversationsWithMembers> = wrapStorageRequest {
-        val entity = memberDAO.getConversationWithUserIdsWithBothDomains(firstDomain, secondDomain)
-        ConversationsWithMembers(
-            oneOnOne = entity.oneOnOne.mapKeys { it.key.toModel() }.mapValues { it.value.map { userIdEntity -> userIdEntity.toModel() } },
-            group = entity.group.mapKeys { it.key.toModel() }.mapValues { it.value.map { userIdEntity -> userIdEntity.toModel() } }
-        )
+    ): Either<CoreFailure, Map<QualifiedID, List<QualifiedID>>> = wrapStorageRequest {
```
yamilmedina, line 792: `typealias GroupConversationMembers = Map<QualifiedID, List<QualifiedID>>` Same as below 🤔

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Members.sq
```diff
@@ -52,7 +52,13 @@ WHERE conversation IN (
 isUserMember:
 SELECT user FROM Member WHERE conversation = :conversationId AND user = :userId;
 
-getMembersFromOneOfTwoDomains:
-SELECT mem.*, con.type FROM Member AS mem
+getFederatedMembersWithOneOfDomainsFromGroupConversation:
```
yamilmedina, line 55: Not relevant, I think we don't have standard, but I see more usages like `select` instead of `get` but not comment on my side

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Members.sq
```diff
@@ -52,7 +52,13 @@ WHERE conversation IN (
...
+SELECT mem.* FROM Member AS mem
 JOIN Conversation AS con ON con.qualified_id = mem.conversation
-WHERE user LIKE ('%@' || :firstDomain) OR user LIKE ('%@' || :secondDomain);
+WHERE con.type == 'GROUP' AND (user LIKE ('%@' || :firstDomain) OR user LIKE ('%@' || :secondDomain));
+
+getFederatedMembersFromOneOnOneConversations:
```
yamilmedina, line 60: Same as above

yamilmedina (approved):
Small comments about naming, but on the core excellent as always 🤖

saleniuk (approved):

PR #1999, opened 2023-08-21
chore: add tests to and update Android NetworkStateObserverImpl [WPB-3470]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

NetworkStateObserver was not properly tested.
Blocked state was not handled, we probably should not assume that with blocked state change, always capabilities are changed as well (it may not be the case for some manufactures for instance).

### Solutions

Simplify handling network events by creating separate StateFlow which contains the recent information about the network, and then the state is determined using all the recent data the app has at the given moment.
Add multiple tests to make sure NetworkStateObserverImpl returns proper state depending on different network events, according to the [documentation](https://developer.android.com/training/basics/network-ops/reading-network-state).

#### Test Coverage (Optional)

----

----

----
##### References


logic/build.gradle.kts
```diff
@@ -93,9 +93,19 @@ kotlin {
                 implementation(libs.work)
             }
         }
+        val androidUnitTest by getting {
```
yamilmedina, line 96: I'm having problems to see this new test execution on the CI reports, can you confirm, or maybe I'm just searching in the wrong place 😅 

I think on CI, we just run `jvmTest` and `jsTest` (tried locally and don't see it either 😓) we might need to add this target exec to our pipeline as well :D

yamilmedina (commented):
Great work, nice to see `robolectric` again. Just left a question over there regarding the test execution.

yamilmedina (approved):
🚀

logic/src/androidMain/kotlin/com/wire/kalium/logic/network/NetworkStateObserverImpl.kt
```diff
@@ -25,40 +25,75 @@ import android.net.NetworkCapabilities
...
+                            "internet:${networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)} " +
+                            "validated:${networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)}"
+                )
+                defaultNetworkDataStateFlow.update { DefaultNetworkData(network, networkCapabilities) }
             }
 
             override fun onLost(network: Network) {
                 kaliumLogger.i("${NetworkStateObserver.TAG} lost connection")
-                networkStateFlow.tryEmit(NetworkState.NotConnected)
+                defaultNetworkDataStateFlow.update { null }
```
mchenani, line 84: question: why null?

saleniuk: `null` is passed when there is no network available at the given moment. I could add another field to the `DefaultNetworkData`, something like `isAvailable` - @mchenani do you think that would be better? I just thought `null` would be more straightforward - if null then no network available, else there is a network connected 😄

mchenani: Thanks for the clarification, IMO I think for  better readability, it's better to avoid having cases as null. Thanks

saleniuk: Ok, thanks @mchenani you're right, I replaced `null` with a sealed class

mchenani (approved):

PR #2006, opened 2023-08-24
feat(monkeys): Allow to split clients per test cases [WPB-4332]

augustocdias:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Adds support to run multiple clients within the test. Each test case in the config will use a separate client for the users in the test.

----
##### References

### DIFF
--- a/monkeys/README.md
+++ b/monkeys/README.md
@@ -37,3 +37,5 @@ An [example](example.json) config is in this repo and the schema can be seen [he
 * Tracing and replaying a test run. For this the order is the important factor, so when replayed it
   won't be executed in parallel.
 * Multi-clients. Right now each user can have just one client within the application
+* Clean-up the data created during the tests
+* Randomising times for action execution
--- a/monkeys/schema.json
+++ b/monkeys/schema.json
@@ -9,6 +9,36 @@
         "backends"
     ],
     "properties": {
+        "conversationDistribution": {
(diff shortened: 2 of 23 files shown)


typfel:
Just to clarify, this PR doesn't add support running multiple clients for each user, or did I miss something? It "only" splits user data by test case?

augustocdias:
> Just to clarify, this PR doesn't add support running multiple clients for each user, or did I miss something? It "only" splits user data by test case?

@typfel It does. For each test case it will instantiate a separate corelogic, which as far as I understood will create separate clients for the users that are used there. The monkey pool is now split per test case so we can track who is logged in per test case as well.

vitorhugods (approved):
Just gotta keep in mind:

### Max clients per account

Once the limit of 6 clients is reached in one account, the 7th client won't be added and it will result in a failure. The monkey should choose one client to remove before adding a new one.


### CoreLogic might misbehave when having multiple instances

I tried looking in the code right now and I couldn't find any suspicious place. But I remember having at least one or two `object` types in CoreLogic that could lead to the results of some operations getting entangled between clients / instances of CoreLogic.

It _might_ be that we removed all of these in the past. I just cannot vouch for it right now.

If any bug happens because of this, we can easily fix it, tho.

typfel (approved):

PR #2035, opened 2023-09-06
feat: Conversation MLS verification status updating (WPB-3872)

borichellow:
# What's new in this PR?

Implementing the machine for calculating the verification state of Conversation. 

### Issues

CoreCrypto can only "say" if MLS conversation is verified or not, while we need also to indicate when the conversation verification state was degraded (was verified but became not_verified). 

### Solutions

- Store the conversation verification state, so we can compare to it the next state.
- When something changes in conversation (adding/removing clients) check its state again and store a proper value.
- Move the whole conversation verification logic into `ConversationVerificationStatusHandler`and call it (start observing the state changes) every time ConversationDetails are observed (by observing that flow in `ObserveConversationDetailsUseCase`)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
```diff
@@ -196,8 +198,8 @@ data class Conversation(
     val supportsUnreadMessageCount
         get() = type in setOf(Type.ONE_ON_ONE, Type.GROUP)
 
-    sealed class ProtocolInfo {
-        object Proteus : ProtocolInfo() {
+    sealed class ProtocolInfo(open val verificationStatus: VerificationStatus) {
```
mchenani, line 201: suggestion: I think tying the conversation status to the protocol info is not a good way, I suggest to have it as an extra field in the conversation data.

typfel:
So checking that've understood the implementation correctly

We have the existing ObserveConversationDetailsUseCase which is the entry point which starts the ConversationVerificationStatusHandler.

ConversationVerificationStatusHandler does:

observes the epoch flow ->
ask CC if verified or not ->
fetch previous verification status ->
derive new verification status ->
insert system message if necessary


Now some questions:
1. Do we only want to update the verification status when conversation if it's displayed on screen? Shouldn't
it be driven from kalium itself and not by the UI? 

2. Wouldn't we potentially lose state transistion for conversations
which are not observed. 

For example, given clients are added add and then removed again:

verified
not verified
verified

Wouldn't we fail to report that the conversation was unverified for time period?

borichellow:
> So checking that've understood the implementation correctly
> 
> We have the existing ObserveConversationDetailsUseCase which is the entry point which starts the ConversationVerificationStatusHandler.
> 
> ConversationVerificationStatusHandler does:
> 
> observes the epoch flow -> ask CC if verified or not -> fetch previous verification status -> derive new verification status -> insert system message if necessary

Yes, you understood the flow perfectly.

> 1. Do we only want to update the verification status when conversation if it's displayed on screen? Shouldn't
>    it be driven from kalium itself and not by the UI?

Well, observing these statuses for ALL the users conversations may have tragical app performance decreasing.
The right solution here is corresponding web-socket event, but we don't have it, so I'm just doing my best with what I have :) 
 
> For example, given clients are added add and then removed again:
> 
> verified 
> not verified 
> verified
> 
> Wouldn't we fail to report that the conversation was unverified for time period?

True, user will lose "not verified" status in that scenario, but I believe it's not so important, as user was not texting anything to that conversation while those changes happened, so verification status is not so important for him for that period of time. 

The main goal is to inform user when the conversation he's going to text/call is not secure anymore.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ProtocolInfoMapper.kt
```diff
@@ -39,7 +39,7 @@ class ProtocolInfoMapperImpl(
                 Conversation.ProtocolInfo.MLS.GroupState.valueOf(protocolInfo.groupState.name),
                 protocolInfo.epoch,
                 protocolInfo.keyingMaterialLastUpdate,
-                Conversation.CipherSuite.fromTag(protocolInfo.cipherSuite.cipherSuiteTag)
+                Conversation.CipherSuite.fromTag(protocolInfo.cipherSuite.cipherSuiteTag),
```
mchenani, line 42:
```suggestion
                Conversation.CipherSuite.fromTag(protocolInfo.cipherSuite.cipherSuiteTag)
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/ConversationVerificationStatusHandler.kt
```diff
@@ -18,45 +18,152 @@
...
+     */
+    private suspend fun fetchVerificationStatusFlow(
+        conversationId: ConversationId,
+        protocol: Conversation.ProtocolInfo
+    ): Flow<VerificationStatus> =
+        observerVerificationStatus(protocol, conversationId)
+            .distinctUntilChanged()
+            .onlyRight()
+            .mapLatest { newStatus ->
+                val currentStatus = conversationRepository.getConversationVerificationStatus(conversationId)
```
mchenani, line 110: question: do we need to call CC anymore? since we update/fetch the state based on the epoch changes? here simply can't we only observe the value in the db?

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/conversation/ConversationEntity.kt
```diff
@@ -76,7 +78,7 @@ data class ConversationEntity(
     enum class MutedStatus { ALL_ALLOWED, ONLY_MENTIONS_AND_REPLIES_ALLOWED, MENTIONS_MUTED, ALL_MUTED }
 
     sealed class ProtocolInfo {
-        object Proteus : ProtocolInfo()
+        data object Proteus : ProtocolInfo()
```
mchenani, line 81:
```suggestion
       object Proteus : ProtocolInfo()
```

mchenani (commented):
added one question and some other tiny comments.

typfel:
> > So checking that've understood the implementation correctly
> > We have the existing ObserveConversationDetailsUseCase which is the entry point which starts the ConversationVerificationStatusHandler.
> > ConversationVerificationStatusHandler does:
> > observes the epoch flow -> ask CC if verified or not -> fetch previous verification status -> derive new verification status -> insert system message if necessary
> 
> Yes, you understood the flow perfectly.
> 
> > 1. Do we only want to update the verification status when conversation if it's displayed on screen? Shouldn't
> >    it be driven from kalium itself and not by the UI?
> 
> Well, observing these statuses for ALL the users conversations may have tragical app performance decreasing. The right solution here is corresponding web-socket event, but we don't have it, so I'm just doing my best with what I have :)

But we will never a web-socket event since this all driven by the clients without any knowledge from the server, no?

Why do think this would have such a big performance impact? Commits are not arriving constantly in conversations. I mean we would need a a separate observer for each conversation we can one observe for all conversations.

> > For example, given clients are added add and then removed again:
> > verified
> > not verified
> > verified
> > Wouldn't we fail to report that the conversation was unverified for time period?
> 
> True, user will lose "not verified" status in that scenario, but I believe it's not so important, as user was not texting anything to that conversation while those changes happened, so verification status is not so important for him for that period of time.
> 
> The main goal is to inform user when the conversation he's going to text/call is not secure anymore.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/ConversationVerificationStatusHandler.kt
borichellow: I'm not sure if I got your question. 
We observe epoch changes, but it only triggers us that something was changed and we need to check if the Verification Status changed (by calling CC).

borichellow:
@typfel Updated the PR, no it run observation in `SessionsScope` once for each user, looks much better i believe :)

mchenani (approved):
🔥

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/MLSConversationsVerificationStatusesHandler.kt
```diff
@@ -0,0 +1,98 @@
...
+                        ) {
+                            return@collect
+                        }
+
+                        if (newStatus == VerificationStatus.NOT_VERIFIED && currentStatus == VerificationStatus.VERIFIED) {
+                            conversationRepository.updateVerificationStatus(VerificationStatus.DEGRADED, conversation.conversation.id)
+                            notifyUserAboutStateChanges(conversation.conversation.id, VerificationStatus.DEGRADED)
+                        } else {
+                            conversationRepository.updateVerificationStatus(newStatus, conversation.conversation.id)
+                        }
```
typfel, line 72: suggestion: extract into private function

typfel (approved):

PR #2048, opened 2023-09-11
feat: avoid race condition when establishing mls 1-1

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Clients will attempt to establish a MLS 1-1 when a connection request is accepted. If more than one client is "online"
 when a connection request is accepted this will result in a race to establish the MLS group, which only one client can win. This is wasteful because the clients who loose the race will also consume key packages but they will be discarded when they realise they have lost the race.

### Solutions

1. Distinguish between live events (Websocket) & catchup events (`GET /notifications`)
2. Add a small delay (3s) before attempting to establish the MLS 1-1 upon receiving a live connection accepted event.

### Testing

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/UserEventReceiver.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/UserEventReceiver.kt
@@ -38,12 +38,7 @@ import com.wire.kalium.logic.functional.map
 import com.wire.kalium.logic.functional.onFailure
 import com.wire.kalium.logic.functional.onSuccess
 import com.wire.kalium.logic.kaliumLogger
-import com.wire.kalium.util.KaliumDispatcher
-import com.wire.kalium.util.KaliumDispatcherImpl
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.delay
-import kotlinx.coroutines.launch
+import kotlin.time.Duration.Companion.ZERO
 import kotlin.time.Duration.Companion.seconds
 
 internal interface UserEventReceiver : EventReceiver<Event.User>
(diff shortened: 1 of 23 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/UserEventReceiver.kt
```diff
@@ -85,7 +96,20 @@ internal class UserEventReceiverImpl internal constructor(
...
+                if (event.connection.status != ConnectionState.ACCEPTED) {
+                    return@flatMap Either.Right(Unit)
+                }
+
+                if (event.live) {
+                    resolveActiveOneOnOneScope.launch {
+                        kaliumLogger.d("Delay resolving active one-on-one since we are live to avoid race conditions")
+                        delay(3.seconds)
+                        oneOnOneResolver.resolveOneOnOneConversationWithUserId(event.connection.qualifiedToId).map { }
+                    }
```
vitorhugods, line 108: Just to confirm.

The idea here is that during these 3 seconds, the protocol might have been updated already. So when it calls `resolveOneOnOneConversationWithUserId`, the resolver will notice that it is already MLS, and shortcircuit successfully.
And checking if `live` or not makes a difference because if it's not live, we can assume that the migration was already solved and we'll know about it shortly after in the coming events.
Correct?

Makes sense to me. I wonder if adding a small jitter would be nice.

Something like $delay = 3 + [0..3] seconds$

Attempting to reduce the chance of clients racing to migrate.

Probably not worth it.

vitorhugods (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/UserEventReceiver.kt
typfel: The idea is that during 3 seconds we will in 99% of the cases receive a welcome message which establishes the MLS group. If there's no errors it's the client which accepts the connection which also establish the group.

Checking if we are `live` or not is to avoid introducing a delays when catching up. But I realised now that we need to post pone resolving the 1-1 also in the non-live case since otherwise we would discard the welcome message.

I'll fix this.

borichellow (approved):

PR #2056, opened 2023-09-13
chore: add archived field to conversations table (WPB-4471)

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We were retrieving `otr_archived` and `otr_archived_ref` from the API but not persisting in the database.

### Solutions

Add `otr_archived` and `otr_archived_ref` to conversation table when getting from API.

#### Test Coverage (Optional)

#### How to Test

Currently the only way to test is:
- Fresh install/login
- check from App Inspection that you have both fields in the `Conversation` table and if you have any archived conversation, it should be represented as such (1 = true, 0 = false)

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
@@ -258,7 +258,7 @@ internal class ConnectionDataSource(
                         messageTimer = null,
                         userMessageTimer = null,
                         archived = false,
-                        archivedRef = ""
+                        archivedDateTime = null
                     )
                 )
             }
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
@@ -258,7 +258,7 @@ internal class ConnectionDataSource(
                         messageTimer = null,
                         userMessageTimer = null,
                         archived = false,
-                        archivedDateTime = null
+                        archivedInstant = null
                     )
                 )
             }
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
@@ -257,8 +257,8 @@ internal class ConnectionDataSource(
                         receiptMode = ConversationEntity.ReceiptMode.DISABLED,
                         messageTimer = null,
                         userMessageTimer = null,
-                        otrArchived = false,
-                        otrArchivedRef = ""
+                        archived = false,
+                        archivedRef = ""
                     )
                 )
             }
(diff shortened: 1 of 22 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
```diff
@@ -256,7 +256,9 @@ internal class ConnectionDataSource(
                         accessRole = emptyList(),
                         receiptMode = ConversationEntity.ReceiptMode.DISABLED,
                         messageTimer = null,
-                        userMessageTimer = null
+                        userMessageTimer = null,
+                        otrArchived = false,
+                        otrArchivedRef = ""
```
typfel, line 261:
```suggestion
                        archived = false,
                        archivedRef = ""
```

I would suggest to drop the `otr` prefix because it's a legacy naming which just doesn't make sense. OTR is an abbreviation for "off the record" but that has nothing to do with we do now. OTR was used in a few places two differentiate between encrypted and non-encrypted features.

alexandreferris: Oh great! I was not aware of that!

Garzas (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
```diff
@@ -68,7 +68,9 @@ data class Conversation(
     val creatorId: String?,
     val receiptMode: ReceiptMode,
     val messageTimer: Duration?,
-    val userMessageTimer: Duration?
+    val userMessageTimer: Duration?,
+    val archived: Boolean,
+    val archivedRef: String
```
vitorhugods, line 73: question: What is `archivedRef`? What is its format? What if it's empty? Wouldn't `null` be clearer than empty?

alexandreferris: This value is rather strange I must say, from Scala we see it being saved as `Long`, although we receive it as DateTime/Instant String `1970-01-01 ...`.

But not used anywhere apart from setting/getting from db/api 🤷🏻‍♂️ 

Thought on having it as nullable but wasn't so sure, happy to change it tho :)

vitorhugods: suggestion: make the name reflect the format. 

Something like `archivedDateTime` or similar, for example.

----

suggestion: better nullable than empty.

If the field is not nullable, I'd expect it to be always parseable. I'd expect `DateTime.fromString(archivedDatetime)` to 100% work if it's not null.
Having it fail with `IllegalArgumentException` because it's an empty string could lead to nasty errors.

So by having it nullable, we can easily differentiate between having this value or not. The compiler will help us.

alexandreferris: Sure! Works out, will do the changes. Thanks!

vitorhugods (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -180,7 +184,9 @@ internal class ConversationMapperImpl(
             creatorId = creatorId,
             receiptMode = receiptModeMapper.fromEntityToModel(receiptMode),
             messageTimer = messageTimer?.toDuration(DurationUnit.MILLISECONDS),
-            userMessageTimer = userMessageTimer?.toDuration(DurationUnit.MILLISECONDS)
+            userMessageTimer = userMessageTimer?.toDuration(DurationUnit.MILLISECONDS),
+            archived = archived,
+            archivedDateTime = archivedInstant?.toIsoDateTimeString()
```
MohamadJaara, line 189: since we now store it as instant we can set the domain model to instant also

MohamadJaara (approved):

PR #2060, opened 2023-09-14
fix: getCallManagerForClient is not thread safe

MohamadJaara:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

`GlobalCallManager.getCallManagerForClient` is not thread safe, this can cause a case where multiple instances of AVS being created 

### Solutions

use `computeIfAbsent` to fix the issue

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
+++ b/logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
@@ -85,22 +85,22 @@ actual class GlobalCallManager(
         conversationClientsInCallUpdater: ConversationClientsInCallUpdater,
         kaliumConfigs: KaliumConfigs
     ): CallManager {
-        return callManagerHolder[userId] ?: CallManagerImpl(
-            calling = calling,
-            callRepository = callRepository,
-            userRepository = userRepository,
-            currentClientIdProvider = currentClientIdProvider,
-            selfConversationIdProvider = selfConversationIdProvider,
-            callMapper = callMapper,
-            messageSender = messageSender,
-            conversationRepository = conversationRepository,
-            federatedIdMapper = federatedIdMapper,
-            qualifiedIdMapper = qualifiedIdMapper,
-            videoStateChecker = videoStateChecker,
-            conversationClientsInCallUpdater = conversationClientsInCallUpdater,
-            kaliumConfigs = kaliumConfigs
-        ).also {
-            callManagerHolder[userId] = it
+        return callManagerHolder.computeIfAbsent(userId) {
+            CallManagerImpl(
+                calling = calling,
+                callRepository = callRepository,
+                userRepository = userRepository,
+                currentClientIdProvider = currentClientIdProvider,
+                selfConversationIdProvider = selfConversationIdProvider,
+                callMapper = callMapper,
+                messageSender = messageSender,
+                conversationRepository = conversationRepository,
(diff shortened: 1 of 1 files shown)


yamilmedina (approved):

logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
```diff
@@ -85,22 +85,22 @@ actual class GlobalCallManager(
...
+                selfConversationIdProvider = selfConversationIdProvider,
+                callMapper = callMapper,
+                messageSender = messageSender,
+                conversationRepository = conversationRepository,
+                federatedIdMapper = federatedIdMapper,
+                qualifiedIdMapper = qualifiedIdMapper,
+                videoStateChecker = videoStateChecker,
+                conversationClientsInCallUpdater = conversationClientsInCallUpdater,
+                kaliumConfigs = kaliumConfigs
+            )
```
ohassine, line 103: Not sure if I understand correctly the fix as `computeIfAbsent` is not thread safe AFAIK.

Also this way we will not have instance by logged in user.

ohassine (changes requested):

logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
MohamadJaara: Previously, there was a process in place. It first checked if a map contained a "callManager" for a specific user ID. If not, it would create a new "callManager" and save its reference in the map. However, there was a problem. If two threads simultaneously performed this check, they could both conclude that there was no "callManager" for the user and both create a new one.

Now, we've improved this process by using a Java data structure called `ConcurrentHashMap`. With it, we use the `computeIfAbsent` method, which ensures that even if multiple threads try to create a "callManager" at the same time, it will be done in a way that is safe and does not result in duplicate "callManagers."

MohamadJaara: you can check out what it does [here](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)

MohamadJaara: > as computeIfAbsent is not thread safe AFAIK.

`computeIfAbsent` in `HashMap` are not thread safe but we are using `ConcurrentHashMap` which it is

ohassine: but now with this changes, we are no longer adding instances to `callManagerHolder`

MohamadJaara: > If the specified key is not already associated with a value, attempts to compute its value using the given mapping function and enters it into this map unless null

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-

ohassine (approved):

PR #2066, opened 2023-09-15
feat: Add user typing indicator configuration [WPB-4589]

gongracr:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In order to be able to switch on and off the global user setting for displaying typing indicator, the right classes needed to be added on the api, repository and use case layers.

#### Test Coverage (Optional)

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
@@ -59,6 +59,8 @@ interface UserConfigRepository {
     fun isSecondFactorPasswordChallengeRequired(): Either<StorageFailure, Boolean>
     fun isReadReceiptsEnabled(): Flow<Either<StorageFailure, Boolean>>
     fun setReadReceiptsStatus(enabled: Boolean): Either<StorageFailure, Unit>
+    fun isTypingIndicatorEnabled(): Flow<Either<StorageFailure, Boolean>>
+    fun setTypingIndicatorStatus(isTyping: Boolean): Either<StorageFailure, Unit>
     fun setGuestRoomStatus(status: Boolean, isStatusChanged: Boolean?): Either<StorageFailure, Unit>
     fun getGuestRoomLinkStatus(): Either<StorageFailure, GuestRoomLinkStatus>
     fun observeGuestRoomLinkFeatureFlag(): Flow<Either<StorageFailure, GuestRoomLinkStatus>>
@@ -200,12 +202,19 @@ class UserConfigDataSource(
     }
 
     override fun isReadReceiptsEnabled(): Flow<Either<StorageFailure, Boolean>> =
-        userConfigStorage.isReadReceiptsEnabled().wrapStorageRequest()
+        userConfigStorage.areReadReceiptsEnabled().wrapStorageRequest()
 
     override fun setReadReceiptsStatus(enabled: Boolean): Either<StorageFailure, Unit> =
         wrapStorageRequest {
             userConfigStorage.persistReadReceipts(enabled)
         }
+
+    override fun isTypingIndicatorEnabled(): Flow<Either<StorageFailure, Boolean>> =
+        userConfigStorage.isTypingIndicatorEnabled().wrapStorageRequest()
+
+    override fun setTypingIndicatorStatus(isTyping: Boolean): Either<StorageFailure, Unit> = wrapStorageRequest {
+        userConfigStorage.persistTypingIndicator(isTyping)
+    }
 
     override fun setGuestRoomStatus(status: Boolean, isStatusChanged: Boolean?): Either<StorageFailure, Unit> =
         wrapStorageRequest {
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/properties/UserPropertyRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/properties/UserPropertyRepository.kt
@@ -33,6 +33,10 @@ interface UserPropertyRepository {
     suspend fun observeReadReceiptsStatus(): Flow<Either<CoreFailure, Boolean>>
     suspend fun setReadReceiptsEnabled(): Either<CoreFailure, Unit>
     suspend fun deleteReadReceiptsProperty(): Either<CoreFailure, Unit>
+    suspend fun getTypingIndicatorStatus(): Boolean
+    suspend fun observeTypingIndicatorStatus(): Flow<Either<CoreFailure, Boolean>>
+    suspend fun setTypingIndicatorEnabled(): Either<CoreFailure, Unit>
+    suspend fun removeTypingIndicatorProperty(): Either<CoreFailure, Unit>
 }
 
 internal class UserPropertyDataSource(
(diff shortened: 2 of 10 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/properties/UserPropertyRepository.kt
```diff
@@ -58,4 +62,23 @@ internal class UserPropertyDataSource(
...
+        propertiesApi.setProperty(PropertiesApi.PropertyKey.WIRE_TYPING_MODE, 1)
+    }.flatMap {
+        userConfigRepository.setTypingIndicatorStatus(true)
+    }
+
+    override suspend fun removeTypingIndicatorProperty(): Either<CoreFailure, Unit> = wrapApiRequest {
+        propertiesApi.deleteProperty(PropertiesApi.PropertyKey.WIRE_TYPING_MODE)
+    }.flatMap {
+        userConfigRepository.setTypingIndicatorStatus(false)
+    }
```
yamilmedina, line 83:
```suggestion
    override suspend fun enableTypingIndicatorProperty(): Either<CoreFailure, Unit> = wrapApiRequest {
        propertiesApi.deleteProperty(PropertiesApi.PropertyKey.WIRE_TYPING_MODE)
    }.flatMap {
        userConfigRepository.setTypingIndicatorStatus(true)
    }
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/properties/UserPropertyRepository.kt
```diff
@@ -58,4 +62,23 @@ internal class UserPropertyDataSource(
...
+            ?.fold({ false }, { it }) ?: false
+
+    override suspend fun observeTypingIndicatorStatus(): Flow<Either<CoreFailure, Boolean>> =
+        userConfigRepository.isTypingIndicatorEnabled()
+
+    override suspend fun setTypingIndicatorEnabled(): Either<CoreFailure, Unit> = wrapApiRequest {
+        propertiesApi.setProperty(PropertiesApi.PropertyKey.WIRE_TYPING_MODE, 1)
+    }.flatMap {
+        userConfigRepository.setTypingIndicatorStatus(true)
+    }
```
yamilmedina, line 77:
```suggestion
    override suspend fun setTypingIndicatorDisabled(): Either<CoreFailure, Unit> = wrapApiRequest {
        propertiesApi.setProperty(PropertiesApi.PropertyKey.WIRE_TYPING_MODE, 0)
    }.flatMap {
        userConfigRepository.setTypingIndicatorStatus(false)
    }
```

yamilmedina: This is an opt-out feature meaning, it's enabled by default if not present or when you opt-out, you have the property as `WIRE_TYPING_MODE = 0`  

https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/29065437/Use+case+Typing+indicator#Syncing-typing-indicator-preference-for-each-account-across-the-devices

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/properties/UserPropertyRepository.kt
yamilmedina: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/29065437/Use+case+Typing+indicator#Syncing-typing-indicator-preference-for-each-account-across-the-devices

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
```diff
@@ -200,13 +202,20 @@ class UserConfigDataSource(
...
 
     override fun setReadReceiptsStatus(enabled: Boolean): Either<StorageFailure, Unit> =
         wrapStorageRequest {
             userConfigStorage.persistReadReceipts(enabled)
         }
 
+    override fun isTypingIndicatorEnabled(): Flow<Either<StorageFailure, Boolean>> =
+        userConfigStorage.isTypingIndicatorEnabled().wrapStorageRequest()
+
+    override fun setTypingIndicatorStatus(isTyping: Boolean): Either<StorageFailure, Unit> = wrapStorageRequest {
```
yamilmedina, line 215:
```suggestion
    override fun setTypingIndicatorStatus(enabled: Boolean): Either<StorageFailure, Unit> = wrapStorageRequest {
```

Or maybe `isTypingEnabled` might be a proper name here, what do you think ?

yamilmedina (commented):
Great work and nice come back!  I left some comments we need to address, looking asombroso 🍷 ⚽

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/properties/PropertiesApi.kt
```diff
@@ -26,7 +26,8 @@ interface PropertiesApi {
     suspend fun deleteProperty(propertyKey: PropertyKey): NetworkResponse<Unit>
 
     enum class PropertyKey(val key: String) {
-        WIRE_RECEIPT_MODE("WIRE_RECEIPT_MODE")
+        WIRE_RECEIPT_MODE("WIRE_RECEIPT_MODE"),
+        WIRE_TYPING_MODE("WIRE_TYPING_MODE")
```
yamilmedina, line 30:
```suggestion
        WIRE_TYPING_MODE("WIRE_TYPING_INDICATOR_MODE")
```

yamilmedina (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventMapper.kt
```diff
@@ -185,9 +187,32 @@ class EventMapper(
...
+                    WIRE_RECEIPT_MODE.key -> ReadReceiptModeSet(
+                        id,
+                        transient,
+                        fieldKeyValue.value == 1
+                    )
+
+                    WIRE_TYPING_INDICATOR_MODE.key -> TypingIndicatorModeSet(
+                        id,
+                        transient,
+                        false
```
saleniuk, line 204: question: shouldn't the `fieldKeyValue.value` be compared here as well instead of the hardcoded `false`? 
I found this: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/29065437/Use+case+Typing+indicator#Syncing-typing-indicator-preference-for-each-account-across-the-devices
I think that only `0` should be treated as disabled, not any value, right? 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/properties/UserPropertyRepository.kt
```diff
@@ -58,4 +62,23 @@ internal class UserPropertyDataSource(
         userConfigRepository.setReadReceiptsStatus(false)
     }
 
+    override suspend fun getTypingIndicatorStatus(): Boolean =
+        userConfigRepository.isTypingIndicatorEnabled()
+            .firstOrNull()
+            ?.fold({ false }, { it }) ?: false
```
saleniuk, line 68: question: this indicator is enabled by default and in `userConfigRepository` it returns a `true` by default, also `ObserveTypingIndicatorEnabledUseCase` returns `true` when `Either.Left`, so shouldn't it also return `true` if it's null here as well?

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
gongracr: Good catch. I think I named it wrongly as I misunderstood the task at the beginning. Changing it to `isEnabled` to keep coherence with other properties.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventMapper.kt
gongracr: I agree this is indeed a confusing way to handle the flag enable/disable value. The thing is that due to the nature of the feature itself (opt-out / enabled by default) the Backend is only sending:

1. A `user.properties-set` event with key `WIRE_TYPING_INDICATOR_MODE` and value `0` when the flag gets disabled on other clients
2. And a `user.properties-delete` event with key `WIRE_TYPING_INDICATOR_MODE` when enabling it on other clients

This means that `updateUserProperties()` will never be called with `WIRE_TYPING_INDICATOR_MODE` key when enabling the feature. But if you think changing it to `fieldKeyValue.value != 0` is less future error prone we can also leave the comparison.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/properties/UserPropertyRepository.kt
gongracr: You are right, I probably didn't check it when copying the logic from the read receipt feature (which is opt-in). 
As usual great eyes 🐛 🦅

saleniuk (approved):

PR #2067, opened 2023-09-15
feat: Lazy init for Scopes and Repositories

borichellow:
# What's new in this PR?

### Issues

getters pattern for getting any Scope or Repository in scopes has a big disadvantage: 
initializing a new instance every time when object is requested which causes performance issues (for example creating a database multiple times takes some time, etc.). 

### Solutions

- remove getters for all Scopes and Repositories
- move initialization of Scopes and Repositories into `lazy` to create instances only when it was requested but only once.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/GlobalKaliumScope.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/GlobalKaliumScope.kt
@@ -105,28 +105,30 @@ class GlobalKaliumScope internal constructor(
         )
     }
 
-    internal val serverConfigRepository: ServerConfigRepository
-        get() = ServerConfigDataSource(
+    internal val serverConfigRepository: ServerConfigRepository by lazy {
+        ServerConfigDataSource(
             unboundNetworkContainer.serverConfigApi,
             globalDatabase.serverConfigurationDAO,
             unboundNetworkContainer.remoteVersion,
             kaliumConfigs.developmentApiEnabled
         )
+    }
 
-    val sessionRepository: SessionRepository
-        get() = SessionDataSource(
+    val sessionRepository: SessionRepository by lazy {
+        SessionDataSource(
             globalDatabase.accountsDAO,
             globalPreferences.authTokenStorage,
             serverConfigRepository,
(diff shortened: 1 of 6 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/GlobalKaliumScope.kt
```diff
@@ -105,28 +105,30 @@ class GlobalKaliumScope internal constructor(
...
             unboundNetworkContainer.serverConfigApi,
             globalDatabase.serverConfigurationDAO,
             unboundNetworkContainer.remoteVersion,
             kaliumConfigs.developmentApiEnabled
         )
+    }
 
-    val sessionRepository: SessionRepository
-        get() = SessionDataSource(
+    val sessionRepository: SessionRepository by lazy {
```
MohamadJaara, line 117: this will make all of those Repository as singelton, and they are already created on demand so not sure if this will change anything

MohamadJaara (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/GlobalKaliumScope.kt
MohamadJaara: by lazy i would assume something like

```
class MyScope(
    val conversationRepo: () -> ConversationRepo 
) {

val conversationRepository by lazy { conversationRepo() }

}
```

this will delay creating the repos inside the scope and still use the same instance inside the scope

MohamadJaara: or we can delegate creating those repos for auser to another class and inject that into the scopes 

```
class UserRepoProvider(
    val userStorage: UserStorage
    val userNetworkContainer: AuthenticatedNetworkContainer
)  {

    fun provideConversationRepository(): ConversationRepository

}


class TeamScope(
    val repoProvider: UserRepoProvider
) {

// create usse cases
}

MohamadJaara:
> creating a database multiple times takes some time

the DB and the shared pref files are created only once on app creation

PR #2068, opened 2023-09-15
feat: limit re-trying commit to 2 retry attemps

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We are currently re-trying MLS commits with recoverable errors indefinitely. This is dangerous we can get stuck in a loop an easily drain someone's key packages. While the errors might re-coverable eventually it might require an action by another client, like uploading new key packages which isn't likely to happen within a few seconds.

### Solutions

Limit re-trying on recoverable errors.

#### Test Coverage

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
@@ -111,10 +111,15 @@ private enum class CommitStrategy {
 }
 
 private fun CoreFailure.getStrategy(
+    retryCount: Int,
     retryOnClientMismatch: Boolean = true,
     retryOnStaleMessage: Boolean = true
 ): CommitStrategy {
-    return if (this is NetworkFailure.ServerMiscommunication && this.kaliumException is KaliumException.InvalidRequestError) {
+    return if (
+        retryCount > 0 &&
+        this is NetworkFailure.ServerMiscommunication &&
+        kaliumException is KaliumException.InvalidRequestError
+    ) {
         if (this.kaliumException.isMlsClientMismatch() && retryOnClientMismatch) {
             CommitStrategy.DISCARD_AND_RETRY
         } else if (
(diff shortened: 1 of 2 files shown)


MohamadJaara (approved):
LGTM

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
```diff
@@ -111,10 +111,15 @@ private enum class CommitStrategy {
 }
 
 private fun CoreFailure.getStrategy(
+    retryCount: Int,
```
vitorhugods, line 114: question: Is `retryCount` the number of tries until now, or the maximum number of tries?

It looks like number of tries, where it starts with 2 and goes down as it fails. The name + `retryCount > 0` looks a bit weird.

The logic is a bit split between `getStrategy` and `handleCommitFailure`, which makes a bit harder to follow and figure out what's going on.

Maybe renaming it to `remainingAttemptsCount`, so it's clearer that if it's `0` we should stop.

vitorhugods (changes requested):
Looks good, there's maybe room for improvement regarding the `retryCount` parameter.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
typfel: Yep confusing name, I'll rename it.

>The logic is a bit split between getStrategy and handleCommitFailure, which makes a bit harder to follow and figure out what's going on.

Any idea on how to improve it? Given the current recursive setup I don't see how avoid also having a parameter in the `handleCommitFailure` function.

vitorhugods: > Yep confusing name, I'll rename it.
> 
> > The logic is a bit split between getStrategy and handleCommitFailure, which makes a bit harder to follow and figure out what's going on.
> 
> Any idea on how to improve it? Given the current recursive setup I don't see how avoid also having a parameter in the `handleCommitFailure` function.

Not really. I think the name change was already an improvement. Made it clearer.

vitorhugods (approved):

PR #2102, opened 2023-09-28
refactor: introduce feature config handlers to avoid code duplication

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Code for handling when a feature config changes is duplicated in `SyncFeatureConfigsUseCase` and in `FeatureConfigEventReceiver`

`SyncFeatureConfigsUseCase` is run during the slow sync and fetches all feature config and updates our local configuration of the feature.

`FeatureConfigEventReceiver` process events when individual feature configs change after the slow sync is completed.

The code in these to classes is almost identical and that it isn't identical comes from the false assumption that `SyncFeatureConfigsUseCase` is only called once after login, which isn't true.

### Solutions

Extract code into  feature config handlers which can be re-used in both classes.

### Testing

I've intentionally made the minimal amount of changes to the tests to validate that I haven't broken anything. Once this PR is merged we should move the testing of each individual feature config into separate test classes.

#### Test Coverage

### Attachments (Optional)

----

----

----
##### References


logic/src/commonTest/kotlin/com/wire/kalium/logic/sync/receiver/FeatureConfigEventReceiverTest.kt
```diff
@@ -270,7 +279,7 @@ class FeatureConfigEventReceiverTest {
             verify(arrangement.userConfigRepository)
                 .suspendFunction(arrangement.userConfigRepository::setTeamSettingsSelfDeletionStatus)
                 .with(matching {
-                    it.hasFeatureChanged == true && it.enforcedSelfDeletionTimer !is TeamSelfDeleteTimer.Disabled
+                    it.hasFeatureChanged == false && it.enforcedSelfDeletionTimer !is TeamSelfDeleteTimer.Disabled
```
typfel, line 282: @gongracr I changed the expectation of `hasFeatureChanged` to `false`. It doesn't make sense to me to flag the feature has having changed if we didn't have any previous persisted configuration of it because that would imply that we would flag the feature as changed after the initial sync.

logic/src/commonTest/kotlin/com/wire/kalium/logic/sync/receiver/FeatureConfigEventReceiverTest.kt
```diff
@@ -291,7 +300,7 @@ class FeatureConfigEventReceiverTest {
             verify(arrangement.userConfigRepository)
                 .function(arrangement.userConfigRepository::setTeamSettingsSelfDeletionStatus)
                 .with(matching<TeamSettingsSelfDeletionStatus> {
-                    it.hasFeatureChanged == true && it.enforcedSelfDeletionTimer is TeamSelfDeleteTimer.Disabled
+                    it.hasFeatureChanged == false && it.enforcedSelfDeletionTimer is TeamSelfDeleteTimer.Disabled
```
typfel, line 303: @gongracr same as above.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/FeatureConfigEventReceiver.kt
```diff
@@ -19,219 +19,77 @@
...
-                    E2EISettings(
-                        isRequired = event.model.status == Status.ENABLED,
-                        discoverUrl = event.model.config.discoverUrl,
-                        gracePeriodEnd = Instant.fromEpochMilliseconds(gracePeriodEndMs)
+            .onFailure {
+                if (it is CoreFailure.FeatureNotImplemented) {
+                    kaliumLogger.logEventProcessing(
+                        EventLoggingStatus.SKIPPED,
+                        event,
+                        Pair("info", "Ignoring unknown feature config update")
```
gongracr, line 67: Why don't we move these logs to each individual handler? I can't see any specific log referencing to each receiver class so we should be good to move it?

gongracr (approved):
Just one small comment to move the logs to each individual handler, so we centralize that piece as well. But other than that, non blocking so great job refactoring 🧹 👏

alexandreferris (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/FeatureConfigEventReceiver.kt
typfel: What do I gain by moving this to each handler? Don't I need to repeat these logs for each handler then?

gongracr: well, as a starter not having to add the same log to 2 different receivers, FeatureConfig and SyncFeatureConfig. Then if for any reason, we decide to start using these handlers in different places as well, then we will not need to add the logs on the invoking class.

typfel: OK, I see your point. We can refactor the again if it becomes and issue. I'll keep it like it is for now since the logging for events and for slow sync are different.

borichellow (approved):

PR #2104, opened 2023-09-28
feat: SystemMessage about MLS verified conversation

borichellow:
# What's new in this PR?

When the MLS conversation becomes Verified, app informs user about it by adding SystemMessage


alexandreferris (approved):

Garzas (changes requested):
Looking good, but can we have some tests for it? 😄

borichellow:
> Looking good, but can we have some tests for it? 😄

there is one test updated in `MLSConversationsVerificationStatusesHandlerTest` ;)

Garzas (approved):

PR #2109, opened 2023-10-02
feat: archived conversation list [WPB-4429]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Changed that we can fetch either conversation without archived ones or only archived


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -112,7 +113,7 @@ interface ConversationRepository {
 
     suspend fun getConversationList(): Either<StorageFailure, Flow<List<Conversation>>>
     suspend fun observeConversationList(): Flow<List<Conversation>>
-    suspend fun observeConversationListDetails(includeArchived: Boolean): Flow<List<ConversationDetails>>
+    suspend fun observeConversationListDetails(fromArchived: Boolean): Flow<List<ConversationDetails>>
```
gongracr, line 116: If we want to use the past participle it would be better to append `xxxConversations` to the variable. Otherwise I'd prefer shorter form `fromArchive` instead. Just personal choice 😄

gongracr (approved):
Looks great. Thanks for the extra set of tests. Just one small grammatically nit, but feel free to ignore it 😅

yamilmedina (approved):

PR #2141, opened 2023-10-17
refactor: unify access token refreshing logic [WPB-5038]

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Users are sometimes reporting being stuck in connecting / decrypting messages and they can only recover after killing and restarting the app.

### Causes

When a token refresh is needed (backend returns a 401), Ktor will give us the tokens currently being used and ask for new tokens.
In case of failure by _any reason_ during the refresh, we should throw an exception. **But instead**, for some reason, we're actually returning _null_ in Ktor's [`refreshToken` function](https://ktor.io/docs/bearer-client.html#step9).

This makes it so Ktor will actually try to use these `null` tokens in the next requests. And the backend will again return `401`. Looping back to the token refresh function, that will now throw a `NullPointerException` because we double-bang the `oldTokens`.

The flow pretty much was:

```mermaid
sequenceDiagram
    loop Invalid tokens
    Kalium->>Ktor: Original Request with expired or null tokens
    Ktor->>Server: Forwarding
    Server->>Ktor: 401
    critical Token Refreshing
    Ktor->>Kalium: "Please refresh the tokens"
    Kalium->>Kalium: New tokens request fail
    Kalium->>Ktor: Take these "Null" tokens
    end
    end
```

#### Unify the access token refreshing
The refreshing is currently being done by two entities (`SessionManagerImpl` and `UpgradeUserSessionUseCase`), that are directly accessing the token API.

- Added a `AccessTokenRepository` to mask the storage and APIs, handling source-specific errors, and also making sure the refresh token is always present to keep consistency.
- Added a specialised `AccessTokenRefresher` that is solely responsible for grabing a new token and storing it.
- Removed dependencies to `AccessTokenApi` and `AuthTokenStorage` from the `SessionManagerImpl` and `UpgradeUserSessionUseCase`, going straight to the `AccessTokenRefresher` for token-refresh purposes.

#### Stop mapping failures to `null`

- It's alright to throw exceptions. Ktor will swallow them and we will catch them, mapping it to a `NetworkFailure`.

So the flow becomes:

```mermaid
sequenceDiagram
    Kalium->>Ktor: Original Request with expired or null tokens
    Ktor->>Server: Forwarding
    Server->>Ktor: 401
    critical Token Refreshing
    Ktor->>Kalium: "Please refresh the tokens"
    Kalium->>Kalium: New tokens request fail
    Kalium->>Ktor: Exception!
    Ktor->>Kalium: Your request failed :(
    end
```

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/network/SessionManagerImpl.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/network/SessionManagerImpl.kt
@@ -22,26 +22,26 @@ import com.wire.kalium.logger.obfuscateId
 import com.wire.kalium.logic.NetworkFailure
 import com.wire.kalium.logic.configuration.server.ServerConfigMapper
 import com.wire.kalium.logic.data.id.QualifiedID
+import com.wire.kalium.logic.data.id.toApi
 import com.wire.kalium.logic.data.id.toDao
 import com.wire.kalium.logic.data.logout.LogoutReason
 import com.wire.kalium.logic.data.session.SessionMapper
 import com.wire.kalium.logic.data.session.SessionRepository
 import com.wire.kalium.logic.di.MapperProvider
+import com.wire.kalium.logic.feature.session.token.AccessTokenRefresherFactory
 import com.wire.kalium.logic.functional.fold
 import com.wire.kalium.logic.functional.map
 import com.wire.kalium.logic.functional.nullableFold
 import com.wire.kalium.logic.functional.onFailure
 import com.wire.kalium.logic.functional.onSuccess
 import com.wire.kalium.logic.kaliumLogger
-import com.wire.kalium.logic.wrapApiRequest
 import com.wire.kalium.logic.wrapStorageRequest
 import com.wire.kalium.network.api.base.authenticated.AccessTokenApi
-import com.wire.kalium.network.api.base.model.AccessTokenDTO
 import com.wire.kalium.network.api.base.model.ProxyCredentialsDTO
-import com.wire.kalium.network.api.base.model.RefreshTokenDTO
 import com.wire.kalium.network.api.base.model.SessionDTO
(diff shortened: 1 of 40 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/network/SessionManagerImpl.kt
```diff
@@ -88,42 +89,37 @@ class SessionManagerImpl internal constructor(
...
+                userId = userId.toApi(),
+                tokenType = refreshResult.accessToken.tokenType,
+                accessToken = refreshResult.accessToken.value,
+                refreshToken = refreshResult.refreshToken.value,
+                cookieLabel = null
             )
-        }.map {
-            sessionMapper.fromEntityToSessionDTO(it)
-        }.onSuccess {
-            session = it
```
typfel, line 104: Where do we assign the new session to `session`? Or should this property be removed?

typfel (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/network/SessionManagerImpl.kt
vitorhugods: Nice catch.

I removed it during debugging and forgot to add again.

This code also had no tests before, and I didn't cover every single thing yet.

I'll fix it and cover this one too.

Makes me want to transform `session` into an observable, instead of whatever this this.

vitorhugods: Fixed and covered with tests

typfel (approved):

alexandreferris (approved):

PR #2186, opened 2023-11-03
feat: count searched message position (WPB-4986)

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There was no way directly from our pager to get the selected searched message position.

### Solutions

This PR aims to add a usecase / repository / DAO to retrieve the message position in DESC order to be used to scroll the user into the right selected message.

There is also the addition of `KaliumOffsetQueryPagingSource` which is an overriden method of `OffsetQueryPagingSource` but with an extra param: `initialOffset` instead of using the given offset provided by :`queryProvider` in order to properly navigate the pager.

### Dependencies (Optional)

AR PR will follow soon.

#### Test Coverage (Optional)

#### How to Test

No way to test this now, only when AR PR will be created.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetSearchedConversationMessagePositionUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetSearchedConversationMessagePositionUseCase.kt
@@ -18,7 +18,6 @@
 package com.wire.kalium.logic.feature.message
 
 import com.wire.kalium.logic.CoreFailure
-import com.wire.kalium.logic.StorageFailure
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.message.MessageRepository
 import com.wire.kalium.logic.functional.fold
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetSearchedConversationMessagePositionUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetSearchedConversationMessagePositionUseCase.kt
@@ -20,14 +20,20 @@ package com.wire.kalium.logic.feature.message
 import com.wire.kalium.logic.StorageFailure
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.message.MessageRepository
-import com.wire.kalium.logic.functional.Either
+import com.wire.kalium.logic.feature.conversation.GetConversationUnreadEventsCountUseCase
+import com.wire.kalium.logic.functional.fold
 
 interface GetSearchedConversationMessagePositionUseCase {
 
     suspend operator fun invoke(
         conversationId: ConversationId,
         messageId: String
-    ): Either<StorageFailure, Int>
+    ): Result
+
+    sealed class Result {
+        data class Success(val position: Int) : Result()
+        data class Failure(val storageFailure: StorageFailure) : Result()
+    }
 }
 
 internal class GetSearchedConversationMessagePositionUseCaseImpl internal constructor(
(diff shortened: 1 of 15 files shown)


persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq
```diff
@@ -518,3 +518,10 @@ WHERE text LIKE ('%' || :searchQuery || '%')
 AND conversationId = :conversationId
 AND contentType = 'TEXT'
 ORDER BY date DESC;
+
+selectSearchedConversationMessagePosition:
+SELECT COUNT(*) FROM Message
+WHERE
+conversation_id = :conversationId
+AND creation_date >= (SELECT creation_date FROM Message WHERE id = :messageId LIMIT 1)
```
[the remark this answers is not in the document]
alexandreferris: If ordered directly through the `id` we won't be getting the correct messages as `messageId` is created using `uuid4` (uuid v4) and not `uuid v1` (which generates the id based on the time as well).

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/message/MessageRepositoryTest.kt
```diff
@@ -521,6 +521,38 @@ class MessageRepositoryTest {
...
+    fun givenSearchedMessages_whenMessageIsSelected_thenReturnMessagePosition() = runTest {
+        // given
+        val qualifiedIdEntity = TEST_QUALIFIED_ID_ENTITY
+        val conversationId = TEST_CONVERSATION_ID
+        val message = TEST_MESSAGE_ENTITY.copy(
+            id = "msg1",
+            conversationId = qualifiedIdEntity,
+            content = MessageEntityContent.Text("message 1")
+        )
+        val expectedResult = 113
```
yamilmedina, line 534: Suggestion: Maybe here we can rename it to have something like `index` or Position in the name ? 
`expectedMessagePosition` or `expectedResultPosition`

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/message/message/GetSearchedConversationMessagePositionUseCaseTest.kt
```diff
@@ -0,0 +1,137 @@
...
+            messageId = MESSAGE_ID
+        )
+
+        assertIs<GetSearchedConversationMessagePositionUseCase.Result.Failure>(result)
+        assertEquals(cause, result.cause)
+    }
+
+    @Test
+    fun givenRepositorySucceeds_whenInvokingUseCase_thenShouldPropagateTheSuccess() = runTest {
+        val expectedPosition = 113
```
yamilmedina, line 87: Like this 🚀  see prev. comment

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetSearchedConversationMessagePositionUseCase.kt
```diff
@@ -0,0 +1,61 @@
...
+interface GetSearchedConversationMessagePositionUseCase {
+
+    suspend operator fun invoke(
+        conversationId: ConversationId,
+        messageId: String
+    ): Result
+
+    sealed interface Result {
+        data class Success(val position: Int) : Result
+        data class Failure(val cause: StorageFailure) : Result
```
yamilmedina, line 42: Question: Do we need to return a `StorageFailure` here? I thought (maybe wrong) that we only bubble up `CoreFailure` to AR.

yamilmedina (commented):
Looking awesome ! all the way down the rabbit hole with the pager 👏 
Added some small suggestions a question

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq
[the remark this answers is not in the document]
MohamadJaara: we keep the order with date but query by ID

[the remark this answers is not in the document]
alexandreferris: Oh sorry I expressed myself wrong, instead of ordering I meant querying, so if we have like 100 messages and we want message on position 90:

if done with `creation_date >= ...` it returns correctly 90
if done with just the id it doesn't return the correct position (rather less value than the correct one)

also, if we just query by id as we are using `COUNT` it will return just 1 always as we will be doing the query WHERE conversation ID = x and message ID = y, there is no count 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/GetSearchedConversationMessagePositionUseCase.kt
alexandreferris: Changed to a default `CoreFailure` as we are not going to use it in AR

yamilmedina (approved):

MohamadJaara (approved):

PR #2189, opened 2023-11-03
fix: handle federation not enabled [WPB-5237]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
One of client servers returns 
```
ErrorResponse(code=400, message=no federator configured, label=federation-not-enabled, cause=null)
```
probably when user tries to fetch other user data with different domain

### Causes (Optional)
Because fetching users is during slow sync, when it returns error the entire slow sync is aborted and user is stuck on connecting

### Solutions
Handle this error and try to fetch again only those users which have the same domain as self user

### DIFF
--- a/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/user/UserRepositoryTest.kt
+++ b/logic/src/commonTest/kotlin/com/wire/kalium/logic/data/user/UserRepositoryTest.kt
@@ -202,7 +202,7 @@ class UserRepositoryTest {
     @Test
     fun givenAnUserIdListWithDifferentDomain_whenApiReturnsFederationDisabledError_thenShouldTryToFetchOnlyUsersWithSelfDomain() = runTest {
         // given
-        val requestedUserIds = setOf(TestUser.OTHER_USER_ID, TestUser.OTHER_USER_ID_2)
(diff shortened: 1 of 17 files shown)


yamilmedina (approved):

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/user/UserRepositoryTest.kt
```diff
@@ -198,6 +199,22 @@ class UserRepositoryTest {
...
+        val (arrangement, userRepository) = Arrangement()
+            .withGetMultipleUsersApiRequestFederationNotEnabledError()
+            .arrange()
+        // when
+        userRepository.fetchUsersByIds(requestedUserIds).shouldFail()
+        // then
+        verify(arrangement.userDetailsApi)
+            .suspendFunction(arrangement.userDetailsApi::getMultipleUsers)
+            .with(any())
+            .wasInvoked(exactly = twice)
```
vitorhugods, line 215: It would be nice to check that it invoked `getMultipleUsers` with the different inputs, instead of `any()` tiwce.

vitorhugods (approved):

PR #2199, opened 2023-11-07
feat: End call on verification degradation (WPB-4487)

borichellow:
# What's new in this PR?

- updated `EndCallOnConversationChangeUseCase` to end an ongoing call if the conversation verification status (Proteus or MLS) became degraded during the call.
- added `ObserveEndCallDialogUseCase` to observe such cases and be able to inform user by dialog or whatever. 
- updated unit tests

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallOnConversationChangeUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallOnConversationChangeUseCase.kt
@@ -29,7 +29,7 @@ class EndCallOnConversationChangeUseCaseImpl(
     private val callRepository: CallRepository,
     private val conversationRepository: ConversationRepository,
     private val endCallUseCase: EndCallUseCase,
-    private val dialogManager: EndOngoingCallManager
+    private val endCallListener: EndCallResultListener
 ) : EndCallOnConversationChangeUseCase {
     override suspend operator fun invoke() {
         val callsFlow = callRepository.establishedCallsFlow().map { calls ->
@@ -88,7 +88,7 @@ class EndCallOnConversationChangeUseCaseImpl(
             }
             .filter { it.shouldFinishCall() }
             .map {
-                dialogManager.onCallEndedBecauseOfVerificationDegraded(conversationId)
+                endCallListener.onCallEndedBecauseOfVerificationDegraded(conversationId)
                 conversationId
             }
 
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallOnConversationChangeUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallOnConversationChangeUseCase.kt
@@ -1,14 +1,21 @@
 package com.wire.kalium.logic.feature.call.usecase
 
 import com.wire.kalium.logic.data.call.CallRepository
+import com.wire.kalium.logic.data.conversation.Conversation
 import com.wire.kalium.logic.data.conversation.ConversationDetails
 import com.wire.kalium.logic.data.conversation.ConversationRepository
+import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.user.ConnectionState
 import com.wire.kalium.logic.functional.fold
+import com.wire.kalium.logic.functional.onlyRight
 import kotlinx.coroutines.flow.cancellable
-import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.emptyFlow
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.merge
+import kotlinx.coroutines.flow.scan
 
 /**
  * End call when conversation is deleted, user is not a member anymore or user is deleted.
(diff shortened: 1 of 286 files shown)


logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallOnConversationChangeUseCaseTest.kt
```diff
@@ -27,97 +29,133 @@ import io.mockative.mock
...
     @Test
-    fun givenAnEstablishedCall_whenUserIsRemovedFromConversation_thenEndTheCurrentCall() = runTest {
-
-        given(conversationRepository)
-            .suspendFunction(conversationRepository::observeConversationDetailsById)
-            .whenInvokedWith(eq(conversationId))
-            .then {
-                flowOf(Either.Right(groupConversationDetail))
-            }
+    fun givenAnEstablishedCall_whenConversationProteusDegraded_thenEndTheCurrentCall() = runTest {
```
mchenani, line 80: suggestion: I know this pr mostly about Proteus, but since the code covers MLS verification status, I suggest to have a test for MLS degraded end call as well.

mchenani (approved):
🔥

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndCallOnConversationChangeUseCase.kt
```diff
@@ -20,34 +28,76 @@ interface EndCallOnConversationChangeUseCase {
 class EndCallOnConversationChangeUseCaseImpl(
     private val callRepository: CallRepository,
     private val conversationRepository: ConversationRepository,
-    private val endCallUseCase: EndCallUseCase
+    private val endCallUseCase: EndCallUseCase,
+    private val dialogManager: EndOngoingCallManager
```
ohassine, line 32: Naming is wrong here, dialogManager for EndOngoingCallManager

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndOngoingCallManager.kt
```diff
@@ -0,0 +1,41 @@
...
+package com.wire.kalium.logic.feature.call.usecase
+
+import com.wire.kalium.logic.data.id.ConversationId
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+
+interface EndOngoingCallManager {
+    suspend fun observeCallEndedBecauseOfVerificationDegraded(): Flow<ConversationId>
+    suspend fun onCallEndedBecauseOfVerificationDegraded(conversationId: ConversationId)
+}
```
ohassine, line 27: I'd suggest to rename this class as we don't manage anything here we are just observing. How about `EndCallResultListener`, `EndCallResultReceiver` ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveEndCallDialogUseCase.kt
```diff
@@ -0,0 +1,32 @@
...
+interface ObserveEndCallDialogUseCase {
+    suspend operator fun invoke(): Flow<Unit>
+}
+
+internal class ObserveEndCallDialogUseCaseImpl(
+    private val dialogManager: EndOngoingCallManager
+) : ObserveEndCallDialogUseCase {
+    override suspend fun invoke(): Flow<Unit> =
+        dialogManager.observeCallEndedBecauseOfVerificationDegraded().map { Unit }
+}
```
ohassine, line 32: Do we need to create a new class, can't we just use `EndOngoingCallManager` directly in other usecases ?
Also for naming, can we remove the term `dialog` as Kalium is purly logic and there is no UI

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/EndOngoingCallManager.kt
borichellow: yeah, i like listener

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveEndCallDialogUseCase.kt
borichellow: this feature is not well documented yet, maybe later we'll need to add some info about the conversation (name, users that caused verification degrade, etc.)
In that case `EndCallResultListener` is only for listening and observing events, and `ObserveEndCallDialogUseCaseImpl` is for collecting that data and propagating it to the platform side.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveEndCallDialogUseCase.kt
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * The useCase for observing when the ongoing call was ended because of degradation of conversation verification status (Proteus or MLS)
+ */
+interface ObserveEndCallDialogUseCase {
+    /**
+     * @return [Flow] that emits only when the call was ended because of degradation of conversation verification status (Proteus or MLS)
+     */
+    suspend operator fun invoke(): Flow<Unit>
+}
```
ohassine, line 31: The name still confusing as there no dialog  here.
maybe `ObserveEndCallDueToConversationDegradationUseCase` ?

borichellow: sounds much better! Thanks

ohassine (approved):

PR #2280, opened 2023-12-01
feat: Set client mls status based on cerificate (WPB-1004)

borichellow:
# What's new in this PR?

### Issues

Devices MLS certificate status was calculated incorrectly: `Revoced` status was absent. 

### Causes (Optional)

CoreCrypto part wasn't ready for that, but now it is.

### Solutions

Update `CertificateStatusChecker` - add calculating if the certificate was revoked.


cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/IDs.kt
```diff
@@ -78,9 +78,14 @@ data class WireIdentity(
     val handle: String,
     val displayName: String,
     val domain: String,
-    val certificate: String
+    val certificate: String,
+    val status: DeviceStatus
 )
 
+enum class DeviceStatus {
```
mchenani, line 85: suggestion: since this is representing the certificate status, I think we can choose a better name here. wdyt?

mchenani (approved):
Looks good buddy, thanks! Minor naming suggestions.

logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/e2ei/CertificateStatusChecker.kt
```diff
@@ -17,20 +17,21 @@
  */
 package com.wire.kalium.logic.feature.e2ei
 
+import com.wire.kalium.cryptography.CertificateStatus
 import java.util.Date
 
 actual interface CertificateStatusChecker {
-    actual fun status(notAfterTimestamp: Long): CertificateStatus
+    actual fun status(notAfterTimestamp: Long, deviceStatus: CertificateStatus): CertificateStatus
```
ohassine, line 24: same here

ohassine (approved):

PR #2290, opened 2023-12-06
fix: self clients list is not updated when a new client event is received (WPB-5791)

borichellow:
# What's new in this PR?

### Issues

When I'm on a call in Proteus-verified conversation and I add a new client (login on new device). Conversation should degrade the verification status and end a call. 

### Causes (Optional)

a `NewClientEvent` handler doesn't save a new Client into the Client DB table, so the trigger for re-checking the conversation verification is not called -> conversation doesn't degrade -> call is not ended.

### Solutions

Save new client into Client DB table on every `NewClientEvent`.


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
```diff
@@ -253,6 +253,7 @@ class ClientDataSource(
         newClientEvent: Event.User.NewClient
     ): Either<CoreFailure, Unit> = wrapStorageRequest {
         newClientDAO.insertNewClient(clientMapper.toInsertClientParam(selfUserID, newClientEvent))
+        clientDAO.insertClient(clientMapper.toInsertClientParam(selfUserID, newClientEvent))
```
mchenani, line 256: What's the different between these two lines?

mchenani (approved):
Lgtm, and one question!

logic/src/commonTest/kotlin/com/wire/kalium/logic/data/client/ClientRepositoryTest.kt
```diff
@@ -400,7 +400,9 @@ class ClientRepositoryTest {
 
     @Test
     fun whenSavingNewClient_thenNewClientSaved() = runTest {
```
MohamadJaara, line 402:
```suggestion
    fun whenSavingNewClient_thenNewClientStoredInClientsAndNewClientsTables() = runTest {
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
```diff
@@ -253,6 +253,7 @@ class ClientDataSource(
...
     }
```
MohamadJaara, line 257: can we do this on the DB level

insertNewSelfClient {

-- insert into new client table 

-- insert into clients table

}

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
borichellow: different DB-tables where we save clients into. 
`NewClient` table containse only self-user clients that was added and user wasn't informed about it yet. So it's used for informing user about new devices by the dialog (kinda "there is a new device, was it you?"). 

And the `Client` table - is a table for all the clients

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
MohamadJaara: No need for triggers in this case, it will overcomplicate it. 
a simple transaction should do the trick.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
```diff
@@ -252,7 +252,7 @@ class ClientDataSource(
     override suspend fun saveNewClientEvent(
         newClientEvent: Event.User.NewClient
     ): Either<CoreFailure, Unit> = wrapStorageRequest {
-        newClientDAO.insertNewClient(clientMapper.toInsertClientParam(selfUserID, newClientEvent))
```
MohamadJaara, line 255: now that the trigger exists we can remove the NewClientDAO.insertNewClient

MohamadJaara (changes requested):
Great work as always, just need to cover with some sweet tests

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Clients.sq
```diff
@@ -120,3 +120,15 @@ BEGIN
...
+AFTER INSERT ON Client
+WHEN new.user_id = (SELECT SelfUser.id FROM SelfUser LIMIT 1)
+BEGIN
+    INSERT INTO NewClient(id, device_type, registration_date, model)
+    VALUES(new.id, new.device_type, new.registration_date, new.model)
+    ON CONFLICT (id) DO UPDATE SET
+    device_type = coalesce(excluded.device_type, device_type),
+    registration_date = coalesce(excluded.registration_date, registration_date),
+    model = coalesce(excluded.model, model);
+END;
```
MohamadJaara, line 134: I also wonder how this will work when logging in for the first time. 
My guess it will insert all of the user other clients as new and they wil always have the the dialog shown after a new login

MohamadJaara (changes requested):

persistence/src/commonMain/db_user/migrations/68.sqm
```diff
@@ -0,0 +1,11 @@
...
+AFTER INSERT ON Client
+WHEN new.user_id = (SELECT SelfUser.id FROM SelfUser LIMIT 1)
+BEGIN
+    INSERT INTO NewClient(id, device_type, registration_date, model)
+    VALUES(new.id, new.device_type, new.registration_date, new.model)
+    ON CONFLICT (id) DO UPDATE SET
+    device_type = coalesce(excluded.device_type, device_type),
+    registration_date = coalesce(excluded.registration_date, registration_date),
+    model = coalesce(excluded.model, model);
+END;
```
MohamadJaara, line 11: one last thing since this fix should target RC having the migration file at 68 will not work we need to add it after the latest migration in RC

MohamadJaara (changes requested):

MohamadJaara (changes requested):
can you please add one more test 
given new client is added, then the same client is added to the clients table

PR #2310, opened 2023-12-14
fix: Duplicated system message about degraded conversation (WPB-5878)

borichellow:
# What's new in this PR?

### Issues

When some other user adds a new client AND conversation with that user was already "proteus-verified", then 3 system messages are added into that conversation: 1. conversation degraded; 2. conversation verified; 3. conversation degraded.

Screenshot 2023-12-14 at 14 13 18

There should be only 1 message "Conversation degraded", as a new client is not verified.

### Causes (Optional)

`MessageSendFailureHandlerImpl` which is called in case when sending message is not prepared for some client (for that new client in our case), calls also `ConversationRepository.fetchConversation()` which calls `MemberDAO.updateFullMemberList`. This `updateFullMemberList` removes all the members from the conversation and adds only the members that should be there. 

But at the moment when all the members are removed from the conversation DB-trigger runs and it "decides" that the conversation is verified as all the members are verified (no members -> no unverified members -> conversation is verified). 
That's why user sees the message "Conversation is verified". 

Then all the valid members are added into the conversation and the same DB-trigger "decides" that conversation is not verified any more -> adds a new system message.

### Solutions

Update `MemberDAO.updateFullMemberList`:
Instead of removing all the members from the conversation before saving only valid - add all the valid members and after that remove only members that should be removed.

### DIFF
--- a/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/member/MemberDAO.kt
+++ b/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/member/MemberDAO.kt
@@ -176,10 +176,17 @@ internal class MemberDAOImpl internal constructor(
     override suspend fun updateFullMemberList(memberList: List<MemberEntity>, conversationID: QualifiedIDEntity) =
         withContext(coroutineContext) {
             memberQueries.transaction {
-                memberQueries.deleteMembersFromConversation(conversationID)
-                for (member: MemberEntity in memberList) {
+                val membersToDelete = memberQueries.selectAllMembersByConversation(conversationID)
+                    .executeAsList()
+                    .filter { member -> memberList.none { it.user == member.user } }
+
+                memberList.forEach { member ->
                     userQueries.insertOrIgnoreUserId(member.user)
                     memberQueries.insertMember(member.user, conversationID, member.role)
+                }
+
+                membersToDelete.forEach { member ->
+                    memberQueries.deleteMember(conversationID, member.user)
                 }
             }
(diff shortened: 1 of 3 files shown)


persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/member/MemberDAO.kt
```diff
@@ -176,11 +176,18 @@ internal class MemberDAOImpl internal constructor(
...
+                    .filter { member -> memberList.none { it.user == member.user } }
+
+                memberList.forEach { member ->
                     userQueries.insertOrIgnoreUserId(member.user)
                     memberQueries.insertMember(member.user, conversationID, member.role)
                 }
+
+                membersToDelete.forEach { member ->
+                    memberQueries.deleteMember(conversationID, member.user)
+                }
```
ohassine, line 190: Can we cover it with test ?

mchenani (approved):
LGTM, please consider @ohassine comment and cover it with tests. Thanks

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/member/MemberDAO.kt
borichellow: your message helped me to understand that my implementation doesn't update existed users (if updating is needed). 
So i've added a new DB-query and update unit-test to cover that case too. 
Thanks :)

ohassine (approved):

PR #2313, opened 2023-12-15
fix: E2EI required: grace period in seconds (WPB-5771)

borichellow:
# What's new in this PR?

### Issues

in `E2EIConfig` API response we were expecting that field `verificationExpiration` is timestamp of the end of the grace period in nano-seconds. In fact it's amount of seconds that left till the end of grace period (stating from the moment when we receive that info). So needed to update `E2EIConfigHandler`

### Causes (Optional)

It wasn't documented well.

### Solutions

Update `E2EIConfigHandler` and tests.

### DIFF
--- a/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/config/UserConfigStorage.kt
+++ b/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/config/UserConfigStorage.kt
@@ -458,7 +458,7 @@ class UserConfigStorageImpl(
             E2EI_NOTIFICATION_TIME,
             timeStamp
         ).also {
-            e2EIFlow.tryEmit(Unit)
+            e2EINotificationFlow.tryEmit(Unit)
         }
     }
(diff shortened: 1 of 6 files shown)


persistence/src/commonMain/kotlin/com/wire/kalium/persistence/config/UserConfigStorage.kt
```diff
@@ -458,7 +458,7 @@ class UserConfigStorageImpl(
             E2EI_NOTIFICATION_TIME,
             timeStamp
         ).also {
-            e2EIFlow.tryEmit(Unit)
+            e2EINotificationFlow.tryEmit(Unit)
```
borichellow, line 461: this doesn't relate to the original task, i just found some copy-past mistake (which causes unknown yet issue) and fixed it :)

mchenani (approved):

ohassine (approved):

PR #2330, opened 2023-12-22
fix: infinite loop on message send failure [WPB-5789]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- added retryCount to avoid infinite loop on message sending failures

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
@@ -246,7 +246,8 @@ internal class MessageSenderImpl internal constructor(
 
     private suspend fun attemptToSendWithProteus(
         message: Message.Sendable,
-        messageTarget: MessageTarget
+        messageTarget: MessageTarget,
+        retryCount: Int = 0
     ): Either<CoreFailure, String> {
         val conversationId = message.conversationId
         val target = when (messageTarget) {
@@ -272,7 +273,7 @@ internal class MessageSenderImpl internal constructor(
                             is MessageTarget.Conversation ->
                                 MessageTarget.Conversation((messageTarget.usersToIgnore + usersWithoutSessions.users).toSet())
                         }
-                        trySendingProteusEnvelope(envelope, message, updatedMessageTarget)
+                        trySendingProteusEnvelope(envelope, message, updatedMessageTarget, retryCount = retryCount + 1)
                     }
             }
     }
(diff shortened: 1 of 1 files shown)


saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -402,7 +411,11 @@ internal class MessageSenderImpl internal constructor(
                     .handleClientsHaveChangedFailure(failure, conversationId)
                     .flatMap {
                         logger.w("Retrying After Proteus $action Failure: { \"message\" : \"${messageLogString}\"}")
-                        retry()
+                        if (retryCount <= MAX_RETRY_COUNT) {
+                            retry()
+                        } else {
+                            Either.Left(failure)
```
saleniuk, line 417: suggestion: maybe add a log describing why it didn't retry to avoid any confusion when analysing logs in the future

MohamadJaara: Question: does this add the sliding indicator on the side?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -391,6 +399,7 @@ internal class MessageSenderImpl internal constructor(
         action: String, // Send or Broadcast
         messageLogString: String,
         conversationId: ConversationId?,
+        retryCount: Int = 0,
```
yamilmedina, line 402: Suggestion: Just a matter of code uniformity for the same concept, what about using the inverse logic like [here called remainingAttempts](https://github.com/wireapp/kalium/blob/d911835d838056c4de0cbbfa99cdda4b6b93527c/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt#L575-L576)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -479,4 +492,8 @@ internal class MessageSenderImpl internal constructor(
             messageRepository.persistRecipientsDeliveryFailure(message.conversationId, message.id, messageSent.failedToConfirmClients)
         }
 
+    companion object {
+        const val MAX_RETRY_COUNT: Int = 1
```
yamilmedina, line 496: Question: The same as prev. comment about uniformity for same concept, why 1 and not 2 ? I believe there is not right or wrong, since they are just arbitrary constants, but leaving here the question anyway :D

yamilmedina (commented):
Awesome work as always, maestro. 🎸

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
```diff
@@ -246,7 +246,8 @@ internal class MessageSenderImpl internal constructor(
 
     private suspend fun attemptToSendWithProteus(
         message: Message.Sendable,
-        messageTarget: MessageTarget
+        messageTarget: MessageTarget,
+        retryCount: Int = 0
```
MohamadJaara, line 250: can you please add a comment about the retry and how it works

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
MohamadJaara: 1 retry should be enough because we send the message and handle all the errors and retry, if that failed then any other retry will also fail

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/MessageSender.kt
Garzas: I think retry attempt is self explanatory

saleniuk (approved):

borichellow (approved):

PR #2338, opened 2024-01-02
fix: Doubled NewConversation System message on accepting connection (WPB-5374)

borichellow:
# What's new in this PR?

### Issues

The unverified user warning shows up twice when the user accepts a connection request.

### Causes (Optional)

`ConversationStartedUnverifiedWarning` system message is added to the conversation in 2 cases: 
1. in `UserEventReceiver` when new connection comes 
2. in `AcceptConnectionRequestUseCase`

### Solutions

remove adding a new message in one of these cases
I picked `UserEventReceiver.onEvent` as there is no conversation for user yet, and maybe he'll ignore it, so no need to add a message that will never be seen.


yamilmedina (approved):
lgtm 🚀

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/UserEventReceiver.kt
```diff
@@ -117,10 +113,6 @@ internal class UserEventReceiverImpl internal constructor(
                             delay = if (event.live) 3.seconds else ZERO
                         )
                         Either.Right(Unit)
-                    }.flatMap {
```
saleniuk, line 120: issue: `AcceptConnectionRequestUseCase` is executed only when user accepts manually the connection request on current device, so if the "connection accepted" event comes from another device then this warning message won't be added to the conversation

saleniuk (approved):

PR #2352, opened 2024-01-08
fix: missing ServerConfig crashes after session expired / logout [WPB-5960]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Top crashes types currently are the ones related to missing `ServerConfig` when making some actions.

### Causes (Optional)

When session is expired (the app receives 403 when trying to refresh a token) there is an infinite loop, because the app tries to logout - deregister token which results in 401 and it triggers the auth token refresh again resulting in 403 and trying to logout again. After each iteration, `CurrentSessionFlowUseCase` emits the same item again, which triggers multiple actions in the whole app and creates a race condition when logging out (session should be invalid but with all the loop iterations in a fraction of a second the app can still get the previous value).
Some logout actions in Android project are not executed when account is logged out from kalium (when session expires, device is removed from another place or account deleted).

### Solutions

- execute `deregister token` request the same way as `logout` request - only if the logout is not because of the session expiration because otherwise it won't be successful anyway since the app doesn't have a valid token anymore - it prevents from falling into an endless 403-401 loop
- provide a `LogoutCallbackManager` which allows the app to register for logout callbacks and triggered when the logout happens in kalium to execute all other required logout-related actions in Android
- make `GlobalKaliumScope` to not create a new instance every time
- make `CurrentSessionFlow` `distinctUntilChanged` so that it doesn’t re-emit the same items
- make `SyncCriteriaResolution.MissingRequirements` a data class to get the `reason` String in logs as right now it’s not visible and we don’t know what is the reason of failed sync
- provide logs for `AccountSwitchUseCase` ✅

#### Test Coverage (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutCallbackManager.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutCallbackManager.kt
@@ -34,4 +34,4 @@ class LogoutCallbackManagerImpl : LogoutCallbackManager {
     override suspend fun logout(userId: UserId, reason: LogoutReason) { callbacks.forEach { it(userId, reason) } }
 }
 
-typealias LogoutCallback = (userId: UserId, reason: LogoutReason) -> Unit
+typealias LogoutCallback = suspend (userId: UserId, reason: LogoutReason) -> Unit
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutCallbackManager.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutCallbackManager.kt
@@ -34,6 +34,4 @@ class LogoutCallbackManagerImpl : LogoutCallbackManager {
     override suspend fun logout(userId: UserId, reason: LogoutReason) { callbacks.forEach { it(userId, reason) } }
 }
 
-interface LogoutCallback {
-    suspend operator fun invoke(userId: UserId, reason: LogoutReason)
-}
+typealias LogoutCallback = (userId: UserId, reason: LogoutReason) -> Unit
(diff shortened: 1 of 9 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutCallbackManager.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.wire.kalium.logic.data.logout.LogoutReason
+import com.wire.kalium.logic.data.user.UserId
+
+interface LogoutCallbackManager {
+    fun register(callback: LogoutCallback)
+    fun unregister(callback: LogoutCallback)
+    suspend fun logout(userId: UserId, reason: LogoutReason)
+}
+
+class LogoutCallbackManagerImpl : LogoutCallbackManager {
```
yamilmedina, line 30:
```suggestion
internal class LogoutCallbackManagerImpl : LogoutCallbackManager {
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/LogoutCallbackManager.kt
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.kalium.logic.feature.auth
+
+import co.touchlab.stately.collections.ConcurrentMutableList
+import com.wire.kalium.logic.data.logout.LogoutReason
+import com.wire.kalium.logic.data.user.UserId
+
+interface LogoutCallbackManager {
```
yamilmedina, line 24: Suggestion: We can add docs here to know the purpose and usages of it since is public API :)

yamilmedina (approved):
Awesome 🙌  Just added some minor comments

MohamadJaara (approved):

PR #2355, opened 2024-01-10
feat: Fetch Federated certificated worker (WPB-5392)

borichellow:
# What's new in this PR?

With Federation of backends, clients need to trust certificates from another domain. There needs to be an extension of the trust from one backend to another by having the certificate authority of client A (on backend A) sign the certificate for client B (on backend B ) as client A can only trust backend A, not backend B.

### Solutions

Added `ACMECertificatesSyncWorker` that will fetch such certificates and register it via CoreCrypto at least once per 24 hours. And run it in `UserSessionScope`

P.S. CoreCrypto is not ready yet, so kalium part just do nothing on registering.


cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/MLSClient.kt
```diff
@@ -371,4 +371,9 @@ interface MLSClient {
      * @return the exist identities for requested clients
      */
     suspend fun getUserIdentities(groupId: MLSGroupId, users: List<CryptoQualifiedID>): Map<String, List<WireIdentity>>
+
+    /**
+     * Register ACME certificates
```
mchenani, line 376:
```suggestion
     * Register CA certificates for E2EI
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/e2ei/E2EIRepository.kt
```diff
@@ -225,6 +227,18 @@ class E2EIRepositoryImpl(
         Either.Right(e2EIClient.getOAuthRefreshToken())
     }
 
+    override suspend fun fetchACMECertificates(): Either<CoreFailure, Unit> = userConfigRepository.getE2EISettings().flatMap {
```
mchenani, line 230: suggestion: fetchFederationCertificates

mchenani (approved):

ohassine (approved):

PR #2359, opened 2024-01-11
test: add support for test video capture and rendering to Kalium CLI (WBP-2101)

c-g-owen:
* Adds support for video capture and rendering with real capturer-renderers (OSX) and fake capturer-renderers (OSX & Linux) in Kalium-cli.
* Updates AVS dependency to 9.7.1 to make use of new kcall API providing above.

#### How to Test

Run Kalium CLI with "console" option, make call, use V command to toggle outgoing video. Incoming video is shown for real renderers automatically.
Add "-T" option to enable testing mode (auto-generated video capture and dummy renderers).


c-g-owen:
Please note that the update to AVS 9.7.1 is necessary for this PR, as there are new AVS functions used in this code (namely the kcall_ functions) that are only present in AVS 9.7.1 and above

cli/src/jvmMain/kotlin/com/wire/kalium/cli/commands/ConsoleCommand.kt
```diff
@@ -62,9 +66,24 @@ class ConsoleCommand : CliktCommand(name = "console") {
         KeyStroke('e', ::endCallHandler),
         KeyStroke('m', ::muteCallHandler),
         KeyStroke('s', ::selectConversationHandler),
+        KeyStroke('V', ::toggleVideoHandler),
```
typfel, line 69: @c-g-owen should it be capital `V`?

alexandreferris (commented):
Great additions!

I just have some nitpick comments.

Also it is missing iOS implementation on `CallManagerImpl.kt`

Another thing, if we can add tests for `SetTestPreviewActiveUseCase`/`SetTestRemoteVideoStatusUseCase`/`SetTestVideoTypeUseCase` even though it is for tests, it would be great to keep the tests tested 😂

ohassine (approved):

typfel:
> Great additions!
> 
> I just have some nitpick comments.
> 
> Also it is missing iOS implementation on `CallManagerImpl.kt`
> 
> Another thing, if we can add tests for `SetTestPreviewActiveUseCase`/`SetTestRemoteVideoStatusUseCase`/`SetTestVideoTypeUseCase` even though it is for tests, it would be great to keep the tests tested 😂

@alexandreferris added the missing tests

alexandreferris (approved):

typfel:
@alexandreferris any idea why the instrumented tests started to fail? Are they broken on develop?

PR #2365, opened 2024-01-12
feat: UseCase for getting all users E2EI certificates

borichellow:
# What's new in this PR?

### Issues

No any UseCase for getting all the E2EI certs for all users client in one "request" (only one-by-one by ClientId, which is not cool).
Would be useful for UserProfile screen, where we display the list of devices and need to show E2EI statuses for each device.

### Causes (Optional)

Wasn't implemented

### Solutions

Added `GetUserE2eiCertificatesUseCase` for that, where we can get them by `UserId` and tests for that.


mchenani (approved):
🔥

alexandreferris (approved):

yamilmedina (approved):

PR #2377, opened 2024-01-16
feat: testservice can send poll message buttons (WPB-2633)

mythsunwind:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR implements poll messages with buttons for the testservice as described in WPB-2633

----

----

----
##### References


mchenani (approved):

MohamadJaara (changes requested):
please add test to this change 
and i wonder if it is worth it to have a separated use case for this change 

the only case i think this will fail is when the conversation have a timer this and we try to send a composet message it will not work since in proto Ephemeral messages can only be 
```
        Text text = 2;
        ImageAsset image = 3; // deprecated in favour of Asset
        Knock knock = 4;
        Asset asset = 5;
        Location location = 6;
        ```

**Review by tmspzz — DISMISSED** [review]
Please make it so that the titles answers the question: If merged, this PR will: ... ³

**Comment by mythsunwind** [comment]
> Please make it so that the titles answers the question: If merged, this PR will: ... ³

"This PR implements poll messages with buttons for the testservice as described in WPB-2633" is not enough?

**Comment by mythsunwind** [comment]
> please add test to this change and i wonder if it is worth it to have a separated use case for this change

I added a test now: https://github.com/wireapp/kalium/pull/2377/commits/6899dd775eccb6d723be45eda170337a88775baa
 
> the only case i think this will fail is when the conversation have a timer this and we try to send a composet message it will not work since in proto Ephemeral messages can only be

Oh, i was not aware that Composite was reflected in the proto itself. I thought it is just a wrapper in kalium. Still not sure what to do with this knowledge. Should we check if the message is not a timed message if buttons are included?

**Comment by mythsunwind** [comment]
After consulting @MohamadJaara we decided on implementing a separate use case for sending a text message with buttons to deal with the Ephemeral specs in the proto.

**Review by MohamadJaara — APPROVED** [review]

File: logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/composite/SendButtonMessageUseCase.kt [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+    private val scope: CoroutineScope
+) {
+
+    suspend operator fun invoke(
+        conversationId: ConversationId,
+        text: String,
+        mentions: List<MessageMention> = emptyList(),
+        quotedMessageId: String? = null,
+        buttons: List<String> = listOf()
+    ): Either<CoreFailure, Unit> = scope.async(dispatchers.io) {
```
yamilmedina, line 67:
```suggestion
    ): Either<CoreFailure, Unit> = withContext(dispatchers.io) {
```

Question: is there any special need, why we need to pass the `coroutineScope` to the UseCase ? Usually we are just doing as suggested ☝️  , just because I guess Use cases are exposed to clients, and I don't know how that will look in other platforms (food for thought) not blocker, just about uniformity

yamilmedina (commented):
Looking pretty good, added a question about it, not blocker.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/composite/SendButtonMessageUseCase.kt
MohamadJaara: this mean the message will be sent async like we do in text message 
1. send a message 
2. close navigate back before it is done sending 
3. message will be sent anyway

PR #2380, opened 2024-01-17
fix: GetUserE2EICertificate in IO dispatcher

borichellow:
# What's new in this PR?

Moved to execute of `GetUserE2eiCertificateStatusUseCase` into IO dispatcher (it should be done from the beginning!)


MohamadJaara (changes requested):
I appreciate your effort in this update. However, I'm not entirely convinced of the necessity for this change, particularly since the functions getUserIdentity relies upon are already main-thread safe. Could you please clarify if there is a specific requirement driving this modification? Additionally, if ensuring main-thread safety is the goal, it might be more effective to verify that each individual function is main-thread safe, rather than broadly applying withContext(dispatchers.io) to every use case. This approach could potentially enhance both the efficiency and clarity of our codebase.

borichellow:
> since the functions getUserIdentity relies upon are already main-thread safe

Ou, i didn't notice that! Thanks! Then this PR could be closed ;P

PR #2386, opened 2024-01-22
feat: e2ei: respect e2ei creating mls client fixed certificate status

borichellow:
# What's new in this PR?

### Issues

Even when the device (client) has E2EI certificates it's displayed as not valid

### Causes (Optional)

`WireIdentity.clientId` that CoreCrypto returns for each Client is not a usual `ClientId` but combination of `ClientId` and `UserId`, so when app was checking if client has a certificate -> there is no certificate for such clientId.

### Solutions

Fix the `GetUserE2eiCertificatesUseCase` to return not just a Map, but a result that allow to get if client has a certificate by our regular `ClientId`.


mchenani (approved):

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/e2ei/GetUserE2eiAllCertificateStatusesUseCaseTest.kt
```diff
@@ -69,10 +70,10 @@ class GetUserE2eiAllCertificateStatusesUseCaseTest {
 
         val result = getUserE2eiAllCertificateStatuses(userId)
 
-        assertEquals(3, result.size)
-        assertEquals(CertificateStatus.VALID, result[identity1.clientId]?.status)
-        assertEquals(CertificateStatus.EXPIRED, result[identity2.clientId]?.status)
-        assertEquals(CertificateStatus.REVOKED, result[identity3.clientId]?.status)
+//         assertEquals(3, result.size)
```
yamilmedina, line 73:
```suggestion

```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/GetUserE2EIAllCertificatesUseCase.kt
```diff
@@ -27,23 +27,28 @@ import com.wire.kalium.logic.data.conversation.ClientId
...
             }
-            result
-        }.getOrElse(mapOf())
+            UsersE2eiCertificates(userId, result)
+        }.getOrElse(UsersE2eiCertificates(userId, mapOf()))
 
 }
+
+data class UsersE2eiCertificates(private val userId: UserId, private val map: Map<String, E2eiCertificate>) {
+    operator fun get(clientId: ClientId): E2eiCertificate? = map["${userId.value}:${clientId.value}@${userId.domain}"]
```
yamilmedina, line 52: question: is this `"${userId.value}:${clientId.value}@${userId.domain}"` the same as `QualifiedClientID`. If that's the case makes sense to return it here?

yamilmedina (commented):
Looking good 🍾 , just left a question and a code commented line removal.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/GetUserE2EIAllCertificatesUseCase.kt
borichellow: well, it's not the same, this ClientId is just a `String` which is kinda combination of `UserId` and `ClientId` values. 
So I could pars it into `QualifiedClientID` but i don't think it makes much sense.

yamilmedina (approved):

PR #2389, opened 2024-01-22
fix: MLS log spam with NPE [WPB-6222]

borichellow:
# What's new in this PR?

### Issues

`NullPointerException ` log is spammed when doing user search even on non MLS enabled enviroments

### Causes (Optional)

redundant `!!` in storage request for getting MLSGroupId. It's not needed there as `wrapStorageRequest` already handles the case when the result is `null` and returns `Either.Left(StorageFailure.DataNotFound)`.

### Solutions

remove `!!`

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
@@ -575,7 +575,7 @@ internal class MLSConversationDataSource(
         userIds: List<UserId>
     ): Either<CoreFailure, Map<UserId, List<WireIdentity>>> =
         wrapStorageRequest {
-            conversationDAO.getMLSGroupIdByConversationId(conversationId.toDao())!!
+            conversationDAO.getMLSGroupIdByConversationId(conversationId.toDao())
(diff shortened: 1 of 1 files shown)


alexandreferris (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
```diff
@@ -575,7 +575,7 @@ internal class MLSConversationDataSource(
         userIds: List<UserId>
     ): Either<CoreFailure, Map<UserId, List<WireIdentity>>> =
         wrapStorageRequest {
-            conversationDAO.getMLSGroupIdByConversationId(conversationId.toDao())!!
+            conversationDAO.getMLSGroupIdByConversationId(conversationId.toDao())
```
MohamadJaara, line 578: suggestion: if nullability is expected then you can use `wrapNullablewStorageRequest`

MohamadJaara: looking at the code i think nullability can be ignored here  LGTM

MohamadJaara (approved):

PR #2445, opened 2024-02-02
feat(monkeys): Add support for 2FA during login [WPB-6289]

augustocdias:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

This adds support to 2FA in the login of the monkeys. A new flag was added in the backend configuration to indicate the need for it during authentication.

----
##### References

### DIFF
--- a/monkeys/schema.json
+++ b/monkeys/schema.json
@@ -18,7 +18,7 @@
             ],
             "properties": {
                 "startCommand": {
-                    "description": "The command to start the process. For example kubectl to start a pod or docker to start a container. It can be templated with the following fields: teamName, email, userId, teamId, monkeyIndex, monkeyClientId, internalUser, internalPassword",
+                    "description": "The command to start the process. For example kubectl to start a pod or docker to start a container. It can be templated with the following fields: teamName, email, userId, teamId, monkeyIndex, monkeyClientId",
                     "type": "string",
                     "examples": [
                         "/opt/app/monkeys/bin/monkey-server -p 80{{monkeyIndex}}",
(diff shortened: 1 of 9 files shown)


monkeys/schema.json
```diff
@@ -198,6 +198,11 @@
                         "description": "How many users should be created in this team",
                         "type": "integer"
                     },
+                    "2FAEnabled": {
+                        "description": "Does this server require 2FA authentication?",
+                        "type": "boolean",
+                        "default": false
+                    },
```
vitorhugods, line 205: Not sure if relevant for Monkeys, but it may also depend on each team. Some teams may have it, some may not, in the same server.

Why store this, if this can be figured out during runtime, based on the login response?

augustocdias: We need it to create the users as well. When we're setting the user handle we need to login. And since I use the backdoor to get the code I don't actually need to make a failed attempt to login to get the code.

augustocdias: And the config is per team. So we can have multiple teams on the same server with different configs.

vitorhugods: Ahh cool.
Also, just to correct myself. It seems it's not per team, but indeed _per backend_  for some technical/business reason?
In any case, it makes sense for Monkeys!

vitorhugods (approved):

monkeys/schema.json
augustocdias: This is inheritance from the first iteration of the monkeys 😅

I should have renamed it. It is used to represent the team and its backend configs.

MohamadJaara (approved):

PR #2457, opened 2024-02-06
feat: e2ei: Respect E2EI creating mls client fixed certificate status

borichellow:
Just a cherry-pick of [closed PR ](https://github.com/wireapp/kalium/pull/2386) . 
Some problems occurred there as original target branch was merged into RC and that one wasn't updated there are a lot of changes and conflics, so it's easier to remove that branch and cherry-pick a new one with only changes that are actually related to the commit message.

# What's new in this PR?

### Issues

Even when the device (client) has E2EI certificates it's displayed as not valid

### Causes (Optional)

`WireIdentity.clientId` that CoreCrypto returns for each Client is not a usual `ClientId` but combination of `ClientId` and `UserId`, so when app was checking if client has a certificate -> there is no certificate for such clientId.

### Solutions

Fix the `GetUserE2eiCertificatesUseCase` to return not just a Map, but a result that allow to get if client has a certificate by our regular `ClientId`.


borichellow:
The issues was fixed in other place by other solution, so no needed anymore :)

PR #2459, opened 2024-02-06
refactor: decouple event content and origin (WPB-6435)

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

1. Events are being processed twice when they are received through the PendingEvents flow and also in the LiveEvents flow
2. `live` and `transient` flags are being passed all around the code

### Causes

The `Event` class has `live` and `transient` flags mixed with the content of the event.
In `PendingEventsBuffer`, when calling `contains`, the event was not being found, because the flag `live` was different, so the `equals` result was "wrong".

### Solutions

Separate the content of the event and the information about the event's origin.

This is two-fold:
1. Cleans up a bit of code, as live/transient were being passed everywhere.
2. Makes sure that when comparing events, their origin won't matter, as we can easily call `equals` between two Events without having `transient` and `live` in the middle

#### Test Coverage

The compiler does the job in this case. There's no difference between a Pending and a Live event, the content is exactly the same.

I'll try to add a couple of integration tests for this later, as I also plan to change `PendingEventsBuffer` to increase its performance.

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/CommitBundleEventReceiver.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/CommitBundleEventReceiver.kt
@@ -21,6 +21,7 @@ package com.wire.kalium.logic.data.conversation
 import com.wire.kalium.logic.CoreFailure
 import com.wire.kalium.logic.MLSFailure
 import com.wire.kalium.logic.data.event.Event
+import com.wire.kalium.logic.data.event.EventDeliveryInfo
 import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.logic.kaliumLogger
 import com.wire.kalium.logic.sync.receiver.EventReceiver
@@ -33,7 +34,7 @@ internal class CommitBundleEventReceiverImpl(
     private val memberJoinEventHandler: MemberJoinEventHandler,
     private val memberLeaveEventHandler: MemberLeaveEventHandler
 ) : CommitBundleEventReceiver {
-    override suspend fun onEvent(event: Event.Conversation): Either<CoreFailure, Unit> {
+    override suspend fun onEvent(event: Event.Conversation, deliveryInfo: EventDeliveryInfo): Either<CoreFailure, Unit> {
         return when (event) {
             is Event.Conversation.MemberJoin -> memberJoinEventHandler.handle(event)
             is Event.Conversation.MemberLeave -> memberLeaveEventHandler.handle(event)
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
@@ -45,12 +45,42 @@ import com.wire.kalium.logic.data.user.Connection
 import com.wire.kalium.logic.data.user.SupportedProtocol
 import com.wire.kalium.logic.data.user.UserId
 import com.wire.kalium.logic.logStructuredJson
+import com.wire.kalium.logic.sync.incremental.EventSource
 import com.wire.kalium.network.api.base.authenticated.conversation.ConversationResponse
 import com.wire.kalium.util.DateTimeUtil
 import com.wire.kalium.util.serialization.toJsonElement
 import kotlinx.serialization.json.JsonNull
 
-sealed class Event(open val id: String, open val transient: Boolean, open val live: Boolean) {
+/**
+ * A wrapper that joins [Event] with its [EventDeliveryInfo].
+ */
+data class EventEnvelope(
+    val event: Event,
+    val deliveryInfo: EventDeliveryInfo,
+)
+
+/**
+ * Data class representing information about the delivery of an event.
+ *
+ * @property isTransient Specifies whether the event is transient.
+ * Transient events are events that only matter if the user is online/active. For example "user is typing",
+ * and call signaling (mute/unmute), which are irrelevant after a few minutes. These are likely to not even
+ * be stored in the backend.
+ * @property source The source of the event.
+ * @see EventSource
+ */
+data class EventDeliveryInfo(
+    val isTransient: Boolean,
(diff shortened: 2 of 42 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
```diff
@@ -45,12 +45,42 @@ import com.wire.kalium.logic.data.user.Connection
...
 import com.wire.kalium.network.api.base.authenticated.conversation.ConversationResponse
 import com.wire.kalium.util.DateTimeUtil
 import com.wire.kalium.util.serialization.toJsonElement
 import kotlinx.serialization.json.JsonNull
 
-sealed class Event(open val id: String, open val transient: Boolean, open val live: Boolean) {
+/**
+ * A wrapper that joins [Event] with its [EventDeliveryInfo].
+ */
+data class EventEnvelope(
```
tmspzz, line 57: does this need to have a `toLogMap()` to properly show up in the logs?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
```diff
@@ -45,12 +45,42 @@ import com.wire.kalium.logic.data.user.Connection
...
+ * Data class representing information about the delivery of an event.
+ *
+ * @property isTransient Specifies whether the event is transient.
+ * Transient events are events that only matter if the user is online/active. For example "user is typing",
+ * and call signaling (mute/unmute), which are irrelevant after a few minutes. These are likely to not even
+ * be stored in the backend.
+ * @property source The source of the event.
+ * @see EventSource
+ */
+data class EventDeliveryInfo(
```
tmspzz, line 72: same here

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
vitorhugods: Yup! Nice catch.
When I refactored I ended up logging only the `event`, so it would keep working. But by doing it, `isTransient` and `isLive` were no longer being logged. Added `toLogMap()` and the like to both.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/CommitBundleEventReceiver.kt
```diff
@@ -33,7 +34,7 @@ internal class CommitBundleEventReceiverImpl(
     private val memberJoinEventHandler: MemberJoinEventHandler,
     private val memberLeaveEventHandler: MemberLeaveEventHandler
 ) : CommitBundleEventReceiver {
-    override suspend fun onEvent(event: Event.Conversation): Either<CoreFailure, Unit> {
+    override suspend fun onEvent(event: Event.Conversation, deliveryInfo: EventDeliveryInfo): Either<CoreFailure, Unit> {
```
yamilmedina, line 37: Question: Where are we using `deliveryInfo` here ? I can not see its usage.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/EventReceiver.kt
```diff
@@ -20,8 +20,9 @@ package com.wire.kalium.logic.sync.receiver
 
 import com.wire.kalium.logic.CoreFailure
 import com.wire.kalium.logic.data.event.Event
+import com.wire.kalium.logic.data.event.EventDeliveryInfo
 import com.wire.kalium.logic.functional.Either
 
 internal fun interface EventReceiver<T : Event> {
-    suspend fun onEvent(event: T): Either<CoreFailure, Unit>
+    suspend fun onEvent(event: T, deliveryInfo: EventDeliveryInfo): Either<CoreFailure, Unit>
```
yamilmedina, line 27: Question: I commented before I came to this eventually. It's not clear to me why every `EventReceiver` implementation needs this `deliveryInfo` since it's not used 🤔  Maybe I'm overseeing something.

yamilmedina (commented):
Titanic job 🥇  ! I raised some questions, all related to the non-usage of the `EventDeliveryInfo` in some `EventReceiver` implementations.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/CommitBundleEventReceiver.kt
vitorhugods: Most event receivers never cared if the event was `live` or `transient`. They were fields inside the `Event`. _Now_ they're explicitly passed to each `EventReceiver`, which may choose to use it or not.

Maybe `EventReceiver` shouldn't even be an `interface` 🤔, this would allow us to have some receiver that need it, and some that don't. But this is a change I wouldn't like to do in this PR.

vitorhugods: `DeliveryInfo` is currently used by:

- [`UserEventReceiver`](https://github.com/wireapp/kalium/pull/2459/files#diff-56f41dcf1357f72c117502a96a9fd685cc0023718f231401c329580688534ffc),
- [`NewMessageEventHandler`](https://github.com/wireapp/kalium/pull/2459/files#diff-d3ebd9cbde3b576faa8e1ecc9b39ac9261383c9dc13f61057ae91ec7719802bd)

yamilmedina: Yep, this makes total sense :D  
> Maybe EventReceiver shouldn't even be an interface

Or another solution playing with inheritance, abstract classes, default implementations, sounds a fun challenge.

Definitely out of scope for this PR.

yamilmedina (approved):

MohamadJaara (approved):

PR #2478, opened 2024-02-08
fix: release: Do not show EnrollE2EI dialog after login (WPB-6364)

borichellow:
# What's new in this PR?

### Issues

The are 2 issues: 

1. After login and enrolling the E2EI certificate user still see the "Enroll E2EI" dialog 
2. Grace period for enrolling E2EI certificate resets on every app opening. 

### Causes (Optional)

1. When user logs in, CC returns an error on getMLSClient, app interprets it as "user doesn't have certificate" and asks user to get it.
2. `E2EIConfigHandler` is called in Syncing and on every sync grace period (which app gets as a period, not exact date) resets

### Solutions

1. 
- update `ObserveE2EIRequiredUseCase` so the error while getting the current client certificate means "Something went wrong, don't show the dialog yet"
- update `MLSConversationRepository.getClientIdentity` to return `null` if client has no certificate (not an error).

2. 
Update `UserConfigStorage.setE2EINotificationTime` -> `UserConfigStorage.setIfAbsentE2EINotificationTime` so the end of grace period is updated only when there is no saved date yet 
AND add `UserConfigStorage.updateE2EINotificationTime` for updating that date when user snoozes the E2EI dialog.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/ObserveE2EIRequiredUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/ObserveE2EIRequiredUseCase.kt
@@ -25,7 +25,6 @@ import com.wire.kalium.logic.feature.e2ei.usecase.GetE2EICertificateUseCaseResul
 import com.wire.kalium.logic.feature.e2ei.usecase.GetE2eiCertificateUseCase
 import com.wire.kalium.logic.featureFlags.FeatureSupport
 import com.wire.kalium.logic.functional.getOrElse
-import com.wire.kalium.logic.functional.getOrNull
 import com.wire.kalium.logic.functional.map
 import com.wire.kalium.logic.functional.onlyRight
 import com.wire.kalium.util.DateTimeUtil
@@ -34,7 +33,6 @@ import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.flowOn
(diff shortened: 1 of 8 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/user/ObserveE2EIRequiredUseCase.kt
```diff
@@ -67,20 +65,26 @@ internal class ObserveE2EIRequiredUseCaseImpl(
 
         return userConfigRepository
             .observeE2EINotificationTime()
-            .map { it.getOrNull() }
-            .filterNotNull()
+            .onlyRight()
```
MohamadJaara, line 68: suggestion: small changes to make this function more readable

```kotlin


    @OptIn(ExperimentalCoroutinesApi::class)
    override fun invoke(): Flow<E2EIRequiredResult> {
        if (!featureSupport.isMLSSupported) return flowOf(E2EIRequiredResult.NotRequired)

        return userConfigRepository
            .observeE2EINotificationTime()
            .onlyRight()
            .delayUntilNotifyTime()
            .flatMapLatest {
                observeE2EISettings().map { setting ->
                    if (!setting.isRequired) return@map E2EIRequiredResult.NotRequired

                    return@map currentClientIdProvider().map { clientId ->
                            val certificateResult = e2eiCertificate(clientId)
                            when {
                                certificateResult.isValid() -> E2EIRequiredResult.NotRequired

                                // TODO check if it was fixed in CC
                                // When user just logged in and app requests the certificate
                                // than MLSClient throws ConversationNotFound exception.
                                // So we do not ask user to create a certificate in that case
                                // as the certificate might be already generated
                                (certificateResult is GetE2EICertificateUseCaseResult.Failure) -> E2EIRequiredResult.NotRequired

                                setting.isGracePeriodLeft() -> {
                                    val timeLeft = setting.gracePeriodEnd!!.minus(DateTimeUtil.currentInstant())
                                    if (certificateResult !is GetE2EICertificateUseCaseResult.Failure)
                                        E2EIRequiredResult.WithGracePeriod.Renew(timeLeft)
                                    else E2EIRequiredResult.WithGracePeriod.Create(timeLeft)
                                }

                                else -> {
                                    if (certificateResult !is GetE2EICertificateUseCaseResult.Failure)
                                        E2EIRequiredResult.NoGracePeriod.Renew
                                    else E2EIRequiredResult.NoGracePeriod.Create
                                }
                            }
                        }.getOrElse { E2EIRequiredResult.NotRequired }

                }
            }
            .flowOn(dispatcher)
    }
```

mchenani (approved):

MohamadJaara (approved):

PR #2497, opened 2024-02-14
fix: release: Remove e2ei settings data on login (WPB-6495)

borichellow:
# What's new in this PR?

### Issues

E2EI settings data is not cleared on user logout

### Causes (Optional)

Wasn't implemented

### Solutions

Implement in

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
@@ -238,7 +238,7 @@ internal class UserConfigDataSource internal constructor(
     override suspend fun clearE2EISettings() {
         wrapStorageRequest {
             userConfigStorage.setE2EISettings(null)
-            userConfigStorage.setE2EINotificationTime(0)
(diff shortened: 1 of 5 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
```diff
@@ -234,6 +235,13 @@ internal class UserConfigDataSource internal constructor(
             }
         }
 
+    override suspend fun clearE2EISettings() {
+        wrapStorageRequest {
+            userConfigStorage.setE2EISettings(null)
+            userConfigStorage.setE2EINotificationTime(0)
+        }
+    }
+
```
MohamadJaara, line 244: do we need to to first check if the user have E2EI or it is safe to call those function anyway ?

borichellow: in fact it's just a records in SharedPrefs, so quit safe to call it :)

mchenani (approved):

MohamadJaara (approved):

PR #2556, opened 2024-02-28
fix(monkeys): Fix removed attribute

augustocdias:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Code broke in previous commits, because an attribute was removed from the KaliumConfig.

----
##### References

### DIFF
--- a/monkeys/src/main/kotlin/com/wire/kalium/monkeys/homeDirectory.kt
+++ b/monkeys/src/main/kotlin/com/wire/kalium/monkeys/homeDirectory.kt
@@ -34,7 +34,6 @@ fun coreLogic(
             encryptProteusStorage = true,
             isMLSSupportEnabled = true,
             wipeOnDeviceRemoval = true,
-            fetchAllTeamMembersEagerly = true,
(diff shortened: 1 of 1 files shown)


MohamadJaara (approved):

vitorhugods (approved):
Just for clarity:

This parameter was replaced with `limitTeamMembersFetchDuringSlowSync` which is `null` by default.

So it will keep the same behaviour. Clients that don't want to fetch _all_ team members should pass a concrete value to it.

monkeys/src/main/kotlin/com/wire/kalium/monkeys/homeDirectory.kt
```diff
@@ -34,7 +34,6 @@ fun coreLogic(
             encryptProteusStorage = true,
             isMLSSupportEnabled = true,
             wipeOnDeviceRemoval = true,
-            fetchAllTeamMembersEagerly = true,
```
MohamadJaara, line 37: btw there is a new value 
`limitTeamMembersFetchDuringSlowSync` if you set it to null is will behave the same as if `          fetchAllTeamMembersEagerly = true,`

PR #2585, opened 2024-03-05
fix: Change return type of getConversationByGroupID to nullable for epoch changes

alexandreferris:
----

- The **PR Title**

- The **PR Description**
  - [X]  is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

While observing epoch changes for calls (SubConversation) it would fail and log an error message as it did not find a conversation with its GroupId.

### Causes (Optional)

Calls (SubConversation) on MLS is only stored in memory and not in the DB, thus resulting in an error log when it did not find.

### Solutions

Change return type of `getConversationByGroupID` to be nullable and change its execution to `executeOneOrNull` so it returns null when it doesn't find it as there is no need to observe epoch changes for calls.

#### Test Coverage (Optional)

#### How to Test

- Have E2EI + MLS enabled
- Have 2 Users [A, B]
- Have a call between both Users
- Start/Join call and keep an eye on the logs, there should be no error log for:

```
java.lang.NullPointerException: ResultSet returned null for Conversations.sq:selectByGroupId
```

### DIFF
--- a/monkeys/src/main/kotlin/com/wire/kalium/monkeys/conversation/Monkey.kt
+++ b/monkeys/src/main/kotlin/com/wire/kalium/monkeys/conversation/Monkey.kt
@@ -313,7 +313,7 @@ private suspend fun getAuthScope(coreLogic: CoreLogic, backend: Backend): Authen
             isOnPremises = true,
             apiProxy = null
         )
-    ).invoke()
+    ).invoke(null)
     if (result !is AutoVersionAuthScopeUseCase.Result.Success) {
         error("Failed getting AuthScope: $result")
     }
--- a/persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/ConversationDAOTest.kt
+++ b/persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/ConversationDAOTest.kt
@@ -1811,7 +1811,7 @@ class ConversationDAOTest : BaseDatabaseTest() {
     }
 
     @Test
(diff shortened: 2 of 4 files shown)


MohamadJaara (approved):

MohamadJaara (approved):

monkeys/src/main/kotlin/com/wire/kalium/monkeys/conversation/Monkey.kt
```diff
@@ -313,7 +313,7 @@ private suspend fun getAuthScope(coreLogic: CoreLogic, backend: Backend): Authen
             isOnPremises = true,
             apiProxy = null
         )
-    ).invoke()
+    ).invoke(null)
```
vitorhugods, line 316: Expected change?

Garzas (approved):

persistence/src/commonTest/kotlin/com/wire/kalium/persistence/dao/ConversationDAOTest.kt
```diff
@@ -1810,6 +1810,22 @@ class ConversationDAOTest : BaseDatabaseTest() {
         )
     }
 
+    @Test
+    fun givenSubConversationGroupIdForCall_whenGettingConversationByGroupId_thenReturnNull() = runTest {
```
vitorhugods, line 1814:
```suggestion
    fun givenInsertedConversations_whenGettingConversationByInexistingGroupId_thenReturnNull() = runTest {
```

The DB has no idea what "subconversation" means. So might as well keep it generic insteaed of over-specific.

vitorhugods (approved):

PR #2590, opened 2024-03-05
fix: showing degradation alert dialog [WPB-6971] [WPB-6972]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The degradation alert dialog is showing when sending first message in verified conversation but it's not showing when sending messages in a group that's degraded.

### Causes (Optional)

Wrong boolean expression leading to persisting the inverted value.

### Solutions

Invert the boolean expression to `setDegradedConversationNotifiedFlag` when the new status is **not** `DEGRADED`, because if this flag is true it means that the user has been already notified about conversation being degraded or there's no need to inform the user because the conversation is not degraded.
Also added another check in `addMessageAfterProteusVerificationStatusChange` trigger to only change the `degraded_conversation_notified` flag if the proteus verification status changes and the conversation still actually uses proteus, otherwise if only `MLS` protocol is used then `proteus_verification_status` should not trigger degradation alert dialog anymore, only MLS' `verification_status` changes should be taken into account in that case.

#### Test Coverage (Optional)

#### How to Test

Have a verified MLS conversation and add new non-verified user (degradation alert dialog should be shown) or send first message in a verified conversation (degradation alert dialog should not be shown).

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/MLSConversationsVerificationStatusesHandler.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/MLSConversationsVerificationStatusesHandler.kt
@@ -127,6 +127,6 @@ internal class MLSConversationsVerificationStatusesHandlerImpl(
 
         persistMessage(conversationDegradedMessage)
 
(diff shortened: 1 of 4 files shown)


MohamadJaara (approved):
great work 🚀

Garzas (approved):

PR #2624, opened 2024-03-11
fix: handle with scheme and domain from corecrypto [WPB-7084]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When getting user identities from core crypto, `handle` field has a different format because it contains also scheme and domain.

### Solutions

Provide another property that returns only the core handle without scheme, `%40` at sign and domain.

#### Test Coverage (Optional)

#### How to Test

Create a new MLS group with all users verified - it should have a badge.

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/IDs.kt
+++ b/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/IDs.kt
@@ -72,7 +72,7 @@ data class CryptoQualifiedClientId(
 
 data class WireIdentity(
     val clientId: CryptoQualifiedClientId,
-    val handle: String,
+    val handle: String, // handle format is "{scheme}%40{handle}@{domain}", example: "wireapp://%40hans.wurst@elna.wire.link"
     val displayName: String,
     val domain: String,
     val certificate: String,
@@ -80,7 +80,10 @@ data class WireIdentity(
     val thumbprint: String,
     val serialNumber: String,
     val endTimestamp: Long
-)
+) {
+    val handleWithoutSchemeAtSignAndDomain: String
+        get() = handle.substringAfter("://%40").removeSuffix("@$domain")
+}
 
 enum class CryptoCertificateStatus {
     VALID, EXPIRED, REVOKED;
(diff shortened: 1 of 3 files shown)


cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/IDs.kt
```diff
@@ -72,15 +72,18 @@ data class CryptoQualifiedClientId(
...
     val domain: String,
     val certificate: String,
     val status: CryptoCertificateStatus,
     val thumbprint: String,
     val serialNumber: String,
     val endTimestamp: Long
-)
+) {
+    val handleWithoutSchemeAtSignAndDomain: String
+        get() = handle.substringAfter("://%40").removeSuffix("@$domain")
```
MohamadJaara, line 85: i assume the %40 is  because it is url encoded,, so can we decode it first and then split?

Garzas (approved):

vitorhugods (approved):
As much as this PR fixes things, I'd love to see a PR that prevents things to be broken in the first place.

Something that properly defines what a `Handle` is, checks for its format, etc.
In general, avoiding using `String` completely. Just once during creation and then we're safe. Single point of failure during creation of the object should be preferred instead of having to know the format of the String everywhere.

```
data class Handle(
    val userId: String,
    val userDomain: String
) {
    companion object {
        fun fromString(rawValue: String): Handle{
            // TODO
        }
    }
}
```

cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/IDs.kt
saleniuk: @MohamadJaara `URLDecoder` is from Android, do we have any multiplatform encoding solution?

saleniuk: I will just merge it like this for now, we can update it later.

PR #2633, opened 2024-03-12
fix: E2EI enroll twice on login

borichellow:
# What's new in this PR?

### Issues

If the E2EI is enabled when user logs in, user has to enroll the certificate 2 times: 
- just after login in EnrollE2EIScreen
- in Dialog on HomeScreen

### Causes (Optional)

mistake in `GetE2eiCertificateUseCaseImpl` which is used in `ObserveE2EIRequiredUseCase`: 
When the user just logged in app strats to observe if it needs to ask user to enroll the certificate. For that it checks if E2EI is required AND If current client has a certificate.
To get "If current client has a certificate" it needs to get MLSGroupId, which is stored in Conversation with that user (SELF conversation in that case).
But the problem is that there is no any conversations in DB when user just logged in. 

So `GetE2eiCertificateUseCaseImpl` gets `StorageFailure.DataNotFound` which is about the conversation, but not a certificate itself. 

### Solutions

In `GetE2eiCertificateUseCaseImpl`: do not map `StorageFailure.DataNotFound` into `GetE2EICertificateUseCaseResult.NotActivated`, but into `GetE2EICertificateUseCaseResult.Failure`


vitorhugods (commented):
Tests that cover this?

borichellow:
> Tests that cover this?

there are a lot of it in `ObserveE2EIRequiredUseCaseTest` :) 
But i can add tests for `GetE2eiCertificateUseCase` too :)

Garzas (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/GetE2EICertificateUseCase.kt
```diff
@@ -37,10 +36,7 @@ class GetE2eiCertificateUseCaseImpl internal constructor(
 ) : GetE2eiCertificateUseCase {
     override suspend operator fun invoke(clientId: ClientId): GetE2EICertificateUseCaseResult =
         mlsConversationRepository.getClientIdentity(clientId).fold(
-            {
-                if (it is StorageFailure.DataNotFound) GetE2EICertificateUseCaseResult.NotActivated
-                else GetE2EICertificateUseCaseResult.Failure
```
MohamadJaara, line 42: are we sure this will fix it but not break other stuff
what if the self conversation is there but the cert is actually `DataNotFound`

MohamadJaara (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/GetE2EICertificateUseCase.kt
borichellow: I'm quit sure about it. 
`DataNotFound` is a child of `StorageFailure`, MLS shouldn't return it. 
If user has no certificate `mlsConversationRepository.getClientIdentity` returns `Either.Right(null)`.

MohamadJaara (approved):

PR #2643, opened 2024-03-13
fix: MLS enabling setting [WPB-7097]

borichellow:
# What's new in this PR?

### Issues

Wrong calculating if MLS is enabled for user or not. 

### Causes (Optional)

Nobody knows how it should work :) 

### Solutions

Fix it: 
- if the MLS is enabled in settings AND MLS is in the supported protocols list, 
- in all other cases MLS is disabled.

AND 
in `com.wire.kalium.logic.data.featureConfig.MLSModel` make field `allowedUsers` nullable as it might be removed in the future.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/featureConfig/FeatureConfigModel.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/featureConfig/FeatureConfigModel.kt
@@ -77,7 +77,7 @@ data class SelfDeletingMessagesConfigModel(
 )
 
 data class MLSModel(
-    val allowedUsers: List<PlainId>,
+    val allowedUsers: List<PlainId>?,
     val defaultProtocol: SupportedProtocol,
     val supportedProtocols: Set<SupportedProtocol>,
     val status: Status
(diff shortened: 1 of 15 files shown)


network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/featureConfigs/FeatureConfigResponse.kt
```diff
@@ -85,7 +85,7 @@ data class ClassifiedDomainsConfigDTO(
 @Serializable
 data class MLSConfigDTO(
     @SerialName("protocolToggleUsers")
-    val protocolToggleUsers: List<String>,
+    val protocolToggleUsers: List<String>?,
```
MohamadJaara, line 88: this can be removed

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/featureConfig/FeatureConfigModel.kt
```diff
@@ -77,7 +77,7 @@ data class SelfDeletingMessagesConfigModel(
 )
 
 data class MLSModel(
-    val allowedUsers: List<PlainId>,
+    val allowedUsers: List<PlainId>?,
```
MohamadJaara, line 80: same for here

MohamadJaara (approved):

ohassine (approved):

PR #2663, opened 2024-03-17
fix: unread mention [WPB-5786]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- unread event with previous message id was not cleared
- updating unread event was not working when it was not previously added

### Causes (Optional)

Unread mention event was not clearing in conversation list

### Solutions

- merge inserting and updating unread event into upsert event
- remove unread event before message id gets updated

### DIFF
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/UnreadEvents.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/UnreadEvents.sq
@@ -18,14 +18,12 @@ DELETE FROM UnreadEvent WHERE id = ? AND conversation_id = ?;
 deleteUnreadEvents:
 DELETE FROM UnreadEvent WHERE creation_date <= ? AND conversation_id = ?;
 
-insertEvent:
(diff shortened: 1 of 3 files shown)


saleniuk:
I think it does the same as the `ON UPDATE CASCADE` which we already have added to the foreign key for this table, of course if we do actions in the right order which I believe is correct:
- the app gets a new message id:A with mention to self
- the app adds new row into `UnreadEvents` table with id:A
- the app gets update to message id:A with new id:B which removes the mention
- the app changes the unread event to from "MENTION" to "MESSAGE" for event id:A
- the app updates the id of the message from id:A to id:B, which also updates this id in `UnreadEvents` table
- both updated message and unread event now have the updated id:B

This change is to make it manually instead of let SQL do that itself when message id changes, but the logic generally is not changed at all, so maybe the problem is somewhere else?

MohamadJaara:
i just tested this PR and compared it to prod and in both apps this bug did not reproduce so it is fixed by something else

PR #2664, opened 2024-03-17
feat: message draft [WPB-1021]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Implemented message draft table to save message drafts when user leaves conversation screen before sending message. It saves 
- text
- mentions
- quoted message id


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/draft/MessageDraft.kt
```diff
@@ -0,0 +1,27 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.kalium.logic.data.message.draft
+
+import com.wire.kalium.logic.data.message.mention.MessageMention
+
+data class MessageDraft(
```
vitorhugods, line 22: What if instead of adding nullables, we create sealed implementations?

`DraftMessage.Reply`

`DraftMessage.Quote`

`DraftMessage.Simple`

Garzas: I think it's not worth it because it can be divided only into 2 types which one doesn't have quoted Id and one which has. Rest variables can be combined (edited message can have quote, mentions) and at the end it will be merged into one entity object saved in db

vitorhugods: Indeed. I forgot about the possibility of it being both Reply and Edit at the same time.

saleniuk (approved):

persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageDrafts.sq
```diff
@@ -0,0 +1,29 @@
+import com.wire.kalium.persistence.dao.QualifiedIDEntity;
+import com.wire.kalium.persistence.dao.message.MessageEntity.Mention;
+import kotlin.collections.List;
+
+CREATE TABLE MessageDrafts (
+      conversation_id TEXT AS QualifiedIDEntity NOT NULL,
+      text TEXT,
+      edit_message_id TEXT,
```
MohamadJaara, line 8: Should the reply and edit also be also a FOREIGN KEY to a message ID ? in that case if the message got deleted or changed we can change how draft will react 
we can for example set the edit_message_id or qouted_message_id to null when deleted on changed 

https://www.sqlite.org/foreignkeys.html

Garzas: Good idea

MohamadJaara (approved):
Great work as always

PR #2667, opened 2024-03-18
fix: False positive E2EI user verification [WPB-7126]

borichellow:
# What's new in this PR?

### Issues

When some user has few clients: 1 is verified with E2EI and others have no any Identity, in that case user marked as Verified in ProfileScreen

### Causes (Optional)

While checking user verification status, we go through each Identities that CC returns for that user and check if all of them are "Verified" and has the same userName and Handler as user has. But if  some client has no Identity at all  -> that case wasn't covered.

### Solutions

Cover such a case by comparing Identities list to Clients (by client id) and make sure that there are Identities for each Client, if not -> that user is not verified.


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/GetMembersE2EICertificateStatusesUseCase.kt
```diff
@@ -36,17 +39,18 @@ interface GetMembersE2EICertificateStatusesUseCase {
...
+    private val clientRepository: ClientRepository
 ) : GetMembersE2EICertificateStatusesUseCase {
     override suspend operator fun invoke(conversationId: ConversationId, userIds: List<UserId>): Map<UserId, CertificateStatus?> =
-        mlsConversationRepository.getMembersIdentities(conversationId, userIds).fold(
-            { mapOf() },
-            {
-                it.mapValues { (_, identities) ->
-                    identities.getUserCertificateStatus(certificateStatusMapper)
+        mlsConversationRepository.getMembersIdentities(conversationId, userIds).flatMap { identitiesMap ->
+            clientRepository.getClientsByConversationId(conversationId).map { clientsMap ->
```
mchenani, line 47: Question: there is a problem here, we don't have the list of clients of the user since sometimes we might get out of sync with the cloud and the list of the clients, there are two solutions, 
one is we make sure we have always the updated list of the users clients
other one: since CC always has the list of the mls clients in the mls tree, we can ask the CC to include an aggregated user status based on all of their existence clients in the mls tree, or return all the clients even some of them doesn't have a wire-identity.

mchenani (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/GetMembersE2EICertificateStatusesUseCase.kt
borichellow: yeah, i like the second too, but it may take some time and iOS and Web already added the fix on their sides.

borichellow:
Fixed on CC side

PR #2673, opened 2024-03-20
fix: null teamId for team member [WPB-5749]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

After changes from this PR https://github.com/wireapp/kalium/pull/2653, `selfTeamIdProvider` can return null `teamId` even if self user belongs to a team.

### Causes (Optional)

When fetching self team member data, during first slow sync the app will try to get self team id from DB before it's persisted. To insert self user data to the DB it needs to fetch team member data, but to fetch team member data it requires team id which is taken from DB by the `selfTeamIdProvider` - it returns null and this null is then used later on in the app.

### Solutions

Do not use `selfTeamIdProvider` until self user data is synced, instead in this case just use `teamId` from `selfUserDTO`.


alexandreferris (approved):

MohamadJaara (approved):

vitorhugods:
Go home codecov, you've drunk

MohamadJaara:
> Go home codecov, you've drunk

yamilmedina:
It must be April fools' 🤔 

slow

PR #2687, opened 2024-03-28
fix: Ignore ConnectionRequest on backend disabled RC [WPB-7087]

borichellow:
----

# What's new in this PR?

### Issues

When user is trying to ignore the Connection request from the user whose backend is unavailable - Error is displayed.
Igrnoring the ConnectionRequest should be available even in that case.

### Causes (Optional)

It was not implemented (quit a tricky case that nobody think about)

### Solutions

Update ConnectionRepository by adding a separate fun for ignoring the connection request. 
In case of `FederatedBackendFailure.FailedDomains` request is still marked as "Ignored" localy.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
@@ -42,7 +42,9 @@ import com.wire.kalium.logic.di.MapperProvider
 import com.wire.kalium.logic.failure.InvalidMappingFailure
 import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.logic.functional.flatMap
+import com.wire.kalium.logic.functional.flatMapLeft
 import com.wire.kalium.logic.functional.isRight
+import com.wire.kalium.logic.functional.left
 import com.wire.kalium.logic.functional.map
 import com.wire.kalium.logic.functional.onFailure
 import com.wire.kalium.logic.functional.onSuccess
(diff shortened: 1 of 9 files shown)


yamilmedina (approved):
❤️

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
```diff
@@ -125,22 +128,36 @@ internal class ConnectionDataSource(
...
-        return wrapApiRequest {
-            connectionApi.updateConnection(userId.toApi(), newConnectionStatus)
-        }.map { connectionDTO ->
-            val connectionStatus = connectionDTO.copy(status = newConnectionStatus)
+        return wrapApiRequest { connectionApi.updateConnection(userId.toApi(), newConnectionStatus) }
+    }
+
+    override suspend fun updateConnectionStatus(userId: UserId, connectionState: ConnectionState): Either<CoreFailure, Connection> =
+        updateRemoteConnectionStatus(userId, connectionState).map { connectionDTO ->
+            val connectionStatus = connectionDTO.copy(status = connectionStatusMapper.toApiModel(connectionState)!!)
```
Garzas, line 143: Should we secure that?

Garzas (approved):
Looking good 👍 Just one small question

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/connection/ConnectionRepository.kt
borichellow: well basically there is checking for null in `fun updateRemoteConnectionStatus` that is called before it. So it's kinda safe

PR #2688, opened 2024-03-28
fix: multiple missed calls system messages (WPB-5854)

ohassine:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The app is showing multiple missed calls system messages, while it should be displayed only once.

Steps to reproduce:
- Start a group call
- Use same account to answer it from web/ios client but not from Android
- Turn off Internet for Android and wait for about 60s
- After 60s turn on the Internet, you should see a missed call system message in the chat.
- Terminate the call on web
- Another missed call message will be displayed on Android.

### Causes (Optional)

An invalid closeCall callback triggered by AVS.
When a user went offline for some time, AVS will consider the ongoing call as terminated(timeout passed) and will trigger OnCloseCall callback while in reality it's not it is just the device went offline and the app can't receive any CONFCHECK message to check the status of the call.

### Solutions

In OnCloseCall call back, we check if the app is connected to Internet before persisting missed calls.

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
```diff
@@ -32,6 +32,7 @@ import com.wire.kalium.logic.data.id.CurrentClientIdProvider
 import com.wire.kalium.logic.feature.call.usecase.ConversationClientsInCallUpdater
 import com.wire.kalium.logic.feature.message.MessageSender
 import com.wire.kalium.logic.featureFlags.KaliumConfigs
+import com.wire.kalium.network.NetworkStateObserver
```
yamilmedina, line 35: I think the failing test is because of this line, we are accessing something inside the feature package (we should have here I think only use cases), We might try to solve it by moving this class to the repository package, since is not needed for AR ? 🤔

yamilmedina (commented):
I left a comment, regarding the failing test 😅

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/GlobalCallManager.kt
ohassine: Yeah but I would preferto not create a data package for each platform, I will just ignore this rule for those classes

alexandreferris (approved):

PR #2689, opened 2024-03-29
refactor: remove self_deletion_start_date [WPB-6219]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

`self_deletion_start_date` is deprecated and should be removed and fully replaced by `self_deletion_end_date` but it is still kept in the database and used in some places.

### Solutions

Use `ALTER TABLE DROP COLUMN` which is available from SQLite 3.35.0 - we provide our own sql binaries and for prod we use 3.44 (sqlcipher), it's needed to also update sqldelight dialect.
Note: this migration will not work on debug builds where we don't use sqlcipher to allow to inspect the db for debugging purposes.

#### Test Coverage (Optional)

#### How to Test

Send some self-deleting messages, update the app to perform DB migration and check if everything is ok with these messages.

----

----

----
##### References


Garzas (approved):

MohamadJaara (changes requested):
Discussed a better solution in 1:1

persistence/src/commonMain/db_user/migrations/81.sqm
```diff
@@ -0,0 +1,169 @@
+DROP INDEX IF EXISTS message_expire_after_millis_index;
+
+DROP VIEW IF EXISTS MessageDetailsView;
+
+ALTER TABLE Message DROP COLUMN self_deletion_start_date;
+
+CREATE INDEX message_expire_after_millis_index ON Message(expire_after_millis);
```
MohamadJaara, line 7: question: is  creating a new index is a bit faster if there it is already included in another composite index?
does it make sense to create the new index before deleting 
```suggestion
CREATE INDEX message_expire_time_index ON Message(expire_after_millis);

DROP INDEX IF EXISTS message_expire_after_millis_index;
DROP VIEW IF EXISTS MessageDetailsView;
ALTER TABLE Message DROP COLUMN self_deletion_start_date;
```

saleniuk: Hmm, at first glance it doesn't look like it takes advantage of another index when creating one, `EXPLAIN` shows exactly the same actions when there's already other composite index and when not, but the only disadvantage of changing the order like so is that we just need to use different name for the new index so I applied this change just in case.

MohamadJaara (approved):

PR #2691, opened 2024-04-02
feat: Update CC to 55 and adopt E2EI changes [WPB-7126]

borichellow:
# What's new in this PR?

1. update CC version to 55
2. update mpping for WireIdentity
3. update calculation of users E2EI verification status: if any WireIdentity has no certificate (no metter if its status is VALID) -> user is not verified.


cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/MLSClientImpl.kt
```diff
@@ -326,14 +326,14 @@ class MLSClientImpl(
             return clientId?.let {
                 WireIdentity(
                     CryptoQualifiedClientId.fromEncodedString(value.clientId)!!,
-                    value.handle,
-                    value.displayName,
-                    value.domain,
-                    value.certificate,
+                    value.x509Identity?.handle ?: "",
```
mchenani, line 329: Question: since the inner object is nullable. Imo we should return null instead of hardcoded values?

mchenani (approved):
LGTM, with a question.

cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/MLSClientImpl.kt
```diff
@@ -326,14 +326,14 @@ class MLSClientImpl(
...
+                    value.x509Identity?.displayName ?: "",
+                    value.x509Identity?.domain ?: "",
+                    value.x509Identity?.certificate ?: "",
```
MohamadJaara, line 332: please do not add default empty strings, if the value is supposed to be nullable now then keep is nullable

MohamadJaara (approved):

MohamadJaara (changes requested):
change to keep nullable values as it is

MohamadJaara (approved):

PR #2698, opened 2024-04-09
fix: SearchUserResult merging local and remote results [WPB-7340]

borichellow:
# What's new in this PR?

### Issues

In creating new group conversation: while searching the user and trying to select some private acc (that i have already connected with), the selecting check-box doesn't react on clicks.

### Causes (Optional)

Searching for the user has 2 sources: local storage (for the users that i'm connected with) and API (for remote users, user from another teams, or private users). 
When kalium gets users from API it's `UserProfileDTO` which needs to be mapped into `OtherUser`, as far as `UserProfileDTO` has no `connectionState` field and `OtherUser` needs it, it's mapped like: "if user is in the same team as me -> he's connected, if not -> not_connected". So even if i'm already connected with some private user he'll be "not_connected". 

Later when these 2 sources (local and remote) are merged: there is a check if some user has status "connected" OR he's in local list (checking by user_id) we replace that user in localList by the user from remoteList and remove it from remoteList. 

That is the problem point: as I'm already connected with some private user he should have status "connected" (and he does in localList) but the user data that comes from API always has "Not_connected" status for private users.

The problem becomes visible later in UI when we check if the CheckBox in contacts list should be checked or not (if user was selected for the new group or not), there is comparing: "does SelectedContacts list contain that Contact": SelectedContacts contains user with connected status and "that Contact" has not_connected status so the `selectedContactsList.contains(contact)` always false.  

### Solutions

in merging remote and local SearchResults: 
1. if localList contains user with the same user_id as in remoteList -> just remove that user from remoteList. 
2. If user in remoteList has "Connected" status -> add it to localList and remove from remoteList (was done before changes)

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/search/SearchUserResult.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/search/SearchUserResult.kt
@@ -32,8 +32,11 @@ data class SearchUserResult(
         ): SearchUserResult {
             val updatedUser = mutableListOf<UserId>()
             remoteSearch.forEach { (userId, remoteUser) ->
-                if ((remoteUser.connectionStatus == ConnectionState.ACCEPTED)) {
+                if (remoteUser.connectionStatus == ConnectionState.ACCEPTED) {
                     localResult[userId] = remoteUser
+                    updatedUser.add(userId)
+                }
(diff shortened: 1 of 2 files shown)


MohamadJaara (changes requested):
one change for how to do the checks

Garzas (approved):

PR #2713, opened 2024-04-23
fix: UseCase for getting if one on one conversation with user exist [WPB-6936]

borichellow:
# What's new in this PR?

Added `IsOneToOneConversationCreatedUseCase` for getting if 1o1 conversation with user exists


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/IsOneToOneConversationCreatedUseCase.kt
```diff
@@ -0,0 +1,44 @@
...
+}
+
+internal class IsOneToOneConversationCreatedUseCaseImpl(
+    private val conversationRepository: ConversationRepository
+) : IsOneToOneConversationCreatedUseCase {
+    override suspend operator fun invoke(otherUserId: UserId): Boolean {
+        return conversationRepository.observeOneToOneConversationWithOtherUser(otherUserId)
+            .first()
+            .isRight()
+    }
```
MohamadJaara, line 43: this query is selecting from the `ConversationDetails` view with quite some joins, if the goal is only to check if there is an established 1:1 then you can 

```sql 
SELECT active_one_on_one_conversation_id 
FROM User 
WHERE qualified_id = 'your_user_id';
```
and a simple chck if this returns a null or not should do the trick

MohamadJaara (changes requested):

vitorhugods (approved):

MohamadJaara (approved):

PR #2736, opened 2024-05-02
fix: repositoriesShouldNotAccessFeature in MLSConversationRepository [WPB-6326]

borichellow:
# What's new in this PR?

### Issues

MLSConversationRepository was accessing the useCase from Feature module

### Causes (Optional)

It was done like that

### Solutions

Move `CheckRevocationListUseCase` into Logic module and update it to not use UseCases from the Feature. 
This UseCase is used only in kalium, so it doesn't affect Android part.

### DIFF
--- a/.github/workflows/benchmarks-check.yml
+++ b/.github/workflows/benchmarks-check.yml
@@ -1,0 +1,55 @@
+name: Benchmark Checks
+on:
+    merge_group:
+    pull_request:
+        types: [ opened, synchronize ]
+    workflow_dispatch:
+
+concurrency:
+    group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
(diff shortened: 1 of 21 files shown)


yamilmedina (approved):

mchenani (approved):

vitorhugods (changes requested):
Nice to dedicate some time to fix this.

But I think just moving the file and adding "internal" to the name isn't enough.

This is just moving a `"UseCase"` to `data`, which is not a pattern we should follow.

Unlike real use cases, this one is not public, it shouldn't be used by the apps.

It's an internal, reusable piece of logic that other use cases (and even repositories) are using. So removing this suffix would be better.

vitorhugods (approved):

PR #2790, opened 2024-05-31
fix: Wrong time of SystemMessage about creating group [WPB-8673]

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8673" title="WPB-8673" target="_blank">BugWPB-8673</a>  [Android] Messages appear before system message about starting conversation
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

Sometimes when you start a new conversation, the system message about the conversation start is displayed before the first messages are being sent.

### Causes (Optional)

System message about creating the group are created with time `DateTimeUtil.currentIsoDateTimeString()`, so if event handled with some delay (ex. was no connection) those messages will have wrong time. 

### Solutions

Use `timestampIso` from `Event.Conversation.NewConversation` for time of such messages.


borichellow:
> Looking good 🚀 Can we just add single test that will check if conversation created system message was received later than other message, then it will return this system message before regular message?

I'm afraid such a test will fail :) 

The fix in this PR is to use date of ConversationCreated Event for the "conversation created messages", instead of "current time" which is the time of handling such even on device. 
So if Event has wrong time -> message has a wrong time, but in that case issue is not on our side, I believe :)

Garzas (approved):

saleniuk (approved):

PR #2800, opened 2024-06-05
feat: CRL proxy [WPB-8793]

borichellow:
# What's new in this PR?

### Issues

The web/desktop frontend use a proxy to fetch CRLs from other domains. However Android clients fetch those CRLs directly.

### Causes (Optional)

It was implemented like this.

### Solutions

Implement usage of proxy url for fetching CRLs when it's required and available.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/E2EISettings.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/E2EISettings.kt
@@ -23,11 +23,13 @@ import kotlinx.datetime.Instant
 data class E2EISettings(
     val isRequired: Boolean,
     val discoverUrl: String?,
-    val gracePeriodEnd: Instant?
+    val gracePeriodEnd: Instant?,
+    val shouldUseProxy: Boolean,
+    val crlProxy: String?,
 ) {
 
     fun toEntity() = E2EISettingsEntity(
-        isRequired, discoverUrl, gracePeriodEnd?.toEpochMilliseconds()
+        isRequired, discoverUrl, gracePeriodEnd?.toEpochMilliseconds(), shouldUseProxy, crlProxy
     )
 
     companion object {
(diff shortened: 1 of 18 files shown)


ohassine (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/E2EISettings.kt
```diff
@@ -23,18 +23,22 @@ import kotlinx.datetime.Instant
...
-        isRequired, discoverUrl, gracePeriodEnd?.toEpochMilliseconds()
+        isRequired, discoverUrl, gracePeriodEnd?.toEpochMilliseconds(), shouldUseProxy, crlProxy
     )
 
     companion object {
         fun fromEntity(entity: E2EISettingsEntity) = E2EISettings(
             entity.status,
             entity.discoverUrl,
             entity.gracePeriodEndMs?.let { Instant.fromEpochMilliseconds(it) },
+            entity.shouldUseProxy == true,
```
mchenani, line 40: question: is that necessary to compare it? why not returning the value it self?

mchenani (approved):
LGTM, added some questions and suggestions.

network/src/commonMain/kotlin/com/wire/kalium/network/api/base/unbound/acme/ACMEApi.kt
```diff
@@ -59,7 +59,7 @@ interface ACMEApi {
      * @return A [NetworkResponse] object containing the certificate chain as a list of strings.
      */
     suspend fun getACMEFederationCertificateChain(discoveryUrl: String): NetworkResponse<List<String>>
-    suspend fun getClientDomainCRL(url: String): NetworkResponse<ByteArray>
+    suspend fun getClientDomainCRL(url: String, proxyUrl: String?): NetworkResponse<ByteArray>
```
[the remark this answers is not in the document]
borichellow: well, we might do that, but this fun is called only from 1 place, so it's not big deal :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/E2EISettings.kt
borichellow: I'd prefer to have as less nullable as possible and in fact this filed should be non-null, but in `Entity` it might be null because of migration from the old versions, that's why i used this ` == true` trick to set `false` as default value even for old-versions.

PR #2806, opened 2024-06-11
feat: Add benchmarks for core logic (WPB-9534)

mythsunwind:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9534" title="WPB-9534" target="_blank">TaskWPB-9534</a>  Benchmark creation of coreLogic object
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Adds benchmarks for instantiation of CoreLogic object with or without In-Memory-Storage.

### DIFF
--- a/benchmarks/build.gradle.kts
+++ b/benchmarks/build.gradle.kts
@@ -25,9 +25,11 @@ kotlin {
         val commonMain by getting {
             dependencies {
                 implementation(project(":persistence"))
+                implementation(project(":logic"))
                 implementation(libs.coroutines.core)
                 implementation(libs.ktxDateTime)
                 implementation(libs.kotlinx.benchmark.runtime)
+                implementation(libs.ktor.mock)
             }
         }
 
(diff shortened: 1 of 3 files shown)


MohamadJaara:
I think this benchmark is a good start, but it will not tell the full story until there is a logged-in user or something that uses the database.

yamilmedina (approved):
Good to see this growing, that was the initial idea to perf. Test different parts of Kalium. 🙌  Great addition as well these scoped suites.

mythsunwind:
> I think this benchmark is a good start, but it will not tell the full story until there is a logged-in user or something that uses the database.

Indeed. This was my initial plan but unfortunately the kotlin benchmark framework makes it currently a bit tricky to load native libraries on microbrenchmark runs. This is a problem because libsodium/corecrypto is needed already in the moment of login and the libraries are not loaded on runtime executing of benchmarks. The issue is described [here](https://github.com/Kotlin/kotlinx-benchmark/issues/180) in their bug tracker.

MohamadJaara (approved):

PR #2811, opened 2024-06-14
fix: send external messages properly

vitorhugods:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
The Android is app is rarely sending external messages, even when the payload is too big.

### Causes

math-hmm

Sum operation instead of multiplication.
The past tests only covered the scenario where the message content is too big, and not when the number of recipients drives the message size above the threshold.

Also, reduce the limit to 200KiB, instead of 256KiB. This matches the behaviour on Web and iOS clients.

### Solutions

Fix the wrong calculation.

#### Test Coverage

----

----
##### References


alexandreferris (approved):

borichellow (approved):

yamilmedina (approved):
Good catch 🤓. Do we have a ticket for this? ,if so can you add it to the description, please

vitorhugods:
> Good catch 🤓. Do we have a ticket for this? ,if so can you add it to the description, please

Not really.
We caught this a few days ago when iOS was investigating a possible bug on their side.

Turns out we identified issues with External Messages on Web and Backend as well. Although it improves performance a bit when sending messages, it's not really the most important feature I guess 😅

yamilmedina:
> > Good catch 🤓. Do we have a ticket for this? ,if so can you add it to the description, please
> 
> Not really. We caught this a few days ago when iOS was investigating a possible bug on their side.
> 
> Turns out we identified issues with External Messages on Web and Backend as well. Although it improves performance a bit when sending messages, it's not really the most important feature I guess 😅

Oh, I see ! thanks for the context. So the question is: Do we want a ticket so we can track back and QA test it ? Or it will go directly to done 🤔  what do you suggest  ?

vitorhugods:
> > > Good catch 🤓. Do we have a ticket for this? ,if so can you add it to the description, please
> > 
> > 
> > Not really. We caught this a few days ago when iOS was investigating a possible bug on their side.
> > Turns out we identified issues with External Messages on Web and Backend as well. Although it improves performance a bit when sending messages, it's not really the most important feature I guess 😅
> 
> Oh, I see ! thanks for the context. So the question is: Do we want a ticket so we can track back and QA test it ? Or it will go directly to done 🤔 what do you suggest ?

The sending/receiving was working. If you send a big-enough message it was being sent as external. It just wasn't sending as external when the message was small and the number of recipients is big.

It's not really testable. It's a protocol-level change. It'd almost be like renaming a JSON field and asking QA to test it. Actually, that would be more testable, as we could at least see if nothing broke 😄 

We already had the condition:
```kotlin
return if (totalEstimatedSize <= MAX_CONTENT_SIZE) {
    // ...
} else {
```

IIRC the current tests for this in QA is just "write a huge message and make sure it goes through". So the `else` branch works.
Now we're changing how frequently the encoding logic goes through the if or the else.

For the user and QA, there's no visible difference between these two but the time it goes from pending to sent. Unless QA can precisely measure time, or create _huge_ groups where this difference can be more easily seen, it's not testable.

yamilmedina:
> > > > Good catch 🤓. Do we have a ticket for this? ,if so can you add it to the description, please
> > > 
> > > 
> > > Not really. We caught this a few days ago when iOS was investigating a possible bug on their side.
> > > Turns out we identified issues with External Messages on Web and Backend as well. Although it improves performance a bit when sending messages, it's not really the most important feature I guess 😅
> > 
> > 
> > Oh, I see ! thanks for the context. So the question is: Do we want a ticket so we can track back and QA test it ? Or it will go directly to done 🤔 what do you suggest ?
> 
> The sending/receiving was working. If you send a big-enough message it was being sent as external. It just wasn't sending as external when the message was small and the number of recipients is big.
> 
> It's not really testable. It's a protocol-level change. It'd almost be like renaming a JSON field and asking QA to test it. Actually, that would be more testable, as we could at least see if nothing broke 😄
> 
> We already had the condition:
> 
> ```kotlin
> return if (totalEstimatedSize <= MAX_CONTENT_SIZE) {
>     // not external
> } else {
>     // external
> }
> ```
> 
> IIRC the current tests for this in QA is just "write a huge message and make sure it goes through". So the `else` branch works. Now we're changing how frequently the encoding logic goes through the if or the else.
> 
> For the user and QA, there's no visible difference between these two but the time it goes from pending to sent. Unless QA can precisely measure time, or create _huge_ groups where this difference can be more easily seen, it's not testable.

Cool, thanks again for the detailed answer. Yup I saw the code, the question it was more about how to track back :) You know, in case things break later or test it since this is part of messaging :D But I understand you don't think it is necessary, so all good here, and happy to hear a simple regression check.

PR #2828, opened 2024-06-20
fix: Wrong E2EI information for other users [WPB-9409]

borichellow:
# What's new in this PR?

### Issues

Wrong e2ei information for other users after self user is kicked out of the group:

- enable MLS for team on bund-next env
- login with user A, B and C
- create an MLS group with all 3 users
- create a 1:1 between A and C
- enable e2ei for all the users
- generate cert for A and B
- delay generating the cert for C
- remove user C from the group conversation
- now take a look as user C at the 1:1 conversation  with user A > details > devices

Result:
device for user A does not have cert information

### Causes (Optional)

To get Identities for each device of user we need to get Id of MLS group this users belongs to. 
To get MLS group kalium finds the conversation that this user belongs to and get MLSGroupId from it. 

The problem comes out when self user is removed from some group-conversation and when kalium is looking for conversation OtherUser belongs to, finds exactly that group-conversation (it happens not in 100% of cases as there are couple of such conversations in DB, but `LIMIT 1` in query is used). So kalium uses MLSGroupId of the conversation SelfUser doesn't belongs to => doesn't have access to MLS group => no Identities found for OtherUser. 

### Solutions

Update DB-query to get MLSGpoupId of conversation that both (SelfUser and OtherUser) belongs to.

### DIFF
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
@@ -393,9 +393,15 @@ AND Conversation.mls_group_state IS 'ESTABLISHED'
 ORDER BY Conversation.type DESC LIMIT 1;
 
 getMLSGroupIdByUserId:
-SELECT Conversation.mls_group_id FROM Member
-JOIN Conversation ON Conversation.qualified_id = Member.conversation
-WHERE Conversation.mls_group_id IS NOT NULL AND mls_group_state IS 'ESTABLISHED' AND Member.user = :userId LIMIT 1;
+SELECT mls_group_id FROM
+	(SELECT COUNT(Member.user) AS users_amout, Conversation.mls_group_id, Conversation.mls_group_state FROM Member
+	JOIN Conversation ON Conversation.qualified_id = Member.conversation
+	WHERE Member.user = :userId OR Member.user = (SELECT SelfUser.id FROM SelfUser LIMIT 1)
+	GROUP BY Conversation.qualified_id)
+WHERE users_amout > 1 -- both (Self and Other users) belongs to that conversation
+AND mls_group_id IS NOT NULL
+AND mls_group_state IS 'ESTABLISHED'
+LIMIT 1;
 
 getMLSGroupIdByConversationId:
 SELECT Conversation.mls_group_id FROM Conversation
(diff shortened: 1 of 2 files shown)


MohamadJaara (changes requested):
please add tests

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
```diff
@@ -393,9 +393,15 @@ AND Conversation.mls_group_state IS 'ESTABLISHED'
 ORDER BY Conversation.type DESC LIMIT 1;
 
 getMLSGroupIdByUserId:
-SELECT Conversation.mls_group_id FROM Member
-JOIN Conversation ON Conversation.qualified_id = Member.conversation
-WHERE Conversation.mls_group_id IS NOT NULL AND mls_group_state IS 'ESTABLISHED' AND Member.user = :userId LIMIT 1;
+SELECT mls_group_id FROM
+	(SELECT COUNT(Member.user) AS users_amout, Conversation.mls_group_id, Conversation.mls_group_state FROM Member
```
mchenani, line 397: Two points:
1. Please let's make sure this query is called whenever it's necessary. Mls and e2ei enabled for example. 
2. Can we think of something better to detect if there is group that we're not part of it? E.g. a state somewhere in the db that we can filter based on it?

borichellow: 1. this is already done in UseCases that are calling this query (`GetUserE2eiCertificatesUseCaseImpl`, `GetUserE2eiCertificateStatusUseCaseImpl`). 

2. I was looking for the best option here. We could make a query to `ConversationDetails` instead of `Conversation` table, and check there `self_role != null` but it's some workaround IMO. 
Or we could store somewhere conversations that user was removed from, BUT:
- this will add more complexity (we'll need to make sure this state is updated properly) and 
- creates additional source of truth, which might add more problems.


P.S. I agree this query is not the most beautiful query in the world, but we have a lot of queries and triggers that are much more complex ;P And this one does exactly what we need, without any additional changes and workarounds.

mchenani: Thanks for the clarification, I would go with this instead of fetching the ConversationDetails [ it's a much bigger boy]

mchenani (approved):

yamilmedina (approved):

MohamadJaara (approved):

PR #2849, opened 2024-06-27
refactor!: use Instant instead of string in most places [WPB-9216]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9216" title="WPB-9216" target="_blank">BugWPB-9216</a>  [Android] out of memory exception when scrolling a group chat with long unread history
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We're using String for date representation in multiple core places, like messages, conversations, events, etc.
This is bad because:
- We can't be sure at compile time that the format is correct
- We do unnecessary operations, considering that the database is storing the dates/times as unix millis, as parsing to String is slower than just instantiating an Instant
- In many places, we need to sum/subtract/compare times, and this requires parsing the string back to a time data format and then back to string

### Solutions

Use Kotlin's date-time Instant.
When receiving from backend -> parse to Instant.
When reading from database -> create an Instant.

Bubble it up all the way to the Android app. Avoid String as much as possible until actually necessary to display in the UI.

Kalium only deals with Instant now 🎉 

### Testing

Not really _added_ tests, but had to adjust some. Should still be alright.

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/androidInstrumentedTest/kotlin/com/wire/kalium/logic/feature/call/CallManagerTest.kt
+++ b/logic/src/androidInstrumentedTest/kotlin/com/wire/kalium/logic/feature/call/CallManagerTest.kt
@@ -49,6 +49,8 @@ import kotlin.test.Ignore
 import kotlin.test.Test
 import com.wire.kalium.logic.feature.call.usecase.ConversationClientsInCallUpdater
 import com.wire.kalium.network.NetworkStateObserver
+import com.wire.kalium.util.DateTimeUtil.toIsoDateTimeString
+import kotlinx.datetime.Instant
 
 class CallManagerTest {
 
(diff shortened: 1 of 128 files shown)


Garzas (approved):
🧑‍🏭

borichellow (approved):

yamilmedina (approved):
Cool improvement across the codebase 🚀 

Just one suggestion before merge, to check (if not done already) for possible breaking changes in `TestService` and `Monkeys` Since they use some classes directly, and they are not part of test suite :D

vitorhugods:
> Cool improvement across the codebase 🚀
> 
> Just one suggestion before merge, to check (if not done already) for possible breaking changes in `TestService` and `Monkeys` Since they use some classes directly, and they are not part of test suite :D

It's all good for monkeys and test service here.

Unless it breaks the runtime, but I find it hard to see it happening 🤔

PR #2867, opened 2024-07-08
feat: Add isReplyAllowed field to notification entity [WPB-7425]

borichellow:
# What's new in this PR?

### Issues

Need to remove option "Reply" from the notifications in conversations that has Degraded verification status.

### Causes (Optional)

Wasn't implemented yet.

### Solutions

Update DB-query to get also `Conversation.degradedConversationNotified` while getting notifications and use it as `isReplyAllowed` flag in creating OS-notifications.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
@@ -34,6 +34,8 @@ import com.wire.kalium.logic.data.id.toDao
 import com.wire.kalium.logic.data.id.toModel
 import com.wire.kalium.logic.data.message.mention.MessageMentionMapper
 import com.wire.kalium.logic.data.notification.LocalNotification
+import com.wire.kalium.logic.data.notification.LocalNotificationMessageMapper
+import com.wire.kalium.logic.data.notification.LocalNotificationMessageMapperImpl
 import com.wire.kalium.logic.data.user.UserId
 import com.wire.kalium.logic.di.MapperProvider
 import com.wire.kalium.logic.failure.ProteusSendMessageFailure
@@ -253,7 +255,7 @@ internal interface MessageRepository {
 
 // TODO: suppress TooManyFunctions for now, something we need to fix in the future
 @Suppress("LongParameterList", "TooManyFunctions")
-internal class MessageDataSource internal constructor (
+internal class MessageDataSource internal constructor(
     private val selfUserId: UserId,
     private val messageApi: MessageApi,
     private val mlsMessageApi: MLSMessageApi,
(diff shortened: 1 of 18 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/LocalNotificationMessageMapper.kt
```diff
@@ -103,4 +111,21 @@ class LocalNotificationMessageMapperImpl : LocalNotificationMessageMapper {
...
+
+    override fun fromEntitiesToLocalNotifications(
+        list: List<NotificationMessageEntity>,
+        messageSizePerConversation: Int,
+        mapMessage: (NotificationMessageEntity) -> LocalNotificationMessage?
+    ): List<LocalNotification.Conversation> =
+        list.groupBy { it.conversationId }
+            .map { (conversationId, messages) ->
+                LocalNotification.Conversation(
+                    // todo: needs some clean up!
```
mchenani, line 123: question: is that todo valid anymore?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/LocalNotificationMessageMapper.kt
```diff
@@ -103,4 +111,21 @@ class LocalNotificationMessageMapperImpl : LocalNotificationMessageMapper {
...
+                    id = conversationId.toModel(),
+                    conversationName = messages.first().conversationName ?: "",
+                    messages = messages.take(messageSizePerConversation).mapNotNull { mapMessage(it) },
+                    isOneToOneConversation = messages.first().conversationType == ConversationEntity.Type.ONE_ON_ONE,
+                    isReplyAllowed = messages.first().degradedConversationNotified
```
mchenani, line 128: question: so if the conversation degraded we allow the reply? based on the logic I see here I think it should be the opposite?

borichellow: No, we allow reply when user was notified about degrading :) 
`degradedConversationNotified = true` means either user was notified about degraded conversation OR conversation is Verified.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/notification/LocalNotificationMessageMapper.kt
borichellow: I didn't add this `TODO` and just moved the whole mapping fun into mapper, so maybe it's still not clear ;P

vitorhugods (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
```diff
@@ -303,17 +305,10 @@ internal class MessageDataSource internal constructor (
...
-                    LocalNotification.Conversation(
-                        // todo: needs some clean up!
-                        id = conversationId.toModel(),
-                        conversationName = messages.first().conversationName ?: "",
-                        messages = messages.take(messageSizePerConversation)
-                            .mapNotNull { message -> messageMapper.fromMessageToLocalNotificationMessage(message) },
-                        isOneToOneConversation = messages.first().conversationType == ConversationEntity.Type.ONE_ON_ONE
-                    )
-                }
+            notificationMapper.fromEntitiesToLocalNotifications(
```
yamilmedina, line 308: Suggestion: maybe we can have a test for this func, since it's changing (I had a look, and it seems we never had one for this)

yamilmedina (approved):
Looking good, just a small suggestion for a test and would be ready to go.
I'll leave it ✅  for now

PR #2868, opened 2024-07-08
feat: add analytics user profile - Part 1 (WPB-8978)

alexandreferris:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8978" title="WPB-8978" target="_blank">TaskWPB-8978</a>  [Android] Countly analytics ID and user properties
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There was no common identifier between same user logged in different devices.

### Causes (Optional)

Not implemented.

### Solutions

This is only Part 1, which solves:
- Receiving the event for `DataTransfer` with tracking identifier
- Saving this value into user metadata table (saving it from Kalium and on a table because it will later be used in the frontend client, and if in the future Kalium is also used by other platforms, only a small tweak on frontend clients will need to be made for receiving/setting the current identifier - from sharedPreferences (Android) to this metadata table)

#### Test Coverage (Optional)

#### How to Test

- Run this branch with AR
- Login from AR and watch metadata table for logged in user
- Login from web (fresh login)
- Metadata table should now have an entry for `analytics_tracking_identifier`

### Notes (Optional)

This is only Part 1

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/handler/DataTransferEventHandler.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/handler/DataTransferEventHandler.kt
@@ -1,0 +1,52 @@
+/*
+ * Wire
+ * Copyright (C) 2024 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.kalium.logic.sync.receiver.handler
+
+import com.wire.kalium.logic.configuration.UserConfigRepository
+import com.wire.kalium.logic.data.message.Message
+import com.wire.kalium.logic.data.message.MessageContent
+import com.wire.kalium.logic.data.user.UserId
+import com.wire.kalium.logic.kaliumLogger
+
+internal interface DataTransferEventHandler {
+    suspend fun handle(
+        message: Message.Signaling,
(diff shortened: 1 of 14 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/handler/DataTransferEventHandler.kt
```diff
@@ -0,0 +1,52 @@
...
+internal class DataTransferEventHandlerImpl(
+    private val selfUserId: UserId,
+    private val userConfigRepository: UserConfigRepository
+) : DataTransferEventHandler {
+
+    override suspend fun handle(
+        message: Message.Signaling,
+        messageContent: MessageContent.DataTransfer
+    ) {
+        kaliumLogger.d("MessageContent.DataTransfer | with Identifier : ${messageContent.trackingIdentifier?.identifier} |end|")
```
saleniuk, line 42: question: shouldn't the identifier be obfuscated in logs? 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/handler/DataTransferEventHandler.kt
```diff
@@ -0,0 +1,52 @@
...
+        // DataTransfer from another user or null tracking identifier shouldn't happen,
+        // If it happens, it's unnecessary
+        // and we can squish some performance by skipping it completely
+        if (message.senderUserId != selfUserId || messageContent.trackingIdentifier == null) return
```
saleniuk, line 46: suggestion: maybe add some info to the logs whether it was handled or ignored or maybe just move creating a log from line 42 to here so that we get the log only when this event is handled

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/handler/DataTransferEventHandler.kt
alexandreferris: Oh yes! this log is moved to only after the if logic on part2 that I'm still working on

alexandreferris: So, if you don't mind (as this is targeting an epic) it will have the changes on Part 2

saleniuk (approved):

Garzas (approved):

PR #2869, opened 2024-07-09
fix: Allow creating empty MLS GroupConversation [WPB-7099]

borichellow:
# What's new in this PR?

### Issues

When user trying to create a new MLS GroupConversation only with users that has no KeyPackages, then conversation is not created and error is displayed. 
We should allow such a scenario and just create GroupConversation with only 1 (self) user.

### Causes (Optional)

Was not implemented (not obvious requirement)

### Solutions

Just do it.

In `KeyPackageDataSource.claimKeyPackages` remove checking if all the users were failed  (and returning `MissingKeyPackages` in that case). 
So `MLSConversationRepository` will take care about returning such an error in case when partial MemberList is not allowed.


vitorhugods (approved):

MohamadJaara (approved):
LGTM

PR #2875, opened 2024-07-12
refactor: create a data module [WPB-8645]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8645" title="WPB-8645" target="_blank">EpicWPB-8645</a>  [Android] Infrastructure code and developer experience
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

If we want to develop a smaller library that can be used by Kalium, but also deployed in a standalone format, we can't add the current classes that exist inside `logic`.

During the hackaton, for example, we created a module called `backup`, and we wanted to:
1. Expose this module so iOS could use it
2. Add this module to logic, so Kalium/Android could use it

We wanted to use classes like ConversationId, UserId, Message, Content, etc. within the `backup` module. But this would cause a circular dependency.

### Causes

The `logic` module is _far too big_. We should modularise it.

### Solutions

Move _most_ of the `data` classes to their own `data` module. Not the perfect solution, but good enough for now.

Some were still left behind, as they were relying on `cryptography`, which shouldn't happen!

Only adjustments necessary were creating local properties when dealing with polymorphism and nullability checks.

### Testing

It builds. Tests pass. Can't go better than that.

----

----
##### References


data/src/commonMain/kotlin/com/wire/kalium/logic/data/web/WebEventContent.kt
```diff
@@ -53,63 +53,71 @@ data class WebConversationContent(
...
+            @SerialName("qualified_from") override val qualifiedFrom: UserId?,
             @SerialName("from") val from: String,
             @SerialName("from_client_id") val fromClientId: String?,
             @SerialName("time") val time: String,
             @SerialName("id") val id: String,
             @SerialName("data") val data: WebTextData,
-            @SerialName("reactions") val reactions: Map<String, String>?
-        ) : Conversation()
+            @SerialName("reactions") val reactions: Map<String, String>?,
+            @SerialName("category") val category: Int? // 16 ?
```
alexandreferris, line 87: where did this come from? 🤔

alexandreferris (approved):
Awesome! Great seeing we are moving stuff out of logic module into its own modules 😍 

Just got a tiny question but it doesn't block its flow of approval :)

data/src/commonMain/kotlin/com/wire/kalium/logic/data/web/WebEventContent.kt
vitorhugods: Oooh.

All of the `WebEventContent` shouldn't be in this PR.

That came by mistake when taking stuff from the Hackaton :D

I'll remove these.

vitorhugods: Done!

data/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientModel.kt
```diff
@@ -18,26 +18,11 @@
...
-    val deviceType: DeviceType?,
-    val label: String?,
-    val capabilities: List<ClientCapability>?,
-    val clientType: ClientType?,
-    val model: String?,
-    val cookieLabel: String?,
-    val secondFactorVerificationCode: String? = null,
-    val modelPostfix: String? = null
-)
-
```
vitorhugods, line 40: Because this uses `PreKeyCrypto`, I left it in `logic/data`.

data/src/commonMain/kotlin/com/wire/kalium/logic/data/message/AssetContent.kt
```diff
@@ -18,8 +18,6 @@
 
 package com.wire.kalium.logic.data.message
 
-import com.wire.kalium.logic.sync.receiver.conversation.message.hasValidRemoteData
```
vitorhugods, line 21: This was just in a very weird package.

Doesn't make sense to import things from `conversation / receiver` in order to check if asset content has valid remote data

mchenani (approved):

PR #2880, opened 2024-07-16
fix: testservice not closing websocket due to wrong logout (WPB-9030)

mythsunwind:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The issue is that the testservice does not allow new logins of accounts after running for a while. The login timed out and did not directly show an error but was just not logging the user in and seems to wait. The machine showed over 220 established connections to the backend even thought no instance was running.

### Causes (Optional)

It looks like the websocket was not closed when instances were destroyed because of using the wrong logout method (which usually triggers closing of the websocket). There is a limit of open websockets either on library side or on system side.

### Solutions

The solution here is to use `logout(reason = LogoutReason.SELF_HARD_LOGOUT, waitUntilCompletes = true)` which triggers a close of the websocket. Not sure if this closing always will happen even if the logout process fails but my testing shows that no connections stay established after it was fired.


yamilmedina (approved):
👏

PR #2932, opened 2024-08-05
fix: Validate other members UserName and DisplayName in E2EI [WPB-10402]

borichellow:
# What's new in this PR?

### Issues

The checking "if user has valid E2EI" was not full in Participants list and UserProfile.
We should also check if users `displayName` and `handle` are the same as in E2EI certificate.

### Causes (Optional)

Was not implemented.

### Solutions

Implement it and add/update tests for new scenarios.


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/FetchMLSVerificationStatusUseCase.kt
```diff
@@ -114,13 +114,13 @@ internal class FetchMLSVerificationStatusUseCaseImpl(
                 // check that all identities are valid and name and handle are matching
                 for ((userId, wireIdentity) in ccIdentity) {
                     val persistedMemberInfo = dbData.members[userId]
-                    val isUserVerified = wireIdentity.firstOrNull {
+                    val isUserVerified = wireIdentity.none {
```
mchenani, line 117: nice!

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/GetUserE2EICertificateUseCase.kt
```diff
@@ -31,12 +33,13 @@ interface IsOtherUserE2EIVerifiedUseCase {
...
 ) : IsOtherUserE2EIVerifiedUseCase {
     override suspend operator fun invoke(userId: UserId): Boolean =
         if (isE2EIEnabledUseCase()) {
-            mlsConversationRepository.getUserIdentity(userId).fold({ false }, { it.isUserMLSVerified() }
-            )
+            val nameHandle = userRepository.getNameAndHandler(userId).getOrNull()
+            mlsConversationRepository.getUserIdentity(userId).fold({ false }, { it.isUserMLSVerified(nameHandle) })
         } else {
             false
         }
```
mchenani, line 45: suggestion:
`
override suspend operator fun invoke(userId: UserId): Boolean =
    isE2EIEnabledUseCase() && 
    userRepository.getNameAndHandler(userId).getOrNull()?.let { nameHandle ->
        mlsConversationRepository.getUserIdentity(userId).fold(
            onSuccess = { it.isUserMLSVerified(nameHandle) },
            onFailure = { false }
        )
    } ?: false
`

mchenani (approved):
LGTM, and one suggestion 🙌

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/GetUserE2EICertificateUseCase.kt
borichellow: I tried even shorter:
```
override suspend operator fun invoke(userId: UserId): Boolean =
    isE2EIEnabledUseCase() && userRepository.getNameAndHandler(userId).flatMap { nameHandle ->
        mlsConversationRepository.getUserIdentity(userId).map { it.isUserMLSVerified(nameHandle) }
    }.getOrElse(false)
```
but i don't like it. Yes it's 3 lines instead of 6, but IMO `if else` is more readable and easier to understand in that case :)

mchenani (approved):

persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/UserDAO.kt
```diff
@@ -306,4 +307,5 @@ interface UserDAO {
     suspend fun upsertConnectionStatuses(userStatuses: Map<QualifiedIDEntity, ConnectionEntity.State>)
     suspend fun isAtLeastOneUserATeamMember(userId: List<UserIDEntity>, teamId: String): Boolean
     suspend fun getOneOnOnConversationId(userId: UserIDEntity): QualifiedIDEntity?
+    suspend fun getNameAndHandler(userId: UserIDEntity): NameAndHandleEntity?
```
alexandreferris, line 310:
```suggestion
    suspend fun getNameAndHandle(userId: UserIDEntity): NameAndHandleEntity?
```

Maybe we keep as `getNameAndHandle`? without the `r` at the end, otherwise it can look like it will return (despite having the specific return type) a name and a handler class?

alexandreferris (approved):
Looking good! I just have a tiny suggestion on naming in `UserDAO`

PR #2950, opened 2024-08-15
fix: Check file extension instead of mimeType [WPB-10605]

borichellow:
# What's new in this PR?

### Issues

in `ValidateAssetMimeTypeUseCase` when we calculate if file can be shared or not we were comparing files mimeType to list of allowed Extensions. 

### Causes (Optional)

Just was implemented in that way long time ago. 

### Solutions

fix it and compare files extension to  list of allowed Extensions.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
@@ -133,7 +133,7 @@ internal class ScheduleNewAssetMessageUseCaseImpl(
                 FileSharingStatus.Value.EnabledAll -> { /* no-op*/
                 }
 
-                is FileSharingStatus.Value.EnabledSome -> if (!validateAssetMimeTypeUseCase(assetMimeType, it.state.allowedType)) {
+                is FileSharingStatus.Value.EnabledSome -> if (!validateAssetMimeTypeUseCase(assetName, it.state.allowedType)) {
                     kaliumLogger.e("The asset message trying to be processed has invalid content data")
                     return ScheduleNewAssetMessageResult.Failure.RestrictedFileType
                 }
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
@@ -110,7 +110,7 @@ internal class ScheduleNewAssetMessageUseCaseImpl(
     private val selfDeleteTimer: ObserveSelfDeletionTimerSettingsForConversationUseCase,
     private val scope: CoroutineScope,
     private val observeFileSharingStatus: ObserveFileSharingStatusUseCase,
-    private val validateAssetMimeTypeUseCase: ValidateAssetMimeTypeUseCase,
+    private val validateAssetFileUseCase: ValidateAssetFileTypeUseCase,
     private val dispatcher: KaliumDispatcher,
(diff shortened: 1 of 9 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ValidateAssetMimeTypeUseCase.kt
```diff
@@ -18,19 +18,22 @@
 package com.wire.kalium.logic.feature.asset
 
 /**
- * Returns true if the mime type is allowed and false otherwise.
- * @param mimeType the mime type to validate.
+ * Returns true if the file extension is present in file name and is allowed and false otherwise.
+ * @param fileName the file name (with extension) to validate.
  * @param allowedExtension the list of allowed extension.
  */
 interface ValidateAssetMimeTypeUseCase {
```
yamilmedina, line 25:
```suggestion
interface ValidateAssetFileTypeUseCase {
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ScheduleNewAssetMessageUseCase.kt
```diff
@@ -133,7 +133,7 @@ internal class ScheduleNewAssetMessageUseCaseImpl(
                 FileSharingStatus.Value.EnabledAll -> { /* no-op*/
                 }
 
-                is FileSharingStatus.Value.EnabledSome -> if (!validateAssetMimeTypeUseCase(assetMimeType, it.state.allowedType)) {
+                is FileSharingStatus.Value.EnabledSome -> if (!validateAssetMimeTypeUseCase(assetName, it.state.allowedType)) {
```
MohamadJaara, line 136: can we change the name of the user case to reflect what it does now

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/ValidateAssetMimeTypeUseCase.kt
```diff
@@ -18,19 +18,22 @@
...
-    operator fun invoke(mimeType: String, allowedExtension: List<String>): Boolean
+    operator fun invoke(fileName: String, allowedExtension: List<String>): Boolean
 }
 
 internal class ValidateAssetMimeTypeUseCaseImpl : ValidateAssetMimeTypeUseCase {
-    override operator fun invoke(mimeType: String, allowedExtension: List<String>): Boolean {
-        val extension = mimeType.split("/").last().lowercase()
-        return allowedExtension.any {
-            it.lowercase() == extension
+    override operator fun invoke(fileName: String, allowedExtension: List<String>): Boolean {
```
yamilmedina, line 30: Question: What if we have multiple extensions hiding in the file or other type of tricks ? Would be good idea to follow some of these tips ? 

https://book.hacktricks.xyz/pentesting-web/file-upload#bypass-file-extensions-checks

yamilmedina (commented):
Looking good :D left some questions and suggestions 🙌

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/asset/AssetMessageHandler.kt
```diff
@@ -53,7 +53,7 @@ internal class AssetMessageHandlerImpl(
                 FileSharingStatus.Value.EnabledAll -> true
 
                 is FileSharingStatus.Value.EnabledSome -> validateAssetMimeTypeUseCase(
-                    messageContent.value.mimeType,
+                    messageContent.value.name ?: "",
```
MohamadJaara, line 56: to avoid any hidden bugs we can pass the name as nullable string and handle the null case by returning not valid

yamilmedina (approved):

MohamadJaara (approved):

PR #2974, opened 2024-08-28
perf: improve ProteusClientCoreCrypto performance [WPB-10000]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10000" title="WPB-10000" target="_blank">SpikeWPB-10000</a>  [Android] Investigate the text last read event handling bottleneck and spickes
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We're a bit slow to process events. 

### Causes

In part, it's nature of E2EE. In part, we are not really being optimal.

In the generic sense, we can improve Proteus/CoreCrypto performance by not saving session (which is unnecessary, as CoreCrypto already does it during decryption), and also avoid checking if the session exists over and over again if we already know it exists.

### Solutions

1. Remove the "saveSession" bit of `ProteusClientCoreCrypto`.
2. Cache known sessions in-memory and check that instead of relying only on CoreCrypto, which uses IO in order to do that

Document my findings in [this notebook](https://github.com/wireapp/kalium/blob/5f1b388245be17bc48b4895522dc441c459ae153/Notebooks/EventPerformanceImprovements/PerformanceImprovements.ipynb).

----

----
##### References


alexandreferris (approved):

MohamadJaara (approved):
this change is a nice candidate for a ADR, just saying it does not need to be the whole notebook, but maybe small description with a link to the notebook

Notebooks/README.md
```diff
@@ -0,0 +1,3 @@
+A place to store [Jupyter Notebook](https://jupyter.org/) files of ongoing studies.
```
yamilmedina, line 1: Suggestion/question: Can we maintain the patterns here for naming folders ? 😬 
Meaning, Is there a reason why is this different from the rest ? Didn't find anything on the docs

yamilmedina (changes requested):
Great work and nice introduction of these `notebooks`
Just left one suggestion/question about naming and breaking patterns. 

Also agree with Mo, this is a good candidate for an ADR, maybe for a next PR 🤔

vitorhugods:
> Great work and nice introduction of these `notebooks` Just left one suggestion/question about naming and breaking patterns.
> 
> Also agree with Mo, this is a good candidate for an ADR, maybe for a next PR 🤔

About ADRs... do we have ADRs on Kalium as well? Because this was the main reason that made me not do it. I am starting to think we should merge the repos more and more.

----

I just did `Notebooks` instead of `notebooks` because it is indeed different from the rest. Maybe `docs` would be better. I'll rename it to something like that.

yamilmedina (approved):

PR #2983, opened 2024-09-02
fix(mls-client): skip MLSClient registration if it has registered

mchenani:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?
Check if the MLS Client has registered, then skip the registration.

### Issues
 When the user logs out without clearing the data, we were trying to register the exist MLSClient again; that lead to an error!

### Solutions

If the client has registered before and the user log in again to continue on a previously established session no need to register the client again and simply skip the api call.

#### Test Coverage (Optional)

#### How to Test

Have a user with MLS Enabled on their team
1- Log in and log out without clearing the data
2- When you try to log in again with the same use, you should be able to proceed successfully

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
@@ -36,6 +36,7 @@ import com.wire.kalium.logic.functional.flatMap
 import com.wire.kalium.logic.functional.map
 import com.wire.kalium.logic.functional.mapLeft
 import com.wire.kalium.logic.functional.onSuccess
+import com.wire.kalium.logic.functional.right
 import com.wire.kalium.logic.kaliumLogger
 import com.wire.kalium.logic.wrapApiRequest
 import com.wire.kalium.logic.wrapStorageRequest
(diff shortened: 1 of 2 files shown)


yamilmedina (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/ClientRepository.kt
```diff
@@ -214,13 +215,16 @@ class ClientDataSource(
         clientId: ClientId,
         publicKey: ByteArray,
         cipherSuite: CipherSuite
-    ): Either<CoreFailure, Unit> =
+    ): Either<CoreFailure, Unit> = if (clientRegistrationStorage.hasRegisteredMLSClient()) {
```
MohamadJaara, line 218: does `clientRegistrationStorage.hasRegisteredMLSClient` take into account if the registered client still exists remotely, meaning if my app session expired and during that time the client was removed remotly, will thisfunction still return true?

mchenani: That's another edge case, if the use was logged out without clearing the data, and delete the client from another device and afterwards tries to login on this device we can't detect it.
What we can do:
Currently if the client is registered, we get this error from the server API:
`InvalidRequestError(errorResponse=ErrorResponse(code=400, message=MLS public key for the given signature scheme already exists, label=mls-duplicate-public-key))`
We can instead of checking the local flag always try to register the client against the server, if we get that error then we ignore it.

Wdyt?

MohamadJaara: For Proteus, we already check if the local client still exists remotely or not, I suggest MLS clients follow the same path as Proteus

mchenani:
Closing this: decided to go with another solution https://github.com/wireapp/kalium/pull/2987

PR #2989, opened 2024-09-03
feat: Save code for EncryptionError [WPB-1795]

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1795" title="WPB-1795" target="_blank">StoryWPB-1795</a>  Display Decryption error message according to Design
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

We want to show error code in DecryptionError message. But it's never saved.

### Causes (Optional)

wasn't implemented.

### Solutions

Update `MessageFailedToDecryptContent` DB table and corresponding data classes to store that code and pass it forward to UI.


cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/exceptions/ProteusException.kt
```diff
@@ -18,9 +18,20 @@
...
 package com.wire.kalium.cryptography.exceptions
 
-class ProteusException(message: String?, val code: Code, cause: Throwable? = null) : Exception(message, cause) {
+class ProteusException(message: String?, val code: Code, val intCode: Int, cause: Throwable? = null) : Exception(message, cause) {
 
-    constructor(message: String?, code: Int, cause: Throwable? = null) : this(message, fromNativeCode(code), cause)
+    constructor(message: String?, code: Code, cause: Throwable? = null) : this(
+        message,
+        code,
+        toProteusCode(code),
```
MohamadJaara, line 26: can we avoid mapping from int code to proteus code and then back to code, by using the int code constructor and then map from int to Code

MohamadJaara (changes requested):

cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/exceptions/ProteusException.kt
```diff
@@ -189,5 +200,24 @@ class ProteusException(message: String?, val code: Code, cause: Throwable? = nul
...
+                Code.INVALID_MESSAGE -> 200
+                Code.DUPLICATE_MESSAGE -> 209
+                Code.TOO_DISTANT_FUTURE -> 211
+                Code.OUTDATED_MESSAGE -> 208
+                Code.IDENTITY_ERROR -> 300
+                Code.PREKEY_NOT_FOUND -> 101
+                Code.PANIC -> 5
+                else -> -1
+            }
+        }
```
MohamadJaara, line 221: this will not be needed since we get the code directly from the native lib

cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/exceptions/ProteusException.kt
borichellow: Actually we can't. 
If you check mappers, you'll find out that `IntCode` that comes from native might be `1,2,3...15`, on that other hand `IntCode` from Proteus: `201, 501, 310, 3, 303...` so these are different ints :)

MohamadJaara: it is fine we can just take the native code as it is and store it, this will always be correct, while over time the mapping can be wrong

MohamadJaara (approved):

saleniuk (approved):

PR #3017, opened 2024-09-17
fix: use server time for calling (WPB-10979)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10979" title="WPB-10979" target="_blank">BugWPB-10979</a>  [Android] No call notifications incoming on graphene device on bund-next-col1 backend
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

No call notifications incoming if device time is not correct

### Causes (Optional)

If device time is different from the server, AVS will return `shouldRing` flag as false causing the app to not show any incoming call notification

### Solutions

On event processing we calculate time offset that will be used later to get server time

Edit:
I changed the logic so now we get serverTime by fetching `/notifications` api with only `querySize` value.

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/util/ServerTimeHandler.kt
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.kalium.logic.util
+
+import kotlinx.datetime.Clock
+
+object ServerTimeHandler {
+    private var timeOffset: Long = 0
```
yamilmedina, line 23: Question: Would this mutable var from the singleton, handle correctly the timezone changes, ie: daylight saving or traveling ? (in case we don't restart the app) food for thought

I have the feeling since we are recalculating every time we receive an event is not big of a deal. But mutable vars on a singletons, sometimes can go wrong

ohassine: Yes and that's the point of recalculating frequently the timeOffset, to not become outdated and handle device's timezone changes.

logic/src/commonMain/kotlin/com/wire/kalium/logic/util/ServerTimeHandler.kt
```diff
@@ -0,0 +1,46 @@
...
+
```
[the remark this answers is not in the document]
ohassine: done

logic/src/commonMain/kotlin/com/wire/kalium/logic/util/ServerTimeHandler.kt
vitorhugods: Super silly thing, maybe, but:
I wonder what if there are two accounts on the same device, and they use two different servers, and there's a mismatch between these servers 😅
Maybe it would be best if it wasn't a singleton, but rather one instance per user.

yamilmedina (approved):

MohamadJaara (approved):

PR #3024, opened 2024-09-24
fix: Catch excecption on Asset uploading [WPB-10700]

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10700" title="WPB-10700" target="_blank">BugWPB-10700</a>  [Android] file not found when sending assets
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

`FileNotFoundException` on sending asset message. 

### Causes (Optional)

Couldn't reproduce it, or find out the actual reason. 

### Solutions

Wrap it to `try catch` to at least not crash the app.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetRepository.kt
@@ -156,29 +156,33 @@ internal class AssetDataSource(
         otrKey: AES256Key,
         extension: String?
     ): Either<CoreFailure, Pair<UploadedAssetId, SHA256Key>> {
+        try {
+            val tempEncryptedDataPath = kaliumFileSystem.tempFilePath("${assetDataPath.name}.aes")
+            val assetDataSource = kaliumFileSystem.source(assetDataPath)
+            val assetDataSink = kaliumFileSystem.sink(tempEncryptedDataPath)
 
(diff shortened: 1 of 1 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/asset/AssetRepository.kt
```diff
@@ -156,29 +156,33 @@ internal class AssetDataSource(
...
+            return if (encryptionSucceeded) {
+                val uploadAssetData = UploadAssetData(tempEncryptedDataPath, encryptedDataSize, mimeType, false, RetentionType.PERSISTENT)
+                uploadAndPersistAsset(uploadAssetData, assetDataPath, extension).map { it to SHA256Key(sha256!!) }
+            } else {
+                kaliumLogger.e("Something went wrong when encrypting the Asset Message")
+                Either.Left(EncryptionFailure.GenericEncryptionError)
+            }
+        } catch (e: IOException) {
+            kaliumLogger.e("Something went wrong when uploading the Asset Message. $e")
+            return Either.Left(EncryptionFailure.GenericEncryptionError)
```
MohamadJaara, line 185: here you can return a generic error and pass the exception to it

```suggestion
            return Either.Left(CoreFailure.Unknown(e))
```

yamilmedina (approved):

MohamadJaara (approved):

PR #3027, opened 2024-09-24
fix: delete asset file when removing self-deleting msg [WPB-1807]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1807" title="WPB-1807" target="_blank">StoryWPB-1807</a>  Self-deleting: Delete assets from internal storage once the timer runs out
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Asset file is not removed from the storage once the self-deleting asset message timer runs out.

### Causes (Optional)

Wrong order of actions:
- the message is marked as deleted, which means that the content is cleared in db
- attempt to send "deleted" messages
- remove message completely from the db
- try to remove the asset from the db and its file from the storage - at this point, because of the first step, the content of the message is already empty so the app doesn't know the id of the asset to remove and no file is being removed. Also, the app tries to remove the asset remotely by making a request which ends up with 403 error, probably because this asset belongs to someone other. 

### Solutions

Change the order of actions, so that now it will look like this:
- get the content of the message
- mark the message as deleted and clear the content from the db
- use the retrieved content data from first step to remove asset, but only locally, together with deleting associated file from the storage
- attempt to send "deleted" messages
- remove message completely from the db
Thanks to that, the asset file and asset data will be removed from the asset table right when the content of the message is being cleared and marked as deleted. Asset is being removed only locally, there's no need to make a request that will probably fail.
Also, `selectPendingEphemeralMessages` is fixed to return only real pending messages, so messages that has yet not started (messages without end time) should be filtered out, to make sure that only messages that has been shown to the user are enqueued.

#### Test Coverage (Optional)

#### How to Test

Send a self-deleting message with asset, wait until that message disappears and check if "used storage" has changed.

### Attachments (Optional)

| Before | After |
| ----------- | ------------ |
| <video src="https://github.com/user-attachments/assets/91603b9f-6abc-44ce-aa83-26ccbb4aa967"/> | <video src="https://github.com/user-attachments/assets/469e4e5d-dddc-48de-ac37-54073c5e9f84"/> |

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/ephemeral/DeleteEphemeralMessageForSelfUserAsReceiverUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/ephemeral/DeleteEphemeralMessageForSelfUserAsReceiverUseCase.kt
@@ -62,30 +62,34 @@ internal class DeleteEphemeralMessageForSelfUserAsReceiverUseCaseImpl(
     private val selfUserId: UserId,
     private val selfConversationIdProvider: SelfConversationIdProvider
 ) : DeleteEphemeralMessageForSelfUserAsReceiverUseCase {
+
     override suspend fun invoke(conversationId: ConversationId, messageId: String): Either<CoreFailure, Unit> =
-        messageRepository.markMessageAsDeleted(messageId, conversationId).flatMap {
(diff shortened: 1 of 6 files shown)


vitorhugods (approved):
Nicely done 💅

Garzas (approved):

PR #3045, opened 2024-09-27
fix: Active speaker highlighting [WPB-11040]

borichellow:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11040" title="WPB-11040" target="_blank">BugWPB-11040</a>  [Android] Active speaker not indicated anymore on latest beta
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

Highlighting of speaking user during a call doesn't work.

### Causes (Optional)

Small mistake in refactoring: 
when we want to know if some participant is speaking during a call we need to check if that participant (pair of userId and clientId) is in a `activeSpeakers` list. The problem is that we were using `participantId` instead of `userId`.

### Solutions

fix the mistake


vitorhugods (commented):
There could be a couple of tests for this.


// Given
Create a CallMetadata with some users. Put some in `ActiveSpeakers`.

// When
Call `getFullParticipants`.

// Then
Only the active speakers have `isSpeaking = true`


-----

// Given
Create a CallMetadata with some users. 
Put some in `ActiveSpeakers`.
Mute some of these.

// When
Call `getFullParticipants`.

// Then
Only the active speakers **not muted** have `isSpeaking = true`

MohamadJaara (approved):

vitorhugods (approved):

PR #3054, opened 2024-10-08
fix: read receipts are not sent when the app is in background (WPB-8756)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8756" title="WPB-8756" target="_blank">BugWPB-8756</a>  [Android] Read receipts not generated when responding from notification popup
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Read receipts are not being sent when the app is in the background.

### Causes (Optional)

In `SendConfirmationUseCase` we wait until app is live to send confirmations.

### Solutions

Add a flag to the use case to ignore this wait if the user replies from notification.
 
Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/UpdateConversationReadDateUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/UpdateConversationReadDateUseCase.kt
@@ -68,11 +68,18 @@ class UpdateConversationReadDateUseCase internal constructor(
      * @param conversationId The conversation id to update the last read date.
      * @param time The last read date to update.
      */
-    operator fun invoke(conversationId: QualifiedID, time: Instant) {
-        workQueue.enqueue(ConversationTimeEventInput(conversationId, time), worker)
+    operator fun invoke(
+        conversationId: QualifiedID,
+        time: Instant,
+        shouldWaitUntilLive: Boolean = true
+    ) {
+        workQueue.enqueue(
+            ConversationTimeEventInput(conversationId, time, shouldWaitUntilLive),
+            worker
+        )
     }
 
-    private val worker = ConversationTimeEventWorker { (conversationId, time) ->
+    private val worker = ConversationTimeEventWorker { (conversationId, time, shouldWaitUntilLive) ->
         coroutineScope {
             conversationRepository.observeConversationById(conversationId).first().onFailure {
                 logger.w("Failed to update conversation read date; StorageFailure $it")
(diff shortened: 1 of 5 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/UpdateConversationReadDateUseCase.kt
```diff
@@ -68,11 +68,18 @@ class UpdateConversationReadDateUseCase internal constructor(
      * @param conversationId The conversation id to update the last read date.
      * @param time The last read date to update.
      */
-    operator fun invoke(conversationId: QualifiedID, time: Instant) {
-        workQueue.enqueue(ConversationTimeEventInput(conversationId, time), worker)
+    operator fun invoke(
+        conversationId: QualifiedID,
+        time: Instant,
+        shouldWaitUntilLive: Boolean = true
```
MohamadJaara, line 74: not sure if this is the right way to solve it, since this use case will send an encrypted message, and messages should only be sent if the app is live, to avoid breaking the session

MohamadJaara: maybe the better approach is to handle it the same way we do when FCM message is received, start incremental sync and send the message when we are live then drop after ?

MohamadJaara (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/UpdateConversationReadDateUseCase.kt
ohassine: I don't see any reason to redo the sync for nothing because the app is already up to date, all messages are synced after getting the event through FCM. 
The user in this case just need to reply from notification without waiting for app to complete sync.

borichellow: but user might reply from notification with some delay, and during that delay some changes (new messages) might happen. For example: user receive notification -> lose connection for 5 minutes -> some messages comes to that conversation, but there is no connection, so the client doesn't know about it -> client goes online and sends reply instantly without sync

PR #3056, opened 2024-10-10
fix: event delay handling [WPB-11490]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11490" title="WPB-11490" target="_blank">BugWPB-11490</a>  Text messages are stuck in handling when there is unstable internet connection
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The use of Mutex in `ConfirmationDeliveryHandlerImpl` led to blocking and concurrency issues when handling message confirmations. This caused message processing to potentially get stuck, especially during high-load scenarios where multiple messages were queued for delivery.

### Causes (Optional)

- The `Mutex` was synchronizing access to the shared map (pendingConfirmationMessages), which led to contention and potential deadlocks, especially when the sendPendingConfirmations coroutine tried to acquire the lock while other coroutines were working with the same messages.
- The original flow of event processing and handling relied on multiple nested flows, which was less efficient and harder to maintain.

### Solutions

- Removed the `Mutex` and replaced the shared map with `ConcurrentMutableMap` to handle concurrent operations in a non-blocking way.
- Extracted the message sending logic into a new use case, SendDeliverSignalUseCase, to improve modularity and separate the concerns of message queue management and sending.
- Refactored the event flow processing (EventRepository) to simplify event handling by using a single flow with collect and emit.
- Improved message removal after sending to ensure that only successfully sent messages are removed from the pending queue.

Added and updated unit tests to verify:
- Proper message removal after sending.
- Separation of message queue management and message delivery logic.
- Handling of concurrency issues using ConcurrentMutableMap.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
@@ -22,27 +22,24 @@ import com.wire.kalium.logic.CoreFailure
 import com.wire.kalium.logic.NetworkFailure
 import com.wire.kalium.logic.StorageFailure
 import com.wire.kalium.logic.data.conversation.ClientId
+import com.wire.kalium.logic.data.id.CurrentClientIdProvider
 import com.wire.kalium.logic.data.user.UserId
 import com.wire.kalium.logic.di.MapperProvider
-import com.wire.kalium.logic.data.id.CurrentClientIdProvider
 import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.logic.functional.flatMap
 import com.wire.kalium.logic.functional.fold
 import com.wire.kalium.logic.functional.map
 import com.wire.kalium.logic.wrapApiRequest
 import com.wire.kalium.logic.wrapStorageRequest
-import com.wire.kalium.network.api.base.authenticated.notification.NotificationApi
 import com.wire.kalium.network.api.authenticated.notification.NotificationResponse
+import com.wire.kalium.network.api.base.authenticated.notification.NotificationApi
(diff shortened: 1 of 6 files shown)


yamilmedina (approved):

borichellow (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
```diff
@@ -104,26 +101,31 @@ class EventDataSource(
...
-                    is WebSocketEvent.Close -> {
-                        flowOf(WebSocketEvent.Close(webSocketEvent.cause))
-                    }
-
-                    is WebSocketEvent.BinaryPayloadReceived -> {
-                        eventMapper.fromDTO(webSocketEvent.payload, true).asFlow().map { WebSocketEvent.BinaryPayloadReceived(it) }
+        wrapApiRequest { notificationApi.listenToLiveEvents(clientId.value) }.map { webSocketEventFlow ->
+            flow {
+                webSocketEventFlow.collect { webSocketEvent ->
+                    when (webSocketEvent) {
```
ohassine, line 107: Why using a flow builder and collect data inside it and not just map the flow as we currently do ?

ohassine (dismissed):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
Garzas: Reverted it and added `flatMapConcat` on top of function

PR #3063, opened 2024-10-16
fix: Asset message resend transfer status #WPB-11035

m-zagorski:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11035" title="WPB-11035" target="_blank">BugWPB-11035</a>  [Android] No error displayed when sending imaged or gifs without internet being availble
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-11035

# What's new in this PR?

### Issues

When we resended asset message, we're not updating its status to uploaded hence the message was stuck with `Uploading` indicator forever

### Solutions

Change the status to Uploaded before trying to again send the message

#### How to Test

- Send any asset message while not having connection, turn on connection again, tap resend, message arrive on the second end and should not be in `Uploading` state

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/RetryFailedMessageUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/RetryFailedMessageUseCase.kt
@@ -178,7 +178,10 @@ class RetryFailedMessageUseCase internal constructor(
 
             else -> handleError("Asset message with transfer status $assetTransferStatus cannot be retried")
         }
-            .onSuccess { retrySendingMessage(it) }
+            .onSuccess {
+                updateAssetMessageTransferStatus(AssetTransferStatus.UPLOADED, message.conversationId, message.id)
+                retrySendingMessage(it)
+            }
             .map { /* returns Unit */ }
     }
 
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/RetryFailedMessageUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/RetryFailedMessageUseCase.kt
@@ -160,6 +160,7 @@ class RetryFailedMessageUseCase internal constructor(
                                 .map { updatedMessage } // we need to persist new asset remoteData and status
                         }
                     }
+                    .onSuccess { updateAssetMessageTransferStatus(AssetTransferStatus.UPLOADED, message.conversationId, message.id) }
                     .onFailure {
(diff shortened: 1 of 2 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/RetryFailedMessageUseCase.kt
```diff
@@ -178,7 +178,10 @@ class RetryFailedMessageUseCase internal constructor(
 
             else -> handleError("Asset message with transfer status $assetTransferStatus cannot be retried")
         }
-            .onSuccess { retrySendingMessage(it) }
+            .onSuccess {
+                updateAssetMessageTransferStatus(AssetTransferStatus.UPLOADED, message.conversationId, message.id)
```
saleniuk, line 182: Question: should the status be changed to `UPLOADED` before the message is actually uploaded? Maybe `UPLOAD_IN_PROGRESS` would be more appropriate and change to `UPLOADED` only after the successful resend 🤔

m-zagorski: I thought about that - but isnt that status only regarding the asset? We have the additional status for message `PENDING` etc. Also if we set up this status like that, we can see:
- Error sending
- Uploading
- Uploaded
- Message pending (until its sent)
- If there will be any connection problem the message will get back to failed state

borichellow: IMO, few line up is better place for this updating. 

```
return when (assetTransferStatus) {
    AssetTransferStatus.FAILED_UPLOAD, AssetTransferStatus.NOT_DOWNLOADED -> {
        updateAssetMessageTransferStatus(AssetTransferStatus.UPLOAD_IN_PROGRESS, message.conversationId, message.id)
        retryUploadingAsset(content)
            ...
            .onSuccess { updateAssetMessageTransferStatus(AssetTransferStatus.UPLOADED, message.conversationId, message.id) }
```

in that case we won't update status if it's already `UPLOADED` (less DB-queries - better :) )

borichellow (approved):
just small suggestion, other then that is great 👍

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/RetryFailedMessageUseCase.kt
m-zagorski: You are right! thanks, changed :)

saleniuk (approved):

PR #3087, opened 2024-11-04
feat: Create use case for calling activated status #WPB-1826

m-zagorski:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1826" title="WPB-1826" target="_blank">StoryWPB-1826</a>  Implement correct alert for non paying team members
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-1826

# What's new in this PR?

### Issues

We needed a way to know when we've updated from being unable to start conference call to when we're able, so when we have upgraded to enterprise.

### Solutions

We observe current value and zip it with the upcoming one to know if we have moved from false->true

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
@@ -93,7 +93,7 @@ interface UserConfigRepository {
     suspend fun getSupportedProtocols(): Either<StorageFailure, Set<SupportedProtocol>>
     fun setConferenceCallingEnabled(enabled: Boolean): Either<StorageFailure, Unit>
     fun isConferenceCallingEnabled(): Either<StorageFailure, Boolean>
-    fun isConferenceCallingEnabledFlow(): Flow<Either<StorageFailure, Boolean>>
+    fun observeConferenceCallingEnabled(): Flow<Either<StorageFailure, Boolean>>
     fun setUseSFTForOneOnOneCalls(shouldUse: Boolean): Either<StorageFailure, Unit>
     fun shouldUseSFTForOneOnOneCalls(): Either<StorageFailure, Boolean>
     fun setSecondFactorPasswordChallengeStatus(isRequired: Boolean): Either<StorageFailure, Unit>
@@ -303,7 +303,7 @@ internal class UserConfigDataSource internal constructor(
             userConfigStorage.isConferenceCallingEnabled()
         }
 
-    override fun isConferenceCallingEnabledFlow(): Flow<Either<StorageFailure, Boolean>> =
+    override fun observeConferenceCallingEnabled(): Flow<Either<StorageFailure, Boolean>> =
         userConfigStorage.isConferenceCallingEnabledFlow().wrapStorageRequest()
 
     override fun setUseSFTForOneOnOneCalls(shouldUse: Boolean): Either<StorageFailure, Unit> = wrapStorageRequest {
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
@@ -93,6 +93,7 @@ interface UserConfigRepository {
     suspend fun getSupportedProtocols(): Either<StorageFailure, Set<SupportedProtocol>>
     fun setConferenceCallingEnabled(enabled: Boolean): Either<StorageFailure, Unit>
     fun isConferenceCallingEnabled(): Either<StorageFailure, Boolean>
+    fun isConferenceCallingEnabledFlow(): Flow<Either<StorageFailure, Boolean>>
     fun setUseSFTForOneOnOneCalls(shouldUse: Boolean): Either<StorageFailure, Unit>
     fun shouldUseSFTForOneOnOneCalls(): Either<StorageFailure, Boolean>
     fun setSecondFactorPasswordChallengeStatus(isRequired: Boolean): Either<StorageFailure, Unit>
@@ -301,6 +302,9 @@ internal class UserConfigDataSource internal constructor(
         wrapStorageRequest {
             userConfigStorage.isConferenceCallingEnabled()
         }
+
+    override fun isConferenceCallingEnabledFlow(): Flow<Either<StorageFailure, Boolean>> =
+        userConfigStorage.isConferenceCallingEnabledFlow().wrapStorageRequest()
 
     override fun setUseSFTForOneOnOneCalls(shouldUse: Boolean): Either<StorageFailure, Unit> = wrapStorageRequest {
         userConfigStorage.persistUseSftForOneOnOneCalls(shouldUse)
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/HasConferenceCallingActivatedUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/HasConferenceCallingActivatedUseCase.kt
@@ -33,18 +33,19 @@ import kotlinx.coroutines.flow.zip
  * This can be used to inform user about the change, for example displaying a dialog about upgrading to enterprise edition.
  */
 interface HasConferenceCallingActivatedUseCase {
-    suspend operator fun invoke(): Flow<Any>
+    suspend operator fun invoke(): Flow<Boolean>
 }
 
 internal class HasConferenceCallingActivatedUseCaseImpl(
     private val userConfigRepository: UserConfigRepository,
 ) : HasConferenceCallingActivatedUseCase {
-    override suspend fun invoke(): Flow<Any> {
+    override suspend fun invoke(): Flow<Boolean> {
         val enabledFlow = userConfigRepository.isConferenceCallingEnabledFlow()
             .map { isEnabled -> isEnabled.fold({ false }, { it }) }
         return enabledFlow
             .zip(enabledFlow.drop(1)) { old, new -> old to new }
             .filter { (old, new) -> !old && new }
+            .map { (_, new) -> new }
             .distinctUntilChanged()
     }
 }
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/HasConferenceCallingActivatedUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/HasConferenceCallingActivatedUseCase.kt
@@ -1,0 +1,50 @@
+/*
+ * Wire
+ * Copyright (C) 2024 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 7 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/HasConferenceCallingActivatedUseCase.kt
```diff
@@ -0,0 +1,50 @@
...
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.zip
+
+/**
+ * Returns [true] only when our conference calling setting changes from false to true, meaning our conference calling
+ * capability has been enabled. Internally we rely on getting event [Event.FeatureConfig.ConferenceCallingUpdated].
+ * This can be used to inform user about the change, for example displaying a dialog about upgrading to enterprise edition.
+ */
+interface HasConferenceCallingActivatedUseCase {
+    suspend operator fun invoke(): Flow<Any>
```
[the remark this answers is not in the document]
m-zagorski: For sure, I had the boolean at first, but then we dont really use it and I didnt want to add additional mapping to `Unit` or something else, hence why I ended up with `Any` :D but I will change it up 👍

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/HasConferenceCallingActivatedUseCase.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.zip
+
+/**
+ * Returns [true] only when our conference calling setting changes from false to true, meaning our conference calling
+ * capability has been enabled. Internally we rely on getting event [Event.FeatureConfig.ConferenceCallingUpdated].
+ * This can be used to inform user about the change, for example displaying a dialog about upgrading to enterprise edition.
+ */
+interface HasConferenceCallingActivatedUseCase {
```
ohassine, line 35: The name would indicate that it's returning a Boolean, how about something like: `ObserveConferenceCallingEnabled` ?

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/call/usecase/HasConferenceCallingActivatedUseCaseTest.kt
```diff
@@ -0,0 +1,113 @@
...
+import io.mockative.mock
+import kotlinx.coroutines.flow.asFlow
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class HasConferenceCallingActivatedUseCaseTest {
+
+    @Test
+    fun givenOnlyDefaultConferenceCallingValue_thenDoNotReturnAnything() = runTest {
```
ohassine, line 34: Let's stick to `given..when..then` pattern for consistency 
https://wearezeta.atlassian.net/wiki/spaces/AR/pages/122520039/Code+Style+Guideline#Test-methods-naming
maybe :
givenOnlyDefaultConferenceCallingValue_whenUseCaseIsCalled_thenDoNotReturnAnything

m-zagorski: I dont know why I didnt put the `when` here, all other cases have it, fixed 👍

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/HasConferenceCallingActivatedUseCase.kt
saleniuk: We usually name use cases that return `Flow` with "Observe", I think Oussama's point is that right now it can suggest that it returns a single value, not a `Flow` that can be observed. 
Maybe something like `ObserveConferenceCallingEnabledChanges` could combine your both ideas. 😄 
You could also return `StateFlow` instead of regular `Flow`, this way everyone would be 100% sure that it would only emit new value if it's different than the previous one. 😄

m-zagorski: Renamed to `ObserveConferenceCallingEnabled` :)
Good idea with state flow, but that then we'd need to change the implementation in `UserConfigRepository` and that'd imply the usage of it, which was not really my point so I left it a default `flow`

saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveConferenceCallingEnabledUseCase.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.zip
+
+/**
+ * Returns [true] only when our conference calling setting changes from false to true, meaning our conference calling
+ * capability has been enabled. Internally we rely on getting event [Event.FeatureConfig.ConferenceCallingUpdated].
+ * This can be used to inform user about the change, for example displaying a dialog about upgrading to enterprise edition.
+ */
+interface ObserveConferenceCallingEnabledUseCase {
+    suspend operator fun invoke(): Flow<Boolean>
```
vitorhugods, line 36: So, in reality it never emits `false`, right?

I guess the `Boolean` could be completely removed, and we just emit `Unit`.

The documentation and the code say it will only emit `true`, but the API itself makes it so something like this feel intuitive:
```kotlin
useCase().collect { isEnabled -> 
    if(isEnabled) doStuff()
    else doAnotherStuff() // Won't ever be called
}
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveConferenceCallingEnabledUseCase.kt
```diff
@@ -0,0 +1,51 @@
...
+    private val userConfigRepository: UserConfigRepository,
+) : ObserveConferenceCallingEnabledUseCase {
+    override suspend fun invoke(): Flow<Boolean> {
+        val enabledFlow = userConfigRepository.isConferenceCallingEnabledFlow()
+            .map { isEnabled -> isEnabled.fold({ false }, { it }) }
+        return enabledFlow
+            .zip(enabledFlow.drop(1)) { old, new -> old to new }
+            .filter { (old, new) -> !old && new }
+            .map { (_, new) -> new }
+            .distinctUntilChanged()
```
vitorhugods, line 49:
```suggestion
```

I suggest removing `distinctUntilChanged`.
If, for some reason, it ever goes:
- disabled -> enabled -> disabled -> enabled 

it should emit twice, right? So the user could be informed twice as well.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveConferenceCallingEnabledUseCase.kt
```diff
@@ -0,0 +1,51 @@
...
+internal class ObserveConferenceCallingEnabledUseCaseImpl(
+    private val userConfigRepository: UserConfigRepository,
+) : ObserveConferenceCallingEnabledUseCase {
+    override suspend fun invoke(): Flow<Boolean> {
+        val enabledFlow = userConfigRepository.isConferenceCallingEnabledFlow()
+            .map { isEnabled -> isEnabled.fold({ false }, { it }) }
+        return enabledFlow
+            .zip(enabledFlow.drop(1)) { old, new -> old to new }
+            .filter { (old, new) -> !old && new }
+            .map { (_, new) -> new }
```
vitorhugods, line 48:
```suggestion
            .map { Unit }
```

as mentioned before, better to remove the Boolean if it will always emit `true` and avoid confusion with the API.

vitorhugods (changes requested):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveConferenceCallingEnabledUseCase.kt
m-zagorski: done

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveConferenceCallingEnabledUseCase.kt
m-zagorski: done

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveConferenceCallingEnabledUseCase.kt
m-zagorski: done

m-zagorski:
@vitorhugods  would you mind taking another look ? :)

vitorhugods (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/call/usecase/ObserveConferenceCallingEnabledUseCase.kt
```diff
@@ -0,0 +1,49 @@
...
+}
+
+internal class ObserveConferenceCallingEnabledUseCaseImpl(
+    private val userConfigRepository: UserConfigRepository,
+) : ObserveConferenceCallingEnabledUseCase {
+    override suspend fun invoke(): Flow<Unit> {
+        val enabledFlow = userConfigRepository.isConferenceCallingEnabledFlow()
+            .map { isEnabled -> isEnabled.fold({ false }, { it }) }
+        return enabledFlow
+            .zip(enabledFlow.drop(1)) { old, new -> old to new }
```
borichellow, line 45: just a suggestion:
```
userConfigRepository.isConferenceCallingEnabledFlow()
    .map {  it.getOrElse(false) }
    .scan(...)
    .drop(2)
    .filter { (old, new) -> !old && new }
    .map { Unit }
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
```diff
@@ -93,6 +93,7 @@ interface UserConfigRepository {
     suspend fun getSupportedProtocols(): Either<StorageFailure, Set<SupportedProtocol>>
     fun setConferenceCallingEnabled(enabled: Boolean): Either<StorageFailure, Unit>
     fun isConferenceCallingEnabled(): Either<StorageFailure, Boolean>
+    fun isConferenceCallingEnabledFlow(): Flow<Either<StorageFailure, Boolean>>
```
borichellow, line 96: I think we have a convention name of `fun`s that returns `Flow` should start with "observe", in other case - "get" / "is"

borichellow (approved):
just a few small suggestion that are not blocking the PR :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/UserConfigRepository.kt
m-zagorski: renamed! I've seen some other `is....Flow` and followed those, but probably they are old ones - thanks for headsup!

PR #3092, opened 2024-11-06
feat: Implement message unknown strategy #WPB-11913

m-zagorski:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11913" title="WPB-11913" target="_blank">TaskWPB-11913</a>  [Android] - Implement RFC - Message handling - Incall reactions / Hand raising
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-11913

# What's new in this PR?

### Issues

Implement Unknown messages RFC 

### Solutions

Add a new optional proto field for GenericMessage.
Basically right now we're handling incoming messages that will support `UnknownStrategy` 
Unless we know the type of added message we cant really implement a strategy to bring back the stored byte data into real content.
Also I did not implement adding the `UnknownStrategy` for current messages as stated in the rfc document.

----

----

----
##### References

### DIFF
--- a/protobuf-codegen/src/main/proto/messages.proto
+++ b/protobuf-codegen/src/main/proto/messages.proto
@@ -46,6 +46,7 @@ message GenericMessage {
         ButtonActionConfirmation buttonActionConfirmation = 22;
         DataTransfer dataTransfer = 23; // client-side synchronization across devices of the same user
     }
+    optional UnknownStrategy unknownStrategy = 24 [default = IGNORE];
 }
 
 message QualifiedUserId {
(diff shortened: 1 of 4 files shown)


protobuf-codegen/src/main/proto/messages.proto
```diff
@@ -360,3 +361,9 @@ enum LegalHoldStatus {
     DISABLED = 1;
     ENABLED = 2;
 }
+
+enum UnknownStrategy {
+    IGNORE = 0;                 // Ignore the message completely.
```
m-zagorski, line 366: should we leave those comments here?

m-zagorski:
@vitorhugods Could you take a looks since you're the author of the RFC, if those changes meet the requirements. I was looking if we could implement the strategy for decoding those stored unknown bytes, but until we have the types it might not be possible - but maybe you had something in mind?

protobuf-codegen/src/main/proto/messages.proto
vitorhugods: Doesn't hurt to have them :D
They were added just to have a bit of documentation in the .proto files

vitorhugods (approved):
Nice :)

MohamadJaara (approved):

PR #3097, opened 2024-11-08
chore: add script for generating new API version #WPB-11975

damian-kaczmarek:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11975" title="WPB-11975" target="_blank">TaskWPB-11975</a>  [Android] Adopt API v7
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

While implementing API v7 I noticed that we need to add many new classes in a repetitive way. Doing it manual is prone to errors.

### Solutions

Added script that automates the process. Added comments about the script in a likely places that develop can start looking in.

#### How to Test

You can either run:
- bash 
```./scripts/generate_new_api_version.sh 5 6 7```
- gradle task: 
```./gradlew :network:generateNewApiVersion -PpreviousApiVersion=5 -PcurrentApiVersion=6 -PnewApiVersion=7```

It should generate NetworkContainers and API classes.
It should add the new api as BackendMetaDataUtil.DevelopmentApiVersions
There could be some unused params copied from previous API version that developer will need to remove, but static analysis will catch it.

----

----

----
##### References

### CODE REVIEW

**Review by MohamadJaara — COMMENTED** [review]
I did run the script and it keeps constructor parameter modifiers like `private val`
it is better to remove them if they are not used inside the class 
so maybe adding to the end of the copy api classes step will help
note: this Regex is from ChatGPT so take it with a grain of salt 

```bash

      # Remove 'private val' from constructor parameters
      new_content=$(echo "$new_content" | perl -0777 -pe '
        s{
          (constructor\s*\()    # Match "constructor(" with optional whitespace
          (                     # Start capture group for parameters
            (?:                 # Non-capturing group
              [^()]             # Any character except parentheses
              |                 # OR
              \([^()]*\)        # Parentheses containing anything but parentheses
            )*?                 # Zero or more times, non-greedy
          )
          (\))                  # Match closing ")"
        }{
          my ($cons, $params, $close) = ($1, $2, $3);
          # Remove visibility modifiers and 'val' or 'var' from parameters
          $params =~ s/\b(private|public|protected|internal)\s+(val|var)\s+//g;
          "$cons$params$close"
        }egsx;
      ')
```

vitorhugods (commented):
I believe the logic within `generate_new_api_version.sh` would be easier to maintain and expand if written in Kotlin, as it's what most of the team is used to. It would also run on Windows :c

We could even leverage tools like [KotlinPoet](https://square.github.io/kotlinpoet/) to generate some pieces of code for us.

------

On the other hand... it's definitely an improvement over not having it at all. So yeah, if the other comments can be addressed, it's fine from my point of view :)

borichellow (approved):
agree with existed comments, other then that: I like the idea, definitely improvement of life  👍 💪

damian-kaczmarek:
I have added removing vals. I did not found a way to remove unused params without installing detekt CLI to run the autofix on a directory. 🤔 

Added a task to rewrite it in Kotlin. I didn't used the KotlinPoet yet, looks fun https://wearezeta.atlassian.net/browse/WPB-14423

yamilmedina (approved):

PR #3098, opened 2024-11-08
fix: extremely slow paginated conversation list query [WPB-11808]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11808" title="WPB-11808" target="_blank">BugWPB-11808</a>  [Android] app stalls when fetching a page of conversations from the local DB
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The paginated conversation list query is unable to show the list in a reasonable time, if someone has many conversations and messages then it can take multiple minutes.

### Causes (Optional)

Left joining messages with multiple columns from different content tables and left joining unread events to just get last message for each conversation and count of different unread events is too far from optimal. We have managed to reduce full table scans but there is still a lot of data to work on by this query.

### Solutions

To have a proper data to work on, I filled a database with ~1150 conversations and over 200 000 messages, biggest conversation had around 35 000 messages and 27 000 unread events.
With this DB, I waited more than 10 minutes using the current paginated query to get the result and eventually gave up. So I checked various ideas, below are the ones that looked most promising.

For unread events table, there are two ways: to left join all unread events into main query and calculate it or to make a subquery or view to first count events on just that table and then left join results of that calculations into the main query.

For message table, first idea was to match the solution implemented some time ago and to make a single subquery to get ids of last messages for each conversation and then left join it with all contents and other things - thanks to that it doesn't left join all messages and their contents first, but just the messages that it actually needs.
Second one was to create a dedicated table which contains ids of last message for each conversation, and just that, so it could replace this subquery and make the query even simpler. The main problem with that was to make sure this table is synced with the message table, so 3 triggers were created - when inserting new message, when updating visibility or type of a message (two parameters that are taken into account when selecting last message) and when deleting a last message (to find the one that from now on should be the last one). Another issue is whether this can affect the event handling times, so I checked that by executing 3000x `processApplicationMessage` using current code and with this additional table which needs to stay synced using triggers when new message is inserted:

image

Results show that there is no difference, so it's safe to use.

To have a perspective, here are the times of three main queries used to fetch conversation list currently without the pagination:
image

Around 175ms to get the list without pagination. 
As I said, with current paginated query I couldn't get a result in over 10 minutes, but with improvements explained above, it started to work in a very reasonable time:
  
image

A, B, C, D are just a combinations of these improvements for joining both message and unread event table, explained here (with average execution times):

image

It looks like the winner is D, and it's unbelievable that the execution of such complex query can take under 20ms, but so far it looks like everything is working correctly. 
So here in this PR, the D approach is implemented - with new dedicated table `LastMessage` and `UnreadEventCountsGrouped` view used in the main query which doesn't require anymore to be GROUPed BY.

#### How to Test

Enable `paginated_conversation_list_enabled` feature flag and open the conversation list.

----

----

----
##### References

### DIFF
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Calls.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Calls.sq
@@ -15,6 +15,7 @@ CREATE TABLE Call (
 CREATE INDEX call_date_index ON Call(created_at);
 CREATE INDEX call_conversation_index ON Call(conversation_id);
 CREATE INDEX call_caller_index ON Call(caller_id);
+CREATE INDEX call_status ON Call(status);
 
 insertCall:
 INSERT INTO Call(conversation_id, id, status, caller_id, conversation_type, created_at, type)
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/ConversationDetailsWithEvents.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/ConversationDetailsWithEvents.sq
@@ -31,7 +31,7 @@ SELECT
     -- last message
     Message.id AS lastMessageId,
     Message.content_type AS lastMessageContentType,
-    Message.creation_date AS lastMessageDatfe,
+    Message.creation_date AS lastMessageDate,
     Message.visibility AS lastMessageVisibility,
     Message.sender_user_id AS lastMessageSenderUserId,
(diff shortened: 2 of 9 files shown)


vitorhugods (commented):
GH4N8Q2XYAAja1L


Amazing work. Truly amazing work!

vitorhugods (approved):

MohamadJaara (approved):

PR #3115, opened 2024-11-20
feat: Delete group as admin [WPB-11559]

m-zagorski:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11559" title="WPB-11559" target="_blank">StoryWPB-11559</a>  [Android] Delete a group as admin
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-11559

# What's new in this PR?

### Issues

We want to allow admins to delete group conversations

### Solutions

In order to delete group as an admin, we have to be in the same team, so we need to know our own team id, to later
check in the client if were matching the team

----

----

----
##### References

### DIFF
--- a/data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
+++ b/data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
@@ -299,7 +299,7 @@ sealed class ConversationDetails(open val conversation: Conversation) {
         override val conversation: Conversation,
         val hasOngoingCall: Boolean = false,
         val isSelfUserMember: Boolean,
-        val isSelfUserCreator: Boolean,
+        val selfUserTeamId: TeamId?,
         val selfRole: Conversation.Member.Role?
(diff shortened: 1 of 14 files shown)


data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
```diff
@@ -299,7 +299,7 @@ sealed class ConversationDetails(open val conversation: Conversation) {
         override val conversation: Conversation,
         val hasOngoingCall: Boolean = false,
         val isSelfUserMember: Boolean,
-        val isSelfUserCreator: Boolean,
+        val selfUserTeamId: TeamId?,
```
MohamadJaara, line 302: There is no need to add the user self team ID to the conversation model, since it is fixed and is not really conversation related, if you need it in the UI part to get whether a conversation can be deleted or not, please expose it as its own use case and use it in the UI

MohamadJaara (changes requested):

PR #3124, opened 2024-11-23
fix: backup and restore progress updates in a more meaningful way.

kl3jvi:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Progress not updating truthfully

### Causes (Optional)

The current implementation was a fake progress that counted with a fixed delay even if we had a small file generated.
I tested with 1GB file compression and it is updating gracefully. 

### Solutions

Implemented a progress tracking system for file compression, which calculates the overall progress by combining the uncompressed and compressed data written. The progress is updated based on both the total bytes of uncompressed files and the size of the compressed output, with a weighted emphasis on the compressed data.

### Dependencies (Optional)

Needs releases with:

- [x] (https://github.com/wireapp/wire-android/pull/3659) [GitHub link to other pull request]

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


MohamadJaara:
@kl3jvi can you please fill the template with a description of how it supposed to work now?

logic/src/jvmMain/kotlin/com/wire/kalium/logic/data/asset/KaliumFileSystemImpl.kt
```diff
@@ -151,4 +152,15 @@ actual class KaliumFileSystemImpl actual constructor(
...
     override suspend fun listDirectories(dir: Path): List<Path> = SYSTEM.list(dir)
+
+    /**
+     * Returns the size of the file at the specified path.
+     * @param path The path to the file whose size is being determined.
+     * @return The size of the file in bytes.
+     */
+    override fun fileSize(path: Path): Long {
+//         val file = File(path.toString())
+//         return if (file.exists() && file.isFile) file.length() else 0
```
kl3jvi, line 163: Will delete this

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/backup/CreateBackupUseCase.kt
```diff
@@ -78,32 +78,33 @@ internal class CreateBackupUseCaseImpl(
...
-                { error -> CreateBackupResult.Failure(error) },
-                { (backupFilePath, backupSize) ->
-                    val isBackupEncrypted = password.isNotEmpty()
-                    if (isBackupEncrypted) {
-                        encryptAndCompressFile(backupFilePath, password)
-                    } else CreateBackupResult.Success(backupFilePath, backupSize, backupFilePath.name)
-                })
-        } finally {
-            databaseExporter.deleteBackupDBFile()
+    override suspend operator fun invoke(password: String, onProgress: (Float) -> Unit): CreateBackupResult =
```
vitorhugods, line 81: I guess a callback works, but it could be a bit safer to instead return a `Flow` that emits progress.

A Flow can be configured to handle back pressure more gracefully. For example, what if the backup creation calls `onProgress` faster than it can be executed?

I think it would be nicer to just return a `Flow<BackupProgress>` instead.
And `BackupProgress` could be something like:

```kotlin
sealed interface BackupProgress {
    data class Complete(val result: CreateBackupResult): BackupProgress
    data class Ongoing(val progress: Float): BackupProgress
}
```

-----

Another argument for Flow:
if we change the implementation of `CreateBackupUseCase` and it runs on a different scope, this different scope would keep a reference to `onProgress`, which could lead to a leak.

It's easier to just return a flow and if it is cancelled, it is cancelled :)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/backup/CreateBackupUseCase.kt
```diff
@@ -78,32 +78,33 @@ internal class CreateBackupUseCaseImpl(
...
+        withContext(dispatchers.default) {
+            val userHandle = userRepository.getSelfUser()?.handle?.replace(".", "-")
+            val timeStamp = DateTimeUtil.currentSimpleDateTimeString()
+            val backupName = createBackupFileName(userHandle, timeStamp)
+            val backupFilePath = kaliumFileSystem.tempFilePath(backupName)
+            deletePreviousBackupFiles(backupFilePath)
+
+            val plainDBPath =
+                databaseExporter.exportToPlainDB(securityHelper.userDBOrSecretNull(userId))?.toPath()
```
vitorhugods, line 90: question: Doesn't this `exportToPlainDB` also consume a long time?

I really don't know 😅.
This is the step that literally takes all the relevant content in the user's DB and exports it. Which can be a ton of data.

vitorhugods (changes requested):
Thanks for the initiative!
We are in progress of refactoring the way we perform/restore backups, and while the progress indication isn't our top priority, it is definitely something we could easily improve :)

I just have a few concerns:

1. The backpressure and cancellation issues I mentioned, so I think a Flow would be more flexible.
2. You mentioned "backup and restore", but I don't see changes in the Restore part.
3. Maybe there is another significant slow step that is being ignored during `exportToPlainDB`

I think progress could _at least_ be split into three steps:
- Exporting/Importing
- Zipping/Unzipping **(covered by this PR)**
- Encrypting/Decrypting **(covered by this PR)**

And, ideally, we should have a progress for each of these steps.

Our current `exportToPlainDB` is actually a single SQL query, so we are unable to track progress. However, the next version of the Backup format will be done in chunks, which should make it possible.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/backup/CreateBackupUseCase.kt
kl3jvi: My initial implementation, was done using a flow. I need to ask about the iOS implementation though. The return of the actual function is Either, so changing that would require to do the same functionality in iOS. 

I know that is the best approach and for sure will try to implement it. 

Thanks for the input man

kl3jvi: Amazing, then i will do the refactoring of this using flow 👍🏻

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/backup/CreateBackupUseCase.kt
kl3jvi: I can only test with our chats, as I don't have any big data to test tho.
Is there any way I can spam and create some random conversations through scripting or api calls?

PR #3154, opened 2024-12-07
chore(backup): add encrypted stream [WPB-10575]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10575" title="WPB-10575" target="_blank">StoryWPB-10575</a>  Cross Platform Backup: Write common backup / restore library
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Backups should be encrypted. We need to add support for this in the multiplatform library

### Solutions

Rely on [libsodium](https://doc.libsodium.org/) and the [Kotlin Multiplatform bindings](https://github.com/ionspin/kotlin-multiplatform-libsodium) for it.

We can use libsodium's [Secret Stream](https://doc.libsodium.org/secret-key_cryptography/secretstream), which allows us to split encrypted messages in chunks, so we don't need to load the whole backup into memory.

This PR adds a wrapper around it, that uses Okio's Sink/Source to read/write data in streams.

On a future PR, I'll add the code that will actually call this, with data being backed up, etc.

This was originally implemented [within `:cryptography`](https://github.com/wireapp/kalium/blob/9ef6a239f5d8c3db4fea1fa8b16567d8d17103e3/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/utils/ChaCha20Decryptor.kt), but was only used for BackUps and the code could use a bit of refactoring.

Maybe in the future it can be in its own `:secret-stream` module.

### Dependencies 

Needs releases with:

- #3153 

#### Test Coverage

----

----
##### References

### DIFF
--- a/backup/src/commonTest/kotlin/com/wire/backup/encryption/XChaChaPoly1305EncryptedStreamTest.kt
+++ b/backup/src/commonTest/kotlin/com/wire/backup/encryption/XChaChaPoly1305EncryptedStreamTest.kt
@@ -1,0 +1,181 @@
+/*
+ * Wire
+ * Copyright (C) 2024 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.backup.encryption
+
+import com.ionspin.kotlin.crypto.pwhash.crypto_pwhash_SALTBYTES
+import com.wire.backup.envelope.cryptography.BackupPassphrase
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.runTest
+import okio.Buffer
+import okio.ByteString.Companion.encodeUtf8
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertContentEquals
+import kotlin.test.assertEquals
+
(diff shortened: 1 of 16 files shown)


backup/src/commonTest/kotlin/com/wire/backup/encryption/XChaChaPoly1305EncryptedStreamTest.kt
```diff
@@ -0,0 +1,188 @@
...
+        assertContentEquals(originalBufferData, decryptedBuffer.readByteArray())
+    }
+
+    /**
+     * Disabled for now as it takes more than 5s to run this on CI / JavaScript / Browser and throws a timeout exception
+     * TODO: Refactor and either split into individual tests, or find a way to increase timeout on browser tests. 5s is too short!
+     */
+    @Test
+    @Ignore
+    fun givenEncryptedMessagesOfMultipleSizesAndCorrectAuthentication_whenDecrypting_thenShouldReturnOriginalMessage() = runTest {
```
Garzas, line 71: Maybe we should use kotest to support parametrized tests?

Garzas (commented):
Looking good 🎖️ Just one question about tests

backup/src/commonTest/kotlin/com/wire/backup/encryption/XChaChaPoly1305EncryptedStreamTest.kt
vitorhugods: I didn't investigate all I could about this, to be honest.

It seems to be a bug with browser tests running async (coroutine) code. It spits out a `Timeout of 5000ms exceeded. For async tests and hooks, ensure “done()” is called; if returning a Promise, ensure it resolves.`

I believe it is within JavaScript's execution of `runTest`.

-----------

Maybe a runBlocking would solve it, as it hides the "async" code? Not 100% sure, to be honest. 

-----------

Not sure if it would solve this issue in particular. But in any case, Kotest and parameterised tests could be a nice addition.

Garzas: If we don't have time for this improvement, can we at least create a task for that to not leave ignore annotation? 😄

vitorhugods: I've split it into smaller tests

PR #3158, opened 2024-12-10
feat: Handle needToRemoveLocally flag [WPB-14603]

m-zagorski:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14603" title="WPB-14603" target="_blank">StoryWPB-14603</a>  [Android] handle needToRemoveLocally flag when reciving a cleared message
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-14603

# What's new in this PR?

### Issues

Handle `needToRemoveLocally` flag for `Cleared` messages

### Solutions

In this PR we only define it and handle its behavior, we do NOT send it or handle sending right now

1. If `needToRemoveLocally` is `false` or `null`, then only the conversation content is cleared, including any assets that are stored locally, but the conversation is not removed from the list.
2. `needToRemoveLocally` is `true` and the user is not a conversation member: same as option 1, but the conversation is removed.
3. `needToRemoveLocally` is `true` and the user is a conversation member: same as option 1

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/handler/ClearConversationContentHandler.kt
```diff
@@ -41,11 +41,18 @@ internal class ClearConversationContentHandlerImpl(
         message: Message.Signaling,
         messageContent: MessageContent.Cleared
     ) {
-        val isMessageComingFromOtherClient = message.senderUserId == selfUserId
+        val isMessageComingFromOtherClient = message.senderUserId != selfUserId
```
vitorhugods, line 44:
```suggestion
        val isMessageComingFromAnotherUser = message.senderUserId != selfUserId
```

vitorhugods (changes requested):
Well-done, just one small suggestion to rename a variable that doesn't reflect its value/usage properly.

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/handler/ClearConversationContentHandler.kt
m-zagorski: Changed 👍

yamilmedina (approved):

saleniuk (approved):

PR #3182, opened 2024-12-17
feat: Delete local conversation use case [#WPB-14601]

m-zagorski:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14601" title="WPB-14601" target="_blank">StoryWPB-14601</a>  [Android] create use case to remove a conversation locally
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-14601

# What's new in this PR?

### Issues

Have an option to delete conversation locally with assets, contents and conversation itself

### Solutions

Create use case for clearing assets and create use case for deleting conversation which involves:
- Clear all assets
- Clear content
- Remove conversation

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -884,6 +885,12 @@ internal class ConversationDataSource internal constructor(
             }
         }
 
+    override suspend fun deleteLocalConversation(conversationId: ConversationId): Either<CoreFailure, Unit> {
```
m-zagorski, line 888: I was thinking between `deleteLocalConversation` and `deleteConversationLocally` - which one would you guys prefer?

m-zagorski: Also there is the `deleteConversation` but it invoked mls method to nuke the conversation which I think we dont want while removing it locally? Am I right?

saleniuk: I would prefer `deleteConversationLocally`, we have other similar methods implemented already like `deleteAssetLocally` or `updateProtocolLocally` so it would make more sense 🤔 
About using `deleteConversation`, I think that would try to delete it globally for everyone, if so then we shouldn't use that in this case 😄

saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
m-zagorski: Gotta rename it to `locally` then 👍

persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageAssetView.sq
```diff
@@ -81,3 +81,13 @@ AND assetMimeType NOT IN :mimeTypes
...
+
+getAllAssetMessagesByConversationId:
+SELECT assetId FROM MessageAssetView
+WHERE conversationId = :conversationId
+AND visibility IN :visibility
+AND contentType IN :contentTypes
+AND assetId IS NOT NULL
+AND isEphemeral = FALSE
+AND dataPath IS NOT NULL
+ORDER BY date DESC;
```
MohamadJaara, line 93: if you want to query for assets to delete there is no need to filter by visibility or isEphemeral jsut get all and start the delete process, also since there is a lot to delete maybe it is worth it to account for interruptions 
1. mark the conversation that it need to be cleared 
2. start a service that does the deletion
3. delete once all data have been deleted

this way it will not block sync the whole time this is running

MohamadJaara (changes requested):

yamilmedina (approved):

sergeibakhtiarov:
Question about Delete and Clear use cases. I believe these usecases can be called from UI view models (from viewmodel scope). What will happen if user closes the screen before the use case is finished with removal the assets and view model scope is cancelled?
This could potentially leave the conversation / assets partially cleared. I would consider using `NonCancellable` job for such operations.

persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageAssetView.sq
m-zagorski: Going to remove the `visibility` and `isEphemeral`. And yes for the handling of the use case its a matter of the `client` to either run it in a service or display a progress until the deletion is done - in our case it will most likely be an service

m-zagorski:
> Question about Delete and Clear use cases. I believe these usecases can be called from UI view models (from viewmodel scope). What will happen if user closes the screen before the use case is finished with removal the assets and view model scope is cancelled? This could potentially leave the conversation / assets partially cleared. I would consider using `NonCancellable` job for such operations.

Yes, when we close up the screen while invoking the use cases it will be partially deleted. I would say that it is on the client side to decide the usage:
- Start a service that does the job - this way it will run until its done, or eventually retry if needed
- Display progress while the deletion is in progress

Personally Im not a big fan of using `NonCancellable` jobs in the business logic directly because it limits a little bit the possible usage. In the current state the client has the full control over the process, it can eventually give user an option to cancel the ongoing deletion process and will be able to do so with the current scoping.

sergeibakhtiarov (approved):

PR #3202, opened 2024-12-30
fix: do not commit User and Client upserts if nothing has changed and fix empty contacts [WPB-15055]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15055" title="WPB-15055" target="_blank">BugWPB-15055</a>  [Android] Crash when opening connection request from paginated list
  </td></table>
  <br />
 

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR contains cherry-picked changes that were first merged  #3188 and then reverted #3198 because contacts data were not persisted properly. Here these changes are brought back along with a fix to persist these data properly. The problem was that in the upsert query, fields were compared using `!=` but some of them are nullable, that's why if the value was set to be null initially and then there was an upsert attempt with proper data, `NULL != 'some string'` equals `NULL` instead of boolean value true/false so it never returns a proper result and when working with nullables, `IS NOT` should be used instead, so in this PR it's changed to use `IS NOT`.

#### Test Coverage (Optional)

----

----

----
##### References


saleniuk:
I tried multiple different accounts and I couldn't reproduce that on my side:
Wire 2024-12-30 at 11-47-07

But I believe comparing nullable using `!=` was the reason here and this commit should fix that https://github.com/wireapp/kalium/pull/3202/commits/95572427ee04e4eb0477652746cbc5a8141b754e

@MohamadJaara could you please verify this using the account on which it happened to you?

MohamadJaara (approved):

vitorhugods (approved):

PR #3228, opened 2025-01-14
feat(backup): support export and import of backups on Web, iOS and Android [WPB-10575]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10575" title="WPB-10575" target="_blank">StoryWPB-10575</a>  Cross Platform Backup: Write common backup / restore library
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

While creating a unified Backup format for iOS, Web and Android, we realised that encryption and compression logic must also be unified.
So, this PR aims to bring the whole solution together:

- Export app data / serialise
- Compress
- Encrypt the file
- Decrypt
- Decompress
- Import app data / de-serialise

The whole flow is joined together here.

### Solutions

Whilst **encryption/decryption** and **serialisation/deserialisation** were already implemented, some changes had to be done in these parts to join the whole flow together. However, the logic implemented in previous PRs stay the same.

#### Pagination

One small change to the current logic is that backup data is now paginated.
As the application exports data, multiple files (pages) are created and memory is released.
Except on Web, where the pages are created but they stay always in memory.

During importing, the process is reversed, and the application can read one file at a time.

#### Compression

**There is no multiplatform zip library for Kotlin**
So I had to come up with some solution to bridge this gap.

##### JS / Web / Browser

The current Web client is using [JSZip](https://stuk.github.io/jszip/) for zipping/unzipping files. It was possible to embed this solution within the library as well using the NPM dependency and some glue / Kotlin bindings + magic to solve dynamic types.

##### iOS

I wasn't able to add compression directly to the library. Our current Kotlin version has some bugs when generating Kotlin bindings for objc and therefore I couldn't use [ZipArchive](https://github.com/ZipArchive/ZipArchive) to implement it.
My solution is to just delegate this responsibility to the consumer.
iOS app will need to provide a `zipper` and `unzipper` in order to use the library.

##### Android

For the sake of reducing code, I left it exactly like iOS. So `kalium:logic` will need to provide a `zipper` and `unzipper` as well.

##### More fields

We need to add more fields to the messages, users, conversations, etc. in order to make sure it all works.

##### File Peeking

I'm raising a smaller PR after this one that will provide "file peeking" capabilities.
This is: read the header of a backup archive and be able to say if self user created it, and if it is encrypted or not.

This way the clients can ask for a password if needed, and abort the importing if self user isn't the one that created the backup.

##### Polishing

There are some possible improvements in terms of performance, and API will probably change a bit as we gather feedback from clients.

##### Publishing

The goal is to publish a npm library for web, and a _spm (?)_ artifact for iOS

### Testing

For testing code that uses compression, it was straight forward on JS: just use JSZip as normal.
But, for Android/iOS, because I left it as a responsibility of callers, I faked it completely.
Instead of actually zipping stuff, all the FakeZip does is:
1. Generate a random ID, let's say "X"
2. Copy all files to a folder called "X"
3. Create a file and write "X" as its content

Then, when unzipping, the reverse process happens:
1. Read the "zip" file, with the content = "X"
2. Find the folder called "X", created during zipping.
3. Return the path to that folder, as if it was unzipped just now 😆

#### Test Coverage

----

----
##### References


backup/src/commonMain/kotlin/com/wire/backup/encryption/SodiumInitializer.kt
```diff
@@ -0,0 +1,26 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.backup.encryption
+
+import com.ionspin.kotlin.crypto.LibsodiumInitializer
+
+internal suspend fun initializeLibSodiumIfNeeded() {
+    if (!LibsodiumInitializer.isInitialized()) {
+        LibsodiumInitializer.initialize()
+    }
```
MohamadJaara, line 25: should this function be thread safe or it will never not be called multiple times

backup/src/commonMain/kotlin/com/wire/backup/dump/MPBackupExporter.kt
```diff
@@ -17,79 +17,169 @@
...
-    public fun addUser(user: BackupUser) {
-        allUsers.add(user)
+    private val usersChunk = mutableListOf<ExportUser>()
+    private val conversationsChunk = mutableListOf<ExportedConversation>()
+    private val messagesChunk = mutableListOf<ExportedMessage>()
+    private var persistedUserChunks = 0
+    private var persistedConversationsChunks = 0
+    private var persistedMessagesChunks = 0
+
+    private val backupInfo by lazy {
```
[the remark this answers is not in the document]
vitorhugods: I wasn't really considering the possibility that the same instance of `MPBackupExporter` could be used for multiple backups.

I was thinking of saying bye bye to it after using it, no reason to keep it around and reuse it.

MohamadJaara (approved):

backup/src/commonMain/kotlin/com/wire/backup/dump/MPBackupExporter.kt
```diff
@@ -17,79 +17,169 @@
...
+            platform = "Common",
+            version = "1.0",
+            userId = selfUserId.toProtoModel(),
+            creationTime = Clock.System.now().toEpochMilliseconds(),
+            clientId = "lol"
+        )
+    }
+
+    @JsName("addUser")
+    public fun add(user: BackupUser) {
```
sbakhtiarov, line 76: Should it be 'suspend' function? With the same pattern as for 'finalize' function, so that Kotlin clients can use it as a suspend function?

backup/src/commonMain/kotlin/com/wire/backup/dump/MPBackupExporter.kt
```diff
@@ -17,79 +17,169 @@
...
+        flushConversations()
+        flushMessages()
+    }
+
+    private fun BackupData.asSource(): Source {
+        val buffer = Buffer()
+        return buffer.write(this.encodeToByteArray())
+    }
+
+    internal suspend fun finalize(password: String?, output: Sink) {
```
sbakhtiarov, line 134: This function should return some result to detect failures

backup/src/commonMain/kotlin/com/wire/backup/ingest/MPBackupImporter.kt
```diff
@@ -18,37 +18,76 @@
...
+                    } else {
+                        // No need to decrypt. We skip the encryption header bytes and copy the zip archive to the destination
+                        source.read(Buffer(), EncryptedStream.XCHACHA_20_POLY_1305_HEADER_LENGTH.toLong())
+                        val buffer = sink.buffer()
+                        buffer.writeAll(source)
+                        buffer.flush()
+                    }
+                }
+                sink.close()
+                BackupImportResult.Success(BackupImportPager(unzipAllEntries()))
```
sbakhtiarov, line 76: If client implementation of `unzipAllEntries` will throw IOException it will be re-thrown from `importBackup`. So we will mix result success/error with exceptions. Maybe we should keep all errors under `Result.Failure`.

backup/src/commonMain/kotlin/com/wire/backup/dump/MPBackupExporter.kt
vitorhugods: Making it `suspendable` adds a tiny bit of complexity, as we need to create a wrapper on `jsMain` to return a `Promise` instead.

But, to be honest, I don't see any benefit in making it suspendable. It just doesn't do anything that requires suspending. Maybe we could switch to IO dispatcher within this function, but I expect whoever is calling this to already be on a worker thread.

For example, fetching stuff from DB, then calling `add`.

We could really benefit from suspendable functions if we try to go fancy, like enqueueing the flushes to be done in parallel while we return from `add`, etc.

backup/src/commonMain/kotlin/com/wire/backup/dump/MPBackupExporter.kt
vitorhugods: Indeed.

I guess mostly IO failures / `UnknownError(cause)` I guess. Sucks that JavaScript won't have IO failures at all, as everything is in memory 😆

backup/src/commonMain/kotlin/com/wire/backup/dump/MPBackupExporter.kt
```diff
@@ -17,79 +17,169 @@
...
+                        header.hashData.hashingMemoryLimit
+                    )
+                )
+            }
+            bufferedOutput
         }
     }
+
+    internal abstract val storage: EntryStorage
+    internal abstract val zipper: Zipper
```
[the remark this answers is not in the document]
vitorhugods: I think it could be done as `archive` indeed.

Comparing with `MPBackupImporter`, for example, I went with `unzipAllEntries` instead of using a `Unzipper` interface. 

I had originally gone for Zipper/Unzipper because it would be easier to create dedicated classes for it, or even a single class that does both if it made sense.

But in the end, it's not needed. I'll simplify a bit and just use an `archive` function.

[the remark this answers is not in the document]
vitorhugods: Done!

PR #3232, opened 2025-01-16
fix: 1:1s end up at the bottom of the list after migration [WPB-15452]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15452" title="WPB-15452" target="_blank">BugWPB-15452</a>  [Android] 1:1s end up at the bottom of the list after migration
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

1:1 conversations, after MLS migration, are moved to the bottom of the conversation list, even if the conversation has some messages sent or received in the meantime. 

### Causes (Optional)

This is due to the fact that when migrating 1:1, a new conversation with different id is created, so in theory this new one doesn’t have any events thus `lastEventTime` and `last_modified_date` are empty (01.01.1970) - that’s why it’s moved to the bottom of the list as the query. It creates confusion and inconsistency in sorting as now, even if a conversation had some messages, which are migrated to the new one.

### Solutions

Update `last_modified_date` when migrating history (messages) - take the value from previous proteus conversation or set a current time otherwise (if there's no previous proteus conversation then it means it's a completely new one so it should be at the top anyway).

#### Test Coverage (Optional)

#### How to Test

Have proteus 1:1, send and receive some messages, migrate to MLS, check whether the conversation is still on the proper position on a list.

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/OneOnOneMigrator.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/OneOnOneMigrator.kt
@@ -55,6 +55,7 @@ interface OneOnOneMigrator {
     suspend fun migrateToMLS(user: OtherUser): Either<CoreFailure, ConversationId>
 }
 
+@Suppress("LongParameterList")
 internal class OneOnOneMigratorImpl(
     private val getResolvedMLSOneOnOne: MLSOneOnOneConversationResolver,
     private val conversationGroupRepository: ConversationGroupRepository,
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/OneOnOneMigrator.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/OneOnOneMigrator.kt
@@ -32,8 +32,11 @@ import com.wire.kalium.logic.functional.Either
 import com.wire.kalium.logic.functional.flatMap
 import com.wire.kalium.logic.functional.fold
 import com.wire.kalium.logic.functional.foldToEitherWhileRight
+import com.wire.kalium.logic.functional.getOrNull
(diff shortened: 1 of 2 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/OneOnOneMigrator.kt
```diff
@@ -140,16 +144,26 @@ internal class OneOnOneMigratorImpl(
...
+                    conversationRepository.getConversationById(proteusOneOnOneConversation).getOrNull()?.lastModifiedDate.let {
+                        listOfNotNull(lastModifiedDate, it).maxOrNull()
+                    }
+                }
+            }.map { lastModifiedDate ->
+                // Fallback to current time if not found as it means that it's completely new conversation without any history
+                lastModifiedDate ?: currentInstant()
             }
+        }.flatMap { lastModifiedDate ->
+            conversationRepository.updateConversationModifiedDate(targetConversation, lastModifiedDate)
```
MohamadJaara, line 163: you can do the same in the query that does the update to avoid updating 
https://github.com/wireapp/kalium/blob/d54cc2881d988b53e362db23fbfe5185a2e0b194/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq#L582

saleniuk: not sure whether `moveMessages` query should have any side effects and we don't update it for every message type so we would need to duplicate the logic from `persistMessage` and `MessageContent.shouldUpdateConversationOrder()` into sql, also, I think it would be harder to set current time as iso string directly in sql 🤔

MohamadJaara (approved):

Garzas (approved):

PR #3237, opened 2025-01-21
fix: mls 1on1 race condition [WPB-15395]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15395" title="WPB-15395" target="_blank">BugWPB-15395</a>  [Android] 1:1 conversation throwing decryption errors after mls was enabled for the team
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What’s New in This PR?

This PR addresses a **race condition** during 1:1 MLS conversation creation and **improves logging** throughout the MLS-related codebase.

## Issues

- **Race condition**: When a user opens a 1:1 conversation at the same time a slow sync process is attempting to (re)join the same MLS conversation, we could end up with errors like:
  - `ConversationAlreadyExists`
  - `Cannot execute operation because a pending commit exists`
- **Lack of logs**: Troubleshooting such concurrency issues was complicated by insufficient logging in some MLS-related operations.

## Causes

- **Concurrent initiation of MLS commits**:  
  - The slow sync step attempts to join existing MLS conversations.  
  - At the same time, the UI notifies that a conversation is open (`NotifyConversationIsOpenUseCase`), which also tries to resolve or establish a 1:1 MLS group.  
  - Because both operations could run in parallel, we ended up with a “pending commit” blocking any subsequent commits.

## Solutions

1. **Waiting for Slow Sync**  
   - In `NotifyConversationIsOpenUseCaseImpl`, we now explicitly wait for `SlowSyncStatus` to reach `Complete` before reevaluating the protocol for a 1:1 conversation.  
   - This ensures no concurrent MLS commits are triggered during slow sync.

2. **Improved Logging**  
   - Added detailed log statements (`kaliumLogger.d(...)`) in `JoinExistingMLSConversationUseCase`, `MLSConversationDataSource`, and `NotifyConversationIsOpenUseCaseImpl`.  
   - These logs help to trace when a commit bundle is sent or processed, and when establishing/joining self and 1:1 MLS groups.

3. **Failure Handling (Wipe on Add Member Failure)**  
   - In `MLSConversationDataSource`, if adding members to an MLS group fails, we attempt a conversation wipe.  
   - This helps to avoid a stuck state if the MLS conversation remains in a “pending commit” state and can’t be recovered via normal commits.


MohamadJaara:
question: in this case do we need it to wait, imo just returning without doing anything is enough since the slow sync will migrate this 1:1 anyway

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
```diff
@@ -485,12 +495,15 @@ internal class MLSConversationDataSource(
                             // We are creating a group with only our self client which technically
                             // doesn't need be added with a commit, but our backend API requires one,
                             // so we create a commit by updating our key material.
+                            kaliumLogger.d("add members to MLS Group: updating keying material for self client")
```
yamilmedina, line 498: Suggestion: Same as prev. we could add tags to improve filtering :D

yamilmedina (approved):
Looks good to me, just would be great to add tags to the logs

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
```diff
@@ -454,7 +456,15 @@ internal class MLSConversationDataSource(
             retryOnStaleMessage = true,
             allowPartialMemberList = false,
             cipherSuite = cipherSuite
-        ).map { Unit }
+        ).onFailure {
```
typfel, line 459: Isnt' this code path also called when adding users to an existing conversation? We certainly don't want wipe any conversation then. We only want to do when we fail to establish the initial MLS group.

MohamadJaara (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/MLSWelcomeEventHandler.kt
```diff
@@ -61,22 +64,33 @@ internal class MLSWelcomeEventHandlerImpl(
...
                     checkRevocationList(it)
                 }
+                kaliumLogger.d("$TAG: Marking conversation as established ${welcomeBundle.groupId.obfuscateId()}")
                 markConversationAsEstablished(GroupID(welcomeBundle.groupId))
             }.flatMap {
+                kaliumLogger.d("$TAG: Resolving conversation if one-on-one ${event.conversationId.toLogString()}")
                 resolveConversationIfOneOnOne(event.conversationId)
             }
+            .flatMapLeft {
+                if (it is MLSFailure.ConversationAlreadyExists) {
```
typfel, line 83: Maybe add warning here then? "Discarding welcome since the conversation already exists "

yamilmedina (approved):

typfel (approved):

PR #3285, opened 2025-02-07
fix: remove folder when conversation list is empty [WPB-15892]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15892" title="WPB-15892" target="_blank">TaskWPB-15892</a>  [Android] It should not be possible to create empty folders
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Because backend doesn't clean folders by itself we need to check if they are empty after conversation from folder removal


borichellow (commented):
I believe task is not only about removing folder after removing the last conversation from it. But also about creating an empty folder in "move to folder" flow. And this PR looks like doesn't solve it. 🤔

borichellow:
> I believe task is not only about removing folder after removing the last conversation from it. But also about creating an empty folder in "move to folder" flow. And this PR looks like doesn't solve it. 🤔

nvm, found Android PR for that :)

saleniuk (approved):

PR #3287, opened 2025-02-07
feat: new login use case chooser for context enterprise vs fallback (WPB-15966)

yamilmedina:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15966" title="WPB-15966" target="_blank">TaskWPB-15966</a>  [Android] Create use case to select old vs new login flow
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- Creates use case to fetch and observe the `LoginContext` (Enterprise or Fallback) according to conditions
  - Api version negotiated
  - Proxy authenticated or not
  - Possible errors in these scenarios
- Create extension of Server config repository, removing duplication and extended.

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/ObserveLoginContextUseCase.kt
```diff
@@ -0,0 +1,70 @@
...
+
+@Suppress("FunctionNaming")
+internal fun ObserveLoginContextUseCase(
+    serverConfigRepository: CustomServerConfigRepository
+): ObserveLoginContextUseCase = object : ObserveLoginContextUseCase {
+    override suspend operator fun invoke(serverLinks: ServerConfig.Links): Flow<LoginContext> {
+        return serverConfigRepository.observeServerConfigByLinks(serverLinks).map {
+            it.flatMap { serverConfig ->
+                if (serverConfig.metaData.commonApiVersion.version >= MIN_API_VERSION_FOR_ENTERPRISE_LOGIN
+                    && serverLinks.apiProxy?.needsAuthentication == false
```
saleniuk, line 47: here second part of this `if` should be:
`&& serverLinks.apiProxy != null && serverLinks.apiProxy.needsAuthentication == false`
or
`&& serverLinks.apiProxy?.needsAuthentication != true`

because right now if `apiProxy` is `null` then it should mean that it doesn't need authentication so it should return `true` but right now it returns `false` because `null == false` is `false`

yamilmedina: Good catch!

saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/server/ServerConfigRepositoryExtension.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.wire.kalium.logic.wrapStorageRequest
+import com.wire.kalium.network.api.base.unbound.versioning.VersionApi
+import com.wire.kalium.network.api.unbound.configuration.ApiVersionDTO
+import com.wire.kalium.persistence.daokaliumdb.ServerConfigurationDAO
+import io.ktor.http.Url
+
+/**
+ * Common operations for the server configuration repository.
+ */
+internal abstract class ServerConfigRepositoryExtension(
```
ohassine, line 37: We missed to cover this class with test

ohassine: Yeah but it's unclear exactly what is being tested, we are only sure that the test runs through those lines of code. 

We need to isolate the class during testing. This way, we can validate its behavior independently, without interference from external classes/dependencies, ensuring that it behaves correctly in all cases.

Garzas (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/configuration/server/ServerConfigRepositoryExtension.kt
yamilmedina: I can take a look at a later point since this is the epic still, but the problem with this is, given is an **abstract class**, it is not straight forward to test, and the functions are covered indeed by the inherited testable classes

PR #3304, opened 2025-02-21
refactor(sync): enable concurrent sync start/stop [WPB-15262]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15262" title="WPB-15262" target="_blank">TaskWPB-15262</a>  [Android] Sync engine refactoring
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

It is a pain to start/stop sync from multiple places using `ConnectionPolicy`.
For instance, what if you want to start sync, perform an operation, and then stop?
You'd need to set the policy, wait, and then reset it to whatever value it was. But what about race conditions? What if another piece of code is changing that too?

PAIN

#### Rename `SyncManager`
`SyncManager` is now `SyncStateObserver`. It had the name of "Manager" but it didn't manage anything. It still has the same function: observe state and **passively** wait for specific states.

#### Move Sync start/stop logic to `SyncExecutor`
Instead of having only autonomous `SlowSyncManager` and `IncrementalSyncManager` that would start/stop on their own and follow the policy: Now `SyncExecutor` will allow multiple requests to be made at the same time. If at least one request is alive, it will start Sync.

Multiple entities can easily call `syncExecutor.request {}`.

Some examples:
```kotlin
// Start/stop sync for a synchronous operation
syncExecutor.request {
    waitUntilLiveOrFailure().onSuccess {
        sendMessage()
    }
}


// Start / Stop sync assynchronously
class LifecycleObserver {

    val job: Job? = null

    fun onBackground()
        job = launch {
            syncExecutor.request {
                awaitCancellation()
            }
        }
    }

    fun onBackground() {
        job?.cancel()
    }
}

// Start sync and never stop (useful for test-service and similar)
@OptIn(DelicateKaliumApi::class)
syncExecutor.request {
    keepSyncAlwaysOn()
}
```

Although not visible in this PR, the code in wire-android regarding sync policy will be reduced and simplified A LOT.

> [!NOTE]
> Maybe `SyncStateObserver` and `SyncExecutor` could be the same class and be named `SyncManager`. HOWEVER, I've identified a few cyclic dependencies during this process. Some classes that are used during event processing have dependencies on `SyncManager`. It is possible to go around it, but in the end I opted to have these two entities: one responsible for passively watching Sync. Another responsible for actively requesting it.

#### Test Coverage

----
##### References


Garzas (commented):
Looking good 🚀  just one comment

logic/src/commonTest/kotlin/com/wire/kalium/logic/fakes/sync/FakeSyncStateObserver.kt
```diff
@@ -0,0 +1,53 @@
...
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+class FakeSyncStateObserver : SyncStateObserver {
+    val mutableSyncState = MutableStateFlow<SyncState>(SyncState.Waiting)
+
+    override val syncState: StateFlow<SyncState> get() = mutableSyncState.asStateFlow()
+
+    override suspend fun waitUntilLive() {
+        TODO("Not yet implemented")
```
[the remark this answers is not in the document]
vitorhugods: These functions aren't used anywhere. The only place using this Fake is in `SyncExecutorTest`. And `SyncExecutor` only needs the `syncState`.

So it's fine to leave them like this.

If someone wants to use this Fake elsewhere, feel free to fill the gaps 😄

Garzas (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncExecutor.kt
```diff
@@ -0,0 +1,169 @@
...
+    private val scope: CoroutineScope,
+    userScopedLogger: KaliumLogger = kaliumLogger,
+) : SyncExecutor() {
+
+    private val syncStateFlow = MutableStateFlow<SyncState>(SyncState.Waiting)
+    private val logger by lazy { userScopedLogger.withFeatureId(SYNC).withTextTag("SyncExecutor") }
+
+    override fun startAndStopSyncAsNeeded() {
+        scope.launch {
+            syncStateObserver.syncState.collect { syncStateFlow.value = it }
```
ohassine, line 109: maybe better rely on `ObserveSyncStateUseCase` abstraction and not directly access syncState.

vitorhugods: UseCases are the top-most level within Kalium.

Something like this:

```mermaid
flowchart TB
subgraph Data
	direction TB
	Repository
	DS[DB/API]
end
subgraph Domain/CoreLogic
	direction TB
	SendMessageUseCase
    ObserveSyncStateUseCase
	SyncStateObserver
    MessageSender
	end
subgraph Presentation
	direction TB
	Compose
	ViewModel
end

Repository-->DS
SendMessageUseCase-->MessageSender
ObserveSyncStateUseCase-->SyncStateObserver
Compose-->ViewModel
ViewModel-->SendMessageUseCase
ViewModel-->ObserveSyncStateUseCase
SyncStateObserver-->Repository
MessageSender-->Repository
```

I am inclined to avoid having these specialised classes (or the "Internal UseCases") depend on client-facing UseCases. Because it can lead to circular dependencies in the future, by having an arrow pointing up (diagonally) in the diagram.

In the context of this PR, I originally had `SyncStateObserver` depend on `SyncExecutor`.
So the single source of truth would be with the one actually executing Sync. I still think this would be the end goal, but this led to circular dependency within Sync itself at the moment (the added TODOs).

ohassine (approved):

PR #3305, opened 2025-02-21
fix: suppress buffer error for mls decrypt transaction [WPB-16233]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16233" title="WPB-16233" target="_blank">BugWPB-16233</a>  [Android] Catch and suppress buffer errors inside transaction
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
- When receiving out-of-order commits (e.g., when a client is missing a preceding commit), the MLS client was throwing a `BufferedFutureMessage` exception instead of buffering and persisting the message.
- This led to transactions always rolling back and no buffered messages actually being saved.

### Causes
- The code did not properly detect or handle `MlsException.BufferedFutureMessage` (wrapped in `CoreCryptoException.Mls`) within the transactional decryption flow.
- As a result, any out-of-order commit triggered a rollback, preventing buffering logic from persisting the messages that should be processed later.

### Solutions
- Updated the `decryptMessage` flow to wrap calls in a transactional API and specifically catch `BufferedFutureMessage` exceptions, allowing the transaction to commit and the message to be buffered rather than rolled back.
- Added tests verifying that out-of-order commits no longer throw exceptions and that previously buffered commits are eventually flushed once the missing epoch is processed.

### DIFF
--- a/cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/MLSClientImpl.kt
+++ b/cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/MLSClientImpl.kt
@@ -140,6 +140,7 @@ class MLSClientImpl(
         return applicationMessage
     }
 
+    @Suppress("TooGenericExceptionCaught")
     override suspend fun decryptMessage(groupId: MLSGroupId, message: ApplicationMessage): List<DecryptedMessageBundle> {
         var decryptedMessage: DecryptedMessage? = null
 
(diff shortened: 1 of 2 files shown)


cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/MLSClientImpl.kt
```diff
@@ -137,22 +140,41 @@ class MLSClientImpl(
...
+                try {
+                    val result = context.decryptMessage(
+                        groupId.decodeBase64Bytes(),
+                        message
+                    )
+                    decryptedMessage = result
+
+                } catch (throwable: Throwable) {
+                    val isBufferedFutureError = throwable.cause is MlsException.BufferedFutureMessage ||
+                            throwable.message?.contains("BufferedFutureMessage") == true ||
```
typfel, line 158: Why do you need this? Isn't it already covered by `Throwable.cause is MlsException.BufferedFutureMessage`?

Garzas: Changed to proper error checking

typfel (approved):

MohamadJaara (approved):
LGTM

PR #3308, opened 2025-02-25
feat: support for local assets preview (#WPB-15738)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15738" title="WPB-15738" target="_blank">TaskWPB-15738</a>  [Android] Video preview
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-15738
https://wearezeta.atlassian.net/browse/WPB-15736
----

# What's new in this PR?

Support for rendering assets preview after downloading / decrypting. To show local preview user needs to download and decrypt the asset file.
- Added `LocalData` field to asset data to have a reference to local file for the asset. This file is used for rendering video / pdf previews.
- Added local data path to the MessageDetailsView to return it with the messages query from the database.
- `SendTextMessageUseCase` removed test code for sending multiple attachments.

### DIFF
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageDetailsView.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageDetailsView.sq
@@ -60,6 +60,7 @@ AssetContent.asset_width AS assetWidth,
 AssetContent.asset_height AS assetHeight,
 AssetContent.asset_duration_ms AS assetDuration,
 AssetContent.asset_normalized_loudness AS assetNormalizedLoudness,
+AssetData.data_path AS assetDataPath,
 MissedCallContent.caller_id AS callerId,
 MemberChangeContent.member_change_list AS memberChangeList,
 MemberChangeContent.member_change_type AS memberChangeType,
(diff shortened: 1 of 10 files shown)


persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageDetailsView.sq
```diff
@@ -148,6 +149,7 @@ FROM Message
 JOIN UserDetails ON Message.sender_user_id = UserDetails.qualified_id
 LEFT JOIN MessageTextContent AS TextContent ON Message.id = TextContent.message_id AND Message.conversation_id = TextContent.conversation_id
 LEFT JOIN MessageAssetContent AS AssetContent ON Message.id = AssetContent.message_id AND Message.conversation_id = AssetContent.conversation_id
+LEFT JOIN Asset AS AssetData ON AssetContent.asset_id = AssetData.key
```
MohamadJaara, line 152: instead of storing path within DB does it make sense to have a fixed path 
something like
/appData/files/{conversaion_id}/{file_type}/{asset_key} 
this way we can avid this left join and we can always get the correct path ?

sbakhtiarov: Good idea, but we are not just storing the path, we are also observing the path changes. So when the app downloads and decrypts the asset the path is updated and UI can immediately use it to show the preview.
The only way to do it with the fixed paths is to set up file observer for the /appData/files/ folder.

MohamadJaara: Ok then we can store the path, but still having fixed paths can help with file management, and it is easier to delete a folder with its constant rather than assets one by one

persistence/src/commonMain/db_user/migrations/99.sqm
```diff
@@ -11,3 +11,170 @@ CREATE TABLE IF NOT EXISTS MessageAttachmentDraft (
     FOREIGN KEY (conversation_id) REFERENCES Conversation(qualified_id) ON DELETE CASCADE ON UPDATE CASCADE,
     PRIMARY KEY (attachment_id)
 );
+
+DROP VIEW IF EXISTS MessageDetailsView;
```
Garzas, line 15: Should this be in separate sqm file because previous version may already migrated through 99?

Garzas (changes requested):

persistence/src/commonMain/db_user/migrations/99.sqm
sbakhtiarov: Version 99 from this branch was never published, so I keep all database updates for the cells support in it.

saleniuk (approved):

Garzas (approved):

PR #3320, opened 2025-03-05
feat: Multipart message support (#WPB-15746)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15746" title="WPB-15746" target="_blank">TaskWPB-15746</a>  [Android] Send message with cell attachments
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-15746

# What's new in this PR?

Support for sending multipart messages for conversations with Wire Cell support. New message type is only supported for sending assets uploaded to wire cell storage. Support for sending regular encrypted assets with new Multipart message will be added later (so there are some TODOs in the code for handling regular assets with Multipart)

Some interactions with the new message type are no supported at the moment:
- Replies/quotes - not supported
- Edit - only for editing text (no attachments edit)
- Delete - work in progress to remove assets when deleting a message.

**1. Protocol update**
New multipart message combining text with multiple attachments (Regular or Cell). Update code for packing / unpacking and all handlers. 

**2. Database**
`99.sqm` migration file is a summary of new/updated tables.
[Updated] `Conversation/ConversationDetails` - added `wire_cell` column to enable/disable wire cell per conversation. Currently just a local client-side value. Once BE is ready it will be synced with the server.
[Updated]`MessageAttachmentDraft`: Asset metadata is now stored in the database for attachment drafts for sending in Multipart message.
[New]`MessageAttachments`: new table for storing message attachments with one-to-many relation (one message/conversation to many asset ids)
[Updated] `MessageDetailsView` : updated to return list of attachments for each message from the `MessageAttachments` table. Implementation is similar to handling reactions/mentions for messages.

**3. Use case**
[New]`SendMultipartMessageUseCase` new use case for sending multipart message. For conversations with Wire Cell support the new Multipart message is used, for regular conversations assets are sent as individual asset messages.
[Updated] `RetryFailedMessageUseCase` updated to support re-sending of the failed Multipart message.


cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/DownloadCellFileUseCase.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.wire.kalium.common.functional.onFailure
+import com.wire.kalium.common.functional.onSuccess
+import com.wire.kalium.logic.data.asset.AssetTransferStatus
+import com.wire.kalium.util.KaliumDispatcher
+import com.wire.kalium.util.KaliumDispatcherImpl
+import kotlinx.coroutines.withContext
+import okio.Path
+
+public class DownloadCellFileUseCase internal constructor(
+    private val cellsRepository: CellsRepository,
```
vitorhugods, line 31: We have some tests on `logic` that check for things like "data layer can't depend on use cases" and "all use cases must be documented", etc.

I just realised that with modularisation, this can be left without check. I don't think it's a rule that is _always_ necessary. But it's better to be 10% annoying than 90% missing in documentation. This usecase seems pretty self-explanatory. 


In general, with this comment, I just want to raise awareness about the tests we had on `logic` and that are no longer being checked in feature modules.

Something that we - as a team - should resolve at some point.

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/GetPreviewUrlUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+                cellsRepository.getPreviews(assetId)
+                    .onFailure { return it.left() }
+                    .getOrElse { emptyList() }
+            )
+
+            if (previewList.isEmpty()) {
+                delay(500)
+            }
+
+        } while (previewList.isEmpty() && retry++ < 20)
```
vitorhugods, line 54: Not a deal-breaker. But I think it's worth going the small effort of extracting this `20` to a `const val MAX_RETRY_COUNT = 20`. Makes it easier to quickly find and change these values if/when needed, also helps a bit with readability in the future.

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/GetPreviewUrlUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        previewList.maxByOrNull { it.density }?.let {
+            cellsRepository.savePreviewUrl(assetId, it.url)
+        }
+
+        return Unit.right()
```
vitorhugods, line 60: Do we ever return Either.Left inthis UseCase?

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/AddAttachmentDraftUseCase.kt
```diff
@@ -45,49 +49,77 @@ public interface AddAttachmentDraftUseCase {
...
         fileName: String,
+        mimeType: String,
         assetPath: Path,
         assetSize: Long,
-    ): Either<NetworkFailure, Unit> {
+        assetMetadata: AssetContent.AssetMetadata?,
+    ): Either<CoreFailure, Unit> {
 
-        val destNodePath = "$ROOT_CELL/$conversationId/$fileName"
+        val cellName = conversationDao.getCellName(QualifiedIDEntity(conversationId.value, conversationId.domain))
```
yamilmedina, line 79: As a prev. comment, we shouldn't be able to use. Entities on the use cases since they can leak by mistake to consumers.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendMultipartMessageUseCase.kt
```diff
@@ -0,0 +1,299 @@
...
+    ): Message.Regular {
+
+        val previews = uploadLinkPreviewImages(linkPreviews)
+        val expectsReadConfirmation = userPropertyRepository.getReadReceiptsStatus()
+        val messageTimer: Duration? = selfDeleteTimer(conversationId, true).first().duration
+
+        return Message.Regular(
+            id = generatedMessageUuid,
+            content = if (isCellEnabled) {
+                MessageContent.Multipart(
```
vitorhugods, line 184: Could also check if there are attachments?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendMultipartMessageUseCase.kt
```diff
@@ -0,0 +1,299 @@
...
+        isCellEnabled: Boolean,
+        attachments: List<MessageAttachment>
+    ) = if (isCellEnabled) {
+        publishAttachments(attachments).onSuccess {
+            messageSender.sendMessage(message)
+        }
+    } else {
+        scope.launch {
+            attachments.forEach { attachment ->
+                with(attachment as CellAssetContent) {
```
vitorhugods, line 227: I wonder if it is possible to add attachments, then the team enables cells, then we have a crash here of some sort, as this cast could fail.

vitorhugods (commented):
LGTM. Just some small details / nits and general questions that reflect my lack of context :P

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/AddAttachmentDraftUseCase.kt
```diff
@@ -45,49 +49,77 @@ public interface AddAttachmentDraftUseCase {
      * @param assetSize size of the attachment asset
      * @return [Either] with [Unit] or [NetworkFailure]
      */
+    @Suppress("LongParameterList")
     public suspend operator fun invoke(
```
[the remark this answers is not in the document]
vitorhugods: It seems Sergei added `explicitApi()` to `cells/build.gradle.kts`, so it requires us to add a visibility modifier, otherwise the compiler throws.

It's actually something we agreed on doing in Kalium a long time ago, but I failed to do so.
Modularising the project has this benefit as well. Incremental adoption of new techniques without having to refactor the whole thing.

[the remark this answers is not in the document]
yamilmedina: Good point ! Completely missed the flag

cells/src/commonMain/kotlin/com/wire/kalium/cells/CellsScope.kt
```diff
@@ -97,17 +104,29 @@ public class CellsScope(
...
         get() = ObserveAttachmentDraftsUseCaseImpl(messageAttachmentsDraftRepository, uploadManager)
 
     public val publishAttachments: PublishAttachmentsUseCase
-        get() = PublishAttachmentsUseCaseImpl(cellsRepository, messageAttachmentsDraftRepository, cellClientCredentials)
+        get() = PublishAttachmentsUseCaseImpl(cellsRepository)
 
     public val observeFiles: ObserveCellFilesUseCase
         get() = ObserveCellFilesUseCaseImpl(conversationsDAO, cellsRepository)
+
+    public val enableWireCell: SetWireCellForConversationUseCase
```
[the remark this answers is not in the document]
sbakhtiarov: This is required by `explicitApi` mode that I set for the module. It requires setting visibility modifiers for each top method/class to ensure that it was a design decision.

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/MessageAttachmentDraftRepository.kt
```diff
@@ -24,13 +24,23 @@ import com.wire.kalium.common.error.CoreFailure
...
 import kotlinx.coroutines.flow.Flow
 
-internal interface MessageAttachmentDraftRepository {
-    suspend fun add(conversationId: QualifiedID, node: CellNode, dataPath: String): Either<CoreFailure, Unit>
-    suspend fun get(uuid: String): Either<CoreFailure, AttachmentDraft?>
-    suspend fun getAll(conversationId: ConversationId): Either<CoreFailure, List<AttachmentDraft>>
-    suspend fun observe(conversationId: QualifiedID): Flow<List<AttachmentDraft>>
-    suspend fun updateStatus(uuid: String, status: AttachmentUploadStatus): Either<CoreFailure, Unit>
-    suspend fun remove(uuid: String): Either<CoreFailure, Unit>
+public interface MessageAttachmentDraftRepository {
```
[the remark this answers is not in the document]
sbakhtiarov: This is required by `explicitApi` mode that I set for the module. It requires setting visibility modifiers for each top method/class to ensure that it was a design decision.

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/AddAttachmentDraftUseCase.kt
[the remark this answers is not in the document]
sbakhtiarov: This is required by `explicitApi` mode that I set for the module. It requires setting visibility modifiers for each top method/class to ensure that it was a design decision.

cells/src/commonMain/kotlin/com/wire/kalium/cells/CellsScope.kt
```diff
@@ -97,17 +104,29 @@ public class CellsScope(
...
+        get() = SetWireCellForConversationUseCase(conversationsDAO)
```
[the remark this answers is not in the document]
sbakhtiarov: I will fix it. I think I just made a shortcut here by using Dao directly, but it is better to do it via repository.

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/AddAttachmentDraftUseCase.kt
sbakhtiarov: Will replace with repository method call.

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/model/NodePreview.kt
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+package com.wire.kalium.cells.domain.model
+
+internal data class NodePreview(
+    val url: String,
+    val density: Int,
+)
```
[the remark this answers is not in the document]
sbakhtiarov: This is a preview thumbnail resolution (max dimension). Server generates multiple 300, 1024 but I am using only the best one with 1024

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/GetPreviewUrlUseCase.kt
sbakhtiarov: This Use Case is work in progress. I should have mentioned it in the comment. We are still working on the final solution for handling asset preview thumbnails.

cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/GetPreviewUrlUseCase.kt
sbakhtiarov: Will refactor it.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendMultipartMessageUseCase.kt
sbakhtiarov: This use case should only be called to send a message with attachments, without attachments client must send a simple Text message. Maybe it could even throw error if it is used to send message without attachments.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/message/SendMultipartMessageUseCase.kt
sbakhtiarov: The implementation for sending regular encrypted attachments is not done yet. I will implement it later and handle the other content type.

yamilmedina (approved):

PR #3325, opened 2025-03-10
feat: refresh cell asset state (#WPB-16447)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-16447

## Refresh cell assets in conversation

Availability and data (contentUrl, previewUrl) of each wire cell attachment must be refreshed when user opens conversation.

1. New data stored for each attachment: 
   -`contentUrl` - to open the content directly without downloading
   -`contentHash` - to detect content updates.
2. `RefreshCellAssetStateUseCase` implements all logic for refreshing asset data and state.
3. CellRepository refactoring. Methods related to conversations and attachments are moved into separate repositories for a clear separation.


cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/RefreshCellAssetStateUseCase.kt
```diff
@@ -0,0 +1,147 @@
...
+ * - Check if the asset is still available.
+ * - Remove local data if the asset is not found.
+ * - Remove local data if the asset is updated (based on contentHash).
+ * - Fetch preview URL with retries.
+ */
+public interface RefreshCellAssetStateUseCase {
+    public suspend operator fun invoke(assetId: String, fetchPreview: Boolean): Either<CoreFailure, Unit>
+}
+
+internal class RefreshCellAssetStateUseCaseImpl internal constructor(
```
saleniuk, line 56: Would be great to cover that with unit tests, not sure if you already plan to do that in the future 😄

saleniuk (approved):

PR #3337, opened 2025-03-19
feat(Channels): Create channel usecase (WPB-15874)

ohassine:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15874" title="WPB-15874" target="_blank">StoryWPB-15874</a>  [Android] Start a new Channel - Create a new Channel
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Create use case to create a channel

### Causes (Optional)

### Solutions

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References

### DIFF
--- a/data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
+++ b/data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
@@ -172,7 +172,6 @@ data class Conversation(
         sealed class Group : Type {
             data object Regular : Group()
             data object Channel : Group()
-            companion object : Group()
         }
     }
 
--- a/data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
+++ b/data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
@@ -172,6 +172,7 @@ data class Conversation(
         sealed class Group : Type {
             data object Regular : Group()
             data object Channel : Group()
+            companion object : Group()
         }
     }
 
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
@@ -113,32 +113,36 @@ internal class ConversationMapperImpl(
         apiModel: ConversationResponse,
         mlsGroupState: GroupState?,
         selfUserTeamId: TeamId?
-    ): ConversationEntity = ConversationEntity(
-        id = idMapper.fromApiToDao(apiModel.id),
-        name = apiModel.name,
-        type = apiModel.toConversationType(selfUserTeamId),
-        teamId = apiModel.teamId,
-        protocolInfo = apiModel.getProtocolInfo(mlsGroupState),
-        mutedStatus = conversationStatusMapper.fromMutedStatusApiToDaoModel(apiModel.members.self.otrMutedStatus),
-        mutedTime = apiModel.members.self.otrMutedRef?.let { Instant.parse(it) }?.toEpochMilliseconds() ?: 0,
-        removedBy = null,
-        creatorId = apiModel.creator ?: selfUserId.value, // NOTE mls 1-1 does not have the creator field set.
-        lastReadDate = Instant.UNIX_FIRST_DATE,
-        lastNotificationDate = null,
-        lastModifiedDate = apiModel.lastEventTime.toInstant(),
-        access = apiModel.access.map { it.toDAO() },
-        accessRole = (apiModel.accessRole ?: ConversationAccessRoleDTO.DEFAULT_VALUE_WHEN_NULL)
-            .map { it.toDAO() },
-        receiptMode = receiptModeMapper.fromApiToDaoModel(apiModel.receiptMode),
-        messageTimer = apiModel.messageTimer,
-        userMessageTimer = null, // user picked self deletion timer is only persisted locally
-        hasIncompleteMetadata = false,
-        archived = apiModel.members.self.otrArchived ?: false,
-        archivedInstant = apiModel.members.self.otrArchivedRef?.toInstant(),
-        mlsVerificationStatus = ConversationEntity.VerificationStatus.NOT_VERIFIED,
-        proteusVerificationStatus = ConversationEntity.VerificationStatus.NOT_VERIFIED,
-        legalHoldStatus = ConversationEntity.LegalHoldStatus.DISABLED
-    )
+    ): ConversationEntity {
+        val type = apiModel.toConversationType(selfUserTeamId)
+        return ConversationEntity(
+            id = idMapper.fromApiToDao(apiModel.id),
+            name = apiModel.name,
+            type = type,
+            teamId = apiModel.teamId,
+            protocolInfo = apiModel.getProtocolInfo(mlsGroupState),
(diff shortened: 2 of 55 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/createconversation/CreateChannelUseCase.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.wire.kalium.logic.data.user.UserId
+
+/**
+ * Use case to create a channel.
+ */
+class CreateChannelUseCase internal constructor(
+    private val createGroupConversation: CreateGroupConversationUseCase
+) {
+    suspend operator fun invoke(name: String, userIdList: List<UserId>, options: ConversationOptions): Result =
+        createGroupConversation.invoke(name, userIdList, options.copy(groupType = ConversationOptions.GroupType.CHANNEL))
```
MohamadJaara, line 31: question: I wonder why create a new use case and add tests to it when we can just reuse the same old one

yamilmedina (approved):
Looks good to me, just left a question, but nothing blocker

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/createconversation/CreateChannelUseCase.kt
ohassine: For channels, in future we could check for permission first

MohamadJaara: permissions will probably be checked in the UI and hide btn if not possible, in this case i suggest for now we reuse the create conversation instead of passing ConversationOptions and then over writing it in the use case

vitorhugods: Channels also have different (extra) parameters when creating them. Like whether the channel is public/private, and if it is private, if regular users can add people or not, etc.

Just because we _can_ check in the UI, doesn't mean we should _only_ check in the UI 😄

----

I think it would be cleaner to have a dedicated class that both UseCases use, instead of one UseCase calling the other.

data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
```diff
@@ -172,6 +172,7 @@ data class Conversation(
         sealed class Group : Type {
             data object Regular : Group()
             data object Channel : Group()
+            companion object : Group()
```
vitorhugods, line 175: Is this intended?

This adds a third "type":

```kotlin
val regularGroup = Conversation(type = Group.Regular)
val channel = Conversation(type = Group.Channel)
val whatIsThis = Conversation(type = Group) // What is this???
```

Also makes this possible:

```kotlin
when(type){
    Group -> {} // What is this case?
    Group.Regular -> {}
    Group.Channel -> {}
}
```

ohassine: It is a type that can be a channel or regular group. 
Added because, in `ConversationEntity`, we have a type `GROUP` that we don't know what it is without fetching db for `isChannel` value.
Mainly added for calling, currently we pass `RegularGroup` for it, which is confusing and I don't think it is worth it to check for `isChannel` value every time. That why `Group` companion object  is added.

I think a cleaner solution would be to add a channel type to ConversationEntity instead of having a separate  boolean value for it.

vitorhugods: For Calling, I guess there is zero distinction between channel or regular groups, so I understand that we would need to pass a generic "Group" to Calling. But this is still odd. It breaks the typing possibilities, adding something else that isn't a channel, nor a regular group. This can lead to weird bugs.

Where is this being passed to Calling? Calling should check `is Group` instead of `== Group`.

Whenever we fetch the data from the Database, we also have access to `isChannel`, so we can already figure out the type.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -573,10 +579,17 @@ internal fun ConversationResponse.toConversationType(selfUserTeamId: TeamId?): C
     }
 }
 
-private fun ConversationEntity.Type.fromDaoModelToType(): Conversation.Type = when (this) {
+private fun ConversationEntity.Type.fromDaoModelToType(isChannel: Boolean?): Conversation.Type = when (this) {
```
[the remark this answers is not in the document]
ohassine: no this is just used for mapping, if we don't pass `isChannel` values, that means it is a `Group`(can be Channel or Regular Group)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/createconversation/CreateChannelUseCase.kt
vitorhugods: It adds a clear separation between them in terms of responsibility and maintenance.

If we need to change something _specific_ about the creation of Regular groups, we change it in the UseCase.

If we need to change something about the _generic_ creation of _any group conversation_, we change it in this `GroupConversationCreator` class.

----

Also, we can inspect the code and clearly understand that "Ok, regular groups have no extra logic to them".

data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
vitorhugods: Maybe with a clearer example of what I mean:

```kotlin
val regularGroup = Conversation(type = Group.Regular)
val channel = Conversation(type = Group.Channel)
val whatIsThis = Conversation(type = Group) // What is this???


assertTrue { regularGroup.type is Group } // true
assertTrue { regularGroup.type == Group } // FALSE

assertTrue { channel.type is Group } // true
assertTrue { channel.type == Group } // FALSE
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -573,10 +579,17 @@ internal fun ConversationResponse.toConversationType(selfUserTeamId: TeamId?): C
...
     ConversationEntity.Type.SELF -> Conversation.Type.Self
     ConversationEntity.Type.ONE_ON_ONE -> Conversation.Type.OneOnOne
-    ConversationEntity.Type.GROUP -> Conversation.Type.Group.Regular
+    ConversationEntity.Type.GROUP -> {
+        when (isChannel) {
+            true -> Conversation.Type.Group.Channel
+            false -> Conversation.Type.Group.Regular
+            null -> Conversation.Type.Group
```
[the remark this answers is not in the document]
ohassine: If we don't check here, we will return `Group.Regular` value as type for all conversations including channels

data/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/Conversation.kt
ohassine: Removed it

vitorhugods (approved):

PR #3360, opened 2025-04-01
fix(MLS): use current client ID when refreshing token [WPB-16726]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16726" title="WPB-16726" target="_blank">BugWPB-16726</a>  [Android] Client not being able to send MLS messages, but receive them
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Some clients have their auth sessions without an associated client ID _somehow_.

### Causes

We are not sure yet. Might have been a bug we had in the past and it's fixed now. But this was dormant for more than a year until now that MLS is coming out, as MLS requests _require_ an "upgraded" session - _i.e._ an associated client ID.

### Solutions

Send the client ID every time we refresh the auth token. This is a broad safety net, in case the auth session wasn't upgraded to have an associated client ID.

This should recover the faulty clients with some delay, and cover our asses in case we break the proper planned upgrade after client registration. Currently the auth token has a 15min expiration on Wire Cloud, so within 15min of an app upgrade, all clients should recover.

On a side-note: Removed `oldAccessToken` from the `SessionManager` refresh process as it wasn't used at all.

Ideally, we should also investigate further to possibly catch the root cause.

#### Test Coverage

----

----
##### References

### DIFF
--- a/logic/src/commonTest/kotlin/com/wire/kalium/logic/network/SessionManagerTest.kt
+++ b/logic/src/commonTest/kotlin/com/wire/kalium/logic/network/SessionManagerTest.kt
@@ -19,6 +19,7 @@ package com.wire.kalium.logic.network
 
 import com.wire.kalium.common.error.CoreFailure
 import com.wire.kalium.common.error.NetworkFailure
+import com.wire.kalium.common.error.StorageFailure
 import com.wire.kalium.logic.configuration.server.ServerConfigMapper
 import com.wire.kalium.logic.data.auth.AccountTokens
 import com.wire.kalium.logic.data.logout.LogoutReason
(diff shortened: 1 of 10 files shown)


logic/src/commonTest/kotlin/com/wire/kalium/logic/network/SessionManagerTest.kt
```diff
@@ -93,15 +95,43 @@ class SessionManagerTest {
...
         }
 
-        val accessToken = "egal"
         val refreshToken = "refreshToken"
-        sessionManager.updateToken(arrangement.accessTokenApi, accessToken, refreshToken)
+        sessionManager.updateToken(arrangement.accessTokenApi, refreshToken)
 
         coVerify {
+            arrangement.accessTokenRefresher.refreshTokenAndPersistSession(eq(refreshToken), any())
+        }.wasInvoked()
```
vitorhugods, line 103: This test didn't have `.wasInvoked()` 💀

MohamadJaara (approved):
LGTM

sbakhtiarov (approved):

yamilmedina (approved):

PR #3394, opened 2025-04-10
fix: mls self conversation join [WPB-17115]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17115" title="WPB-17115" target="_blank">BugWPB-17115</a>  [Android] Second client in a Self MLS conversation stuck on connecting
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
- When a second client (belonging to the same user) attempted to join a self-type MLS conversation, the app would get stuck in a `connecting` state. After a restart, the conversation list was displayed, but the user could not send messages.
- A `409 InvalidRequestError` was thrown due to the second client trying to establish a brand-new self MLS conversation instead of joining the existing one.

### Causes
- In the original code, any zero-epoch `SELF` conversation would always call `establishMLSGroup` with an empty list of members, regardless of whether the user already had an existing MLS group for that conversation.
- This mismatch caused MLS conflicts, preventing the second client from syncing properly.

### Solutions
- **Modified `JoinExistingMLSConversationUseCaseImpl`:**
  - For a `SELF` conversation with zero epoch, we now check the number of clients mapped to the current user in that conversation:
    - If there is only one client, we call `establishMLSGroup` with an empty list (original behavior).
    - If there are more, we call `establishMLSGroup` with a list containing the user ID, effectively “adding” the new client to the existing MLS group.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/JoinExistingMLSConversationUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/JoinExistingMLSConversationUseCase.kt
@@ -189,20 +189,10 @@ internal class JoinExistingMLSConversationUseCaseImpl(
 
             type == Conversation.Type.Self -> {
                 kaliumLogger.d("$TAG: Establish Self MLS Conversation ${conversation.id.toLogString()}")
-                clientRepository.getClientsByConversationId(conversation.id)
-                    .flatMap { clients ->
-                        if ((clients[selfUserId]?.size ?: 0) > 1) {
-                            mlsConversationRepository.establishMLSGroup(
-                                protocol.groupId,
-                                listOf(selfUserId)
-                            )
-                        } else {
-                            mlsConversationRepository.establishMLSGroup(
-                                protocol.groupId,
-                                listOf()
-                            )
-                        }
-                    }
+                mlsConversationRepository.establishMLSGroup(
+                    protocol.groupId,
+                    listOf(selfUserId)
+                )
                     .onSuccess {
                         kaliumLogger.logStructuredJson(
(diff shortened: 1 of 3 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/JoinExistingMLSConversationUseCase.kt
```diff
@@ -187,21 +189,32 @@ internal class JoinExistingMLSConversationUseCaseImpl(
...
-                        level = KaliumLogLevel.INFO,
-                        leadingMessage = "Establish Group",
-                        jsonStringKeyValues = mapOf(
-                            "conversationId" to conversation.id.toLogString(),
-                            "conversationType" to type,
-                            "protocol" to ConversationOptions.Protocol.MLS.name,
-                            "protocolInfo" to conversation.protocol.toLogMap(),
+                clientRepository.getClientsByConversationId(conversation.id)
+                    .flatMap { clients ->
+                        if ((clients[selfUserId]?.size ?: 0) > 1) {
```
yamilmedina, line 194:
```suggestion
                    .flatMap { clients ->
                        val hasOtherClients = (clients[selfUserId]?.size ?: 0) > 1
                        if (hasOtherClients) {
```

saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/JoinExistingMLSConversationUseCase.kt
Garzas: After discussing with @typfel we can get rid of this 😄

PR #3404, opened 2025-04-14
fix: capturing errors during mls message decryption [WPB-17192]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17192" title="WPB-17192" target="_blank">BugWPB-17192</a>  [Android] Broken error handling when decrypting messages
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- MLS conversations were not being repaired correctly after they went out of sync.
- Exceptions like `WrongEpoch` were not handled properly and were being wrapped in a `ClientError`, which caused recovery logic to fail silently.

### Causes (Optional)

- The `coreCrypto.transaction` block suppresses internal exceptions if they are not caught inside the transaction itself.
- Any unhandled exception within the transaction (e.g. `MlsException.WrongEpoch`) is replaced by a `ClientError`, losing the original context and making targeted recovery impossible.

### Solutions

- Refactored `decryptMessage(...)` to capture exceptions inside the transaction.
- If an exception occurs, it is stored in a `capturedError` variable and rethrown after the transaction, preserving the original context.
- Special case handling is still respected: `BufferedFutureMessage` is filtered out as before, while all other errors are surfaced properly.

### DIFF
--- a/cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/MLSClientImpl.kt
+++ b/cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/MLSClientImpl.kt
@@ -146,31 +146,31 @@ class MLSClientImpl(
         var decryptedMessage: DecryptedMessage? = null
         var capturedError: Throwable? = null
 
-        coreCrypto.transaction(object : CoreCryptoCommand {
-            override suspend fun execute(context: CoreCryptoContext) {
-                try {
-                    decryptedMessage = context.decryptMessage(
-                        groupId.decodeBase64Bytes(),
-                        message
-                    )
-                } catch (throwable: Throwable) {
-                    val isBufferedFutureError = (
-                            throwable is CoreCryptoException.Mls && throwable.v1 is MlsException.BufferedFutureMessage
-                            ) || throwable.message?.contains(
-                        "Incoming message is a commit for which we have not yet received all the proposals"
-                    ) == true
+        try {
+            coreCrypto.transaction(object : CoreCryptoCommand {
+                override suspend fun execute(context: CoreCryptoContext) {
+                    try {
+                        decryptedMessage = context.decryptMessage(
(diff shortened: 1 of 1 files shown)


cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/MLSClientImpl.kt
```diff
@@ -144,28 +144,33 @@ class MLSClientImpl(
...
-                    decryptedMessage = result
-
                 } catch (throwable: Throwable) {
-                    val isBufferedFutureError = (
-                            throwable is CoreCryptoException.Mls && throwable.v1 is MlsException.BufferedFutureMessage
-                            ) || throwable.message
-                        ?.contains("Incoming message is a commit for which we have not yet received all the proposals") == true
-                    if (!isBufferedFutureError) {
-                        throw throwable
-                    }
```
typfel, line 164: You need to capture and swallow the buffer errors inside transactions otherwise messages won't get buffered.

typfel (changes requested):

MohamadJaara (approved):

cryptography/src/commonJvmAndroid/kotlin/com.wire.kalium.cryptography/MLSClientImpl.kt
```diff
@@ -144,28 +144,34 @@ class MLSClientImpl(
...
+                        return
                     }
+
+                    capturedError = throwable
+                    throw throwable
                 }
             }
         })
 
+        capturedError?.let { throw it }
```
typfel, line 173: I think you need a try catch around `coreCrypto.transaction` otherwise you'll never reach this code when an error is thrown.

typfel (changes requested):

typfel (approved):

PR #3414, opened 2025-04-15
fix: keep and fetch MLS public keys every 24h [WPB-17161]

saleniuk:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently we do not keep the MLS public keys, they are being fetched every time.
MLS public keys should be treated the same way as feature-config and api-version, so should be fetched once every 24h and when the app is put into foreground (cannot be done in kalium itself, but the worker is provided to do it on the specific platform).

### Solutions

Keep MLS public keys in-memory. 
Create a worker to fetch them once every 24h.
Update repository to be thread safe and try to fetch keys when cannot be found for the given cipherSuite.

#### Test Coverage (Optional)

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
@@ -284,6 +284,8 @@ import com.wire.kalium.logic.feature.message.PersistMigratedMessagesUseCaseImpl
 import com.wire.kalium.logic.feature.message.StaleEpochVerifier
 import com.wire.kalium.logic.feature.message.StaleEpochVerifierImpl
 import com.wire.kalium.logic.feature.migration.MigrationScope
+import com.wire.kalium.logic.feature.mls.MLSPublicKeysSyncWorker
+import com.wire.kalium.logic.feature.mls.MLSPublicKeysSyncWorkerImpl
 import com.wire.kalium.logic.feature.mlsmigration.MLSMigrationManager
 import com.wire.kalium.logic.feature.mlsmigration.MLSMigrationManagerImpl
 import com.wire.kalium.logic.feature.mlsmigration.MLSMigrationWorkerImpl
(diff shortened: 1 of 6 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -2258,6 +2269,10 @@ class UserSessionScope internal constructor(
             proteusSyncWorker.execute()
         }
 
+        launch {
+            mlsPublicKeysSyncWorker.schedule()
+        }
```
MohamadJaara, line 2274: doing this from the background is not great if it is not 100 needed, since this will run even when waking uop the app due to fcm, maybe we can stick to the app start only ?

vitorhugods: It is only scheduling on app wakeup. Internally it has a delay and will wait. I am not seeing a problem with this 🤔

Having a check outside for `isOnForeground` or collecting a flow for foreground activity would be pretty much the same.

saleniuk: That's the requirement from doc: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/160858877/Feature+flags+and+feature+configuration
> For mobile: When the app comes to the foreground (iOS, Android), and on Android every 24 hours that the app has been in the foreground
We know users use Android tablets as PC, so there is a case where Android users never put the app in the background. Therefore the 24h requirement.

That's why app start only is not enough.

This `MlsPublicKeysSyncWorker` is implemented in the same way as `ProteusSyncWorker` in this regard, so after scheduling it delays until 24h passes from previous fetch and then waits until incremental sync is live.

Yes, this can be executed while syncing due to fcm push, but only once per 24h, and that's the requirement - each time user opens the app and once every 24h so both cases should be treated separately and the app should fetch it as soon as it realises that 24h has passed, so waiting until incremental sync is live is a good option as it's the moment the app has the internet connection.

PR #3418, opened 2025-04-18
fix: E2EE info system message when adding users and creating MLS 1:1 [WPB-17287][WPB-17202]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17287" title="WPB-17287" target="_blank">BugWPB-17287</a>  [Android] E2EE info system message when adding users and creating MLS 1:1
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There are two found cases when this E2EE initial system message behaves wrongly:

- it’s being added every time when adding someone else to a group - it should be added only if you are being added to a group and you haven’t been in that group yet (so that this message is added only once initially)

- it’s not being added when you create 1:1 MLS conversation with someone from your contacts (so that there is no connection request needed, e.g. from your team)

Also, there's a problem with our inserting message logic - for this particular type of message it should check whether such message already exists for the given conversation and not add it again, but looks like it's not working properly.

### Solutions

- use `LocalId.generate()` to create system message with id having "local" prefix so that our inserting message logic can check whether such message already exists - we check only messages with local ids to not mess up any message from remote and this message type was being created with non-local id so our logic of checking for duplicate was failing
- for adding someone to the conversation, check if the user being added is self user and add this system message only in that case
- for creating MLS 1:1 conversation, add it right after migration - if it's a newly created conversation then it will add, otherwise, if it's migrated with all other messages from proteus conversation then our logic of checking for duplicates for this specific type of message will prevent from adding it again

#### Test Coverage (Optional)

#### How to Test

- create a conversation and add someone else to it - this system message should be added only once at the beginning of the conversation
- create a 1:1 MLS conversation with someone new from your contacts - this system message should be added at the beginning of the conversation just like for proteus conversations

----

----

----
##### References

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/NewGroupConversationSystemMessagesCreator.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/NewGroupConversationSystemMessagesCreator.kt
@@ -19,6 +19,8 @@ package com.wire.kalium.logic.data.conversation
 
 import com.benasher44.uuid.uuid4
 import com.wire.kalium.common.error.CoreFailure
+import com.wire.kalium.common.functional.Either
+import com.wire.kalium.common.functional.fold
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.id.QualifiedIdMapper
 import com.wire.kalium.logic.data.id.SelfTeamIdProvider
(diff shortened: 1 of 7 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/NewGroupConversationSystemMessagesCreator.kt
```diff
@@ -201,7 +202,7 @@ internal class NewGroupConversationSystemMessagesCreatorImpl(
     ): Either<CoreFailure, Unit> =
         persistMessage(
             Message.System(
-                uuid4().toString(),
+                LocalId.generate(),
```
vitorhugods, line 205: Not exactly related to this PR, but:

In the future we will need to agree with iOS and Web on how to generate IDs for system messages, if we want to backup and restore system messages as well without having duplicates.

In fact, this is something that we can already have issues with on Android.
When restoring a backup from client A onto client B, system messages that exist on both will be duplicated as we don't have a single ID for both.

vitorhugods (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/NewGroupConversationSystemMessagesCreator.kt
MohamadJaara: for backups we can ignore system messages not sure if those are needed anyway

MohamadJaara (approved):

PR #3431, opened 2025-04-29
feat: cross platform backup (#WPB-10579)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10579" title="WPB-10579" target="_blank">StoryWPB-10579</a>  [Android] Cross Platform Backup: Implement common library in clients
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-10579
----

# Integration of cross platform backup library for Android

### Issues

Support for export and import backups in new format for multiplatform support.
Supported backup data:
- Users
- Conversations
- Messages
  - Text
  - Location
  - Asset

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/backup/RestoreMPBackupUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/backup/RestoreMPBackupUseCase.kt
@@ -1,0 +1,129 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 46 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/backup/RestoreMPBackupUseCase.kt
```diff
@@ -0,0 +1,129 @@
...
+interface RestoreMPBackupUseCase {
+    /**
+     * Restores a valid previously created backup file in multiplatform format into the current database, respecting the current data
+     * if there is any overlap.
+     * @param backupFilePath The absolute file system path to the backup file.
+     * @param password the password used to encrypt the original backup file. Null if the file was not encrypted.
+     * @return A [RestoreBackupResult] indicating the success or failure of the operation.
+     */
+    suspend operator fun invoke(backupFilePath: Path, password: String?): RestoreBackupResult
+}
```
vitorhugods, line 51: Will the Android app have to call the correct UseCase? We could have a single useCase for both. Just like the `VerifyBackupUseCase` is handling both. Less logic in the Android App.

sbakhtiarov: My idea was that sometime after releasing the cross platform backup we could remove the use case for creating the old backup. But the verify\restore will probably stay for compatibility.

vitorhugods (approved):

Garzas (approved):

PR #3433, opened 2025-04-29
chore: update kotlin to 2.1 and mockative to 3.0.1 [WPB-17116]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17116" title="WPB-17116" target="_blank">TaskWPB-17116</a>  [Android] migrate to kotlin 2.1
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- updated kotlin to 2.1
  - used kotlin 2.0 branch as a base and changed to 2.1
  - updated ksp to 2.1.0-1.0.28
  - changed our KMP expect classes that extend interfaces to have expect functions declared as well, otherwise getting error:
  ```
  [ACTUAL_WITHOUT_EXPECT] Actual class X has no corresponding expected declaration
  The following declaration is incompatible because modality is different
  ```
- updated mockative to 3.0.1 (as this version is compatible with kotlin 2.0)
  - replaced `@Mock` for properties in your test source sets with `@Mockable` for types in main source sets
  - replaced `classOf<T>()` with `of<T>()`
  - changed `AssetDAO.getAssets` to use `AssetEntity` instead of generated `Asset`, otherwise getting error:
  ` Asset is not resolvable in the current round of processing.` (trying to use element that is also being generated while generating other)
  - due to a duplication error in mockative 3.0.1 and not being able to run most of the tests because of that, raised an issue (https://github.com/mockative/mockative/issues/143) and PR with proposed fix for that issue (https://github.com/mockative/mockative/pull/144) and built a new temporary version (3.0.1-fix) that contains this fix and can be used by kalium until mockative is patched properly
  - added suppressing `OPTIONAL_DECLARATION_USAGE_IN_NON_COMMON_SOURCE` for some elements marked as @Mockable and containing some other annotations that are `@OptionalExpectation`, because mockative can't handle that cases (https://github.com/mockative/mockative/issues/142)
  - for `backup` module it couldn't be possible to add mockative plugin as it collides with `skie` plugin (`Cannot change dependencies of dependency configuration ':backup:jvmMainImplementation' "after it has been included in dependency resolution.`) so to mock some backup interfaces, another interfaces with mockable annotation that extend these backup ones were added to the module that needs these mocks

----
##### References


build.gradle.kts
```diff
@@ -43,6 +42,7 @@ repositories {
     wireDetektRulesRepo()
     google()
     mavenCentral()
+    maven(url = "https://oss.sonatype.org/content/repositories/snapshots")
```
MohamadJaara, line 45: afaik this will cause issues when releasing to fdroid, lets be careful and wait for the fix to be merged first

MohamadJaara (changes requested):

build.gradle.kts
```diff
@@ -66,6 +67,9 @@ tasks.withType<Test> {
 
 allprojects {
     repositories {
+        // temporary repo containing mockative 3.0.1 with a fix for a bug https://github.com/mockative/mockative/issues/143 is uploaded
+        // until mockative releases a new version with a proper fix
+        maven(url = "https://raw.githubusercontent.com/saleniuk/mockative/fix/duplicates-while-merging-dex-archives-mvn/release")
```
MohamadJaara, line 72: afaik this will cause issues when releasing to fdroid, lets be careful and wait for the fix to be merged first

saleniuk: Can you tell a bit more how would it cause problems? It's just another maven repo url so it should work just like any other 🤔 
Other options are: wait with updating to kotlin 2.1 until mockative releases version with that fix or disabling big part of our tests (ios, js, most of android instrumented)

saleniuk: I think I found what you mean - is it because f-droid has this policy of trusting only some "well known" repositories so using some custom one may be restricted?
So back to my question - should I just revert and disable all these tests for the time being?

MohamadJaara (approved):

vitorhugods (approved):

PR #3448, opened 2025-05-15
fix: Call Not Connect When SFT 1-1 Is Enabled on Proteus with MLS backend [WPB-17404]

MohamadJaara:
### Issues
The call connection was failing when trying to establish 1-1 calls between users when Secure Forwarding Trust (SFT) was enabled on Proteus protocol. This affected users trying to make direct calls in these specific protocol conditions.

### Solutions
Introduced `ConversationMetaDataRepository` to provide a clean way to retrieve conversation type and protocol details. This new component:

1. Streamlines call type determination based on conversation metadata
2. Simplifies MLS protocol integration in conversation handling
3. Provides a consistent way to access conversation properties needed for call establishment

This change allows the calling code to properly identify the protocol being used and apply the appropriate connection logic when SFT is enabled, ensuring calls connect successfully in all protocol configurations.

### Testing
The fix was tested by:
1. Setting up test accounts with SFT for 1:1 enabled on Proteus with MLS client created and no migration
2. Attempting 1-1 calls between these accounts
3. Verifying call connection and audio/video transmission worked properly

After implementing the fix, calls connected successfully in all tested scenarios where they previously failed.


ohassine (approved):

vitorhugods (approved):
🦾

PR #3465, opened 2025-06-03
chore: exclude self-deleting messages from backup (#WPB-17978)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-17978

# What's new in this PR?
Do not include self-deleting messages in backup

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/backup/BackupRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/backup/BackupRepository.kt
@@ -86,9 +86,11 @@ internal class BackupDataSource(
                 ),
                 offset = offset,
                 pageSize = PAGE_SIZE.toLong(),
-            ).map {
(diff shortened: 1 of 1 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/backup/BackupRepository.kt
```diff
@@ -86,9 +86,11 @@ internal class BackupDataSource(
                 ),
                 offset = offset,
                 pageSize = PAGE_SIZE.toLong(),
-            ).map {
-                messageMapper.fromEntityToMessage(it)
-            }
+            )
+                .filter { it.expireAfterMs == null }
```
MohamadJaara, line 90: let's do it on the query level and not fetch those messages from the DB, it will be much faster than sorting after selecting

MohamadJaara (changes requested):
one request about moving this loginc to the DB

saleniuk (approved):

PR #3487, opened 2025-06-19
feat: keep conversations order after restoring backup (#WPB-18290)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18290" title="WPB-18290" target="_blank">TaskWPB-18290</a>  [Android] Keep conversations order after restoring backup
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-18290

# What's new in this PR?

### Issues
Conversations stay in alphabetical order after restoring backup on a new app install with empty conversations.

### Causes (Optional)
LastModifiedTime field for all conversations is 0 and they are sorted by name.

### Solutions
Store lastModifiedTime for each conversation in backup and update it when restoring backup. It is important to overwrite the lastModifiedTime only if it is 0 in case the backup is restored after new messages/events were received in the conversation and lastModifiedTime is already updated accordingly.

### DIFF
--- a/backup/src/commonMain/kotlin/com/wire/backup/data/BackupData.kt
+++ b/backup/src/commonMain/kotlin/com/wire/backup/data/BackupData.kt
@@ -86,6 +86,7 @@ public data class BackupConversation(
     val id: BackupQualifiedId,
     @SerialName("name")
     val name: String,
+    @SerialName("lastModifiedTime")
     val lastModifiedTime: BackupDateTime? = null,
 )
 


persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/conversation/ConversationDAOImpl.kt
```diff
@@ -133,11 +133,11 @@ internal class ConversationDAOImpl internal constructor(
         }
     }
 
-    override suspend fun insertOrIgnoreConversations(conversationEntities: List<ConversationEntity>) = withContext(coroutineContext) {
+    override suspend fun insertOrUpdateLastModified(conversationEntities: List<ConversationEntity>) = withContext(coroutineContext) {
```
vitorhugods, line 136: It would be nice to cover this function with a couple of tests.
Making sure it only updates the `lastModifiedDate`, does not overwrite any other data, etc.

vitorhugods (approved):

PR #3492, opened 2025-06-24
fix: local events handling order [WPB-18313]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18313" title="WPB-18313" target="_blank">BugWPB-18313</a>  [Android] Missing messages on latest beta
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Solved race condition causing incorrect event source state (`LIVE` appearing prematurely) during event synchronization.
- Prevented WebSocket live events from interrupting the processing order of pending events fetched in batches.

### Causes (Optional)

- The previous implementation didn't distinguish properly between live events arriving via WebSocket and events fetched from the pending batch, causing unexpected race conditions and incorrect processing order.

### Solutions

- Introduced a mechanism to explicitly mark events fetched via batch as pending (`is_live = false`) and WebSocket events as live (`is_live = true`).
- Implemented logic to remove duplicate live events from the database when corresponding pending events are inserted, preserving the correct order of event processing.
- Added control for event source (`currentSource`) updates, ensuring it remains accurate and reflective of the actual event synchronization state.


yamilmedina (commented):
Looks great great work 🎉 left one question before leaving aprroved.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
```diff
@@ -132,24 +140,42 @@ class EventDataSource(
...
+    private val eventMapper: EventMapper = MapperProvider.eventMapper(selfUserId),
+    private val dispatcher: KaliumDispatcher = KaliumDispatcherImpl
 ) : EventRepository {
 
     private val clearOnFirstWSMessage = MutableStateFlow(false)
+    private val localEventsTrigger = MutableSharedFlow<Unit>(
+        extraBufferCapacity = 1,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST
+    )
+    private val processedEventIds = mutableSetOf<String>()
```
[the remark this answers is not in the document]
MohamadJaara: also using HashSet to improve access

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventRepository.kt
```diff
@@ -74,8 +82,8 @@ interface EventRepository {
      * Performs an acknowledgment of the missed event after performing a slow sync.
      */
     suspend fun acknowledgeMissedEvent(): Either<CoreFailure, Unit>
-    suspend fun fetchEvents(): Flow<Either<CoreFailure, EventEnvelope>>
-    suspend fun liveEvents(): Either<CoreFailure, Flow<WebSocketEvent<Unit>>>
+    suspend fun fetchEvents(): Flow<Either<CoreFailure, PendingEventInfo>>
+    suspend fun liveEvents(): Either<CoreFailure, Flow<WebSocketEvent<Boolean>>>
```
vitorhugods, line 86: I am struggling a bit with `fetchEvents`, `observeEvents` and `liveEvents`.

Maybe their names could be a bit clearer? Is `observeEvents` just for already-stored events? `fetchEvents` gets only pending events maybe, without returning them? And `liveEvents` only from live storage without storing them?

I wonder if we could simplify and abstract all of this into a single flow of events. I see flows interacting with eachother in a way that can be a bit hard to follow and understand how they are connected.

For example, it seems that `observeEvents` will only emit if `liveEvents` is collected separately? This is a bit odd 🤔 Seems to me that it could be a single flow.

Maybe the responsibility is too spread between `EventRepository` and `EventGatherer`, and it could live in a single entity. I don't have recommendations yet, but I will take some time to think about this with more calm in the coming days.

logic/src/commonTest/kotlin/com/wire/kalium/logic/framework/TestEvent.kt
```diff
@@ -274,25 +274,13 @@ object TestEvent {
     )
 
     fun Event.wrapInEnvelope(
-        isTransient: Boolean = false,
+        isTransient: Boolean = false, // TODO Kubaz remove
```
vitorhugods, line 277: TODO

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
```diff
@@ -76,51 +75,25 @@ data class EventEnvelope(
...
- * @property source The source of the event.
- * @see EventSource
+ * @property isLive Indicates whether the event was received in real-time via WebSocket (`true`)
+ * or fetched in batch as a pending event (`false`).
  */
-sealed class EventDeliveryInfo(
-    open val isTransient: Boolean,
-    open val source: EventSource
+data class EventDeliveryInfo(
+    val isLive: Boolean,
```
vitorhugods, line 82: Might as well use the `EventSource` to improve readability.
I see some places doing `EventDeliveryInfo(true)`, but `EventDeliveryInfo(EventSource.LIVE)` is more readable.

vitorhugods (approved):

MohamadJaara (approved):

saleniuk (approved):

PR #3499, opened 2025-07-01
fix: notifications not showing (#WPB-18531)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18531" title="WPB-18531" target="_blank">BugWPB-18531</a>  [Android] Notification not showing if app process was not running
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-18531
# What's new in this PR?

### Issues
No notifications are shown if push message comes when application process is not running.

### Causes
Here is the flow of notification handling events.
1. Push notification handler starts NotificationFetchWorker
2. NotificationFetchWorker starts observing notification events, request sync and wait for Live SyncStatus.
3. Once Live status is received app waits for 1 extra second
4. Notification event observers are cancelled, worker is complete

The problem is that Live sync status comes instantly after starting sync status observation because there are no pending events available for processing. After that application starts opening the socket to receive actual messages but in 1 second it is cancelled since worker stops observer, cancels it's coroutine scope and stops.

### Solutions

Option 1: Increase delay after getting the Live sync status to 10 seconds. This will keep the worker running and will let the app connect to the socket, fetch events and show notification.

Option 2: Ignore the first Live event when waiting for Live sync status in WireNotificationsManager. This will let the worker wait until the app opens the socket, fetches messages and switch to Live state again.

Proposed solution in this PR is option 2.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncExecutor.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SyncExecutor.kt
@@ -37,6 +37,7 @@ import kotlinx.coroutines.flow.cancellable
 import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.drop
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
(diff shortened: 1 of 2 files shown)


Garzas (approved):

vitorhugods:
It sounds to me that the Status is the issue, not the waiting for it.

What if the app is already `Live` and we want to `waitUntilLive` to send a message? Wouldn't it be stuck?

If it isn't connected through the websocket, it shouldn't be considered `Live`. 
If it is connected and no events pending for processing, THEN it should be considered `Live`.

PR #3501, opened 2025-07-02
refactor: get rid of mls dependencies in conversation repository [WPB-14735]

Garzas:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14735" title="WPB-14735" target="_blank">StoryWPB-14735</a>  [Android] Batch decrypt using CC transaction
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR introduces a structural refactor aimed at removing the dependency on `MLSClient` from the `ConversationRepository` layer. The motivation behind this change is to decouple MLS-specific logic from data persistence, especially in the context of batch decryption and clean separation of concerns.

### Key changes:

- **MLS group state resolution** (`conversationExists(...)`) has been moved out of `ConversationRepository` and into `PersistConversationsUseCaseImpl`. This change ensures that repository methods remain data-centric and protocol-agnostic.
- Methods in `ConversationRepository` that deal with data coming from or going to the backend (`fetchConversation`, `fetchConversations`, `persistConversations`, `updateConversationMembers`) are now annotated with `@ConversationPersistenceApi`. This opt-in annotation enforces explicit usage and highlights the sensitive nature of these methods.
- Special handling in places like `insertConversationFromConnection(...)` is now guarded with `@OptIn(ConversationPersistenceApi::class)`, with inline comments explaining the limitations and temporary workarounds related to missing server-side support (e.g., WPB-3560).
- `PersistConversationsUseCaseImpl` now determines the correct MLS group state (`ESTABLISHED`, `PENDING_JOIN`, or `PENDING_WELCOME_MESSAGE`) based on the result of `MLSClient.conversationExists(...)` and the origin of the conversation data (`originatedFromEvent` flag).
- Documentation has been added to all relevant use case interfaces to clarify their responsibilities and expected behavior.
- Functional responsibilities between `PersistConversationUseCase`, `FetchConversationsUseCase`, and related components have been clarified and aligned with the updated repository contract.

This refactor establishes a cleaner layering boundary between backend integration and business logic, making the system easier to maintain, test, and extend.


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/FetchMLSOneToOneConversationUseCase.kt
```diff
@@ -0,0 +1,83 @@
...
+            }
+    }
+
+    private fun addOtherMemberIfMissing(
+        conversationResponse: ConversationResponse,
+        otherMemberId: UserId
+    ): ConversationResponse {
+        val currentOtherMembers = conversationResponse.members.otherMembers
+        val hasOtherUser = currentOtherMembers.any { it.id == otherMemberId.toApi() }
+        val otherMembers = if (hasOtherUser) {
```
sbakhtiarov, line 66: Two questions here:
1. We can just return conversationResponse if hasOtherUser == true, and create a copy with updated list otherwise.
2. I do not see any "add" operations here. If otherMemberId is not in the list of currentMembers then we are not adding it, we are setting it as the only member.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/PersistConversationUseCase.kt
```diff
@@ -0,0 +1,77 @@
...
+    private val conversationRepository: ConversationRepository,
+    private val persistConversations: PersistConversationsUseCase,
+) : PersistConversationUseCase {
+
+    override suspend fun invoke(
+        conversation: ConversationResponse,
+        originatedFromEvent: Boolean
+    ): Either<CoreFailure, Boolean> {
+        val existingConversation = conversationRepository.getConversationDetails(conversation.id.toModel()).getOrNull()
+        val isNewConversation = existingConversation?.let { conversationDetails ->
```
sbakhtiarov, line 61: some simple extension function with a meaningful name could improve readability here.
```val isNewConversation = existingConversation.isGroupStateEstablished()```

**Review by sbakhtiarov — APPROVED** [review]

File: logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/FetchMLSOneToOneConversationUseCase.kt (hunk 1 above)
Comment (Garzas) [reply logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/FetchMLSOneToOneConversationUseCase.kt:66]: Good catch 👍  I was just moving all those functions to be sure it will work as previously. But I can add your suggestions

File: logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/PersistConversationsUseCase.kt [hunk 3]
```diff
@@ -0,0 +1,126 @@
...
+    private val idMapper: IdMapper = IdMapperImpl(),
+    private val conversationMapper: ConversationMapper = MapperProvider.conversationMapper(selfUserId),
+) : PersistConversationsUseCase {
+
+    override suspend fun invoke(
+        conversations: List<ConversationResponse>,
+        invalidateMembers: Boolean,
+        originatedFromEvent: Boolean
+    ): Either<CoreFailure, Unit> {
+        val conversationEntities = conversations
```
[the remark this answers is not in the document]
Garzas: In this case, we intentionally do **not** propagate the error from `mlsGroupState(...)`.  
The reason is that we want to persist the conversation regardless of whether the MLS group state can be resolved.

This situation can occur when another user has created an MLS conversation, but the current client hasn't yet fully initialized MLS support.  
Failing the whole persist operation in such cases would result in incomplete sync, so instead we fallback to `GroupState.PENDING_JOIN` and proceed.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/FetchConversationIfUnknownUseCase.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    override suspend fun invoke(conversationId: ConversationId): Either<CoreFailure, Unit> {
+        return conversationRepository.getConversationById(conversationId)
+            .run {
+                if (isLeft()) {
+                    return fetchConversation(conversationId)
+                } else {
+                    Either.Right(Unit)
+                }
+            }
```
saleniuk, line 50: Just a nitpick but there is a `return` at the top so no need to add one later, moreover it's not consistent because `Either.Right(Unit)` doesn't have it 😄

saleniuk (approved):

PR #3504, opened 2025-07-04
fix: read receipt race condition (#WPB-8756)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8756" title="WPB-8756" target="_blank">BugWPB-8756</a>  [Android] Read receipts not generated when responding from notification popup
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-8756

# What's new in this PR?

### Issues
Message status is inconsistent. Changing from read to delivered.

### Causes (Optional)
Delivered status receipt from one client overwrites current Read status received from another client. 

### Solutions
Check message status before updating and do not allow changing from Read to Delivered.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
@@ -259,17 +259,11 @@ internal interface MessageRepository {
 
     suspend fun getSenderNameByMessageId(conversationId: ConversationId, messageId: String): Either<CoreFailure, String>
     suspend fun getNextAudioMessageInConversation(conversationId: ConversationId, messageId: String): Either<CoreFailure, String>
-
-    suspend fun getMessagesStatus(
-        messageIds: List<String>,
-        conversationId: ConversationId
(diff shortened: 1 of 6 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/message/MessageRepository.kt
```diff
@@ -728,4 +733,12 @@ internal class MessageDataSource internal constructor(
...
     ): Either<CoreFailure, String> =
         wrapStorageRequest { messageDAO.getNextAudioMessageInConversation(messageId, conversationId.toDao()) }
+
+    override suspend fun getMessagesStatus(
+        messageIds: List<String>,
+        conversationId: ConversationId
+    ): Either<CoreFailure, List<Pair<String, MessageEntity.Status>>> = wrapStorageRequest {
+        messageDAO.getMessagesStatus(messageIds, conversationId.toDao())
+            .map { it.id to it.status }
+    }
```
MohamadJaara, line 743: it is better to do this in one transaction get value, compare with new and set

saleniuk (approved):

PR #3506, opened 2025-07-09
fix(sync): consider Live when local events are exhausted [WPB-18531]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18531" title="WPB-18531" target="_blank">BugWPB-18531</a>  [Android] Notification not showing if app process was not running
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Sync is being reported `LIVE` before it actually is Live.

### Causes

Before, we had Fetching events + Processing together. We would do in order. When we fetch, we update the sync status. Simple and easy.
It was simple to assume that fetching pending = processing pending = NOT LIVE.
Or that fetching live = processing live = WE'RE LIVE.

Now we can't do that!
We might be fetching live, but still processing pending, so what's the status?

### Solutions

Make it so Sync is considered Live starting from the first moment that the local event Storage has 0 events.
This means: all events are consumed.

#### How?

1. Changed the `EventGatherer` to emit `EventStreamData.NewEvents`, when there are events to process, or `EventStreamdata.IsUpToDate` when it is exhausted.
2. Changed the `IncrementalSyncWorker` to emit `EventSource.PENDING` once the flow starts, and `EventSource.LIVE` once it sees `IsUpToDate`, add a `distinctUntilChanged()` to avoid duplications and state repetition.

This eliminates the need for the `currentSource` StateFlow, which was only used for this. And, by consequence, eliminates the artificial `LiveSourceChangeHandlerTest`, used to mimic `Pending->Live` source transition.

This is a ridiculously simple approach and I don't know how I haven't thought of this before. 🙃

### Notes

This makes me think that we should also consider removing completely the concept of "Pending" vs "Live" events. This distinction is artificial and serves no purpose. Pending events can be weeks or milliseconds old, depending on their creation date and when the device has become online.

We could remove this field from the Event itself, the Event table, etc. in order to simplify our logic.
If we need to know whether the app is Online or not, we can use th

#### Test Coverage

----

----
##### References


yamilmedina (approved):

MohamadJaara (approved):
did run some tests to repro and looks solved 🚀

PR #3518, opened 2025-07-14
chore: set compile SDK version to 35 [WPB-18871]

MohamadJaara:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18871" title="WPB-18871" target="_blank">TaskWPB-18871</a>  [Android] bump compile sdk to 35
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Set compile SDK to 35

#### ✅ Summary of changes

- **StoreServerConfigUseCaseImpl**: Replaced `removeLast()` with `removeAt(lastIndex)` to avoid `NoSuchMethodError` on Android runtimes using Java 8 core libraries.
- **MLSClientTest.kt**: Removed unnecessary `Mutex` guarding `sendMessageResponses` and `sendCommitResponses`, and replaced `removeFirst()` guarded by `withLock` with `removeFirstOrNull() ?: default` to simplify test logic.
- **Android.kt**: Raised `compileSdk` and `targetSdk` to **API 35** in preparation for Android 15.

#### 📄 From Android official docs

> Collision with [MutableList.removeFirst()](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/remove-first.html) and [MutableList.removeLast()](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/remove-last.html) extension functions in kotlin-stdlib  
> The [List](https://developer.android.com/reference/java/util/List) type in Java is mapped to the `MutableList` type in Kotlin. Because the [List.removeFirst()](https://developer.android.com/reference/java/util/List#removeFirst()) and [List.removeLast()](https://developer.android.com/reference/java/util/List#removeLast()) APIs have been introduced in Android 15 (API level 35), the Kotlin compiler resolves function calls, for example `list.removeFirst()`, statically to the new `List` APIs instead of to the extension functions in `kotlin-stdlib`.  
> If an app is re-compiled with `compileSdk` set to 35 and `minSdk` set to 34 or lower, and then the app is run on Android 14 and lower, a runtime error is thrown.

#### 📦 Android 15 Compatibility

Targeting API 35 (Android 15) enables stricter enforcement of Android-compatible core library APIs and exposes edge cases like this that only crash on older OS versions at runtime.

#### 🔗 References

- Stack Overflow: [Why do I keep getting an error on removeLast()?](https://stackoverflow.com/a/79566737/1819287)  
- Android Docs: [Android 15 Behavior Changes](https://developer.android.com/about/versions/15/behavior-changes-15)  
- Kotlin Docs: [kotlin.collections.removeLast()](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/remove-last.html)


saleniuk (approved):

yamilmedina (approved):

MohamadJaara:
Updating the target SDK version for the app caused `addClientsToConversation` to stall indefinitely in one of the test tests only when running on android 31 emulator, i tried adding break points and the app calls the rust code and then there it never finishes, the same test was working fine on target sdk 34

PR #3527, opened 2025-07-17
refactor: clean-up and unify 24h workers and foreground actions [WPB-17161]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17161" title="WPB-17161" target="_blank">BugWPB-17161</a>  [Android] Excessive frequency of fetching MLS public keys
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Gather 24h workers into a single one called `UserConfigSyncWorker`. This is a `WorkManager` work for Android instead of running a coroutine and delaying for 24h. It waits for the sync to be live after being started but with a timeout of 10s, after that it returns `Result.Retry` so that it’s retried later by the backoff mechanism. Also, new function is added to the scheduler to reset the backoff for that worker - it’s used when the work is already retried and it’s waiting until the backoff duration passes - in this case it means that 24h has already passed but the sync wasn’t live during the run so if the app knows that the sync is now live it can notify the worker to not wait for longer but to try again now.

Gather all actions that need to be executed when the app comes into foreground into a single use case called `ForegroundActionsUseCase` and share only this use case with the app.

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/androidMain/kotlin/com/wire/kalium/logic/sync/WorkSchedulerImpl.kt
```diff
@@ -90,43 +66,97 @@ internal actual class GlobalWorkSchedulerImpl(
...
+        .setRequiredNetworkType(NetworkType.CONNECTED)
+        .build()
+    return OneTimeWorkRequest.Builder(workerClass)
+        .setConstraints(connectedConstraint)
+        .setExpedited(OutOfQuotaPolicy.RUN_AS_NON_EXPEDITED_WORK_REQUEST)
+        .setInputData(inputData)
+        .build()
+}
+
+private const val TIME_OF_EXECUTION = 4  // schedule at 4AM
```
vitorhugods: I wonder if we won't create a spike at 4AM on the backend because of this 😄

Although it is spread through different timezones, it could lead to some peaks. I won't advocate for over-engineering, but if we do need to fix it in the future, I guess we could schedule to be `lastExecutionDate + 24h` or something like that. So it will be spread across different users more randomly.

vitorhugods (approved):
Pretty awesome solution! Nice one!

Just a couple of small comments that are food for thought / possible issues that could arrive in the future, but I don't think they need to be tackled right now.

logic/src/jvmMain/kotlin/com/wire/kalium/logic/sync/WorkSchedulerImpl.kt
```diff
@@ -59,4 +59,16 @@ internal actual class UserSessionWorkSchedulerImpl(
...
+        kaliumLogger.withFeatureId(SYNC).w(
+            "Scheduling periodic user config sync is not supported on JVM."
+        )
+    }
+
+    actual override fun resetBackoffForPeriodicUserConfigSync() {
+        kaliumLogger.withFeatureId(SYNC).w(
+            "Resetting backoff for user config sync is not supported on JVM."
+        )
+    }
```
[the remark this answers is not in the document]
saleniuk: Ok so I added some basic intervalFlows for jvm and ios, I hope that's enough

logic/src/androidMain/kotlin/com/wire/kalium/logic/sync/WorkSchedulerImpl.kt
saleniuk: It wouldn't be exactly 4AM for everyone, work manager uses some flex windows (I think it's like 15mins by default) but I added some randomizer to have some deviation (+/- 60mins so it gives a 2h window for each time zone) to spread it more evenly

PR #3532, opened 2025-07-18
fix: clearing content actions order and deleted_locally state for conversation [WPB-18778]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18778" title="WPB-18778" target="_blank">BugWPB-18778</a>  [Android] Fix order of “delete locally” worker and “clear content” actions
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The order of actions when clearing content was invalid and because of that, asset files wasn't being cleared.
Removing conversation locally can take some time and but the actual deletion happens at the very end so the app still needs to wait until all is executed and block the user from making any actions to the conversation that's being deleted locally.

### Solutions

Fix order of actions when clearing content - move removing assets to be executed before clearing messages.
Split deleting conversation locally into two steps. First mark the conversation as deleted locally using new use case which removes this conversation from the app (not appearing on lists and cannot fetch conversation details) and then execute the deletion using the same use case as before, but now it can be executed in the background.

#### Test Coverage (Optional)

----

----

----
##### References


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationRepository.kt
```diff
@@ -759,6 +761,12 @@ internal class ConversationDataSource internal constructor(
 
     override suspend fun deleteConversationLocally(conversationId: ConversationId) = wrapStorageRequest {
         conversationDAO.deleteConversationByQualifiedID(conversationId.toDao())
+        Unit
```
sbakhtiarov, line 764: Why not return a Boolean here from the DAO function result? (same for the next one)

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/delete/MarkConversationAsDeletedLocallyUseCase.kt
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Marks a conversation as deleted locally, which is a first step in the local deletion process.
+ * After this call, the conversation will not be visible in the app.
+ * To complete the local deletion, call [DeleteConversationUseCase], which can already be done asynchronously in the background.
+ */
+interface MarkConversationAsDeletedLocallyUseCase {
+    suspend operator fun invoke(conversationId: ConversationId): Result
+
+    sealed class Result {
```
sbakhtiarov, line 33: Is there a reason why we cannot use Either here instead of a custom Result class?

sbakhtiarov (approved):

Garzas (approved):

PR #3547, opened 2025-07-28
fix: crash during transaction when client is removed [WPB-18140]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- fixed crash when current client is removed during cc transaction
- added tests

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/CryptoTransactionProvider.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/CryptoTransactionProvider.kt
@@ -18,11 +18,12 @@
 package com.wire.kalium.logic.data.client
 
 import com.wire.kalium.common.error.CoreFailure
+import com.wire.kalium.common.error.MLSFailure
 import com.wire.kalium.common.functional.Either
 import com.wire.kalium.common.functional.flatMap
-import com.wire.kalium.common.functional.getOrFail
 import com.wire.kalium.common.functional.getOrNull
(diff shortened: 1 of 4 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/CryptoTransactionProvider.kt
```diff
@@ -80,8 +81,17 @@ class CryptoTransactionProviderImpl(
...
         return proteusClientProvider.getOrError()
             .flatMap { proteus ->
-                proteus.transaction(name.toTransactionName("proteus")) { proteusContext ->
-                    block(proteusContext)
+                try {
+                    proteus.transaction(name.toTransactionName("proteus")) { proteusContext ->
+                        block(proteusContext)
+                    }
+                } catch (e: IllegalStateException) {
+                    if (e.message?.contains("already been destroyed") == true) {
```
yamilmedina, line 89: Question: Is there an error code or something that can be caught instead of the string in the stacktrace ?

yamilmedina (approved):
Looks good! Just a question that perhaps can make it more secure in case it's possible

saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/CryptoTransactionProvider.kt
Garzas: No, it's not an core crypto error 🫤

PR #3548, opened 2025-07-29
feat: support for conversation endpoints api version 10 (WPB-19107)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-19107
# What's new in this PR?

## Support for API version 10 for conversation endpoints
- added new skip_creator parameter to create conversation request
- set members.self as optional
- moved mls/reset-conversation to api v10


yamilmedina:
Looks good, it would be ideal to tidy up, to add a test for the v10 parsing of the response 💡 so we can have it tested from v10+

PR #3552, opened 2025-07-29
fix: not able to unmute incoming or outgoing call until established [WPB-19112]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19112" title="WPB-19112" target="_blank">BugWPB-19112</a>  [Android] Not able to mute/unmute incoming or outgoing calls until established
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

No use case to observe the last non-ended call if its state is different than ESTABLISHED, so when its state is STARTED, INCOMING, ANSWERED or STILL_ONGOING).

### Solutions

Create the use case to be able to observe the last call for the given conversation if it's still active (so when the state is one of: STARTED, INCOMING, ANSWERED, STILL_ONGOING, ESTABLISHED), otherwise (if last call for the given conversation is already inactive) it returns null. This way the app can observe the active call during status changes and before it's fully established.

#### Test Coverage (Optional)

----

----

----
##### References

### DIFF
--- a/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Calls.sq
+++ b/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Calls.sq
@@ -100,3 +100,18 @@ SET status = 'CLOSED'
 WHERE
     status != 'CLOSED'
     AND status != 'MISSED';
+
+selectLastCallIfActiveByConversationId:
+SELECT *
+FROM Call
+WHERE -- consider only the last call and return it only if it has one of the active statuses
+    id = (
+            SELECT id
+            FROM Call
+            WHERE conversation_id = :conversationId
+            ORDER BY created_at DESC
+            LIMIT 1
+        )
+    AND conversation_id = :conversationId
+    AND status IN ('STARTED', 'INCOMING', 'ANSWERED', 'ESTABLISHED', 'STILL_ONGOING')
+LIMIT 1;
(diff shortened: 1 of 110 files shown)


persistence/src/commonMain/db_user/com/wire/kalium/persistence/Calls.sq
```diff
@@ -100,3 +100,18 @@ SET status = 'CLOSED'
...
+    id = (
+            SELECT id
+            FROM Call
+            WHERE conversation_id = :conversationId
+            ORDER BY created_at DESC
+            LIMIT 1
+        )
+    AND conversation_id = :conversationId
+    AND status IN ('STARTED', 'INCOMING', 'ANSWERED', 'ESTABLISHED', 'STILL_ONGOING')
+LIMIT 1;
```
yamilmedina, line 117: Question: I'm wondering if this query is the same as just matching the `conversationId` and `status` filter in the outer query 🤔 

```sql
SELECT *
FROM Call
WHERE conversation_id = :conversationId
  AND status IN ('STARTED', 'INCOMING', 'ANSWERED', 'ESTABLISHED', 'STILL_ONGOING')
ORDER BY created_at DESC
LIMIT 1;
```

yamilmedina (approved):
LGTM! ✅ Just left a question regarding the query, it **might** be not necessary to do the subquery 🤔

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Calls.sq
saleniuk: It's not exactly the same because I wasn't sure whether it's possible to for instance have last call be "not active" but previous one be still "active" so I wanted to make sure that this query takes last call for the given conversation id and then checks **if** it has "active" state, but I see that we usually just update the state of the last call so I guess it's safe to assume that only the last call can be "active" and simplify the query as you proposed. 😃

PR #3561, opened 2025-08-05
fix: user availability status [WPB-18697]

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18697" title="WPB-18697" target="_blank">BugWPB-18697</a>  [Android] Status not sent to other clients
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-18697

# What's new in this PR?

### Issues
User availability status is not sent to other users.

### Causes (Optional)
When sending status update via Proteus broadcast message app is using `report_only` strategy to discover missing clients but not preparing correct list of recipient user ids.

### Solutions
1. When preparing a list of all recipients of a broadcast message we need to include all users that have conversation with current user.
2. When setting up broadcast message app must add user ids of all recipients to the `report_only` strategy to receive 412 error message when some clients are missing.

### Documentation
https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/106660748/Use+case+setting+the+status+Proteus
https://staging-nginz-https.zinfra.io/v10/api/swagger-ui/#/default/post-proteus-broadcast

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
@@ -76,6 +76,7 @@ import com.wire.kalium.persistence.dao.UserDAO
 import com.wire.kalium.persistence.dao.UserIDEntity
 import com.wire.kalium.persistence.dao.UserTypeEntity
 import com.wire.kalium.persistence.dao.client.ClientDAO
+import com.wire.kalium.persistence.dao.member.MemberDAO
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.map
@@ -163,6 +164,7 @@ interface UserRepository {
 internal class UserDataSource internal constructor(
     private val userDAO: UserDAO,
     private val clientDAO: ClientDAO,
+    private val memberDAO: MemberDAO,
     private val selfApi: SelfApi,
     private val userDetailsApi: UserDetailsApi,
     private val upgradePersonalToTeamApi: UpgradePersonalToTeamApi,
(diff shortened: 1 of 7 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -508,15 +510,27 @@ internal class UserDataSource internal constructor(
...
                 allRecipients.forEach {
                     if (teamMateIds.contains(it.id)) teamRecipients.add(it)
                     else otherRecipients.add(it)
                 }
+
+                // Ensure we add all users in the list even if they do no have a client id stored locally.
+                otherRecipients.addAll(
+                    allUsersWithConversations
+                        .filter { otherUserId -> teamRecipients.none { it.id == otherUserId } }
+                        .filter { otherUserId -> otherRecipients.none { it.id == otherUserId } }
```
harrytmthy, line 530: These `.none { ... }` within `.filter { ... }` creates O(N²) complexity, since `.none { ... }` does a linear scan for every user in the outer list, which could lead to noticeable slowdowns if the list is large.

One way to improve this would be:

```kotlin
val teamRecipientIds = teamRecipients.mapTo(mutableSetOf()) { it.id }
val otherRecipientIds = otherRecipients.mapTo(mutableSetOf()) { it.id }

otherRecipients.addAll(
    allUsersWithConversations
        .filter { it !in teamRecipientIds && it !in otherRecipientIds }
        .map { Recipient(it, emptyList()) }
)
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/user/UserRepository.kt
```diff
@@ -508,15 +510,27 @@ internal class UserDataSource internal constructor(
...
             }.map { allRecipients ->
                 val teamRecipients = mutableListOf<Recipient>()
                 val otherRecipients = mutableListOf<Recipient>()
+
                 allRecipients.forEach {
                     if (teamMateIds.contains(it.id)) teamRecipients.add(it)
                     else otherRecipients.add(it)
                 }
+
+                // Ensure we add all users in the list even if they do no have a client id stored locally.
```
harrytmthy, line 526: Should be `if they do not have...`

saleniuk (approved):

yamilmedina (approved):

PR #3612, opened 2025-09-03
chore(conversation-history): persist retention settings for channels [WPB-18408]

vitorhugods:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18408" title="WPB-18408" target="_blank">TaskWPB-18408</a>  [Android] Persist conversation history settings
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

We need to store the configurations of a history sharing for a conversation.

## Data structure to represent History Sharing:

- sealed interface: `ConversationHistorySharing`, which can be enabled or disabled (`Private` or `ShareWithNewMembers`).
- Currently, this value is only accessible for `Channel` `ConversationDetails`.

## Store history retention into the DB

- Add a column to the Conversation table: `historySharingRetentionSeconds` to represent the amount of seconds that a conversation history should be shared.

## Testing

Added integration tests for `ConversationDataSource`. Instead of mocking all the dependencies, just use a test database and do the proper end-to-end test: store and get the expected results back.

## Summary by CodeRabbit

- New Features
  - Per-channel conversation history sharing added. Channel owners can keep history private or share recent messages with new members for a configurable retention period.
  - Conversation and channel details now reflect the current history-sharing setting.
  - Existing conversations default to private (no history sharing).

- Chores
  - Background data updates to support history-sharing settings (no user action required).

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
@@ -43,6 +43,7 @@ import com.wire.kalium.logic.di.MapperProvider
 import com.wire.kalium.network.api.authenticated.conversation.ChannelAddPermissionTypeDTO
 import com.wire.kalium.network.api.authenticated.conversation.ConvProtocol
 import com.wire.kalium.network.api.authenticated.conversation.ConvTeamInfo
+import com.wire.kalium.network.api.authenticated.conversation.ConversationHistorySettingsDTO
 import com.wire.kalium.network.api.authenticated.conversation.ConversationResponse
 import com.wire.kalium.network.api.authenticated.conversation.CreateConversationRequest
 import com.wire.kalium.network.api.authenticated.conversation.GroupConversationType
@@ -66,6 +67,7 @@ import com.wire.kalium.util.time.UNIX_FIRST_DATE
 import io.mockative.Mockable
 import kotlinx.datetime.Instant
 import kotlinx.datetime.toInstant
+import kotlin.time.Duration.Companion.seconds
 import kotlin.time.DurationUnit
 import kotlin.time.toDuration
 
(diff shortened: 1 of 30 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -156,6 +159,10 @@ internal class ConversationMapperImpl(
             channelAccess = null, // TODO: implement when api is ready
             channelAddPermission = apiModel.channelAddUserPermissionTypeDTO?.toDAO(),
             wireCell = conversationId.toString().takeIf { apiModel.cellEnabled() }, // TODO refactor to boolean in WPB-16946
+            historySharingRetentionSeconds = when (val historySharingSettings = apiModel.historySharingSettings) {
```
sbakhtiarov, line 162: Should this piece of logic be placed in a use case rather than a mapper?

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
```diff
@@ -49,7 +49,8 @@ CREATE TABLE Conversation (
     channel_access TEXT AS ConversationEntity.ChannelAccess DEFAULT NULL,
     channel_add_permission TEXT AS ConversationEntity.ChannelAddPermission DEFAULT NULL,
     wire_cell TEXT,
-    deleted_locally INTEGER AS Boolean NOT NULL DEFAULT 0
+    deleted_locally INTEGER AS Boolean NOT NULL DEFAULT 0,
+    history_sharing_retention_seconds INTEGER NOT NULL DEFAULT 0 CHECK (history_sharing_retention_seconds >= 0)
```
sbakhtiarov, line 53: Should we expect more channel-specific data we need to store? Maybe we should consider moving these into a new table (ChannelDetails)?

sbakhtiarov (approved):

persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
vitorhugods: Not that I know of.
It's an artificial boundary, to be honest... _Currently_ channels are the only kind of conversation that we allow to have history. But there is no real difference / nothing preventing other MLS conversations to have it enabled as well.

The DB doesn't need to be super modeled around this artificial distinction. We can do that on "domain" layer above. If we start allowing other conversations to have history sharing, we don't need to update the DB.

saleniuk (approved):

PR #3620, opened 2025-09-08
fix: missed call notification for builds (WPB-19996)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19996" title="WPB-19996" target="_blank">BugWPB-19996</a>  [Android] Federated 1:1 call no missed call notification
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-19996

# What's new in this PR?

### Issues
When 1:1 call is cancelled by caller before the call is established receiver does not get a "Missed call" message.

### Causes (Optional)
1:1 calls on c env are handled as conference calls. The client does not get a MISSED (AVS NORMAL) call status and call close logic does not handle this.

### Solutions
This is a workaround just for c builds, must not be merged back to develop.

1. Do not check the conversation type when checking the persist missed call condition.
2. Avoid overwriting REJECTED call status by STILL_ONGOING for 1:1 calls


saleniuk (approved):
Would be nice to add a few tests for these cases, but well done overall 👏

MohamadJaara (approved):

PR #3651, opened 2025-10-07
feat: asset audit log feature flag (WPB-20296)

sbakhtiarov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20296" title="WPB-20296" target="_blank">EpicWPB-20296</a>  [Android] Adopt new API with Enterprise File Upload Metadata
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-20296

# What's new in this PR?

- Support for new feature flag for Asset Audit Log feature.
- Use case for checking if extra parameters need to be included
- No metadata changes, so we can merge this without any impact on asset upload

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/IsAssetAuditLogEnabledUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/IsAssetAuditLogEnabledUseCase.kt
@@ -1,0 +1,45 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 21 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/IsAssetAuditLogEnabledUseCase.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+package com.wire.kalium.logic.feature.client
+
+import com.wire.kalium.common.functional.fold
+import com.wire.kalium.logic.configuration.UserConfigRepository
+import com.wire.kalium.logic.configuration.server.ServerConfigRepository
+import com.wire.kalium.logic.data.user.UserId
+
+/**
+ * Use case to check if Simplified Connection Request feature is enabled.
```
yamilmedina, line 26:
```suggestion
 * Use case to check if the Assets audit log feature is enabled.
```

yamilmedina (approved):
lgtm 🚀 ! Just a nitpick for the comment of the use case `IsAssetAuditLogEnabledUseCase` as it slipped from another one.

MohamadJaara (approved):

PR #3764, opened 2025-12-09
refactor: build waveform after fetching audio asset [WPB-22118]

saleniuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22118" title="WPB-22118" target="_blank">BugWPB-22118</a>  [Android] Some audio files to not generate an audio wave when shared on android
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Waveforms generation can be executed each time the audio message is shown on the screen.

### Causes (Optional)

Generation is executed from the ViewModel so it can be called multiple times. It can cause some problems and unnecessary use of resources when user sends an audio message - then first waveform build attempt is done while sending, and then another attempt can be called right after when the message is shown.

### Solutions

Change the logic so that the waveform is generated right after downloading the audio asset. This way it's executed only once and never retries if it failed (no need - if it failed the first time then it will fail each next time for that same exact audio file). Generation is executed inside a newly created worker which is scheduled as an one-time job right when the successful download is finished, so that it can be executed in the background (it's not crucial for the app to have it instantly). It required one small change in `AssetRepository` for "fetch asset" methods to return info whether the asset was downloaded during this execution or was already available locally - if just downloaded and doesn't contain a waveform then schedule building it.

#### Test Coverage (Optional)

#### How to Test

Send audio message from app that doesn't yet support sending waveform along with the audio message, so some old Android app like 4.17.0 or iOS. After opening that conversation and receiving that message on new Android app, it should fetch the audio file and then generate a waveform for it.

----

----

----
##### References


yamilmedina (approved):
I left approved 🚀  Just a small suggestion about the Result type. If it makes sense we could do it and in case is not big of a change :D

Garzas (approved):

PR #3858, opened 2026-02-10
fix: keep decrypting banner until pending events are fully drained [WPB-23396]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-23396

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- The "Decrypting messages" banner could disappear too early during incremental sync catch-up, even when pending events were still waiting to be processed/decrypted.
- This was especially visible with batched catch-up traffic (temporary gaps between pending pages) and unstable/slower network conditions.

### Causes (Optional)

- Sync was allowed to transition to `LIVE` when `IsUpToDate` was observed, but `IsUpToDate` can appear in temporary gaps between pending batches.
- The transition condition did not enforce a hard check that pending (`is_live = 0`) events were actually drained.
- After entering live mode, repeated `IsUpToDate` emissions could still trigger unnecessary checks.

### Solutions

- Made the `LIVE` transition state-based:
  - On `IsUpToDate`, we now verify whether unprocessed pending events still exist.
  - We only emit `LIVE` when pending events are truly empty.
- Added a dedicated DAO/repository check for pending backlog using SQL `EXISTS(...)` for minimal overhead (`hasUnprocessedPendingEvents`).
- Added a guard to avoid re-running pending checks after `LIVE` has already been reached in the same worker session.

### DIFF
--- a/data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Events.sq
+++ b/data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Events.sq
@@ -51,3 +51,9 @@ LIMIT 500;
 deleteUnprocessedLiveEventsByIds:
 DELETE FROM Events
 WHERE is_processed = 0 AND is_live = 1 AND event_id IN ?;
+
+hasUnprocessedPendingEvents:
+SELECT EXISTS(
+    SELECT 1 FROM Events
+    WHERE is_processed = 0 AND is_live = 0
+);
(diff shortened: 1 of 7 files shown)


data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Events.sq
```diff
@@ -51,3 +51,9 @@ LIMIT 500;
 deleteUnprocessedLiveEventsByIds:
 DELETE FROM Events
 WHERE is_processed = 0 AND is_live = 1 AND event_id IN ?;
+
+hasUnprocessedPendingEvents:
+SELECT EXISTS(
+    SELECT 1 FROM Events
+    WHERE is_processed = 0 AND is_live = 0
```
MohamadJaara: i have no idea how big this table can get, but is it worth it to add an index for is_processed and is_live ?

maybe one for where whose 2 values are 0 ?

sbakhtiarov (approved):

PR #3898, opened 2026-02-23
fix(auth): drain UserSessionScope before wiping DB on hard logout [WPB-8643]

mastaab:
https://wearezeta.atlassian.net/browse/WPB-8643

## Summary

- Cancel and join `UserSessionScope` before calling `wipeAllData()` on any logout path that nukes the database
- Remove the `delay(CLEAR_DATA_DELAY)` (1 s) band-aid that was attempting to paper over the same race

## Root Cause

On any wipe path (`SELF_HARD_LOGOUT`, `REMOVED_CLIENT`/`DELETED_ACCOUNT` with `wipeOnDeviceRemoval`, `SESSION_EXPIRED` with `wipeOnCookieInvalid`), `wipeAllData()` → `ClearUserDataUseCase` → `UserDatabaseBuilder.nuke()` was called while `UserSessionScope` coroutines were still in-flight.

`nuke()` correctly closes the SQLite driver and deletes `main.db`. However, coroutines in the session scope still holding JDBC connections to that path would subsequently prepare a statement — causing SQLite JDBC to **auto-create a new empty 0-byte `main.db`** at the deleted path. The next login found this ghost file, skipped schema creation (file exists), and failed with:

```
SQLITE_ERROR: no such table: SelfUser
```

This was confirmed by per-file size instrumentation at every `nuke()` and `userDatabaseBuilder` call:

```
[UserDB] nuke() called — deleting DB files
[UserDB] deleteUserDatabaseFiles: main.db existed=true deleted=true
[UserDB] deleteUserDatabaseFiles: main.db-wal existed=false deleted=true   ← no stale sidecars
[UserDB] nuke() result=true
...
[UserDB] Opening: main.db=0B wal=absent ...   ← ghost file created post-nuke
```

The `delay(1000L)` in `wipeAllData()` was an attempt to avoid this race but is not reliable under load.

## Fix

Before calling `wipeAllData()`, cancel the `UserSessionScope` and await full completion:

```kotlin
if (willWipeData) {
    userSessionScopeProvider.get(userId)?.let { scope ->
        scope.cancel()
        scope.coroutineContext.job.join() // wait for all in-flight coroutines to stop
    }
}
// DB is now safe to nuke — no coroutines can write to the path
wipeAllData()
```

After the fix, re-login finds `main.db=absent` (no ghost file) and creates a fresh schema cleanly.

## Relation to PR #3897

PR #3897 (`fix(persistence): delete SQLite sidecar files on JVM user DB nuke`) is an independent correctness fix for stale WAL/SHM files after a process kill. It does **not** address the race fixed here. Both fixes are orthogonal and both are worth merging.

## Testing

- All existing `LogoutUseCaseTest` tests pass
- Added `givenHardLogout_whenScopeExists_thenScopeGetIsCalledToAllowCancellationBeforeWipe` to verify `userSessionScopeProvider.get()` is invoked (scope cancel path taken) before `clearUserDataUseCase` on the wipe path

Closes wireapp/wiretui#19


MohamadJaara:
closed for https://github.com/wireapp/kalium/pull/4080

PR #3960, opened 2026-03-12
fix: sync established state/epoch after orphan welcome recovery [WPB-24051]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-24051

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Android could attempt `join/establish` for an MLS conversation even when the local MLS group was already established in core-crypto.
- After `OrphanWelcome` recovery via external commit, local conversation metadata could stay partially out of sync (`groupState`/`epoch` in DB), which increased chances of wrong follow-up decisions.

### Causes (Optional)

- The decision path in `JoinExistingMLSConversationUseCase` could rely on stale local metadata for join/establish branching.
- `joinGroupByExternalCommit` updated only `groupState`, without syncing the current local MLS epoch from core-crypto to DB.

### Solutions

- Added a guard in join/establish flow: if the MLS group already exists locally in core-crypto, skip join/establish.
- Added local state sync in that guard: when DB `groupState != ESTABLISHED`, update it to `ESTABLISHED`.
- Updated `joinGroupByExternalCommit` to also read current local epoch from core-crypto and persist it to DB (`updateMLSGroupIdAndState`) when the conversation exists locally.
- Added/updated unit tests for:
  - skipping join/establish when local group exists,
  - syncing DB `groupState` to `ESTABLISHED` in that case,
  - persisting epoch after successful external commit join,
  - fallback behavior when local conversation row is missing.


MohamadJaara (approved):
LGTM i have few comments about updating with the same gorup id that can be nicer

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/MLSConversationRepository.kt
```diff
@@ -368,12 +368,27 @@ internal class MLSConversationDataSource(
...
+            val existingConversation = conversationDAO.getConversationByGroupID(localGroupId)
+            if (existingConversation != null) {
+                conversationDAO.updateMLSGroupIdAndState(
+                    existingConversation.id,
+                    localGroupId,
+                    localEpoch.toLong(),
+                    ConversationEntity.GroupState.ESTABLISHED
+                )
+            } else {
+                conversationDAO.updateConversationGroupState(
```
sbakhtiarov, line 387: We have this fallback here if existingConversation is null. Should we have the same fallback if mlsContext.conversationEpoch returns Failure?

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/JoinExistingMLSConversationUseCase.kt
```diff
@@ -153,6 +153,25 @@ internal class JoinExistingMLSConversationUseCaseImpl(
         return when {
             protocol !is Conversation.ProtocolInfo.MLSCapable -> Either.Right(Unit)
 
+            transactionContext.wrapInMLSContext { mlsContext ->
+                mlsConversationRepository.hasEstablishedMLSGroup(mlsContext, protocol.groupId)
+            }.fold({ false }, { it }) -> {
```
sbakhtiarov, line 158: This use case worth some refactoring. It took me too long to figure out what is going on.
I suggest:
Removing hard to read conditions in when block. Could use a new function:
hasEstablishedMLSGroup(context, groupId) -> {
...
}

Creating a function for each when block, so that it is easier to follow the main logic:

```
when {
  protocol !is Conversation.ProtocolInfo.MLSCapable -> Either.Right(Unit)

  hasEstablishedMLSGroup(context, groupId) -> handleAlreadyEstablishedConversation()

  protocol.epoch != 0UL -> joinByExternalCommit()
}

```

sbakhtiarov (approved):

PR #3973, opened 2026-03-17
fix: sync MLS metadata from core-crypto in join and stale verification [WPB-24051]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-24051

----

- The **PR Title**

- The **PR Description**
----

### Issues

- MLS recovery could still rely on stale local conversation metadata even when the MLS group already existed in core-crypto.
- `JoinExistingMLSConversationUseCase` only synced local DB state partially in that case, which left room for `groupState` / `epoch` drift.
- `StaleEpochVerifier` did not consistently use the freshest MLS metadata source when deciding whether a local group was actually stale.

### Causes (Optional)

- Local persistence, backend metadata, and core-crypto state are not updated atomically and can temporarily drift apart.
- When a conversation already existed locally in core-crypto, Android previously updated only part of the local metadata, instead of fully synchronizing it from the actual local MLS state.
- Some recovery decisions were therefore made from stale or incomplete local metadata rather than from the most reliable available source.

### Solutions

- Updated `JoinExistingMLSConversationUseCase` so that when the MLS group already exists locally, Android synchronizes conversation metadata from core-crypto instead of only forcing `ESTABLISHED`.
- Added retrieval of the local MLS epoch from `MLSConversationRepository` and used it to update persisted metadata consistently.
- Updated `StaleEpochVerifier` integration and related wiring/tests so stale verification and join/recovery logic use the synchronized metadata model.
- Added/adjusted tests covering:
  - syncing established MLS metadata from local core-crypto state,
  - avoiding unnecessary metadata updates when local and persisted state already match,
  - stale verification behavior after the metadata synchronization changes.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/JoinExistingMLSConversationUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/JoinExistingMLSConversationUseCase.kt
@@ -45,6 +45,7 @@ import com.wire.kalium.logic.featureFlags.FeatureSupport
 import com.wire.kalium.logic.sync.receiver.conversation.message.MLSMessageFailureHandler
 import com.wire.kalium.logic.sync.receiver.conversation.message.MLSMessageFailureResolution
 import com.wire.kalium.network.api.base.authenticated.conversation.ConversationApi
+import com.wire.kalium.persistence.dao.conversation.ConversationEntity
 import com.wire.kalium.util.KaliumDispatcher
 import com.wire.kalium.util.KaliumDispatcherImpl
 import io.mockative.Mockable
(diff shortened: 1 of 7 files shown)


sbakhtiarov (approved):

MohamadJaara:
One thing I noticed: the old flow used `FetchConversationUseCase` which fetched *and* persisted the conversation to DB before reading protocol info back. The new flow calls `conversationRepository.fetchConversation()` directly and uses the response without persisting anything.

So stale epoch verification no longer keeps the local DB in sync with backend state as a side effect. I assume `joinExistingMLSConversation` handles that eventually, but there's a wider window where DB can be stale now. Was this intentional? If so, might be worth a quick comment explaining why we skip persistence here.

MohamadJaara (changes requested):
one comment about a case where this change may not work

MohamadJaara (approved):

PR #3985, opened 2026-03-20
fix: remote muting not working [WPB-3185]

saleniuk:
https://wearezeta.atlassian.net/browse/WPB-3185

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When an admin tries to remotely mute a participant, it doesn't work in Android app and the participant is not muted.

### Causes (Optional)

Android app, when it was implemented, was making use of the fact that AVS had the logic of handling such type of message, so this message was just passed to AVS, but after some time this logic was removed from AVS as it was not in accordance to the specs: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/969605169/Use+case+conversation+admin+mutes+a+remote+participant#Step-3%3A-Targeted-client-receives-a-message-and-mutes-itself

### Solutions

App needs to call mute itself when receiving `REMOTEMUTE` message type. In specs there's also a part of passing this message to AVS, but it results in a failure and after discussing it with the AVS team, it turned out that this step isn't necessary, web doesn't do that and passing it to AVS results in a failure as it isn't actually a calling message, it just happens to have the same content type, maybe just so that it's grouped with calling messages as it's correlated.
In this PR, handling of messages with `Calling` content is extracted to a dedicated handler, just like we have for all other message content types. Inside this handler, it's determined whether it's actually a calling message - in that case pass it to AVS, or `REMOTEMUTE` - then execute the `MuteCallUseCase`.

#### Test Coverage (Optional)

#### How to Test

Join a call, unmute yourself and let admin mute you remotely.

----

----

----
##### References

### DIFF
--- a/logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
+++ b/logic/src/commonJvmAndroid/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
@@ -49,7 +49,6 @@ import com.wire.kalium.logic.data.call.mapper.CallMapper
 import com.wire.kalium.logic.data.call.mapper.ParticipantMapperImpl
 import com.wire.kalium.logic.data.conversation.ClientId
 import com.wire.kalium.logic.data.conversation.Conversation
-import com.wire.kalium.logic.data.conversation.ConversationRepository
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.id.CurrentClientIdProvider
 import com.wire.kalium.logic.data.id.FederatedIdMapper
@@ -93,12 +92,10 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
 import kotlinx.coroutines.cancel
-import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.joinAll
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.datetime.Instant
-import kotlinx.serialization.json.Json
 import java.util.Collections
 
 @Suppress("LongParameterList", "TooManyFunctions")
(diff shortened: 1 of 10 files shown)


MohamadJaara (commented):
in the `ShouldRemoteMuteCheckerImpl` is see without any checks just wrongly  assuming if there is not target then mute self, what is preventing someone from sending a message with no targets to an MLS/proteus  group and muting everyone or i am missing something here or this is something we want to preserver to mute app 


```kotlin
            } ?: run {
                // If there are no targets, we should mute. It's a proteus message with no targets.
                true
            }
```

saleniuk:
> in the `ShouldRemoteMuteCheckerImpl` is see without any checks just wrongly assuming if there is not target then mute self, what is preventing someone from sending a message with no targets to an MLS/proteus group and muting everyone or i am missing something here or this is something we want to preserver to mute app
> 
> ```kotlin
>             } ?: run {
>                 // If there are no targets, we should mute. It's a proteus message with no targets.
>                 true
>             }
> ```

@MohamadJaara there are two things:
- only an admin can mute others, and it's already being checked before that in `ShouldRemoteMuteCheckerImpl`, so it's not like anyone can mute others in a call
- for proteus message, it's needed to pass targets when sending a message so it's only sent to the clients which need to be muted, for MLS it's being sent to all and then client itself needs to verify whether this message targets itself or should be ignored, so if the client receives a proteus mute message then it's sure that it's meant for that client so no additional check should be required

But you're right with one thing - it's possible to send MLS message with no targets and our logic will treat it as a proteus message and mute even if the current client is not a target - this one needs to be changed. 
Docs don't say that the `targets` should be added and checked only for MLS conversations: 
https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/969605169/Use+case+conversation+admin+mutes+a+remote+participant
https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/969605169/Use+case+conversation+admin+mutes+a+remote+participant
I checked and targets are being sent also for proteus, so I adjusted the code to check targets for proteus as well.

MohamadJaara (approved):

Garzas (approved):

PR #3989, opened 2026-03-25
fix: harden one-on-one MLS resolve failure handling [WPB-24176]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Slow Sync could leave one-on-one MLS resolution in an inconsistent state when batch resolution hit recoverable failures.
- In particular, `MessageRejected(StaleMessage)` during `RESOLVE_ONE_ON_ONE_PROTOCOLS` could be skipped at batch level even though the conversation might still be recoverable.
- Batch handling for one-on-one resolution was also too broad for some failure types such as retryable federated errors and generic `ServerMiscommunication`.

### Causes (Optional)

- One-on-one resolution during Slow Sync is processed as a batch, and some failures were treated as “skip and continue” too early.
- This could allow Slow Sync to complete while a specific one-on-one conversation remained unresolved locally.
- Some error handling in this flow was less strict or less specific than in other MLS join/recovery paths.

### Solutions

- Added targeted retry scheduling for one-on-one resolution when batch processing hits `MessageRejected(StaleMessage)`.
- Added the same retry behavior for `MissingKeyPackages` and retryable federated failures.
- Tightened `ServerMiscommunication` handling:
  - invalid request errors are still skipped
  - other server miscommunication failures now fail the batch instead of being silently ignored
- Hardened the delayed retry path so it exits safely if incremental sync state completes before reaching `Live`.
- Added regression tests covering stale message, retryable federated failures, and server miscommunication handling.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/OneOnOneResolver.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/OneOnOneResolver.kt
@@ -98,6 +98,7 @@ internal class OneOnOneResolverImpl(
     private val oneOnOneProtocolSelector: OneOnOneProtocolSelector,
     private val oneOnOneMigrator: OneOnOneMigrator,
     private val incrementalSyncRepository: IncrementalSyncRepository,
+    private val pendingOneOnOneResolutionsRepository: PendingOneOnOneResolutionsRepository,
     kaliumDispatcher: KaliumDispatcher = KaliumDispatcherImpl
 ) : OneOnOneResolver {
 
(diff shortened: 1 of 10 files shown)


yamilmedina (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/OneOnOneResolver.kt
```diff
@@ -130,19 +130,70 @@ internal class OneOnOneResolverImpl(
...
+            }
+            Either.Right(Unit)
+        }
+
+        is NetworkFailure.FederatedBackendFailure.RetryableFailure -> {
+            kaliumLogger.w(
+                "Resolving one-on-one failed $failure for ${user.id.toLogString()}, " +
+                    "scheduling retry after incremental sync is live"
+            )
+            scheduleResolveOneOnOneConversationWithUserId(transactionContext, user.id)
```
MohamadJaara: would passing the transactionContext here create a case where a transaction that is used in slow sync is leaked and can be used after it is closed? 
e.g. 
1. Start slow sync
2. scheduleResolveOneOnOneConversationWithUserId(transactionContext, user.id) is called for any reason
3. Slow sync commit the transaction
4. the schedule is called after the commit KaBoom!

MohamadJaara (changes requested):

Garzas:
Improved solution for develop: https://github.com/wireapp/kalium/pull/4001

PR #3993, opened 2026-03-25
fix: restart slow sync after MLS client registration and E2EI finalize [WPB-20462]

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

During MLS migration testing on Android, a user who was logged out during migration could not properly recover access to an MLS group after logging back in.

- The user was not able to receive messages in the migrated group.
- The group started working only after the user sent a message from Android.
- In practice, the missing recovery after login left the local MLS state incomplete until another action forced it forward.

### Causes (Optional)

- After the client became MLS-capable again, Android did not reliably trigger the recovery path that depends on rerunning `SlowSync`.
- Because of that, the local MLS group state was not rebuilt early enough after login.
- In the E2EI path, the same missing `SlowSync` restart could happen after MLS client finalization.
- `MLSClientManager` could also restart `SlowSync` in the wrong moment, even when MLS registration was still blocked by `E2EICertificateRequired`.

### Solutions

- Restart `SlowSync` after successful MLS client verification/registration in the login flow by clearing `lastSlowSyncCompletionInstant`.
- Restart `SlowSync` after `FinalizeMLSClientAfterE2EIEnrollment`, so the E2EI path uses the same recovery trigger.
- Ensure `MLSClientManager` clears the slow sync timestamp only for real MLS registration success, not for `E2EICertificateRequired`.
- Add logs around the slow sync reset points to make this recovery trigger visible during debugging.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/Event.kt
@@ -522,7 +522,7 @@ internal sealed class Event(open val id: String) {
             val model: MLSMigrationModel
         ) : FeatureConfig(id) {
             override fun toLogMap(): Map<String, Any?> = mapOf(
-                typeKey to "FeatureConfig.MLSUpdated",
+                typeKey to "FeatureConfig.MLSMigrationUpdated",
                 idKey to id,
                 featureStatusKey to model.status.name,
                 "startTime" to model.startTime,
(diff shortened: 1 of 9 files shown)


MohamadJaara:
> Restart SlowSync after successful MLS client verification/registration in the login flow by clearing lastSlowSyncCompletionInstant.

I feel like this should be regardless of the MLS status after login always make sure slow sync is triggered.

MohamadJaara (approved):
Resolved in 1:1 for 4.22 we will do it this way, and for develop, we will trigger slow sync when a new client is created regardless.

CinnamonNinja (approved):

PR #4001, opened 2026-03-27
fix: queue mls conversations recovery retries for foreground processing [WPB-24176]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-24176

----

- The **PR Title**

- The **PR Description**
----

**What's new in this PR?**

**Issues**

- MLS conversation recovery during Slow Sync could leave some conversations unresolved when recoverable failures happened in batch processing.
- Recovery state for skipped MLS recovery work was not persisted, so retries could be lost after process death.
- Foreground recovery stopped on first failure, which could starve later pending items.

**Causes (Optional)**

- Recoverable MLS recovery failures were not isolated from non-recoverable flow.
- No durable queue existed for deferred MLS recovery actions.
- Recovery execution used fail-fast semantics and acknowledged only full-batch success.

**Solutions**

- Improved batch failure handling for recoverable MLS cases:
  - MissingKeyPackages
  - FederatedBackendFailure.RetryableFailure
  - MLSFailure.MessageRejected(StaleMessage)
- Replaced in-memory pending recovery queue with persisted pending actions in user DB.
- Added dedicated PendingActions table + DAO/repository flow for deferred MLS recovery.
- Added foreground recovery use cases that process pending MLS recovery actions in fresh transactions:
  - one-on-one MLS conversation resolution
  - MLS group conversation join
- Changed recovery semantics to best-effort:
  - attempt all pending items,
  - acknowledge only successfully recovered items.
- Wired pending MLS recovery into foreground actions.


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/PendingOneOnOneResolutionsRepository.kt
```diff
@@ -0,0 +1,44 @@
...
+    override suspend fun enqueue(userId: UserId) {
+        mutex.withLock {
+            pendingUserIds.add(userId)
+        }
+    }
+
+    override suspend fun dequeueAll(): Set<UserId> = mutex.withLock {
+        pendingUserIds.toSet().also { pendingUserIds.clear() }
+    }
+}
```
MohamadJaara: this means the resolving will be lost to the void in case of process death is it needed to be in memory or we can persist the q in the DB for example?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/mls/OneOnOneResolver.kt
```diff
@@ -156,26 +157,71 @@ internal class OneOnOneResolverImpl(
...
             kaliumLogger.e("Resolving one-on-one failed $failure, skipping")
             Either.Right(Unit)
         }
 
+        failure is NetworkFailure.ServerMiscommunication -> {
+            if (failure.kaliumException is KaliumException.InvalidRequestError) {
+                kaliumLogger.e("Resolving one-on-one failed $failure, skipping")
+                Either.Right(Unit)
+            } else {
+                kaliumLogger.e("Resolving one-on-one failed $failure, retrying")
```
sbakhtiarov: Is this retry on purpose?

sbakhtiarov (approved):

MohamadJaara:
we just need to make sure MLS groups that the user leave are not retried as part of this since they will technically not be established in CC, but other than this all good

MohamadJaara (approved):

PR #4004, opened 2026-03-30
fix: retry MLS conversation reset on stale-message [WPB-24434]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-24434

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Previously, reset used the local MLS/CoreCrypto epoch, which could trigger `409 mls-stale-message` on `POST /mls/reset-conversation` when the backend had already advanced. The new flow keeps the normal path unchanged, but if reset fails specifically with `mls-stale-message`, it refetches the conversation from the backend and retries once with the remote epoch.


logic/src/commonTest/kotlin/com/wire/kalium/logic/data/conversation/ResetMLSConversationUseCaseTest.kt
```diff
@@ -317,7 +380,11 @@ class ResetMLSConversationUseCaseTest {
 
             coEvery {
                 conversationRepository.resetMlsConversation(any(), any())
-            } returns Unit.right()
+            }.thenReturnSequentially(*resetConversationResults.toTypedArray())
+
+            coEvery {
+                conversationRepository.fetchConversation(any())
+            } returns remoteConversationResponse.right()
```
MohamadJaara: Should this be as a fallback since, as I understand it, in most cases this epoch would be old? or this is only a rare case ?

typfel: I'd suggest to always go for the remote epoch if it would simplify the code. 

MLS resets should be rare so there shouldn't be a performance concern here.

Garzas: This should stay a fallback. While reviewing the Confluence use case in more detail, I checked the full reset and rejoin flow and a stale epoch does not look like the expected normal path. It should only happen in out-of-sync or race-condition cases.

During that review I also found a separate issue in the PENDING_AFTER_RESET rejoin flow: we were deciding from local metadata without refreshing it first. I fixed that too.

MohamadJaara: still imho we should relay less and less on local state for epoch in recovery cases it can always be a little issue, and getting it from remote in this case is quite rare and should not affect it a lot.

typfel: > This should stay a fallback. While reviewing the Confluence use case in more detail, I checked the full reset and rejoin flow and a stale epoch does not look like the expected normal path. It should only happen in out-of-sync or race-condition cases.

The green line here is android doing resets supplying an out of date epoch so I mean it does happen. MLS reset are fairly rare (like 10 per day from all android clients on wire cloud) IMO this doesn't justify having a fallback logic when we can simplify the code by always taking the `getRemoteResetInfo` path.

Screenshot 2026-03-31 at 09 15 39

Garzas: Using the remote epoch every time is fine and it does simplify the code. I would just treat it more as making the flow safer than as fixing the real issue. The Android spike still suggests that we should understand why local MLS state gets out of date this often on Android in the first place

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ResetMLSConversationUseCase.kt
```diff
@@ -167,6 +167,94 @@ internal class ResetMLSConversationUseCaseImpl(
             }
     }
 
+    private suspend fun getLocalResetInfo(
```
typfel, line 170: I thought we agreed to only rely on remote reset info?

typfel (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ResetMLSConversationUseCase.kt
typfel: But as implemented now it doesn't add that much complexity so it's fine IMO.

MohamadJaara (approved):

PR #4007, opened 2026-03-31
fix: remove local asset path from message read model [WPB-18914]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-18914

----

- The **PR Title**

- The **PR Description**
----

### Issues

- Scrolling long conversations could trigger repeated Paging refreshes and viewport churn while media assets were being fetched in the background.
- Message read models were carrying local asset file paths, which mixed persistent message state with ephemeral local storage state.
- Audio and media-related flows depended on local asset paths being present in message models, which made the system fragile and caused UI side effects such as preview/waveform flicker.

### Causes (Optional)

- `MessageDetailsView` depended on the `Asset` table, so writes to locally cached assets invalidated the conversation message query and caused PagingSource refreshes during scroll.
- `assetDataPath` was propagated through the read model stack (`MessageDetailsView` -> persistence message entity -> asset mapping), even though it represents local runtime/cache state rather than message content.
- Some consumers, especially audio-related paths, still assumed local asset data could be read directly from message models instead of being resolved on demand.

### Solutions

- Removed the dependency between conversation message paging and local asset cache state by dropping the `Asset` join from `MessageDetailsView` and adding a forward database migration.
- Removed local asset path from message read models, including `MessageEntityContent.Asset` and `AssetContent`.
- Updated asset and audio flows to resolve local files through `GetMessageAssetUseCase` on demand instead of relying on local paths embedded in message models.
- Adjusted audio normalized loudness generation and attachment handling paths so they no longer depend on local asset data being stored inside the message read model.
- Tightened Cells attachment cleanup logic to operate explicitly on `CellAssetContent` where local file paths are still valid and expected.


saleniuk (approved):

MohamadJaara (approved):
one less join this the view 🚀

PR #4009, opened 2026-04-01
feat: support web auth transport cookies and versionless access [WPB-23964]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-23964

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Web auth transport was not fully aligned with browser cookie handling.
- The access-token refresh flow was going through the default API versioning path, even though the correct browser contract is the versionless `/access` endpoint.

### Causes (Optional)

- Kalium applies `/v{commonApiVersion}/` automatically to most REST requests.
- That is correct for regular API endpoints, but not for the browser refresh/access-token flow.
- Cookie handling also differs between browser and native clients, so the existing inline refresh-cookie logic was too rigid for web.

### Solutions

- Added a small platform-specific auth helper layer for browser credentials and refresh-cookie handling.
- Introduced a JS implementation for browser-managed cookie behavior, while keeping native platforms aligned with the previous manual cookie flow.
- Added an explicit way to skip API version prefixing for special-case endpoints.
- Updated the access-token refresh flow to use the correct versionless `/access` path.
- Switched the affected auth-related requests to the new helpers for more consistent web-compatible behavior.

### DIFF
--- a/data/network/src/commonMain/kotlin/com/wire/kalium/network/api/v0/authenticated/AccessTokenApiV0.kt
+++ b/data/network/src/commonMain/kotlin/com/wire/kalium/network/api/v0/authenticated/AccessTokenApiV0.kt
@@ -33,7 +33,7 @@ import io.ktor.client.request.post
 internal open class AccessTokenApiV0(private val httpClient: HttpClient) : AccessTokenApi {
     override suspend fun getToken(refreshToken: String, clientId: String?): NetworkResponse<Pair<AccessTokenDTO, RefreshTokenDTO?>> =
         wrapKaliumResponse<AccessTokenDTO> {
-            httpClient.post(PATH_ACCESS) {
+            httpClient.post("/$PATH_ACCESS") {
                 skipApiVersion()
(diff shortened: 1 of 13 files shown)


data/network/src/commonMain/kotlin/com/wire/kalium/network/api/v0/authenticated/AccessTokenApiV0.kt
```diff
@@ -53,6 +54,6 @@ internal open class AccessTokenApiV0(private val httpClient: HttpClient) : Acces
         }
 
     private companion object {
-        const val PATH_ACCESS = "access"
+        const val PATH_ACCESS = "/access"
```
saleniuk: We usually keep these const vals without any slashes and we add slashes in the httpClient method where we use these const vals, maybe we should keep it unified 😄

saleniuk (approved):

sbakhtiarov (approved):

PR #4010, opened 2026-04-01
feat: add browser-safe js file util shims [WPB-23964]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-23964

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Some JS runtime helpers still assumed native filesystem behavior.
- On web, this could lead to unsupported file operations and incorrect persistent-storage cleanup.
- JS tests also needed an IndexedDB bootstrap in the Node-based test environment.

### Causes (Optional)

- Native targets use file-based storage, while web uses browser-managed storage like IndexedDB.
- The existing JS helpers were either unimplemented or too close to native filesystem semantics.
- The shared `FileUtil` API had no web-specific persistent cleanup path.

### Solutions

- Added `deletePersistentDirectory(...)` to the shared `FileUtil` API.
- Implemented the JS version using IndexedDB cleanup.
- Mapped the new API to existing directory deletion on native targets.
- Replaced unsupported web filesystem helpers in CoreCrypto with safe inert shims.
- Added a one-time warning log for unexpected JS filesystem shim usage.
- Added an IndexedDB bootstrap for JS tests.


saleniuk (approved):

MohamadJaara (approved):
should be fine-ish for now as POC but we need to rethink the whole assets handling if we want to support browsers

PR #4023, opened 2026-04-08
fix(mls): avoid external-commit joins during slow sync when a last event ID exists [WPB-24665]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-24665

# What's new in this PR?

### Issues

During slow sync, pending MLS conversations could be joined via external commit even when the client already had a last processed event ID.

In that situation, there may still be unprocessed pending events locally or remotely, including the commit that should establish the group naturally. Triggering an external-commit join too early can race with the normal event-driven establishment flow.

### Causes (Optional)

The slow sync flow already knows whether a last processed event ID exists, but that signal was not used when deciding how to handle pending MLS conversations.

As a result, the MLS rejoin step treated all `PENDING_JOIN` conversations the same way and allowed external-commit joins regardless of whether the client was resuming from an existing event position or starting fresh.

### Solutions

Slow sync now derives whether external-commit joins should be allowed from the presence of a last saved event ID.

If a last saved event ID already exists, slow sync disables external-commit joins for pending MLS conversations and leaves those conversations to be established by the pending commit/event flow.

If no last saved event ID exists, the existing behavior is preserved and pending MLS conversations may still be joined via external commit.

The `allowJoinByExternalCommit` decision is now propagated through the MLS join use cases, and the per-conversation join logic skips the external-commit path when that flag is disabled.

### Dependencies (Optional)

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

login as user A as nomad device and B can be whatever
logout as user A 
have B add A to a new group
login again as user A as nomad device
expected after this change 
the new group should have the messages that were send when A was logged out
actual before this change 
the new group would be empty since A would join it via external commit before processing pending events
 
### Notes (Optional)

This change is intentionally scoped to the slow sync path. Other callers keep the default behavior unless they explicitly opt out of external-commit joins.


sbakhtiarov:
Could you please tell where do we join the conversations after slow sync if there were no pending commits for these conversations?

MohamadJaara:
> Could you please tell where do we join the conversations after slow sync if there were no pending commits for these conversations?

yes it turnes out there is MLSConversationsRecoveryManagerImpl that does this after slow sync

sbakhtiarov:
I looked into MLSConversationsRecoveryManager.

One of the conditions for starting recovery is needsToRecoverMLSGroups, which is set on incremental sync failure. Even if this recovery starts, it only covers ESTABLISHED conversations — not PENDING_JOIN.

I think we need a safety net: if we skip joining a PENDING_JOIN conversation during slow sync and no welcome events are received afterward, we should have a fallback to handle those conversations (e.g., retry joining).

Garzas (approved):

sbakhtiarov (approved):

PR #4025, opened 2026-04-09
feat: Search for Apps - Part 1 (Database changes) [WPB-20357]

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

No Apps table available to store Apps data

### Causes (Optional)

Not implemented yet.

### Solutions

Add `App` table (and its migration and tests) so on future PRs when receiving data from remote it will update `App` table.

### Dependencies (Optional)

Following PRs to come:
- DB related changes (current)
- API related changes (next)
- Wiring changes (last)

#### Test Coverage (Optional)

#### How to Test

No way to test yet

### DIFF
--- a/data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Apps.sq
+++ b/data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Apps.sq
@@ -1,0 +1,35 @@
+import com.wire.kalium.persistence.dao.AppCategoryEntity;
+import com.wire.kalium.persistence.dao.QualifiedIDEntity;
+
+CREATE TABLE App (
+    id TEXT AS QualifiedIDEntity PRIMARY KEY,
+    name TEXT NOT NULL,
+    description TEXT NOT NULL,
+    category TEXT AS AppCategoryEntity,
+    preview_asset_id TEXT AS QualifiedIDEntity,
+    complete_asset_id TEXT AS QualifiedIDEntity
+);
+
+insert:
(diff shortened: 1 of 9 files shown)


data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Apps.sq
```diff
@@ -0,0 +1,35 @@
...
+byId:
+SELECT *
+FROM App
+WHERE App.id = :id;
+
+getUserIdFromMember:
+SELECT User.qualified_id
+FROM User
+JOIN Member ON Member.user = User.qualified_id AND Member.conversation = :conversation
+WHERE User.qualified_id = :id;
```
MohamadJaara, line 26: i am confused about this one since you pass a user id and it can return the same user id or null if that user is not a member of the group conv ? can we rename it to is a member and it return boolean ?

alexandreferris: yeah you're right, it does return the same user id that we pass in 😅

the reason it's structured like this is to stay consistent with the existing pattern we have for old bots/services, where we pass a serviceId and resolve it to a userId via the Member table, in this new flow we already have the userId (App Id), so it ends up looking a bit redundant.

functionally it’s acting as a member check + passthrough of the id, we could change it to return a boolean, but that would diverge from the current pattern while old bots/services and new Apps still need to coexist.

i’ll add a comment in the code to make this clearer so it’s not confusing for the next person

MohamadJaara (approved):

yamilmedina (approved):
🧩 🚀

PR #4034, opened 2026-04-09
feat: Search for Apps - Part 2 (API changes) [WPB-20357]

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Given the existence of Apps table from Part 1 PR, we are now detecting Users with `type`==`APP` and saving their data into the `Apps` table.

Added AppRepository and AppMapper for DB<>Domain class (`AppDetails`)

### Dependencies (Optional)

Following PRs to come:

DB related changes (previous)
API related changes (current)
Wiring changes (next)

Needs releases with:

- [ ] #4025 

#### Test Coverage (Optional)

#### How to Test

"No way to test yet"

If running this branch (after Part 1 was merged) with Android, in an team where Apps are enabled and doing a sync, it should save existing Apps from the team into the Apps table.

### Notes
Needs #4025 merged before reviewing as it contains changes from Part 1.

### DIFF
--- a/data/network-model/src/commonMain/kotlin/com/wire/kalium/network/api/model/UserDTO.kt
+++ b/data/network-model/src/commonMain/kotlin/com/wire/kalium/network/api/model/UserDTO.kt
@@ -19,13 +19,8 @@
 package com.wire.kalium.network.api.model
 
 import com.wire.kalium.network.api.unauthenticated.register.NewBindingTeamDTO
-import kotlinx.serialization.KSerializer
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.descriptors.PrimitiveKind
-import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
-import kotlinx.serialization.encoding.Decoder
-import kotlinx.serialization.encoding.Encoder
 
 @Serializable
 sealed class UserDTO {
--- a/data/network-model/src/commonMain/kotlin/com/wire/kalium/network/api/model/UserDTO.kt
+++ b/data/network-model/src/commonMain/kotlin/com/wire/kalium/network/api/model/UserDTO.kt
@@ -54,7 +54,8 @@ data class UserProfileDTO(
     @SerialName("service") override val service: ServiceDTO?,
     @SerialName("supported_protocols") override val supportedProtocols: List<SupportedProtocolDTO>?,
     @SerialName("legalhold_status") val legalHoldStatus: LegalHoldStatusDTO,
-    @SerialName("type") val type: UserTypeDTO?
+    @SerialName("type") val type: UserTypeDTO?,
+    @SerialName("app") val app: AppDTO?
 ) : UserDTO()
 
 fun UserProfileDTO.isLegacyBot() = this.service != null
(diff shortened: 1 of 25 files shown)


data/network-model/src/commonMain/kotlin/com/wire/kalium/network/api/model/UserDTO.kt
```diff
@@ -147,3 +148,50 @@ enum class UserTypeDTO {
         return this.name.lowercase()
     }
 }
+
+enum class AppCategoryDTO {
+    @SerialName("security")
+    SECURITY,
+
```
MohamadJaara: what would happen when the backend add a new catigory ? this enum will probably break instead it is better to have it as a sealed class and have unknown that takes whatever value the server responds with and you store it as it is in the DB, or maybe just use a string and store it as it is and we just have internal enum when sorting and selecting

alexandreferris: The initial idea was to have a backend endpoint where we could retrieve all categories for Apps but it hasn't being done yet, so yes, it would break.

The idea seemed simple in the beginning but having another look at the usage of `category` from an App, its not even needed to be a sealed class, as it will only for display purposes, no filtering nor searching for the category field, so I will be changing its type to String / TEXT only. So we can also get rid of extra mappings and classes.

alexandreferris:
Some more files were "Added" as the changes for App category being changed from data class to `String` involve some files from the Part 1 PR (db changes), but mostly minimal

cc @MohamadJaara @yamilmedina

data/network-model/src/commonMain/kotlin/com/wire/kalium/network/api/model/UserDTO.kt
```diff
@@ -147,3 +153,9 @@ enum class UserTypeDTO {
         return this.name.lowercase()
     }
 }
+
+@Serializable
+data class AppDTO(
+    @SerialName("description") val description: String,
+    @SerialName("category") val category: String?
```
yamilmedina, line 155: Question: Does it make sense for category to be an enum ?

data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Apps.sq
```diff
@@ -1,11 +1,10 @@
-import com.wire.kalium.persistence.dao.AppCategoryEntity;
 import com.wire.kalium.persistence.dao.QualifiedIDEntity;
 
 CREATE TABLE App (
     id TEXT AS QualifiedIDEntity PRIMARY KEY,
     name TEXT NOT NULL,
     description TEXT NOT NULL,
-    category TEXT AS AppCategoryEntity,
+    category TEXT,
```
yamilmedina, line 7: Question: Same about maybe makes sense for category to be an enum ?

yamilmedina (commented):
Looks good 🚀 ! Just left a small question before moving forward.

data/network-model/src/commonMain/kotlin/com/wire/kalium/network/api/model/UserDTO.kt
alexandreferris: We were using enum in the beginning but after talking with Vina about the category usage and how its also done in other clients (iOS and Web), the category field is only for display purposes.

In this case we don't have a set values of categories, they can change later meaning if its an enum it can then crash, and if we have it as a sealed class, we will be adding some extra unnecessary complexity as it will not be used anywhere else rather than its raw value

yamilmedina (approved):

MohamadJaara (approved):

PR #4035, opened 2026-04-09
fix: join via guest link (WPB-23481)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-23481

https://wearezeta.atlassian.net/browse/WPB-23481
----

# What's new in this PR?

### Issues
1. Android user does not receive messages in MLS conversation after joining via guest link on Android.
2. Android user cannot be removed from conversation by admin on Web client.

### Causes (Optional)
Join by external commit was called instantly after making join API call and sending local event by async handling of join.

### Solutions
 1. Removed join by external commit from ConversationGroupRepository. Instead, MemberJoinEventHandler now calls JoinExistingMLSConversationUseCase when it detects that the self user was added to a group conversation whose MLS group state is PENDING_JOIN. 

To avoid error log when opening conversation in PENDING_JOIN state:
  2. Added a PENDING_JOIN guard to FetchConversationMLSVerificationStatusUseCase so it skips the verification query until the device has completed the external commit and the group is ESTABLISHED.


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/FetchConversationMLSVerificationStatusUseCase.kt
```diff
@@ -37,8 +37,15 @@ internal class FetchConversationMLSVerificationStatusUseCaseImpl(
     override suspend fun invoke(conversationId: ConversationId) {
         conversationRepository.getConversationById(conversationId).onSuccess {
             val protocol = it.protocol
-            if (protocol is Conversation.ProtocolInfo.MLSCapable)
+            // Skip verification check when the group is PENDING_JOIN: the local device has not yet
+            // performed the external commit, so the group does not exist in CoreCrypto and any
+            // attempt to query its state (e.g. e2eiConversationState) will fail with "Couldn't
+            // find conversation".
+            if (protocol is Conversation.ProtocolInfo.MLSCapable &&
+                protocol.groupState != Conversation.ProtocolInfo.MLSCapable.GroupState.PENDING_JOIN
```
MohamadJaara: instead or relaying on PENDING_JOIN maybe would be better to call cc and see if the conv is established or not ?

sbakhtiarov: Agree, that is better. Updated.

yamilmedina (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/e2ei/usecase/FetchConversationMLSVerificationStatusUseCase.kt
```diff
@@ -31,14 +33,21 @@ public interface FetchConversationMLSVerificationStatusUseCase {
...
 
     override suspend fun invoke(conversationId: ConversationId) {
         conversationRepository.getConversationById(conversationId).onSuccess {
             val protocol = it.protocol
-            if (protocol is Conversation.ProtocolInfo.MLSCapable)
-                fetchMLSVerificationStatusUseCase(protocol.groupId)
+            if (protocol is Conversation.ProtocolInfo.MLSCapable) {
+                val groupId = protocol.groupId.value
+                transactionProvider.mlsTransaction { mlsContext ->
+                    wrapMLSRequest { mlsContext.conversationExists(groupId) }
```
ohassine: Suggestion: move transaction handling to a lower layer.
The use case directly calls `mlsTransaction` and works with `mlsContext`, which leaks crypto concerns into the domain layer. It would be better to move this logic into the repository so the use case stays focused on business logic

sbakhtiarov: Right, we even have a function for that in MLSConversationRepository. But I am afraid we do work with CC transactions on the use case level, all MLS repository functions take context as a parameter.

Garzas (approved):

PR #4041, opened 2026-04-13
feat: Search for Apps - Part3 (Wiring Changes to UseCases) [WPB-20357]

alexandreferris:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- Add usecase to get an App by Id, returning ServiceDetails
- Add usecase to observe all Apps returning a list of ServiceDetails
- Add usecase to observe if App is member of a conversation returning its id
- Add usecase to search Apps by name, returning a list of found ServiceDetails

Usecases returning `ServiceDetails` instead of `AppDetails` in order to keep using the existing logics for showing old bots and new Apps as they still need to coexist.

### Dependencies (Optional)

Needs releases with:

- [X] #4025 
- [X] #4034 

#### Test Coverage (Optional)

#### How to Test

"No way to test yet"

If running this branch (after Part 1 was merged) with Android, in an team where Apps are enabled and doing a sync, it should save existing Apps from the team into the Apps table.

### DIFF
--- a/core/data/src/commonMain/kotlin/com/wire/kalium/logic/data/service/ServiceDetails.kt
+++ b/core/data/src/commonMain/kotlin/com/wire/kalium/logic/data/service/ServiceDetails.kt
@@ -17,7 +17,6 @@
  */
 package com.wire.kalium.logic.data.service
 
-import com.wire.kalium.logic.data.id.QualifiedID
 import com.wire.kalium.logic.data.user.UserAssetId
 
 data class ServiceDetails(
@@ -35,9 +34,3 @@ data class ServiceId(
     val id: String,
     val provider: String
 )
-
-fun ServiceId.toQualifiedID(): QualifiedID =
-    QualifiedID(
-        value = this.id,
-        domain = this.provider
-    )
--- a/core/data/src/commonMain/kotlin/com/wire/kalium/logic/data/service/ServiceDetails.kt
+++ b/core/data/src/commonMain/kotlin/com/wire/kalium/logic/data/service/ServiceDetails.kt
@@ -17,6 +17,7 @@
  */
 package com.wire.kalium.logic.data.service
 
+import com.wire.kalium.logic.data.id.QualifiedID
 import com.wire.kalium.logic.data.user.UserAssetId
 
 data class ServiceDetails(
@@ -34,3 +35,9 @@ data class ServiceId(
     val id: String,
     val provider: String
 )
+
+fun ServiceId.toQualifiedID(): QualifiedID =
+    QualifiedID(
+        value = this.id,
+        domain = this.provider
+    )
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/app/ObserveIsAppMemberUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/app/ObserveIsAppMemberUseCase.kt
@@ -18,7 +18,7 @@
 package com.wire.kalium.logic.feature.app
 
 import com.wire.kalium.common.error.StorageFailure
-import com.wire.kalium.common.functional.isRight
+import com.wire.kalium.common.functional.fold
 import com.wire.kalium.logic.data.app.AppRepository
 import com.wire.kalium.logic.data.id.ConversationId
 import com.wire.kalium.logic.data.id.QualifiedID
@@ -51,10 +51,11 @@ internal class ObserveIsAppMemberUseCaseImpl internal constructor(
             appId = appId,
             conversationId = conversationId
         ).map {
-            when (it.isRight()) {
-                true -> ObserveIsAppMemberResult.Success(it.value)
-                false -> ObserveIsAppMemberResult.Failure(it.value)
-            }
+            it.fold({ failure ->
+                ObserveIsAppMemberResult.Failure(failure)
+            }, { success ->
+                ObserveIsAppMemberResult.Success(success)
+            })
         }
 }
 
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/app/ObserveIsAppMemberUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/app/ObserveIsAppMemberUseCase.kt
@@ -1,0 +1,64 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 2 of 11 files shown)


core/data/src/commonMain/kotlin/com/wire/kalium/logic/data/service/ServiceDetails.kt
```diff
@@ -34,3 +35,9 @@ data class ServiceId(
     val id: String,
     val provider: String
 )
+
+fun ServiceId.toQualifiedID(): QualifiedID =
+    QualifiedID(
+        value = this.id,
+        domain = this.provider
+    )
```
alexandreferris: To be used in Android

MohamadJaara: what is provider in this case ? 
better to not mix domain with other type of ids, if bots need to have its own id struct then lets just let it have one

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/app/ObserveIsAppMemberUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+        conversationId: ConversationId
+    ): Flow<ObserveIsAppMemberResult> =
+        appRepository.observeIsAppMember(
+            appId = appId,
+            conversationId = conversationId
+        ).map {
+            when (it.isRight()) {
+                true -> ObserveIsAppMemberResult.Success(it.value)
+                false -> ObserveIsAppMemberResult.Failure(it.value)
+            }
```
MohamadJaara: There is already a fold function that does this
```suggestion
.fold(
          { failure -> ObserveIsAppMemberResult.Failure(failure) },
          { userId -> ObserveIsAppMemberResult.Success(userId) }
      )
  }
```

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/app/ObserveIsAppMemberUseCaseTest.kt
```diff
@@ -0,0 +1,88 @@
...
+import io.mockative.mock
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertIs
+import kotlin.uuid.Uuid
+
+class ObserveIsAppMemberUseCaseTest {
```
MohamadJaara, line 36: netpick: this only add coverage for the happy path without considering the error path 
e.g. repository returns Either.Left(StorageFailure) → should return ObserveIsAppMemberResult.Failure

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/app/GetAppByIdUseCaseTest.kt
```diff
@@ -0,0 +1,102 @@
...
+import io.mockative.coEvery
+import io.mockative.eq
+import io.mockative.mock
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertIs
+import kotlin.uuid.Uuid
+
+class GetAppByIdUseCaseTest {
```
MohamadJaara, line 36: same as the other tests missing error case coverage

Repository returns Either.Left(StorageFailure) → should return null
Repository returns Either.Right(null) (app not found) → should return null

MohamadJaara (changes requested):

core/data/src/commonMain/kotlin/com/wire/kalium/logic/data/service/ServiceDetails.kt
yamilmedina: Question: Is it guaranteed that `provider` = `domain`?

alexandreferris:
> what is provider in this case ? 
> better to not mix domain with other type of ids, if bots need to have its own id struct then lets just let it have one

Provider is the domain, the reason is because of how its being used in Android side.

As `Service` is everything related to old bots and they will need to coexist, I did not wanted to change much and kept re-using what we had for old bots until we can properly remove everything and have only Apps available.

From the Android side if we want to have a hard split between both we would need to add verifications for Apps and old bots in every place, such as when doing a click to open the Service details screen or other places.

The reason as is now, doesn't matter if its old bots or Apps, its ID will be passed as ServiceId for all Screens related and only be mapped to `QualifiedId` in `ServiceDetailsViewModel` where we are already doing the verification if we have Apps or old bots enabled for the conversation/team, and only then it will be passed to the respective UseCases (the ones being added in this PR)

Let me know what you think about it

cc @MohamadJaara

core/data/src/commonMain/kotlin/com/wire/kalium/logic/data/service/ServiceDetails.kt
alexandreferris: The reason is not to re-use just the ID from bots solely but because its the one being used to pass across screens in Android.

The mapping is introduced here in Kalium but used in Android and solely on `ServiceDetailsViewModel` where we have the verification for Apps or not, and then only mapped if its Apps.

Any usage before that will continue using what was there before.

After the Android PR is merged from this ticket (WPB-20357), there is another one explicit for Android to rework the UI/Screen, which then some changes will be done in regards to the value that is being passed to ServiceDetailsScreen, as some flows changed slightly (From new Conversation when clicking an App it should lead to ServiceDetailsScreen and not OtherUserProfileScreen anymore)

alexandreferris: Removed the extension function as the work will be done from Android with a sealed class for new Apps or old bots ID handling.

cc @MohamadJaara @yamilmedina

yamilmedina (approved):

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/app/GetAppByIdUseCaseTest.kt
```diff
@@ -0,0 +1,147 @@
...
+import com.wire.kalium.common.error.StorageFailure
+import com.wire.kalium.common.functional.Either
+import com.wire.kalium.logic.data.app.AppDetails
+import com.wire.kalium.logic.data.app.AppRepository
+import com.wire.kalium.logic.data.id.QualifiedID
+import com.wire.kalium.logic.data.service.ServiceDetails
+import com.wire.kalium.logic.data.service.ServiceId
+import io.mockative.coEvery
+import io.mockative.eq
+import io.mockative.mock
```
MohamadJaara: all new tests should use mokkery instead of Mockative (we are in the process or migrating existing tests) you can use the llm skill in the repo to do the migration for you if you want

MohamadJaara (changes requested):

logic/src/commonTest/kotlin/com/wire/kalium/logic/feature/app/GetAppByIdUseCaseTest.kt
alexandreferris: done ✅

MohamadJaara (approved):

PR #4044, opened 2026-04-16
fix: failure message for adding user with no mls client (WPB-17105)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-17105

https://wearezeta.atlassian.net/browse/WPB-17105
----

# What's new in this PR?

### Issue

When creating a new MLS group conversation where some invited users lack MLS-capable clients (no key packages available), the system message shown in the conversation incorrectly indicated a federation failure instead of a missing key packages failure. Users would see misleading error text and a "Learn More" link pointing to federation documentation rather than MLS documentation.

### Cause
When establishing an MLS group, key packages are fetched for each member via claimKeyPackages. If a user has no MLS client, an empty key packages response is returned. Users who return empty key packages are collected in noKeyPackageUsers, but any API error during that call is collected in federationFailedUsers — regardless of whether the error is actually a federation issue. In ConversationGroupRepository, noKeyPackageUsers were then mapped to FailedToAdd.Type.Federation, producing the wrong system message type and "Learn more" link.
  
  
### Solution
Changed the failure type from FailedToAdd.Type.Federation to FailedToAdd.Type.MissingKeyPackages for users skipped during MLS group establishment (allowSkippingUsersWithoutKeyPackages = true). This surfaces the correct system message with a "Learn more" link pointing to the MLS support article — even though the client cannot distinguish between "no MLS client" and "no key packages available".

### Open question
In KeyPackageDataSource.claimKeyPackages, the when block handling per-user API failures has a structural issue:

  when (failure) {
      is NetworkFailure.NoNetworkConnection,
      is NetworkFailure.ProxyError -> return@flatMap Either.Left(failure)
      is NetworkFailure.FederatedBackendFailure -> federationFailedUsers.add(userId)
      else -> federationFailedUsers.add(userId)  // ← same bucket, wrong label
  }

The else branch catches any other error (e.g. ServerMiscommunication, unknown failures) and adds the user to federationFailedUsers, even though the error may not be related to federation.

A more accurate fix would introduce a separate bucket for unclassified errors and map them to FailedToAdd.Type.Unknown. However, given this is an edge case with low user-visible impact, the additional complexity may not be justified at this time.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationGroupRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationGroupRepository.kt
@@ -40,6 +40,7 @@ import com.wire.kalium.logic.data.id.TeamId
 import com.wire.kalium.logic.data.id.toApi
 import com.wire.kalium.logic.data.id.toDao
 import com.wire.kalium.logic.data.id.toModel
+import com.wire.kalium.logic.data.conversation.mls.MLSAdditionResult
 import com.wire.kalium.logic.data.message.MessageContent.MemberChange.FailedToAdd
 import com.wire.kalium.logic.data.mls.CipherSuite
 import com.wire.kalium.logic.data.service.ServiceId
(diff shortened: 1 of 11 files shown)


Garzas (approved):

MohamadJaara (changes requested):
need to be tested with federation errors before merging

Garzas (approved):

yamilmedina (approved):

MohamadJaara (approved):
LGTM

PR #4071, opened 2026-04-23
fix: avoid double external-commit race and sync welcome metadata

Garzas:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- In MLS flows after logout without local data wipe, the app could end up with inconsistent local MLS state (DB vs CoreCrypto), leading to decryption failures (`MessageEpochTooOld` / `StaleCommit`).
- In the `MemberJoin -> MLSWelcome` sequence for the same conversation, we could trigger two external commits, which artificially advanced the local epoch and caused subsequent messages to be treated as stale.

### Causes (Optional)

- After successful `MLSWelcome`, we were updating only `groupState`, without fully syncing `conversationId -> groupId + epoch` from local CoreCrypto state.
- `MemberJoinEventHandler` attempted auto-join for `PENDING_JOIN` even when the user was added by someone else (`addedBy != self`), although in that flow a `Welcome` is expected and no preemptive external commit is needed.

### Solutions

- After successful `MLSWelcome`, we now fully sync conversation MLS metadata with CoreCrypto by persisting `groupId`, `epoch`, and `ESTABLISHED` state by `conversationId`.
- We restricted `MemberJoinEventHandler` auto-join to self-initiated joins only (`addedBy == selfUserId`) to avoid duplicate external commits in flows where `Welcome` will establish the group.
- Added/updated unit tests to cover both fixes.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
@@ -1935,7 +1935,8 @@ public class UserSessionScope internal constructor(
             revocationListChecker = checkRevocationList,
             certificateRevocationListRepository = certificateRevocationListRepository,
             joinExistingMLSConversation = joinExistingMLSConversationUseCase,
-            fetchConversationIfUnknown = fetchConversationIfUnknownUseCase
+            fetchConversationIfUnknown = fetchConversationIfUnknownUseCase,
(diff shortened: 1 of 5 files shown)


MohamadJaara (changes requested):
i reverted the commit causing the issue is this needed after t he revert ?

Garzas:
Closing, because we reverted change causing that https://github.com/wireapp/kalium/pull/4079

PR #4083, opened 2026-04-24
fix: wipe conversation properly on leave and clear content events (WPB-25058)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25058

https://wearezeta.atlassian.net/browse/WPB-25058

----

# What's new in this PR?

### Issue

After User A (Web) leaves an MLS conversation and User B re-adds them, User A's Android client shows message decryption errors and "device not used for a while" warnings. The Android client holds stale MLS group state (epoch and key material) that is no longer valid after re-joining.
  
### Cause

The `MemberLeaveEventHandler` relied on a "delete queue" persisted in `MetadataDAO` to defer conversation deletion until after the member leave event arrived.  This had two fundamental flaws:
  - It called `deleteConversation` (DB-level deletion) instead of properly wiping the MLS group cryptographic state via CoreCrypto.
  - If the member leave event arrived *without* a prior clear event (the normal leave-without-clearing path), no wipe happened at all, leaving stale crypto state on the device.

### Solution

  - **`ClearConversationContentHandler`:** When `needToRemoveLocally && isSelfSender`, delete the conversation immediately. The membership check and delete-queue mechanism are removed entirely.
  - **`MemberLeaveEventHandler`:** When the self user leaves an MLS conversation (reason: `Left`), always call `MLSConversationRepository.leaveGroup()` to properly wipe the MLS group state through CoreCrypto, ensuring stale epoch and key material are cleared before the user can re-join.
  - `ConversationRepository` no longer depends on `MetadataDAO` for queue management.


MohamadJaara (changes requested):
can we please add one db migration that will clear the metadata value for the q since it is removed

MohamadJaara (approved):

MohamadJaara (approved):

Garzas (approved):

PR #4011, opened 2026-04-01
Mo/feat/custom notify query

MohamadJaara:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

in the current model sqldelight `.asFlow()` from select queries will emit an update for any mutaion select that changes any table in the select without control of wither this mutaion need to update the flow of not 
for example 
```sql
SELECT * FROM Message WHERE conversation_id = 'cool_boiz';

INSERT INTO Message(id, conversation_id, text)
VALUES (:id, `not_cool_boiz`, :text);
```
the select will update even tho the conversaiton is not related and there is 0 data change for the 'cool_boiz' conv

### Causes (Optional)

this is how it works with sqldelight

### Solutions

fork sqldelight and add a feature where you can have annotaions in a comment inside the .sq file, those annotaions will resolve to a custom flow listner that follow the annotaions

```sql
selectConversationMessages:
-- @CustomKey conversation_:conversation_id
SELECT ...

insertMessage:
-- @NotifyCustomKey conversation_:conversation_id
INSERT ...
```

How it works:
- `@CustomKey` tells a query which keys to listen for
- `@NotifyCustomKey` tells a mutation which keys to notify after it runs
- Placeholders like `:conversation_id` get resolved from the actual arguments at runtime, so `conversation_:conversation_id` becomes something like `conversation_conv123`
- When custom keys are used, they replace the default table-based invalidation for that statement

This behavior can be configured at compile time so we can adapt it only in internal apps first

what i am looking for in this PR is to start the conversation if this is something we want to adapt 

### Dependencies (Optional)

_If there are some other pull requests related to this one (e.g. new releases of frameworks), specify them here._

Needs releases with:

- [ ] GitHub link to other pull request

#### Test Coverage (Optional)

- [ ] I have added automated test to this contribution

#### How to Test

_Briefly describe how this change was tested and if applicable the exact steps taken to verify that it works as expected._

### Notes (Optional)

_Specify here any other facts that you think are important for this issue._

### Attachments (Optional)

_Attachments like images, videos, etc. (drag and drop in the text box)_

----
[...]

PR #4087, opened 2026-04-28
fix: prevent zombie SharedFlow coroutines in FlowCache and MetadataDAO [WPB-25125]

sbra0902:
- Fix memory/coroutine leak in FlowCache: each shareIn() now runs in a child SupervisorJob scope that is cancelled on cache eviction, preventing launchSharing coroutines from accumulating as zombies
- Fix memory/coroutine leak in MetadataDAOImpl.observeSerializable(): replace SharingStarted.Lazily (never stops) with WhileSubscribed(5s) + child scope cleanup via onCompletion
- Add regression test verifying that repeated cache evictions don't leave zombie coroutines behind

**Problem**
Root Cause 1 — FlowCache: shareIn() internally launches a launchSharing coroutine in the provided scope. When WhileSubscribed sends STOP, the upstream collection is cancelled, but the launchSharing coroutine stays alive waiting for the next command. On cache eviction (remove()), only the flow reference was removed from the map — the coroutine in cacheScope lived on forever. Over time, zombie coroutines accumulated with each eviction cycle.
Root Cause 2 — observeSerializable(): Every call created a new SharedFlow in databaseScope with SharingStarted.Lazily, which starts on first subscriber and never stops. No caching meant repeated calls for the same key (e.g. appLockFlow(), isFileSharingEnabledFlow()) each spawned a permanent coroutine + SQLDelight query listener.

**Impact**
In our case, the accumulating zombie coroutines caused steadily increasing CPU load over time. After applying this fix, the rising CPU usage was eliminated and CPU load returned to expected levels.

**Changes**
FlowCache.kt

Each cached flow now gets its own SupervisorJob (child of cacheScope's job) and a derived sharingScope
remove() now cancels the associated sharingJob, killing the launchSharing coroutine
New sharingJobs map tracks jobs by key alongside the existing storage map

MetadataDAOImpl.kt

SharingStarted.Lazily → SharingStarted.WhileSubscribed(5_000L): upstream stops 5s after last subscriber leaves
databaseScope → child sharingScope with its own SupervisorJob
.onCompletion { sharingJob.cancel() }: when upstream completes after STOP, the child job is cancelled → launchSharing coroutine dies
Removed stale // TODO: Cache comment

FlowCacheTest.kt

New test givenMultipleCacheEvictions_thenSharingCoroutinesShouldNotAccumulate: repeats 5 get/evict cycles and asserts 0 remaining child coroutines

**Review Request**
Please review this PR carefully with regard to potential negative side effects — in particular around the new scope/job lifecycle management, cancellation behavior on cache eviction, and the change from Lazily to WhileSubscribed(5s). 
Root Cause 2 might be also resolvable by applying the Flowcache here aswell.


MohamadJaara:
Hi, thanks for the contribution, but we enforce signed commits on this repo, and it can be merged only with signed commits. Can you please enable it and rebase your branch, and we can get it merged right after?

MohamadJaara:
also for the sqlDelight listner this is something on our radar for a while and we have a POC fix for this that should make it better not only for metadata but for messages and conv lists as well.

MohamadJaara:
you can try the branch from this draft PR https://github.com/wireapp/kalium/pull/4011 to see if this also fixes the other part of the issue

sbra0902:
> Hi, thanks for the contribution, but we enforce signed commits on this repo, and it can be merged only with signed commits. Can you please enable it and rebase your branch, and we can get it merged right after?

done

sbra0902:
> you can try the branch from this draft PR #4011 to see if this also fixes the other part of the issue

nice i suppose that will address the unecessary emits <3

But i think i won't fix the root cause for the zombies

MohamadJaara:
sorry that it took me some time to get the review 

Worth noting that `FlowCache` is used in 5 places, not just metadata:
- `userCache` (UserDAO)
- `conversationCache` + `conversationDetailsCache` (ConversationDAO)
- `membersCache` (MemberDAO)
- `metadataCache` (MetadataDAO)

The user/conversation/member ones are keyed by IDs, so keys churn a lot as you navigate around the app, every eviction leaked a zombie coroutine into `databaseScope` (which lives for the whole session). That's probably the main source of the CPU creep, more than metadata.

Also, after the recently merged SQLDelight fork update (#4135) we now have per-key custom query notifications on the Metadata table, writes to one key no longer wake listeners on other keys. So the metadata caching layer is less valuable than it was.

Plan: let's merge this PR as-is to fix the leak across all 5 caches, and I'll remove the `metadataCache` in a follow-up since custom notifications make it mostly redundant there. The `FlowCache` fix stays useful for user/conversation/member which still benefit from shared upstreams.

MohamadJaara (approved):

MohamadJaara:
hmmm i (my claude code) just noticed something after 5s with zero subscribers the upstream now stops. A new subscriber arriving later:
  - gets the replay=1 cached value first (potentially stale by N minutes),
  - then gets a fresh DB read.
  
this can be an issue if a function is calling one of those flows .first() which can be an issue for conversations users and other places where this is used

yamilmedina (approved):

PR #4092, opened 2026-04-29
fix: join MLS conv with link race condition [WPB-25121]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-25121

# What's new in this PR?

### Issues

After joining an MLS conversation via guest link on native Android, messages sent by another member before the joiner sent their first message could be lost on the joiner.

### Causes

The guest-link join path emitted the local member-join event asynchronously, then immediately tried to prepare MLS state from local conversation storage. This could race before the joined conversation was persisted.

### Solutions

Fetch and persist the joined conversation synchronously after `joinViaInviteCode` succeeds, before reading protocol info and running MLS setup for the joined conversation.

#### Test Coverage

Manual repro was also verified before and after the fix: after the patch, the Android joiner receives messages sent immediately after guest-link join and before sending its first message.


sbakhtiarov:
It seems like the the race-condition issue is solved with duplication.
Now the same event is processed twice:
- via local event manager and EventProcessor
- immediately inside the prepareJoinedConversationForMLS function

MohamadJaara:
> It seems like the the race-condition issue is solved with duplication. Now the same event is processed twice:
> 
>     * via local event manager and EventProcessor
> 
>     * immediately inside the prepareJoinedConversationForMLS function

kinda yes, yes it duplicate the fetch of conv info from remote but the local event still process the system message and if there is something needed

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/JoinConversationViaCodeUseCase.kt
```diff
@@ -64,6 +84,42 @@ public class JoinConversationViaCodeUseCase internal constructor(
...
+    private suspend fun onConversationChanged(response: ConversationMemberAddedResponse.Changed): Result {
+        val event = eventMapper.fromEventContentDTO(
+            LocalId.generate(),
+            response.event
+        ) as Event.Conversation.MemberJoin
+        val conversationId = response.event.qualifiedConversation.toModel()
+        transactionProvider.transaction("joinViaInviteCode") { transactionContext ->
+            memberJoinEventHandler.handle(transactionContext, event)
+            joinMLSGroupIfNeeded(transactionContext, conversationId)
+        }
```
sbakhtiarov: I think we need to add error handling here. Both memberJoinEventHandler.handle() and joinMLSGroupIfNeeded() can fail and we still return Result.Success.Changed(conversationId)

sbakhtiarov (changes requested):

sbakhtiarov (approved):

yamilmedina (approved):

saleniuk (approved):

PR #4096, opened 2026-04-30
fix: ios sqlite test db isolation [WPB-8645]

Garzas:

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Fixed flaky iOS simulator persistence tests failing with `SQLITE_IOERR` / `disk I/O error`.
- Fixed local JVM persistence DAO/global DB tests hanging or failing due to mismatched coroutine test schedulers.

### Causes (Optional)

- Apple persistence tests reused the same cache directory and deterministic database file names across test instances.
- A following test could delete a database file while a previous native SQLite driver or statement still had a handle to it.
- JVM persistence test fixtures used their own `StandardTestDispatcher`, while some tests used `runTest {}` without passing that dispatcher, causing DAO calls to run on a scheduler that the test was not advancing.
- JVM user database test paths created a new temp directory on each access, so delete/create/path calls could point at different files.

### Solutions

- Isolated Apple test database storage by giving each `BaseDatabaseTest` instance a unique cache subdirectory.
- Applied the same isolation to `GlobalDBBaseTest` for global database tests.
- Created the isolated Apple test storage directories before calling SQLiter's `deleteDatabase`, because SQLiter expects the parent directory to exist.
- Set the JVM persistence test dispatcher as `Dispatchers.Main` during each test so `runTest {}` and database DAO dispatchers share the same scheduler.
- Reused a stable temp storage directory per JVM `BaseDatabaseTest` instance.


MohamadJaara (approved):
LGTM

sbakhtiarov (approved):

PR #4104, opened 2026-05-05
fix: filter stale MLS jobs by active self membership [WPB-24981]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-24981

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- MLS conversations could remain in a stale local state after the self user was removed from a group.
- Background MLS jobs could continue acting on conversations where the self user was no longer a member.
- Pending proposal retries could keep repeating when the backend returned `no-conversation`.
- MLS keying material and proposal timer queries did not filter out locally deleted conversations or conversations without self membership.

### Causes (Optional)

- Self-removal handling did not cover all relevant member leave cases.
- MLS background queries were based mostly on MLS state and timers, without verifying that the local self user was still an active member.
- `no-conversation` from the backend was not treated as an unrecoverable pending proposal failure, so the proposal timer could remain scheduled.

### Solutions

- Clean up MLS state when the self user is removed from an MLS conversation.
- Filter MLS keying material and proposal timer queries by `deleted_locally = 0` and active self membership.
- Clear proposal timers when commit proposals fail with backend `no-conversation`.
- Add indexes matching the updated MLS query filters.
- Add regression tests for self removal, stale proposal cleanup, and filtered MLS background queries.

### Side note

There is a difference between being removed by another user and leaving from another self client.

When another user removes us from an MLS conversation, we receive `Conversation.MemberLeave` followed by an MLS handshake message. Processing that MLS message appears to advance CoreCrypto state and then CoreCrypto reports `Couldn't find conversation`, which effectively prevents the stale MLS group from continuing in the same way.

However, when one of our other clients leaves the conversation, we should not rely on a follow-up MLS message/error path to clean up local state on this client. In that case the local conversation can remain `ESTABLISHED`/MLS locally even though the self user is no longer a member, so cleanup needs to happen from the member leave event itself.

We should still decide how to handle `ConversationNotFound` from the incoming MLS path. It can be expected after self removal, but we should avoid turning every such error into a silent success unless we can verify that the self user is no longer a member or the conversation is locally deleted.


logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/conversation/MemberLeaveEventHandler.kt
```diff
@@ -142,8 +142,9 @@ internal class MemberLeaveEventHandlerImpl(
     }
 
     private suspend fun wipeMLSConversationIfNeeded(transactionContext: CryptoTransactionContext, event: Event.Conversation.MemberLeave) {
-        val isSelfUserLeftConversation = event.removedList == listOf(selfUserId) && event.reason == MemberLeaveReason.Left
-        if (!isSelfUserLeftConversation) return
+        val isSelfUserNoLongerMember = selfUserId in event.removedList &&
+                (event.reason == MemberLeaveReason.Left || event.reason == MemberLeaveReason.Removed)
```
MohamadJaara: this condition is always true or i am missing something here

data/persistence/src/commonMain/db_user/migrations/133.sqm
```diff
@@ -0,0 +1,16 @@
...
+    mls_group_id
+) WHERE mls_group_id IS NOT NULL;
+
+CREATE INDEX IF NOT EXISTS conversation_proposal_timer_index ON Conversation(
+    deleted_locally,
+    protocol,
+    qualified_id,
+    mls_group_id,
+    mls_proposal_timer
+) WHERE mls_group_id IS NOT NULL AND mls_proposal_timer IS NOT NULL;
```
MohamadJaara, line 16: needs PRAGMA optimize; at the end

MohamadJaara (approved):

sbakhtiarov (approved):

PR #4129, opened 2026-05-11
feat: disable team-wide app lock when using nomad - WPB-25543

marcoconti83:
https://wearezeta.atlassian.net/browse/WPB-25543

When an account is configured as a nomad profile, that account should not see any app lock prompt, even if the team administrator enforces app lock for the team.
Testing

I've manually tested this on Fulu on a team with app lock enabled


yamilmedina (approved):
lgtm !

MohamadJaara (approved):

PR #4130, opened 2026-05-12
feat: iOS AVS calling support [WPB-24959]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-24959

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Adds iOS AVS calling support in Kalium’s Apple calling implementation.
- Enables iOS call signaling and call control paths that were previously no-op on Apple targets.
- Adds iOS video-related AVS flow manager plumbing.
- Wires `avs-kmp` only into iOS source sets to avoid resolving unavailable macOS variants.

### Causes (Optional)

Apple calling code was still using placeholder/no-op implementations for call manager, media manager, and flow manager paths. Adding `avs-kmp` at the wider Apple source set level also caused KMP variant resolution to request macOS artifacts, while the current AVS KMP artifact is iOS-only.

### Solutions

- Added `AppleAvsInterop` to bridge Kalium’s Apple calling code to AVS KMP / AVS C APIs.
- Replaced the Apple no-op `CallManagerImpl` with a real implementation for:
  - incoming signaling message handling
  - start / answer / reject / end call
  - mute state
  - video send state
  - remote video stream requests
  - MLS epoch updates
  - conversation clients updates
  - background / notification / network quality reporting
- Updated `GlobalCallManager` on Apple to create real `CallManagerImpl`, `FlowManagerServiceImpl`, and `MediaManagerServiceImpl` instances.
- Updated Apple `FlowManagerServiceImpl` to start AVS media/flow managers, attach iOS video views, and switch camera capture device.
- Extended Apple `PlatformView` to carry a `UIView?` for AVS video attachment.
- Added `avs-kmp` dependency alias and scoped it to `ios*Main` source sets only, avoiding unsupported macOS variant resolution.

### Notes

- This PR wires the Kalium-side iOS AVS call control path, but full end-to-end video rendering still requires iOS UI integration that provides and manages native `UIView` instances for AVS video preview/rendering.
- iOS speaker routing is still not implemented in `MediaManagerServiceImpl`; speaker state currently remains a no-op/false path.
- Video view lifecycle still needs follow-up validation/cleanup handling for attach/detach, call teardown, rejoin, and background/foreground transitions.
- UI rotation is not wired because the current AVS KMP `AVSFlowManager` API does not expose the equivalent operation.
- `avs-kmp` is scoped to iOS source sets only because the current artifact does not publish macOS variants.


domain/calling/build.gradle.kts
```diff
@@ -65,6 +65,13 @@ kotlin {
                 implementation(libs.jna)
             }
         }
+        matching { sourceSet ->
+            sourceSet.name.startsWith("ios") && sourceSet.name.endsWith("Main")
+        }.all {
+            dependencies {
+                api(libs.avsKmp)
+            }
+        }
```
saleniuk: Can't we use the same approach as for other platforms?
Something like:
```
val iosMain by getting {
    dependencies {
        api(libs.avsKmp)
    }
}
```

gradle/libs.versions.toml
```diff
@@ -228,6 +228,7 @@ difference-jvm = { module = "dev.andrewbailey.difference:difference-jvm", versio
 
 # avs
 avs = { module = "com.wire:avs", version.ref = "avs" }
+avsKmp = { module = "com.wire:avs-kmp", version.ref = "avs" }
```
saleniuk: Wouldn't be better to separate the version numbers for regular and kmp?

logic/src/appleMain/kotlin/com/wire/kalium/logic/feature/call/AppleAvsInterop.kt
```diff
@@ -0,0 +1,441 @@
...
+
+    fun setBackground(handle: UInt, background: Boolean): Int = wcall_set_background(handle, background.toAvsInt())
+
+    fun setNetworkQualityInterval(handle: UInt, callbacks: Callbacks, intervalInSeconds: Int) {
+        val arg = handles.values.firstOrNull { it.stableRef.get() === callbacks }?.stableRef?.asCPointer()
+        wcall_set_network_quality_handler(handle, networkQualityHandler, intervalInSeconds, arg)
+    }
+
+    fun notifyNetworkChangedIfAvailable(): Boolean {
+        if (!startIfAvailable()) return false
```
saleniuk, line 442: Why only some functions are secured by this ` if (!startIfAvailable()) return false`? 
Others, like `setNetworkQualityInterval` can still be called before it's started and fail in that case, right?
Is it because some of them are called inside the `withCalling` block? So maybe either we should secure all of them just to be sure or maybe we should make it so that they are all scoped and can be called only inside the `withCalling` block?

logic/src/appleMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
```diff
@@ -16,24 +16,171 @@
...
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withTimeoutOrNull
+import kotlinx.datetime.Instant
+import kotlinx.serialization.json.Json
+import kotlin.time.Duration.Companion.seconds
+import kotlin.uuid.Uuid
+
+@Suppress("LongParameterList", "TooManyFunctions")
+internal class CallManagerImpl internal constructor(
```
saleniuk, line 92: Is it basically almost 1:1 copy of JvmAndroid's `CallManagerImpl`? It will take quite a bit more work to maintain them separately and make sure they stay consistent, so would it be possible to extract the common code and reuse it?

saleniuk (commented):
I have a few (possibly silly) questions 🙇

logic/src/appleMain/kotlin/com/wire/kalium/logic/feature/call/CallManagerImpl.kt
Garzas: I will try to tackle that in another PR, when we will have stable release of AVS KMP

saleniuk (approved):

sbakhtiarov (approved):

PR #4143, opened 2026-05-15
feat: show admin promotion message (WPB-25283)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25283

https://wearezeta.atlassian.net/browse/WPB-25283
----

# What's new in this PR?

Show new system message when self user get promoted to group admin

### DIFF
--- a/data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageEntity.kt
+++ b/data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageEntity.kt
@@ -200,7 +200,12 @@ sealed interface MessageEntity {
         /**
          * A member(s) was removed from the team.
          */
-        REMOVED_FROM_TEAM
+        REMOVED_FROM_TEAM,
+
+        /**
+         * Self user was assigned the admin role in the conversation. Local-only.
+         */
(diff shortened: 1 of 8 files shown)


data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageEntity.kt
```diff
@@ -200,7 +200,12 @@ sealed interface MessageEntity {
         /**
          * A member(s) was removed from the team.
          */
-        REMOVED_FROM_TEAM
+        REMOVED_FROM_TEAM,
+
+        /**
+         * Self user was assigned the admin role in the conversation. Local-only.
+         */
+        SELF_USER_PROMOTED_TO_ADMIN
```
MohamadJaara, line 208: do we show it for self user only or there is a chance that it will be X user became admin ?

sbakhtiarov: No, this is for self user only, other members are not notified.

Garzas (approved):

yamilmedina (approved):

PR #4145, opened 2026-05-18
refactor: Migrate message attachments asset data to cell file (WPB-23968)

ohassine:
https://wearezeta.atlassian.net/browse/WPB-23968

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Description

Restructure MessageAttachments into a lean link table

`MessageAttachments` previously carried all asset metadata inline (mime type, dimensions, URLs, transfer status, etc.). This migration extracts that data into the existing CellFile table, which becomes the single source of truth for cell asset metadata.

**What the migration does:**
- Copies all asset metadata from `MessageAttachments` into `CellFile` (INSERT OR IGNORE + UPDATE COALESCE to safely merge with any rows already present)
- Rebuilds `MessageAttachments` as a slim link table with only (asset_id, message_id, conversation_id, asset_index) and proper ON DELETE CASCADE FKs to both Message and CellFile
- Adds an index on MessageAttachments(asset_id)
- Recreates `MessageDetailsView` to read attachment metadata via CellFile join instead of directly from `MessageAttachments`
- Removes cell assets from the legacy Asset table

**Why:** Separates file storage/state from message references — a file can now exist independently of any message (offline saves, future standalone uploads), while message→file links stay normalized in the link table.

Needs releases with:

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageDetailsView.sq
```diff
@@ -112,25 +112,26 @@ IFNULL(
...
+        "content_hash":"' || IFNULL(cf.contentHash,'') || '",
+        "content_url":"' || IFNULL(cf.contentUrl,'') || '",
+        "preview_url":"' || IFNULL(cf.previewUrl,'') || '",
+        "content_url_expires_at":"' || IFNULL(cf.contentUrlExpiresAt,0) || '",
+        "edit_supported":"' || IFNULL(cf.editSupported,0) || '"}') || ']'
+    FROM MessageAttachments ma
+    LEFT JOIN CellFile cf ON cf.uuid = ma.asset_id
+    WHERE (ma.message_id = Message.id
+        AND ma.conversation_id = Message.conversation_id)
     ),
```
MohamadJaara: adding yet another join to the MessageView is not the right move imho, i wonder why this is needed and what is the pros and cons of this refactor

MohamadJaara (changes requested):

data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageAttachments.sq
```diff
@@ -1,82 +1,119 @@
...
-
-      asset_transfer_status TEXT NOT NULL DEFAULT 'NOT_DOWNLOADED',
-      asset_index INTEGER AS Int,
-
-      content_url_expires_at INTEGER DEFAULT(NULL),
-      edit_supported INTEGER AS Boolean NOT NULL DEFAULT 0,
+      asset_index INTEGER,
 
       FOREIGN KEY (message_id, conversation_id) REFERENCES Message(id, conversation_id) ON DELETE CASCADE ON UPDATE CASCADE,
+      FOREIGN KEY (asset_id) REFERENCES CellFile(uuid) ON DELETE CASCADE ON UPDATE CASCADE,
```
MohamadJaara, line 10: hmmm this can not be added this way without deleting and recreating the table afaik 🤔

data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageDetailsView.sq
ohassine: Before this refactor, `MessageAttachments` stored all asset metadata, and `MessageDetailsView` read those fields directly from that table.

After the refactor, that metadata is normalized into `CellFile`, so a LEFT JOIN CellFile is needed to read the same fields from their new location.
It's primarily a change to avoid caller-side changes

But yeah still, the split attachments model is cleaner overall, and I’ll follow up with that change

data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageAttachments.sq
ohassine: Yes and I am already doing the migration in `134.sqm`
renaming table to `MessageAttachments_legacy`, creating new one, copying data and then dropping the old table

data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
```diff
@@ -83,6 +87,20 @@ internal class MessageDAOImpl internal constructor(
...
+        val multipartContent = regularMessage.content as? MessageEntityContent.Multipart ?: return message
+        val attachments = attachmentsQueries.getAttachments(
+            regularMessage.id,
+            regularMessage.conversationId,
+            MessageAttachmentMapper::toDao
+        ).executeAsList()
+        return regularMessage.copy(content = multipartContent.copy(attachments = attachments))
+    }
+
+    private fun withMultipartAttachments(messages: List<MessageEntity>): List<MessageEntity> = messages.map(::withMultipartAttachments)
```
sbakhtiarov: I am a bit concerned that this query is now executed for every single message in the conversation.
So we load 30 messages with one query to MessageDetailsView and then do one extra query per each message to fetch multipart attachments.
So 1 query for loading messages becomes 31 queries?

ohassine: Valid point, I wanted to avoid a LEFT JOIN, but I ended up with an N+1 problem.
I will address it

data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageDetailsView.sq
```diff
@@ -110,30 +110,6 @@ IFNULL(
...
-        "asset_width":"' || IFNULL(asset_width,0) || '",
-        "asset_height":"' || IFNULL(asset_height,0) || '",
-        "asset_transfer_status":"' || asset_transfer_status || '",
-        "asset_duration_ms":"' || IFNULL(asset_duration_ms,0) || '",
-        "content_hash":"' || IFNULL(content_hash,'') || '",
-        "content_url":"' || IFNULL(content_url,'') || '",
-        "preview_url":"' || IFNULL(preview_url,'') || '",
-        "content_url_expires_at":"' || IFNULL(content_url_expires_at,0) || '",
-        "edit_supported":"' || edit_supported || '"}') || ']'
-    FROM MessageAttachments
```
MohamadJaara, line 131: can we check if the MessageAttachments join can be removed with this change ?

data/persistence/src/commonMain/db_user/migrations/134.sqm
```diff
@@ -8,7 +8,304 @@ CREATE TABLE IF NOT EXISTS CellFile (
...
+        WHEN assetTransferStatus = 'NOT_DOWNLOADED' THEN COALESCE((SELECT ma.asset_transfer_status FROM MessageAttachments ma WHERE ma.asset_id = CellFile.uuid LIMIT 1), assetTransferStatus)
+        ELSE assetTransferStatus
+    END,
+    contentUrlExpiresAt = COALESCE(contentUrlExpiresAt, (SELECT ma.content_url_expires_at FROM MessageAttachments ma WHERE ma.asset_id = CellFile.uuid LIMIT 1)),
+    editSupported = MAX(editSupported, COALESCE((SELECT ma.edit_supported FROM MessageAttachments ma WHERE ma.asset_id = CellFile.uuid LIMIT 1), 0))
+WHERE uuid IN (SELECT asset_id FROM MessageAttachments);
+
+PRAGMA legacy_alter_table = ON;
+ALTER TABLE MessageAttachments RENAME TO MessageAttachments_legacy;
+PRAGMA legacy_alter_table = OFF;
```
MohamadJaara, line 102: legacy_alter_table means it will not rename indexes referance is this what you wanted ?

data/persistence/src/commonMain/db_user/migrations/134.sqm
```diff
@@ -8,7 +8,304 @@ CREATE TABLE IF NOT EXISTS CellFile (
...
+
+CREATE TABLE IF NOT EXISTS MessageAttachments (
```
MohamadJaara: why adding if it doesn't exist? since you rename the old one first you don't need the "IF NOT EXISTS" just to make sure the table is created with the correct schema

data/persistence/src/commonMain/db_user/migrations/134.sqm
ohassine: Yes, it is intentional.

We only enable `legacy_alter_table` during the temporary `_legacy` rename step in the table rebuild migration, so we don’t have to update schema references to the temporary name. After that, we recreate `MessageAttachments` and its index, and then turn it back off.

data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageDetailsView.sq
ohassine: Yes, checked. `MessageDetailsView` no longer depends on `MessageAttachments` as you see.
The join in `MessageAttachments.sq` cannot be removed, because attachment queries still need `CellFile` fields

yamilmedina (approved):

data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/MessageDetailsView.sq
MohamadJaara: fair enough as long as we do not add a new join fine by me

MohamadJaara (approved):

PR #4149, opened 2026-05-19
refactor: store acrtive calls state in memory

MohamadJaara:
rn active calls are stored in DB which means on app init we need to clear them from memory doing not needed work

this data should be in memory only and this is a POC of how this can look like it can be improved but first want to create a POC


data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/ConversationDetailsWithEvents.sq
```diff
@@ -91,7 +90,11 @@ WHERE archived = :fromArchive
     )
     AND CASE WHEN :onlyInteractionsEnabled THEN interactionEnabled = 1 ELSE 1 END
 ORDER BY
-    CASE WHEN :newActivitiesOnTop THEN hasNewActivitiesToShow ELSE 0 END DESC,
+    CASE
+        WHEN :newActivitiesOnTop THEN
+            CASE WHEN qualifiedId IN :activeCallConversationIds THEN 1 ELSE hasNewActivitiesToShow END
+        ELSE 0
+    END DESC,
```
MohamadJaara, line 97: for now this is needed just for order but moving forward we can change this to not move a conv to the top of the list because of calls

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
```diff
@@ -188,6 +189,8 @@ internal class CallDataSource(
     private val _recentlyEndedCallFlow = MutableSharedFlow<RecentlyEndedCallMetadata>(
         extraBufferCapacity = 1
     )
+    private val activeCallEntities = MutableStateFlow<List<CallEntity>>(emptyList())
+    private val activeCallCreatedAt = ConcurrentMutableMap<String, String>()
```
saleniuk: Is it split into two collections just because `CallEntity` doesn't have a `createdAt` field?
I think we could add it to `CallEntity` anyway, it will be needed sooner or later (other platforms show duration on the screen during the call), and then maybe `activeCallCreatedAt` wouldn't be needed anymore? 🤔

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
```diff
@@ -328,31 +359,52 @@ internal class CallDataSource(
...
+                }
+            }
+            if (!callEntityStatus.isActive && !activeCall.shouldKeepGroupCallJoinable(callEntityStatus)) {
+                wrapStorageRequest {
+                    callDAO.insertCall(
+                        call = updatedActiveCall,
+                        createdAt = activeCallCreatedAt.remove(updatedActiveCall.id)
+                    )
+                }
+            }
```
saleniuk, line 392: Active calls are not added to the database, only closed ones?
I think we need to add all calls to keep track of the person who initiated the call and the time when the call actually started.
Although `createdAt` is being added here, but in case the app is killed and restarted - then `activeCallCreatedAt` will also be cleared and filled with the time when the app got info about that call again, so we lose the original start time and original caller from before the app was killed and restarted, but these are actually the main two things that we need to keep in the database for active calls 😅

PR #4166, opened 2026-05-25
fix: register MLS client before finalizing E2EI enrollment [WPB-25837]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-25837

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Users with E2EI required could finish certificate enrollment during login, but the newly enrolled client was not guaranteed to become a registered MLS client before the E2EI registration block was cleared.

This could leave the app in an inconsistent state where sync starts while `hasRegisteredMLSClient` is still `false`.

### Causes (Optional)

The E2EI enrollment flow initialized the local MLS client with the certificate, but finalization did not explicitly register that MLS client on the backend before clearing the E2EI block and resetting slow sync.

MLS registration could happen later through another asynchronous path, which made the login flow timing-dependent.

### Solutions

Register the MLS client explicitly as part of E2EI enrollment finalization before clearing the E2EI registration block, joining existing MLS conversations, and resetting slow sync.

Added regression coverage for the finalization order and for the failure case where the E2EI block must not be cleared if MLS registration fails.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/FinalizeMLSClientAfterE2EIEnrollment.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/FinalizeMLSClientAfterE2EIEnrollment.kt
@@ -17,11 +17,16 @@
  */
 package com.wire.kalium.logic.feature.client
 
-import com.wire.kalium.logic.data.client.ClientRepository
-import com.wire.kalium.logic.data.conversation.JoinExistingMLSConversationsUseCase
-import com.wire.kalium.logic.data.sync.SlowSyncRepository
+import com.wire.kalium.common.error.CoreFailure
+import com.wire.kalium.common.functional.Either
+import com.wire.kalium.common.functional.flatMap
+import com.wire.kalium.common.functional.left
 import com.wire.kalium.common.functional.map
 import com.wire.kalium.common.logger.kaliumLogger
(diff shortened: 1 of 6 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/FinalizeMLSClientAfterE2EIEnrollment.kt
```diff
@@ -17,24 +17,42 @@
...
+import com.wire.kalium.common.functional.map
+import com.wire.kalium.common.logger.kaliumLogger
 import com.wire.kalium.logic.data.client.ClientRepository
 import com.wire.kalium.logic.data.conversation.JoinExistingMLSConversationsUseCase
+import com.wire.kalium.logic.data.id.CurrentClientIdProvider
 import com.wire.kalium.logic.data.sync.SlowSyncRepository
-import com.wire.kalium.common.functional.map
-import com.wire.kalium.common.logger.kaliumLogger
 
 public interface FinalizeMLSClientAfterE2EIEnrollment {
```
saleniuk: Shouldn't this class name end with `...UseCase`? 😄 Not exactly a part of this fix but we could take this opportunity to fix the name as well

saleniuk (approved):

yamilmedina (approved):

sbakhtiarov (approved):

PR #4167, opened 2026-05-26
fix: read metadata values directly from database [WPB-25889]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-25889

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Fixed a race where sync could get stuck after slow sync because `last_processed_event_id` was read as missing even though it had just been stored.

### Causes (Optional)

- `MetadataDAO.valueByKey()` used `valueByKeyFlow(key).first()`, so one-shot reads could receive a replayed cached value.
- If the metadata flow had previously cached `null` for a key, a fast read right after storing the value could still observe that stale `null` before the DB flow emitted the update.

### Solutions

- Changed `MetadataDAO.valueByKey()` to read directly from the database instead of going through the cached flow.
- Kept `valueByKeyFlow()` unchanged for observing metadata changes.
- Added a regression test covering the case where a missing cached value is followed by a DB write and then a one-shot read.


data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/MetadataDAOImpl.kt
```diff
@@ -68,7 +68,9 @@ class MetadataDAOImpl internal constructor(
             .flowOn(readDispatcher.value)
     }
 
-    override suspend fun valueByKey(key: String): String? = valueByKeyFlow(key).first()
+    override suspend fun valueByKey(key: String): String? = withContext(readDispatcher.value) {
+        metadataQueries.selectValueByKey(key).executeAsOneOrNull()
+    }
```
MohamadJaara, line 64: how about removing the whole cached flow from metadata dao since now we notify the query iff that key changes not when any key changes

MohamadJaara (approved):

yamilmedina (approved):

PR #4170, opened 2026-05-27
chore: create automated third party license notice - WPB-25929

marcoconti83:
https://wearezeta.atlassian.net/browse/WPB-25929

This PR adds the scripts to automatically add the third party notice file


AGENTS.md
```diff
@@ -113,6 +113,26 @@ make                                    # Build from source, outputs to ./native
...
+This runs `:collectSbomArtifacts` to materialise every third-party runtime
+artifact (JVM jars, Android AARs, native klibs, resolved npm packages, prebuilt
+AVS libs) under `build/sbom/artifacts/`, then drives `extractcode` (to unpack
+AARs/JARs in place) and `scancode` to emit JSON, SPDX, CycloneDX, and HTML
+reports under `build/sbom/`.
+
+Scope is restricted to licensable modules — `:sample:*`, `:tools:*`, `:test:*`,
+and `:data:persistence-test` are excluded. Run on an **Apple Silicon Mac** when
+iOS/macOS native dependencies must be included; on other hosts those targets
+resolve empty and the SBOM omits them.
```
MohamadJaara: i would not add this to the AGENTS.md since it does not need to be in the context for every request maybe a SKILL.md for it would be enugh tho

scripts/generate-sbom.sh
```diff
@@ -0,0 +1,414 @@
...
+    BREW_PREFIX="$(brew --prefix)"
+    export PATH="$BREW_PREFIX/opt/icu4c/bin:$BREW_PREFIX/opt/icu4c/sbin:$PATH"
+    export PKG_CONFIG_PATH="$BREW_PREFIX/opt/icu4c/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
+fi
+
+if ! command -v scancode >/dev/null 2>&1; then
+    echo "  Installing scancode-toolkit..."
+    pip install --upgrade pip setuptools wheel >/dev/null
+    pip install scancode-toolkit >/dev/null
+else
```
yamilmedina: question: Just thinking out loud about reproducibility on CI, maybe these dependencies and other 3rd party can be pinned in a requirements file?

marcoconti83: great point

scripts/sbom-requirements.txt
```diff
@@ -0,0 +1,84 @@
...
+license-expression==30.4.4
+lxml==6.1.1
+MarkupSafe==3.0.3
+more-itertools==10.8.0
+normality==3.0.0
+packageurl-python==0.17.6
+packaging==26.2
+packvers==21.5
+parameter-expansion-patched==0.3.1
+pdfminer.six==20251107
```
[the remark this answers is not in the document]
marcoconti83: /ar this is just a built-time script, it's not used in production

marcoconti83:
@yamilmedina I've addressed your comment in the last commit

yamilmedina (approved):

scripts/audit-license-files.py
```diff
@@ -0,0 +1,157 @@
...
+# than over-reporting misses.
+LICENSE_FILE_PREFIXES = ("license", "licence", "notice", "copying", "unlicense", "copyright", "attribution", 
+                         "legal", "third-party-notice", "third_party_notice", "third-party-license",
+                         "third_party_license", "AL20", "AL2", "apache", "MIT", "GPL", "LGPL", "BSD", 
+                         "MPL", "CDDL", "EPL", "EULA")
+
+
+def is_license_file(name):
+    lower = name.lower()
+    return any(lower.startswith(p) for p in LICENSE_FILE_PREFIXES)
```
MohamadJaara, line 43: here the name is converted to lowercase but some PREFIXES in the list are uppercase

scripts/sbom-requirements.txt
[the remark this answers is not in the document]
MohamadJaara: Hey @marcoconti83 sorry to reopen this one; I would appreciate your read.

I want to make sure I'm thinking about it right: scancode invokes pdfminer to extract text from any `.pdf` files it finds in `build/sbom/artifacts/`, and that tree is populated from third-party Maven/npm packages. Some upstream artifacts do ship PDFs (docs bundled in jars), so the vulnerable code path can run on inputs we don't fully control. If this script ends up running in CI with publication/signing credentials in scope, a crafted PDF in a transitive dep would put that environment at risk.

Happy to defer if I'm missing context. Otherwise the fix looks essentially free bumping the pin in `scripts/sbom-requirements.txt` to `pdfminer.six>=20251230` should be a transparent change since it's only a transitive of scancode-toolkit. Want me to push that bump, or would you prefer to keep the `/ar` as-is?

Thanks!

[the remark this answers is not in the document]
marcoconti83: Good catch, yes this needs to be fixed. Bumping the pin is the right approach.

MohamadJaara (approved):

PR #4180, opened 2026-06-01
feat: retry sending text, location and asset messages (WPB-25291)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25291

https://wearezeta.atlassian.net/browse/WPB-25291
----

# What's new in this PR?

### Issues

- Text, location, and asset messages that fail to send because of a transient network outage are not retried when connectivity returns; the user must manually re-send each one.
- For asset messages, retry was particularly broken: the local file was deleted on first attempt, so even manual retry could not re-upload the original bytes, and audio waveform (normalized loudness) was recomputed (or lost) on retry.

### Solutions

- Flip `scheduleResendIfNoNetwork = true` for `SendTextMessageUseCase`, `SendLocationUseCase`, and both failure branches in `ScheduleNewAssetMessageUseCase` / `UploadAssetUseCase`, so transient network failures enqueue the message for the pending-messages worker.
- Introduce `SendPendingAssetMessageUseCase`, which on retry: re-fetches the local asset, recomputes audio normalized loudness only if it's missing, re-uploads, updates the message with the new remote data, persists, deletes the local copy, and dispatches via `MessageSender`.
- `PendingMessagesSenderWorker` now branches: regular asset messages go through `SendPendingAssetMessageUseCase`; everything else stays on `messageSender.sendPendingMessage`.
- On upload failure, `UploadAssetUseCase` now persists any already-computed audio loudness so retry doesn't recompute (or lose) it.
- `MessageScope` exposes `sendPendingAssetMessage` and the worker is wired up via `UserSessionScope`.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/upload/UploadAssetUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/upload/UploadAssetUseCase.kt
@@ -32,6 +32,7 @@ import com.wire.kalium.logic.data.message.MessageContent
 import com.wire.kalium.logic.data.message.PersistMessageUseCase
 import com.wire.kalium.logic.feature.asset.AudioNormalizedLoudnessBuilder
 import com.wire.kalium.logic.feature.asset.UpdateAssetMessageTransferStatusUseCase
+import com.wire.kalium.logic.feature.asset.UpdateAudioMessageNormalizedLoudnessUseCase
 import com.wire.kalium.logic.feature.message.MessageSendFailureHandler
 import com.wire.kalium.logic.util.fileExtension
 import com.wire.kalium.messaging.sending.MessageSender
@@ -49,6 +50,7 @@ internal class UploadAssetUseCaseImpl(
     private val messageSender: MessageSender,
     private val messageSendFailureHandler: MessageSendFailureHandler,
     private val updateAssetMessageTransferStatus: UpdateAssetMessageTransferStatusUseCase,
+    private val updateAudioNormalizedLoudness: UpdateAudioMessageNormalizedLoudnessUseCase,
     private val persistMessage: PersistMessageUseCase,
     private val audioNormalizedLoudnessBuilder: AudioNormalizedLoudnessBuilder,
     private val dispatcher: KaliumDispatcher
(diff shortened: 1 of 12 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/asset/upload/UploadAssetUseCase.kt
```diff
@@ -76,7 +78,22 @@ internal class UploadAssetUseCaseImpl(
...
+                scheduleResendIfNoNetwork = true
+            )
+
+            audioNormalizedLoudnessDeferred.await()?.let { normalizedLoudness ->
+                updateAudioNormalizedLoudness(
+                    conversationId = message.conversationId,
+                    messageId = message.id,
+                    normalizedLoudness = normalizedLoudness
+                )
+            }
```
saleniuk, line 95: With this order: first set failed status and then update waveform, there can be an edge case where the "resend" worker starts and resends that message before the waveform gets updated here, so maybe we could either flip the order here or add `getOrBuildAudioNormalizedLoudnessIfNeeded` to the `SendPendingAssetMessageUseCase` as well, otherwise it can be sent without the waveform 🤔

sbakhtiarov: SendPendingAssetMessageUseCase already calls audioNormalizedLoudnessBuilder if normalized loudness is missing for the audio asset.
I have added this extra loudness update to UploadAssetUseCase to avoid the problem with normalized loudness not persisted if asset upload fails. In this case app was showing audio without waveform.
So the worst case here is the normalized loudness builder will be called twice.
I also think the asset upload state update is more important than loudness, so it is better to update the state and schedule retry as soon as possible after detecting the failure.
Wdyt?

saleniuk: > SendPendingAssetMessageUseCase already calls audioNormalizedLoudnessBuilder if normalized loudness is missing for the audio asset.

Oh, sorry, I don't know why I missed that part 🙈 
I agree with you, it's ok, just one suggestion to maybe also do it in a deferred way in `SendPendingAssetMessageUseCase`, just like here, so that it executes both generating waveform and uploading asset in parallel rather than sequentially - would that be ok?

sbakhtiarov: Updated

sbakhtiarov: Updated

saleniuk (approved):

yamilmedina (approved):

PR #4181, opened 2026-06-02
feat: show pending messages in the end of messages list (WPB-25292)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25292

https://wearezeta.atlassian.net/browse/WPB-25292
----

# What's new in this PR?

### Issues

Messages in "PENDING" state (scheduled for retry) are sent after Internet connection is established and sync is finished. New messages from sync may push the PENDING message from the screen and it will jump back to bottom after sending.

### Solutions

Key issues: 
Need to avoid all sorting all conversation messages (ORDER BY CASE solution)
Need to avoid expensive database migration (new index or column)

- Added a partial idx_msg_pending index and migration (fast to create due to low amount of PENDING messages).
- Added pending/non-pending count and select queries.
- Introduced PendingMessagesPagingSource to page over a virtual merged list: pending messages first, then non-pending messages.


data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Messages.sq
```diff
@@ -444,7 +444,7 @@ selectOldestVisibleMessageTimestampByConversationId:
 SELECT MIN(creation_date) FROM Message WHERE conversation_id = ? AND visibility = "VISIBLE";
 
 selectByConversationIdAndVisibility:
-SELECT * FROM MessageDetailsView WHERE conversationId = :conversationId AND visibility IN :visibility ORDER BY date DESC LIMIT :limit OFFSET :offset;
+SELECT * FROM MessageDetailsView WHERE conversationId = :conversationId AND visibility IN :visibility ORDER BY CASE WHEN status = 'PENDING' THEN 1 ELSE 0 END DESC, date DESC LIMIT :limit OFFSET :offset;
```
MohamadJaara: hmmm not sure about this one tho do we have the correct indexes for this new order ?

sbakhtiarov: It seems like we cannot even have the correct index for it. Currently I do not think we will notice a few milliseconds difference for regular conversations.
But this may have a performance impact on large 50-100k messages conversations.
The alternative is to run a second query just for pending messages and put the to the end of the list.

MohamadJaara: i can test it on my full backup, but i can tell you it will be noticeably slower.

saleniuk: We could add `isPending` column to `Message` table, this way we would be able to create index for that

sbakhtiarov: Please review the other solution: the inner query still runs sort but over a plain Message table and the outer query with MessageDetailsView will only sort one page, not the full message list.

saleniuk: What do you think about doing it using CTE with JOIN instead of a subquery?
```
WITH page AS (
    SELECT
        id,
        conversation_id,
        CASE WHEN status = 'PENDING' THEN 1 ELSE 0 END AS pendingSort,
        creation_date
    FROM Message
    WHERE conversation_id = :conversationId
      AND visibility IN :visibility
    ORDER BY pendingSort DESC, creation_date DESC
    LIMIT :limit OFFSET :offset
)
SELECT MessageDetailsView.*
FROM page
JOIN MessageDetailsView
  ON MessageDetailsView.id = page.id
 AND MessageDetailsView.conversationId = page.conversation_id
ORDER BY page.pendingSort DESC, page.creation_date DESC;
```

It carries the sort keys from the limited `Message` page, so the outer sort does not need to recompute CASE from `MessageDetailsView` (it just makes one `USE TEMP B-TREE FOR ORDER BY` instead of two). 
Also, JOIN is usually easier for the planner to optimize than id IN (...).

saleniuk (approved):

MohamadJaara (commented):
LGTM

MohamadJaara (approved):
LGTM

PR #4202, opened 2026-06-09
feat: support message editing in offline (WPB-25295)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25295

https://wearezeta.atlassian.net/browse/WPB-25295
----

# What's new in this PR?

Allow editing messages when no Internet connection is available.
The following scenarios are covered:
1. Online, message status = sent: no changes, current edit feature.
2. Online, message status = pending: edit not allowed to avoid race conditions, message is supposed to be sent shortly.
3. Offline, message status = sent: editing already sent message. Edit will be sent once connected.
4. Offline, message status = pending: editing pending message, only update the message content, message will be sent as regular message once connected.

Key changes in app code:
Enable "Edit" menu for pending messages

Key changes in kalium code:
SendEditTextMessageUseCase - support Pending and Regular message edit flows.
EditMessageBuilder - reuse TextEdited message builder between use case and worker.
MessageDAO \ Repository - new updateTextMessage() function to update only the contents without edit status for pending message editing.

### DIFF
--- a/data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
+++ b/data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
@@ -450,18 +450,36 @@ internal class MessageDAOImpl internal constructor(
         }
     }
 
+    override suspend fun updateTextMessageContent(
+        conversationId: QualifiedIDEntity,
+        messageId: String,
+        textContent: MessageEntityContent.Text,
+    ): Unit = withContext(writeDispatcher.value) {
+        queries.transaction {
+            updateTextMessageContentInDB(
+                conversationId = conversationId,
+                currentMessageId = messageId,
+                newTextContent = textContent,
+                newMessageId = messageId,
+                editInstant = null,
+            )
+        }
+    }
+
     /**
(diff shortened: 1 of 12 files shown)


saleniuk (approved):

data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/message/MessageDAOImpl.kt
```diff
@@ -450,18 +450,36 @@ internal class MessageDAOImpl internal constructor(
...
+            )
+        }
+    }
+
     /**
      * Be careful and run this operation in ONE wrapping transaction.
      */
     private suspend fun updateTextMessageContentInDB(
-        editInstant: Instant,
+        editInstant: Instant?,
```
ohassine, line 473: Instead of making `editInstant` nullable, could we split this into two explicit methods (e.g. `applyTextEdit` and `updatePendingTextContent`) so we make the intent clearer and also reduce the risk of future callers accidentally passing null and skipping the edit

sbakhtiarov: I was thinking about it, but this is private method, not exposed as part of the DAO.

ohassine (approved):

PR #4203, opened 2026-06-10
feat: restrict login to different backend (WPB-26152)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-26152

https://wearezeta.atlassian.net/browse/WPB-26152
----

# What's new in this PR?

New IsCrossBackendLoginBlockedUseCase for supporting login restrictions if new and current backends do not match.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/server/IsCrossBackendLoginBlockedUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/server/IsCrossBackendLoginBlockedUseCase.kt
@@ -1,0 +1,59 @@
+/*
+ * Wire
+ * Copyright (C) 2024 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/server/IsCrossBackendLoginBlockedUseCase.kt
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    @Suppress("ReturnCount")
+    private suspend fun currentSessionConfig(): ServerConfig? {
+        val session = currentSession() as? CurrentSessionResult.Success ?: return null
+        val valid = session.accountInfo as? AccountInfo.Valid ?: return null
+        val result = serverConfigForAccount(valid.userId) as? ServerConfigForAccountUseCase.Result.Success
+        return result?.config
+    }
```
ohassine, line 58: we could make it chained to avoid ReturnCount Suppress
```
return (currentSession() as? CurrentSessionResult.Success)
    ?.accountInfo
    ?.let { it as? AccountInfo.Valid }
    ?.let { serverConfigForAccount(it.userId) as? Result.Success }
    ?.config
```

ohassine (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/server/IsCrossBackendLoginBlockedUseCase.kt
sbakhtiarov: I usually try to avoid these chains as they get complicated and not easy to follow the logic if it is more than two calls.

PR #4215, opened 2026-06-12
feat: exclude not connected users when searching [WPB-25049]

saleniuk:
https://wearezeta.atlassian.net/browse/WPB-25049

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Currently, search use cases always execute both remote and local and return both connected contacts and the ones that we are not connected with. 

### Solutions

Add an option to exclude not connected users from search. 
The logic is that it first executes remote, then gets users locally and removes local ones from remote result to avoid duplicates, and all remaining remote results are considered "not connected". Adding a parameter to exclude not connected ones makes it so that there's no need to run remote search so it runs only local search.

#### Test Coverage (Optional)

----

----

----
##### References


Garzas (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/search/SearchUsersByNameUseCase.kt
```diff
@@ -60,18 +63,21 @@ internal class SearchUsersUseCaseImpl internal constructor(
...
         } else {
-            handleSearch(searchQuery, excludingMembersOfConversation, customDomain)
+            handleSearch(searchQuery, excludingMembersOfConversation, excludingNotConnected, customDomain)
         }
     }
 
     private suspend fun handleSearch(
         searchQuery: String,
         excludingConversation: ConversationId?,
+        excludingRemote: Boolean,
```
sbakhtiarov: Looks like some inconsistency in new argument names.
There is also an assumption made that local repository will always contain connected users. It is hidden behind "excludeNotConnected".
Suggestion: use "skipRemoteSearch" name for the new argument.
- It easier to read
- It explains what exactly happens when you set it.

sbakhtiarov (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/search/SearchUsersByNameUseCase.kt
saleniuk: It was just a mental shortcut - in our app we keep all connected contacts locally and then we subtract it from remote results so remotes are equal to not connected
but you are right, it's what we do in the app, but in kalium we should keep it clean, so I changed these parameters to skipRemoteSearch as you suggested

PR #4220, opened 2026-06-16
feat: foreground service for sending pending messages (WPB-26422)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-26422

https://wearezeta.atlassian.net/browse/WPB-26422
----

# What's new in this PR?

### Issues

Current solution for sending pending messages after restoring Internet connection does not work when application is not in opened by user.

### Causes (Optional)

Caused by recent Android restrictions for background work:
- WorkManager work started when the application is in background is not allowed call setForeground.
- Without "setForeground" call the work cannot make network calls.

### Solutions

Use foreground service which is waiting for connection and retry sending pending messages.


saleniuk (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/SendPendingMessagesUseCase.kt
```diff
@@ -26,55 +29,50 @@ import com.wire.kalium.logic.data.user.UserId
...
-import com.wire.kalium.common.functional.onSuccess
-import com.wire.kalium.common.logger.kaliumLogger
 
 /**
- * This worker attempts to send all pending messages created by this user.
- * @see [PendingMessagesSenderWorker.doWork]
+ * Attempts to send all pending messages created by this user.
  */
-internal class PendingMessagesSenderWorker(
+public interface SendPendingMessagesUseCase {
```
ohassine, line 36: Can we have it covered with unit test ?

ohassine (approved):

Garzas (approved):

PR #4260, opened 2026-07-01
refactor: get calls status from in memory store [WPB-22834] [WPB-26730]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-22834

- Move ongoing/joinable call status out of conversation state and into an in-memory call store keyed by conversation ID.
- Use the in-memory joinable call data for conversation list/search ordering and join-call row markers.
- Remove public conversation call-state fields from Kalium conversation models.
- Keep existing DB call updates for compatibility with current consumers.
- Add a changelog fragment for the breaking API/source change.
- fix issue where when the status is set to away, the app does not ring and no join btn is visible.
## Data Flow

```mermaid
flowchart LR
    AVS["AVS call events"]
    CallRepo["CallRepository"]
    Memory["In-memory calls map<br/>ConversationId -> Call"]
    DB["Call DB writes<br/>(kept for existing consumers)"]
    Ordering["Conversation list/search ordering"]
    AppUI["Android conversation UI<br/>join button / row marker"]

    AVS --> CallRepo
    CallRepo --> Memory
    CallRepo --> DB
    Memory --> Ordering
    Memory --> AppUI
```

## Why
Conversation list state should not depend on persisted calling rows or session-init reset work. Keeping active AVS call state in memory avoids DB pressure during Kalium init and gives consumers faster `ConversationId -> Call` lookup.

## Breaking Change
Removed call-state fields from public Kalium conversation models:
- `ConversationDetails.Group.hasOngoingCall`
- `ConversationDetailsWithEvents.hasOngoingCall`

Consumers should derive join-call UI from call APIs / the in-memory joinable-call map instead.


data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/ConversationDetailsWithEvents.sq
```diff
@@ -97,6 +95,10 @@ WHERE archived = :fromArchive
     )
     AND CASE WHEN :onlyInteractionsEnabled THEN interactionEnabled = 1 ELSE 1 END
 ORDER BY
+    CASE
+        WHEN :newActivitiesOnTop AND type = 'GROUP' AND qualifiedId IN :ongoingCallConversationIds THEN 1
+        ELSE 0
+    END DESC,
     CASE WHEN :newActivitiesOnTop THEN hasNewActivitiesToShow ELSE 0 END DESC,
```
saleniuk: can these two cases be combined? something like:
```suggestion
    CASE
        WHEN :newActivitiesOnTop AND type = 'GROUP' AND qualifiedId IN :ongoingCallConversationIds 
        THEN 1
        WHEN :newActivitiesOnTop THEN hasNewActivitiesToShow
        ELSE 0
    END DESC,
```

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/call/CallRepository.kt
```diff
@@ -282,30 +296,33 @@ internal class CallDataSource(
...
-            val callNewStatus = if (isGroupCall) CallStatus.STILL_ONGOING else CallStatus.CLOSED
-            if (lastCallStatus in activeCallStatus) { // LAST CALL ACTIVE
-                callingLogger.i("[CallRepository][createCall] -> Update.1 | callNewStatus: [$callNewStatus]")
-                // Update database
-                updateCallStatusById(
-                    conversationId = conversationId,
-                    status = callNewStatus
-                )
-            }
+        }
```
saleniuk, line 320: I think this still has one disadvantage over current solution - when the app is reopened and we receive `onIncomingCall` for the same call again, it will save the metadata from that moment, so we lose the original start time and original caller id.
Start time is not that used currently, but with caller id being changed, we will see more "missed call" system messages with the wrong name again, because `onIncomingCall` always gives us id of user who is the current keygenerator for that call, not the original caller. I checked whether passing original caller id in AVS would work and can confirm that, but that would work only if all participants use new AVS version that has this update.
On the other hand, it would happen only when app is killed and restarted, not sure to what extent we can accept this regression.

saleniuk: Ok, I was wrong, I just checked and currently we also don't carry over the original caller id from database after reopening the app, it just gets overwritten by the id from next onIncomingCall, so there is no regression with this solution.

saleniuk (approved):

MohamadJaara:
@claude review

yamilmedina (approved):

PR #4262, opened 2026-07-01
feat: add shared KMP S3 client for Cells transfers [WPB-26729]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-26729

## Stack
- PR 1 of 2.
- Follow-up hardening: #4348.

## Intent
Replace the JVM/Android AWS SDK path and Apple no-ops with a shared internal Ktor S3-compatible client across Apple, JVM, and Android.

## Changes
- shared internal KMP client for object download, regular upload, multipart create/parts/complete
- SigV4 signing and S3 URL canonicalization
- Ktor/Okio streaming, progress, retry classification, and HTTP timeout configuration
- shared commonMain platform wiring and pre-signed URL downloader
- removal of the AWS SDK dependency and helper classes
- focused signer, transfer, and timeout tests

## Scope
Strict XML parsing, download-length validation, typed credential failures, and extended edge-case coverage are in #4348.


MohamadJaara:
@claude review

MohamadJaara:
> ## Review: Shared KMP S3 client for Cells transfers
> 
> Reviewed the full diff plus the final state of the changed files. Overall this is a nicely scoped change: it removes the heavy AWS SDK Kotlin dependency (and its version catalog) from `:domain:cells`, replaces the Apple no-op stubs with a real implementation, and unifies Apple/JVM/Android onto one small SigV4 + S3 client. The SigV4 signer is validated against the official AWS test vector, and the retry-classification tests are thoughtful.
> 
> I confirmed the removed `awssdk` catalog is not referenced by any other module, so that removal is safe, and that `getPreSignedUrl` was only removed from the impl (it was never part of the `CellsAwsClient` interface), so nothing dangles.
> 
> Findings ordered by impact.
> ### High
> 
> **1. A full token refresh is triggered on every signed request (incl. each multipart part and each retry)**
> 
> `signedRequest()` calls `credentialsProvider()` unconditionally on every request, and on all three platforms that lambda awaits the cells credentials and then calls `sessionManager.updateToken(accessTokenAPI, ...)` to produce the S3 access key from the fresh access token.
> 
> `SessionManagerImpl.updateToken()` always calls `refreshTokenAndPersistSession(...)` — a network round-trip to the token endpoint that rotates and persists a new refresh token every time. Consequences:
> 
>     * A 1 GB multipart upload (~100 x 10 MB parts) forces ~102 token refreshes — one per part, plus create/complete — each a blocking network call. Big latency/throughput regression.
> 
>     * Because the refresh token is rotated and persisted on each call, rapid back-to-back refreshes risk racing or invalidating each other.
> 
> 
> The previous AWS-SDK path used `TokenRefreshingCredentialsProvider`, but the SDK cached resolved credentials until near expiry, so it did not refresh per request. Recommend caching the resolved `S3Credentials` and only refreshing when expired (or reactively on a 401), rather than on every `signedRequest()`.
> 
> **2. Retry backoff is far too short to be useful against throttling**
> 
> `S3_FIRST_RETRY_MAX_DELAY_MILLIS = 10` and `S3_SECOND_RETRY_MAX_DELAY_MILLIS = 15`. Retryable conditions include `429 TooManyRequests`, `503 ServiceUnavailable`, and S3 `SlowDown`/`Throttling` error codes — but the max backoff is 10 ms then 15 ms (randomized 0..max), with only 3 total attempts. Against throttling this is effectively an immediate re-hammer and will not let the server recover. Suggest exponential backoff in the hundreds-of-ms-to-seconds range with jitter, and make it (and `S3_MAX_ATTEMPTS`) configurable via `CellsS3ClientConfig` so tests can keep it fast.
> ### Medium
> 
> **3. No HTTP timeouts configured**
> 
> Both actuals build a bare `HttpClient()` with no `HttpTimeout` plugin. As a result:
> 
>     * `HttpRequestTimeoutException` can never be thrown, so that `catch` branch in `performRequestAttempt` is dead code.
> 
>     * Uploads/downloads can hang indefinitely except for whatever default socket timeouts the underlying engine happens to apply (OkHttp has 10s read/write on JVM/Android; the Darwin engine defaults differ). Recommend installing the `HttpTimeout` plugin with request/connect/socket timeouts.
> 
> 
> **4. Failed multipart uploads are never aborted**
> 
> If `uploadPart` or `completeMultipartUpload` fails terminally, the exception propagates but no `AbortMultipartUpload` is issued, leaving orphaned parts consuming storage on the backend indefinitely. Consider wrapping `uploadMultipart` so a failure after `createMultipartUpload` triggers an abort.
> 
> **5. XML parsing is naive and inconsistent**
> 
> `xmlTagValue()` matches only bare tags and won't handle namespace-prefixed or attributed tags (e.g. `s3:UploadId` or `UploadId` with an xmlns attribute), whereas the error/result detection regexes (`S3_ERROR_ELEMENT`, `S3_COMPLETE_MULTIPART_RESULT_ELEMENT`) explicitly do handle namespace prefixes — so a namespaced response could be detected as valid yet yield a null `UploadId`/`Code`. Also, `createMultipartUpload` does not check for an embedded error element in a 200 body (only `completeMultipartUpload` does), so an error-in-200 would surface as the less helpful "did not include an UploadId". Since the target is arbitrary S3-compatible backends (Cells/MinIO), tightening this is worthwhile.
> ### Low / Tests / Nits
> 
> **6. Test gaps**
> 
>     * The multipart test uses a 9-byte file with `maxRegularUploadSize = 1`, which produces a single part — the actual chunking loop (part numbering, offset math, `readPart` across multiple 10 MB chunks) is never exercised with more than one part. Worth a test that produces several parts and asserts sequential PartNumbers and correct reassembled bytes.
> 
>     * `AwsSigningDate.now()` / `GMTDate.toAwsSigningDate()` (month +1 offset, zero-padding) has no test.
> 
>     * `escapeXml` and namespaced/attributed `xmlTagValue` parsing are untested.
> 
> 
> **7. Download body is streamed outside the retry wrapper** — a mid-stream connection drop during `copyToSink` is not retried (only connection setup is). Acceptable, but note it. There is also no verification that bytes received match `Content-Length`.
> 
> **8. Nits**
> 
>     * `download`'s override uses fully-qualified `okio.Sink` while other okio types are imported — minor inconsistency.
> 
>     * `awaitOrThrow()` throws `IllegalStateException` via `error(...)`; a domain/`NetworkFailure`-typed failure would integrate better with the surrounding `Either` flow.
> 
> 
> The retry classification (retryable vs terminal, exception ordering with `RetryableS3Exception` before `S3RequestException`, `UploadSourceException`/`RetryableTransportException` handling) is correct and well-tested. The two High items — per-request token refresh and the ineffective backoff — are the ones I would want addressed before merge; the rest are improvements.
> 
> Automated review

Follow-up on the AI review:
Fixed issues introduced by the Ktor refactor:
Added connection and socket timeouts.
Improved S3 XML parsing and error handling.
Added download size validation and safe streaming behavior.
Added multipart, signing-date, timeout, and XML tests.
Improved missing-credentials errors and minor cleanup.
Not changed because they already existed with the AWS SDK:
Short retry delays.
Multipart uploads are not aborted after failure.

Garzas (approved):

ohassine (approved):

PR #4263, opened 2026-07-02
feat(network): implement Apple network monitoring [WPB-27082]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-27082

Replaces the Apple `NetworkStateObserverImpl` stub, which always reported “connected”, with a real iOS/macOS implementation using Apple’s `Network.framework` and `nw_path_monitor`.

### What changed

- The observer now watches the default network path on iOS and macOS.
- `nw_path_status_satisfied` is mapped to `ConnectedWithInternet`.
- All other states are treated as `NotConnected`, because the path is not currently usable.
- `observeCurrentNetwork()` now reports:
  - network type: `WIFI`, `CELLULAR`, or `OTHER`
  - a stable network id based on the active interfaces
  - whether the network has internet access
- Network updates are handled on a dedicated serial queue, so state changes are applied in order.
- The mapping logic is covered by unit tests in `appleTest`.

This brings iOS/macOS closer to Android behavior and unblocks connectivity-aware sync and `CallNetworkChangeManager`.

### Known limitation

The network id is based on the interface name and index. This means switching between two networks on the same interface, for example changing Wi-Fi SSID, will not be detected as a network change. @johnxnguyen how does the iOS app deal with this?

Detecting that would require reading the SSID, which needs location permission. We are intentionally not requesting that here.


johnxnguyen (approved):
👍🏻

PR #4277, opened 2026-07-07
feat: support meeting type groups [WPB-20268]

saleniuk:
https://wearezeta.atlassian.net/browse/WPB-20268

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Supporting new `MEETING` group type.
Changed db schema and models to support that, so instead of having `is_channel` now there is a `group_type` enum containing three types: GROUP, CHANNEL, MEETING, just like we get from backend.
MEETING types should not be included in queries that are used for showing conversations on list.

#### Test Coverage (Optional)

----

----

----
##### References


data/network-model/src/commonMain/kotlin/com/wire/kalium/network/api/authenticated/conversation/ConversationResponse.kt
```diff
@@ -141,6 +141,9 @@ data class ConversationResponse(
 
         @SerialName("channel")
         CHANNEL,
+
+        @SerialName("meeting")
+        MEETING,
     }
```
MohamadJaara: would be nice to make this a sealed class with support to unknown types

MohamadJaara (changes requested):

data/network-model/src/commonMain/kotlin/com/wire/kalium/network/api/authenticated/conversation/ConversationResponse.kt
saleniuk: @MohamadJaara ok, changed it to sealed class, if null then it handles it as unknown as well. Unknown types are stored in database with the received value, but are filtered out in queries used on conversation lists (same with meeting types).

MohamadJaara (approved):

data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/ConversationDetails.sq
```diff
@@ -138,6 +138,7 @@ selectAllConversationDetails:
 SELECT * FROM ConversationDetails
 WHERE
     type IS NOT 'SELF'
+    AND group_type IN ('GROUP', 'CHANNEL')
```
sbakhtiarov: Why are we dropping all ONE_ON_ONE conversations from this query?

sbakhtiarov: Ok, it looks like there is no ONE_ON_ONE group type, but this is very confusing.

sbakhtiarov: Also not sure how it works for "CONNECTION_PENDING". If CONNECTION_PENDING conversations have Unknown group_type (from ConnectionRepository) they are filtered out by this query and OR condition below for CONNECTION_PENDING is useless.

saleniuk: Yes, you are right, it was wrong, I forgot to include `null`s there, but since then we already discussed these filters with Mohamad and came up with some better idea, it's just that I forgot to change it here as well, thanks for spotting that! 🙇 

This `AND group_type IN ('GROUP', 'CHANNEL')` should be dropped here, and instead we apply it only for the filtered queries, so for 'ALL' filter should be changed like so:
```
-- When filter is ALL, include all types, but for GROUP, only include GROUP or CHANNEL group_type, filter out MEETINGS and unknown
WHEN :conversationFilter = 'ALL' THEN (type IS NOT 'GROUP' OR group_type IN ('GROUP', 'CHANNEL'))
```
for 'ALL' we take all other types, but for `GROUP` type we only take `GROUP` and `CHANNEL` `group_type`, so we filter `MEETING` and all unknown ones
so we receive `ONE_ON_ONE`, `CONNECTION_PENDING` and `GROUP`, but for `GROUP` we only have regular `GROUP` and `CHANNEL`

It is a bit confusing, because we have two fields:
-`type` - determines the type of conversation item, it can be one of `SELF`, `ONE_ON_ONE`, `GROUP`, `CONNECTION_PENDING`
-`group_type` - determines the type of `GROUP` conversation item, so it only applies to `GROUP` type, for others it is unknown, but for `GROUP` type it can be one of `Group`, `Channel`, `Meeting`, `Unknown`

We get it like that from the backend (as "group_conv_type") so at least it's unified across all platforms

I updated that now

sbakhtiarov (approved):

PR #4279, opened 2026-07-08
fix: call updates resetting conversation paging [WPB-22834] [WPB-26730]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-22834

update conversation list pagination so call status changes refresh the
current paging source instead of recreating the whole Pager.

Previously, joinable-call changes could cause the use case to emit a new Pager,
which risked disrupting scroll position and paging state. Removing the call flow
entirely would keep the Pager stable, but would also make the conversation list
stale when new calls started or ended.

The new approach keeps one Pager for the current query/filter and passes the
ongoing-call ID flow into the paging layer. When that flow emits a new value, the
active PagingSource is invalidated and the next load reads the latest call IDs.

## Changes

- Added optional invalidation support to `KaliumPager`.
- Passed `ongoingCallConversationIdsFlow` through repository extensions into DAO paging config.
- Updated conversation paging source creation to read latest ongoing-call IDs on invalidation.
- Removed Pager recreation on call status changes from the search pagination use case.
- Added/updated tests for initial call IDs and call-flow updates.

### DIFF
--- a/data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/conversation/ConversationExtensions.kt
+++ b/data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/conversation/ConversationExtensions.kt
@@ -19,12 +19,19 @@ package com.wire.kalium.persistence.dao.conversation
 
 import androidx.paging.Pager
 import androidx.paging.PagingConfig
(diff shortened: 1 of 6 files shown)


MohamadJaara:
@claude review

saleniuk (approved):

PR #4286, opened 2026-07-10
fix(persistence): use conversation type for group kinds [WPB-27024]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-27024

## Intent

the #4277 with the new group type column will cause the app to hand when parsing 1:1s since the gorup type is empty string and the app will fail to map it and throw  
Use the existing persisted conversation type to distinguish regular groups, channels, and meetings instead of introducing a separate non-null `group_type` column.

## Changes

- extend the persisted conversation type with `CHANNEL` and `MEETING`
- correct unreleased migration 136 in place so legacy channel groups become `CHANNEL` while 1:1, self, connection-pending, and regular group types are preserved
- remove empty-string group subtype persistence and obsolete group type adapter usage
- update conversation queries and group-family behavior for regular groups, channels, and meetings
- keep meetings excluded from the normal conversation list
- add migration and regression coverage for type conversion, filtering, cells, and mappings

## Root cause

The original migration replaced `is_channel` with a second discriminator and populated missing group subtype values with an empty string, including non-group conversations. This duplicated the existing conversation type and created invalid persisted state.


logic/src/commonMain/kotlin/com/wire/kalium/logic/data/conversation/ConversationMapper.kt
```diff
@@ -622,7 +611,11 @@ internal fun ConversationResponse.toConversationType(selfUserTeamId: TeamId?): C
             if (isTeamOneOne) {
                 ConversationEntity.Type.ONE_ON_ONE
             } else {
-                ConversationEntity.Type.GROUP
+                when (conversationGroupType) {
+                    is ConversationResponse.GroupType.Channel -> ConversationEntity.Type.CHANNEL
+                    is ConversationResponse.GroupType.Meeting -> ConversationEntity.Type.MEETING
+                    else -> ConversationEntity.Type.GROUP
```
saleniuk: With that, we map all `group_conv_type` that we don't recognize into GROUP so we treat it as a a regular group, show on the list and have no record whatsoever that it is not a group type but some unknown type

MohamadJaara: fair i will update it

saleniuk (approved):

yamilmedina (approved):

PR #4287, opened 2026-07-12
feat(network): add opt-in HttpTrafficObserver - WPB-28188

marcoconti83:
https://wearezeta.atlassian.net/browse/WPB-28188

Internal consumers (e.g. a debugging tool) can now pass an HttpTrafficObserver via KaliumConfigs to see real requests/responses, including bodies and unredacted headers/paths. This is unlike KaliumHttpLogger, which redacts them by design for production log safety. 

It's wired through CoreLogic -> UserSessionScope/AuthenticationScope/GlobalKaliumScope into defaultHttpEngine, where it's installed as an OkHttp network interceptor on JVM/Android (no-op on Apple/JS, which don't use OkHttp).

This change is required for Kalium integration with internal debugging tools. For this reason, it's enough to support this on JVM targets.


MohamadJaara:
Hi @marcoconti83 if you still need this change can you please make it build so we can review it ❤️

data/network-util/src/commonMain/kotlin/com/wire/kalium/network/HttpTrafficObserver.kt
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Optional, opt-in hook to observe real outgoing HTTP requests and their responses at the
+ * transport level, including bodies and unredacted headers/paths.
+ *
+ * This is distinct from [KaliumHttpLogger]/[KaliumKtorCustomLogging], which obfuscate paths and
+ * headers and never read bodies, by design, since those are meant to be safe to enable in
+ * production logs. [HttpTrafficObserver] is for consumers (e.g. a developer debugging tool) that
+ * need full visibility and can be trusted with unredacted traffic.
+ */
+interface HttpTrafficObserver {
```
yamilmedina, line 32: I think this belongs to `data:network-util` rather than `data:network`, given other examples we have, i.e. `NetworkStateObserver`.

Also, this will allow us to just consume `:logic` as we do now, since logic declares `data:network-util` as an API dependency, therefore, it is transitively provided to consumers.

data/network-util/src/commonMain/kotlin/com/wire/kalium/network/HttpTrafficObserver.kt
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Optional, opt-in hook to observe real outgoing HTTP requests and their responses at the
+ * transport level, including bodies and unredacted headers/paths.
+ *
+ * This is distinct from [KaliumHttpLogger]/[KaliumKtorCustomLogging], which obfuscate paths and
+ * headers and never read bodies, by design, since those are meant to be safe to enable in
+ * production logs. [HttpTrafficObserver] is for consumers (e.g. a developer debugging tool) that
+ * need full visibility and can be trusted with unredacted traffic.
+ */
```
yamilmedina, line 31: Another suggestion here it would be to use `@InternalKalium` so consumers must explicitly opt-in for this feature, rather than being available for everyone by default.

yamilmedina (commented):
Looks good. I left a couple of observations about location and api for consumers.

marcoconti83:
@yamilmedina I tried to address your comments but I could not figure out how to propagate the "internal" annotation properly. I've let the AI do it instead and now it has added it to build.gradle.kts for network for all. Does it sound correct to you?

yamilmedina (approved):

saleniuk (approved):

PR #4302, opened 2026-07-15
feat(mls): add explicit conversation recovery [WPB-8645]

mastaab:

## Summary

This PR contains two related recovery changes:

1. **Dedicated explicit recovery API**
   - adds a user-session use case for explicit MLS conversation recovery
   - scans established groups first, then joins pending groups in a fresh transaction with external commits enabled
   - includes `PENDING_AFTER_RESET` in batch recovery
   - keeps ordinary slow-sync safeguards unchanged

2. **Self-membership safety fix**
   - filters pending recovery candidates through `filterSelfMemberConversations` before attempting external-commit joins
   - checks self membership before inspecting established conversations in CoreCrypto
   - skips conversations the self user has already left, because missing local CoreCrypto state is expected for those rows
   - prevents left conversations from being incorrectly changed to `PENDING_AFTER_RESET` or rejoined

## Motivation

Manual recovery previously had to restart the full slow-sync pipeline. That also caused the normal saved-event safeguard to suppress the external commit needed by groups already known to be missing locally. The dedicated operation performs only the two required recovery phases with separate crypto transaction boundaries.

The established-group recovery scan also treated every database `ESTABLISHED` row missing from CoreCrypto as recoverable. Persisted rows for conversations the self user has left can legitimately be absent from CoreCrypto, so both the established scan and pending batch join now require persisted self membership.

## Verification

- `:logic:checkKotlinAbi`
- focused `RecoverMLSConversationsForUserUseCaseTest`, `RecoverMLSConversationsUseCaseTests`, and `JoinExistingMLSConversationsUseCaseTest`
- root `detekt`
- JVM runtime recovery successfully external-committed a confirmed `PENDING_AFTER_RESET` replacement group


MohamadJaara:
isn't this PR doing the same github.com/wireapp/kalium/pull/4269 ?

mastaab:
> isn't this PR doing the same github.com/[/pull/4269](https://github.com/wireapp/kalium/pull/4269) ?

I think you wanted to comment on #4313 - you're right in this case, did not see above PR. #4313 is now closed.
This PR here is about something else, happy if you take a 2nd look :)

MohamadJaara:
> > isn't this PR doing the same github.com/[/pull/4269](https://github.com/wireapp/kalium/pull/4269) ?
> 
> I think you wanted to comment on #4313 - you're right in this case, did not see above PR. #4313 is now closed.
> This PR here is about something else, happy if you take a 2nd look :) 

my bad i had both open and commented on the wrong one 😆

MohamadJaara:
@Garzas and @sergeibakhtiarov can you please take a look at this one, i would love to review it but i am swamped atm

sbakhtiarov (approved):

Garzas (approved):

PR #4309, opened 2026-07-17
feat: use SQLCipher raw keys for Android databases [WPB-27002]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-27002

## Goal

Use 32-byte raw SQLCipher keys for Android databases while keeping existing installations readable and making key creation and migration crash-safe.

## Changes

This epic contains three changes:

1. **Persist database keys durably**
   - Store database keys synchronously before creating or rekeying a database.
   - Serialize database-key transitions across SDK instances in the same process.

2. **Use raw keys for new user databases**
   - Generate new user databases with SQLCipher's `x'<64 hex characters>'` raw-key format.
   - Keep existing user databases on their legacy keys to avoid an unnecessary migration.

3. **Migrate the global database**
   - Create new global databases with raw keys.
   - Rekey existing global databases from their legacy key to a fresh raw key.
   - Checkpoint WAL and temporarily switch journal mode before rekeying.
   - Validate the cipher and database integrity before promoting the new key.
   - Recover if rekeying completed but key promotion was interrupted.
   - Continue with the legacy key and retry later when migration fails but the database remains readable.

## Key flow

```mermaid
flowchart TD
    A["Open encrypted database"] --> B{"Database exists?"}
    B -- "No" --> C["Create with a new raw SQLCipher key"]
    B -- "Yes: user database" --> D["Continue with the existing legacy key"]
    B -- "Yes: global database" --> E["Generate and durably store a pending raw key"]
    E --> F["Checkpoint WAL and rekey"]
    F --> G{"Raw key and integrity valid?"}
    G -- "Yes" --> H["Promote the pending key"]
    G -- "No, legacy key works" --> I["Continue with legacy key and retry later"]
    G -- "Neither key works" --> J["Propagate the migration failure"]
```

## Test coverage

- Raw-key formatting and key-selection behavior.
- New user databases opening with raw keys.
- Existing user databases remaining accessible with legacy keys.
- Global database rekeying and integrity validation.
- Recovery when rekeying succeeds before the new key alias is promoted.
- Durable key storage behavior.

## Startup benchmark

Measured Wire Android cold startup after login against the production app and production backend.

- 5 cold-start iterations per revision and device
- `CompilationMode.None()` (`run-from-apk`)
- No baseline profile
- Metric: time to initial display (TTID)

| Device | Kalium revision | Median TTID | Min–max | CoV |
| --- | --- | ---: | ---: | ---: |
| Pixel 6 (Android 17 / API 37) | This PR (`c576e443`) | 694.6 ms | 686.2–710.7 ms | 1.3% |
| Pixel 6 (Android 17 / API 37) | `develop` (`4a9e999e`) | 2,584.4 ms | 2,210.4–2,857.7 ms | 9.3% |
| Samsung A50 (Android 11 / API 30) | This PR (`c576e443`) | 1,391.7 ms | 1,274.0–5,763.3 ms | 87.4% |
| Samsung A50 (Android 11 / API 30) | `develop` (`4a9e999e`) | 7,894.1 ms | 7,350.1–8,706.0 ms | 6.5% |

- **Pixel 6:** median TTID decreased by **1,889.8 ms (73.1%)**, making startup **3.72× faster**.
- **Samsung A50:** median TTID decreased by **6,502.4 ms (82.4%)**, making startup **5.67× faster**.

The A50 PR result includes a slower first run at 5,763.3 ms; the following four runs were 1,273.9–1,496.8 ms. Every run from this PR was faster than every matching `develop` run on both devices.


data/persistence/src/androidMain/kotlin/com/wire/kalium/persistence/kmmSettings/GlobalPrefProvider.kt
```diff
@@ -28,14 +29,23 @@ import com.wire.kalium.persistence.dbPassphrase.PassphraseStorageImpl
...
+    private val encryptedSharedPreferences = buildSharedPreferences(settingsOptions, platformParam)
+
     private val encryptedSettingsHolder: KaliumPreferences = KaliumPreferencesSettings(
-        buildSettings(SettingOptions.AppSettings(shouldEncryptData), EncryptedSettingsPlatformParam(context))
+        SharedPreferencesSettings(encryptedSharedPreferences, commit = false)
+    )
+
+    // Database creation/rekey must not commit before its key alias has reached durable storage.
+    private val durablePassphraseSettingsHolder: KaliumPreferences = KaliumPreferencesSettings(
+        SharedPreferencesSettings(encryptedSharedPreferences, commit = true)
```
Garzas, line 42: Could we lose the database key here if `SharedPreferences.Editor.commit()` returns `false`? `SharedPreferencesSettings` uses `commit()` in this mode, but appears to ignore its boolean result. Would that allow database creation or rekeying to continue even though the new key was not persisted, potentially making the database unreadable after a process restart?

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
```diff
@@ -60,7 +63,15 @@ internal fun UserSessionScope(
     val securityHelper = SecurityHelperImpl(globalPreferences.passphraseStorage)
     val platformUserStorageProperties =
         PlatformUserStorageProperties(applicationContext) { userId ->
-            securityHelper.userDBSecret(userId)
+            // The existence check has to be inside the lock: it decides whether a legacy or a raw
+            // key is used, so reading it outside would race with a concurrent database creation.
+            DatabaseKeyLock.withLock {
+                val userIdEntity = UserIDEntity(userId.value, userId.domain)
+                val databaseExists = applicationContext
```
Garzas, line 70: Could a zero-length database file left behind by an interrupted creation be treated as an existing database here? In that case, if no v2 alias was persisted, we would select or generate a legacy key and create the new database with it. Should this check require `exists() && length() > 0`, as the global database migration does?

data/persistence/src/androidMain/kotlin/com/wire/kalium/persistence/kmmSettings/GlobalPrefProvider.kt
MohamadJaara: what case whould caus the commit to fail?

sbakhtiarov (approved):

logic/src/androidMain/kotlin/com/wire/kalium/logic/feature/UserSessionScope.kt
MohamadJaara: in theory this might happen however i have my doubt if this is a real issue tho 😅 and if checking the file size will help at all

data/persistence/src/androidMain/kotlin/com/wire/kalium/persistence/db/support/SqlCipherKey.kt
```diff
@@ -0,0 +1,160 @@
...
+            SQLiteDatabase.OPEN_READWRITE,
+            null
+        )
+        hasReadableSchema(database) && (!verifyIntegrity || hasValidCipherIntegrity(database))
+    } catch (_: SQLiteException) {
+        false
+    } finally {
+        database?.close()
+    }
+}
```
ohassine, line 145: `canOpenDatabase()` is duplicated between SqlCipherKey.kt and UserDatabaseRawKeyTest. Would it make sense to move it to a shared utility so both the production code and tests use the same implementation?

data/persistence/src/androidMain/kotlin/com/wire/kalium/persistence/kmmSettings/GlobalPrefProvider.kt
Garzas: For example insufficient storage, some I/O error -> of course edge case scenario

ohassine (approved):

Garzas (approved):

data/persistence/src/androidMain/kotlin/com/wire/kalium/persistence/kmmSettings/GlobalPrefProvider.kt
MohamadJaara: the global DB for one user logged in is 30kb if you add 3 each on their own server it will be around 70-80kb ading more code to handle this isn't worth making is complicated since in practice this will never happen, happy to add it tho if you think it is needed

PR #4310, opened 2026-07-17
fix(sync): recover immediately when connectivity returns [WPB-27272]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-27272

## Intent

Fix sync getting stuck on `Waiting for network` after a device stays offline for a long time, reconnects, and returns to the foreground.

Android consumer coverage and submodule update: [wire-android PR #5069](https://github.com/wireapp/wire-android/pull/5069).

## What the device logs showed

The F-Droid device lost its network and websocket connection, then incremental sync retried with exponential delays of 1s, 2s, 4s, 8s, 16s, 32s, and 1m4s.

When Wi-Fi returned, Android first reported an available network and shortly afterwards a validated network. Sync retried immediately, but DNS was not ready yet and resolving the backend failed with `UnknownHostException`. The retry state retained the old exponential history, so this single DNS warm-up failure scheduled a 2m8s wait. Later Wi-Fi reconnects repeated the same pattern and increased the delay to 4m16s and then 8m32s.

Opening the app did not recover sync. A foreground request was added while another request was already active, but `SyncExecutor` reduced requester count to a boolean. Since `true` remained `true`, the running sync collection was not restarted and its existing backoff continued. Force-stopping recreated the process, reset that state, opened the websocket, and completed sync immediately.

## Root causes

- A validated network interrupted the current retry delay, but did not reset the exponential delay before retrying. A likely DNS warm-up failure therefore inherited the stale, already-large backoff.
- `SyncExecutor` only reacted to transitions between zero and non-zero requesters. A new foreground request could not restart an already-requested sync that was failed and backing off.
- Request demand was inferred from subscriptions to the sync-state flow. Internal state waiters were therefore also counted as requesters.
- Restart decisions could read an asynchronously mirrored sync state instead of the authoritative `SyncStateObserver` value.

## What changed

- Make the network wait report whether a newly validated connection or the timeout ended the wait.
- Reset exponential retry state when connectivity returns before retrying slow or incremental sync.
- Reset stale retry state when a new slow-sync collection starts.
- Track logical request lifetimes independently from sync-state collectors.
- Restart failed sync when a genuinely new logical request arrives, while leaving an already-live connection undisturbed.
- Read and expose the authoritative observer state directly, removing the mirrored-state race.

## Sync flow now

```mermaid
flowchart TD
    A["Logical sync request starts"] --> B{"Any active requests?"}
    B -->|"first request"| C["Start slow sync, then incremental sync"]
    B -->|"additional request while Live"| D["Keep current sync running"]
    B -->|"additional request while Failed"| E["Cancel backoff collection and restart sync"]
    C --> F{"Sync attempt"}
    E --> F
    F -->|"success"| G["Live"]
    F -->|"failure"| H["Wait for retry delay or a newly validated network"]
    H -->|"delay elapsed"| F
    H -->|"network returned"| I["Reset exponential delay"]
    I --> F
    F -->|"DNS still warming up"| J["Fail and back off again from the minimum"]
    J --> H
    A --> K["State waiters observe authoritative SyncState"]
    K --> L["Waiters do not change request demand"]
```

When the final logical request ends, sync is stopped. A temporary state waiter still observes `Live` or `Failed`, but cannot accidentally start or restart sync by changing a subscription count.

## Coverage

Regression coverage includes network reconnection versus timeout, ignoring an already-connected initial state, DNS not being ready immediately after reconnection, retry-delay resets for slow and incremental sync, failed-state request restarts, authoritative-state delivery races, waiter isolation, and exactly-once restart behavior.


MohamadJaara:
@claude review

saleniuk (approved):

Garzas (approved):

PR #4313, opened 2026-07-18
fix: suppress redundant admin promotion messages

mastaab:
## Summary
- Read the stored conversation member role before applying a role-change event.
- Persist an admin-promotion system message only for a non-admin to admin transition.
- Keep processing repeated role assignments without adding duplicate history entries.

## Why
The backend can emit distinct near-simultaneous MemberChangedRole(Admin) events for the same member. Each event has a unique ID, so event-ID deduplication cannot identify the redundant user-visible state change.

## Verification
- Focused MemberChangeEventHandlerTest JVM suite passes.


mastaab:
Closing as superseded by #4269, which already implements the same stored-role transition check with dedicated DAO/repository coverage and has merged.

PR #4316, opened 2026-07-21
fix: log buffered MLS messages and epoch updates (WPB-23206)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-23206

https://wearezeta.atlassian.net/browse/WPB-23206

----

# What is new in this PR?

### Issues
1. MLS events buffered by CoreCrypto are mapped to MLS handshake messages without exposing the buffering reason to higher layers.
2. Conversation epoch recovery lacks structured diagnostics for local and remote epoch updates.

### Solutions
- Surface `MLSFailure.BufferedFutureMessage` and `MLSFailure.BufferedCommit` so buffered MLS handshake events can be logged at the event-processing layer.
- Add structured buffering logs with the conversation, group, local epoch, and buffer type.
- Add structured local/remote epoch comparison logs, including the epoch gap, state, trigger, and recovery action.

### DIFF
--- a/core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/MLSClient.kt
+++ b/core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/MLSClient.kt
@@ -121,6 +121,12 @@ data class DecryptedMessageBundle(
     }
 }
 
+sealed interface MLSDecryptResult {
+    data class Success(val messages: List<DecryptedMessageBundle>) : MLSDecryptResult
+    data object BufferedFutureMessage : MLSDecryptResult
+    data object BufferedCommit : MLSDecryptResult
+}
+
 @JvmInline
 value class ExternalSenderKey(
     val value: ByteArray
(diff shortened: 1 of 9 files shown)


core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/MLSClient.kt
```diff
@@ -121,6 +121,12 @@ data class DecryptedMessageBundle(
     }
 }
 
+sealed interface MLSDecryptResult {
+    data class Success(val messages: List<DecryptedMessageBundle>) : MLSDecryptResult
+    data object BufferedFutureMessage : MLSDecryptResult
+    data object BufferedCommit : MLSDecryptResult
+}
```
MohamadJaara, line 128: is there a way to know that it was a proposal where we are not inclided/removed ?

sbakhtiarov: No, as far as I can see Core Crypto does not expose required proposal metadata.

MohamadJaara (approved):

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/MLSClient.kt
MohamadJaara: ok good to know anyway this is a good improvement.

Garzas (approved):

PR #4318, opened 2026-07-21
fix(client): restore MLS registration after client removal [WPB-27228] [WPB-27362]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-27228

## Goal

Restore MLS client registration after remote client deletion and safely stop session work before CoreCrypto is destroyed.

## Reproduction

1. Log in with an MLS-capable client.
2. Delete that client remotely.
3. Let the app receive its own client-removal event and log out.
4. Confirm the removed-client dialog and log in again without restarting the process.
5. MLS conversations can remain in `PENDING_JOIN`; pending session work can also crash with `CoreCrypto object has already been destroyed` during teardown.

## Root cause

The forced-logout metadata wipe cleared the current client ID but preserved the retained client ID. The next login therefore tried to verify a client that had already been deleted remotely.

When verification failed, stale-client cleanup cleared locally persisted feature configuration after the initial feature-config sync. Client registration then consulted the cleared local MLS value and created a Proteus-only client. MLS registration happened later, after the first slow sync had saved its event ID, so a subsequent slow sync could not use external commits to join the pending conversations.

The pending-proposal scheduler also owned an independent coroutine scope. It could survive `UserSessionScope` cancellation and start a combined crypto transaction after logout had destroyed CoreCrypto. Proteus wraps the underlying destroyed-object `IllegalStateException` in `ProteusException`, so the existing fallback did not recognize it.

```mermaid
flowchart LR
    A["Remote client removal"] --> B["Cancel UserSessionScope"]
    B --> C["Stop pending session work"]
    C --> D["Destroy CoreCrypto"]
    D --> E["Clear retained client metadata"]
    E --> F["Read MLS eligibility remotely"]
    F --> G["Register MLS-capable client"]
    G --> H["First slow sync joins MLS conversations"]
```

## Changes

- Clear the retained client ID during removed-client and deleted-account metadata wipes.
- Resync feature configuration after stale retained-client cleanup.
- Use the backend MLS feature status and supported protocols as the authoritative registration decision.
- Fail closed when remote configuration cannot be fetched instead of using stale local state.
- Make the pending-proposal scheduler a child of `UserSessionScope` so logout drains it before CoreCrypto teardown.
- Treat wrapped destroyed-CoreCrypto failures as missing client registration instead of allowing a fatal background exception.
- Cover retained-client cleanup, sync ordering, remote MLS configuration, scheduler cancellation, and wrapped CoreCrypto failures.

## Impact

A replacement client is registered with MLS before its first slow sync, allowing pending MLS conversations to join through external commits. Confirming the removed-client flow no longer leaves background session work using destroyed cryptographic state.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/IsAllowedToRegisterMLSClientUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/IsAllowedToRegisterMLSClientUseCase.kt
@@ -18,19 +18,23 @@
 
 package com.wire.kalium.logic.feature.client
 
-import com.wire.kalium.logic.configuration.UserConfigRepository
+import com.wire.kalium.logic.data.featureConfig.FeatureConfigRepository
+import com.wire.kalium.logic.data.featureConfig.Status
 import com.wire.kalium.logic.data.mlspublickeys.MLSPublicKeysRepository
+import com.wire.kalium.logic.data.user.SupportedProtocol
 import com.wire.kalium.logic.featureFlags.FeatureSupport
 import com.wire.kalium.common.functional.getOrElse
 import com.wire.kalium.common.functional.isRight
+import com.wire.kalium.common.functional.map
 import com.wire.kalium.util.DelicateKaliumApi
 
 /**
  * Answers the question if the self user allowed to register an MLS client
  *
  * Which we are allowed to do if:
- * - MLS support is enabled.
- * - MLS supported is configured on the backend, which can be verified by looking for the existence of MLS public keys.
+ * - This build supports MLS.
+ * - MLS is enabled and supported in the current backend feature configuration.
+ * - MLS public keys are available on the backend.
  */
 @DelicateKaliumApi("This use case performs network calls, consider using IsMLSEnabledUseCase.")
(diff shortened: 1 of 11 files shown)


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/client/IsAllowedToRegisterMLSClientUseCase.kt
```diff
@@ -41,12 +45,19 @@ internal interface IsAllowedToRegisterMLSClientUseCase {
...
-                && userConfigRepository.isMLSEnabled().getOrElse(false)
-                && mlsPublicKeysRepository.getKeys().isRight()
+        if (!featureSupport.isMLSSupported) return false
+
+        val isMLSEnabledRemotely = featureConfigRepository.getFeatureConfigs()
+            .map { featureConfig ->
+                featureConfig.mlsModel.status == Status.ENABLED &&
+                        SupportedProtocol.MLS in featureConfig.mlsModel.supportedProtocols
+            }
+            .getOrElse(false)
```
MohamadJaara, line 59: never relay on the locally stored value for MLS when registing a new client

saleniuk (approved):

Garzas (approved):

PR #4327, opened 2026-07-22
fix: last read event not sent (WPB-26157)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-26157

https://wearezeta.atlassian.net/browse/WPB-26157
----

# What's new in this PR?

### Issues
Last read is not sent to other clients if conversation with unread messages is closed within 3 seconds.

### Causes (Optional)

When closed within three seconds:
- Reading schedules UpdateConversationReadDateUseCase in the three-second debounce queue.
- Closing triggers onConversationClosed() and immediately persists the same timestamp locally.
- The queued worker later observes stored == requested.
- The old stored >= requested condition returned before sending LastRead.

When left open longer, the queued worker runs first, sees an older stored timestamp, and sends LastRead.

### Solutions

Skip only strictly older queued timestamps.
Allow equal, already-persisted timestamps to synchronize with other self-clients.
Avoid duplicate database writes and persistence hooks.

Refactored the code to make it more readable.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/UpdateConversationReadDateUseCase.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/UpdateConversationReadDateUseCase.kt
@@ -93,55 +93,75 @@ public class UpdateConversationReadDateUseCase internal constructor(
         doWork(conversationId, time)
     }
 
-    private suspend fun doWork(conversationId: QualifiedID, time: Instant) {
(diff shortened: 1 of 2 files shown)


MohamadJaara (approved):
LGTM - a bit related to this something i noticed on web and wonder how does android handle it 
if i am offline and i open a few conversations, will they be marked as read only locally or the message is sent when going online later ?

saleniuk (approved):

sbakhtiarov:
@MohamadJaara No, we do not have any mechanism to re-try / sync failed LastRead events. They will stay unread on other clients.

PR #4331, opened 2026-07-24
perf(datetime): avoid current time string round trip [WPB-10090]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-10090

`currentInstant()` now truncates through epoch milliseconds instead of formatting and parsing a string.

This keeps the same millisecond precision while avoiding the extra formatter/parser work and allocations. It also removes the unused `String.toEpochMillis()` helper.

Added coverage for fixed sub-millisecond input and the real system clock.

### Benchmarks

The direct path was faster in both runs:

| Environment | String round trip | Epoch milliseconds | Improvement |
| --- | ---: | ---: | ---: |
| JVM 21 | 676.7 ns/op, 2,752 B/op | 29.6 ns/op, 40 B/op | 22.9x faster |
| Pixel 6, Android 17 | 7,443 ns/op, 130 allocations/op | 104 ns/op, 4 allocations/op | 71.5x faster |

On the Pixel 6, the string path also triggered a blocking allocation GC and a 73 ms GC wait. The device CPU was not locked, so the absolute timings can move around, but the difference between the two paths is clear.

### DIFF
--- a/core/util/src/commonMain/kotlin/com.wire.kalium.util/DateTimeUtil.kt
+++ b/core/util/src/commonMain/kotlin/com.wire.kalium.util/DateTimeUtil.kt
@@ -95,12 +95,14 @@ object DateTimeUtil : PlatformDateTimeUtil() {
     fun currentSimpleDateTimeString(): String = fromInstantToSimpleDateTimeString(Clock.System.now())
 
     /**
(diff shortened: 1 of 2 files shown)


saleniuk (approved):

yamilmedina (approved):
🚀

PR #4336, opened 2026-07-24
refactor(cells): use Instant for version modified time PR 4 [WPB-10090]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-10090

## What changed

- Keep the API modification time as a string in `NodeVersionDTO`.
- Convert it once to `Instant?` when mapping to `NodeVersion`.
- Treat invalid API values as `null`.

## Why

The app parsed every timestamp once for grouping and again for formatting. For `N` versions, that meant `2N` string conversions per load.

Parsing at the DTO boundary reduces that to `N` conversions and keeps the domain model type-safe. No DB migration is needed.

## Stack

- #4332
- #4334
- #4335


saleniuk (approved):

domain/cells/src/commonMain/kotlin/com/wire/kalium/cells/data/model/VersionMapper.kt
```diff
@@ -46,7 +47,7 @@ internal fun NodeVersionDTO.toModel() = NodeVersion(
     editorUrls = editorUrls,
     filePreviews = filePreviews,
     isHead = isHead,
-    modifiedTime = modifiedTime,
+    modifiedTime = modifiedTime?.toLongOrNull()?.let { Instant.fromEpochSeconds(it) },
```
saleniuk, line 50: This is a bit weird, perhaps we should comment that somewhere in code, maybe in `NodeVersionDTO` model, so that no one is confused why this string `val modifiedTime: String?,` is actually a number representing timestamp, however not in millis but in seconds 😅

MohamadJaara: agree this is weird i will add comment to reflect this and also change the variable name

PR #4337, opened 2026-07-24
feat: recover MLS conversations after key packages refill (WPB-27077)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-27077

https://wearezeta.atlassian.net/browse/WPB-27077
----

# What's new in this PR?

### Summary
This PR adds MLS conversation membership recovery for clients that were not added to MLS groups because they had no available key packages or had not yet registered an MLS identity during MLS migration.

After key packages become available and synchronization completes, the client audits its active MLS-capable conversations against CoreCrypto and rejoins any missing groups using external commits.

The implementation keeps normal key-package refill behavior separate from MLS membership-audit orchestration.

#### Scenario 1 (MLS):

User has two MLS clients (Client A, Client B)
Client B is offline with no key packages left.
User is added to a new conversation, Client A is added, Client B cannot be added (no key packages).
Client B goes online and refills key packages.

Expected:
Client B joins new MLS conversation

Actual:
Client B cannot receive messages in a new conversation.

#### Scenario 2 (Proteus -> MLS migration):

User has two Proteus clients (Client A, Client B)
Client B is offline.
Proteus to MLS migration is started.
Existing proteus conversation is migrated to MLS.
Client A creates MLS client, uploads key packages and is added to MLS conversation, Client B cannot be added.
Client B goes online and refills key packages.

Expected:
Client B joins migrated MLS conversation.

Actual:
Client B cannot receive messages in migrated conversation.

#### Durable membership-audit state
Added a user-scoped MLSMembershipAuditRepository backed by the existing Metadata table.
The repository supports three states:

- NOT_REQUIRED
- REQUIRED
- REQUIRED_AFTER_SLOW_SYNC

The marker survives application restarts and is removed only after the complete membership audit succeeds.
The existing "true" metadata value remains compatible with the new state representation. No database schema migration is required.

#### Key-package depletion detection
RefillKeyPackagesUseCase now treats backend key packages count of zero as a recovery signal.
When zero packages are observed, it:
- Persists the audit-required marker.
- Aborts the refill if marker persistence fails.
- Uploads replacement key packages only after the marker is durable.
- Leaves the marker set if the upload fails.

RefillKeyPackagesResult.Success now includes:
- availableCountBeforeRefill
- refilled
These values allow the audit flow to determine whether key packages are available after the check:
refilled == true means this pass successfully uploaded replacements.

availableCountBeforeRefill > 0 handles restart recovery where a previous pass uploaded packages before the application stopped. Normal refills from a low but nonzero count do not independently schedule an audit.

#### Post-registration recovery
RegisterMLSClientUseCase now persists REQUIRED_AFTER_SLOW_SYNC before registering the MLS client or uploading its initial key packages.
This prevents the audit from starting while registration or its subsequent slow sync is still in progress.
After the manager observes a real post-registration slow-sync cycle completing and incremental sync returning to Live, it promotes the marker to REQUIRED.
This covers existing Proteus clients that receive their MLS identity only after conversations have already migrated.

#### Dedicated audit orchestration
Added MLSConversationMembershipAuditManager to separate membership recovery from normal key-package management.

The manager:
1. Observes incremental sync, slow sync, and durable audit state.
2. Waits until incremental sync is Live and slow sync is complete.
3. Handles deferred post-registration audit state.
4. Exposes a force-check signal so a pending audit can wake KeyPackageManager even when incremental sync is already Live.
5. Verifies that key packages are available before auditing.
6. Runs the audit in a crypto transaction separate from the key-package refill transaction.
7. Clears the marker only after a successful audit.
8. Retains the marker after refill, audit, transaction, or marker-clear failures.

#### Simplified KeyPackageManager
KeyPackageManager remains focused on its original refill flow:
1. Wait for incremental sync to become Live.
2. Verify that MLS is supported.
3. Verify that this client has registered MLS.
4. Check whether:
- the regular 24-hour interval elapsed;
- the cached count indicates refill is needed; or
- membership recovery requests a forced check.
5. Run the refill/check inside an MLS transaction.
6. Reset the last-check timestamp.
7. Pass the successful result to the membership-audit manager.
Normal refill behavior remains independent of slow-sync status. Slow-sync gating applies only to the audit.

#### Membership audit
Added AuditMLSConversationMembershipUseCase.
The audit queries every active MLS-capable conversation where the self user is still a member:
- MLS conversations
- Mixed-protocol conversations
- All local MLS group states
- Self conversations

It excludes:
- Proteus-only conversations
- Locally deleted conversations
- Conversations the self user has left

For each conversation, the audit:
1. Checks whether its MLS group actually exists in CoreCrypto.
2. Skips groups that already exist.
3. Rejoins missing groups with external-commit joining enabled.
4. Continues auditing remaining conversations after individual failures.
5. Returns overall failure if any conversation could not be checked or repaired.
The operation is idempotent: subsequent attempts skip groups repaired by an earlier partial audit.

#### Stale migration metadata recovery
Updated JoinExistingMLSConversationUseCase to handle migrated conversations whose local metadata incorrectly contains:
```
groupState = ESTABLISHED
epoch = 0
```
Previously, epoch zero caused the use case to establish a new local group instead of joining the already-existing migrated group.
When external-commit recovery is enabled, the use case now refreshes authoritative conversation metadata before choosing the recovery strategy if the local conversation is:
- PENDING_AFTER_RESET; or
- ESTABLISHED at epoch zero.
After refresh, a migrated conversation with a nonzero backend epoch is joined using an external commit.
This additional refresh is restricted to external-commit recovery so normal pending Welcome processing remains unchanged.

### DIFF
--- a/data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
+++ b/data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
@@ -320,6 +320,20 @@ SELECT * FROM Conversation WHERE mls_group_id = ? AND deleted_locally = 0;
 selectByGroupState:
 SELECT * FROM Conversation WHERE mls_group_state = ? AND (protocol IS 'MLS' OR protocol IS 'MIXED') AND deleted_locally = 0;
 
+selectActiveMLSConversationsForMembershipAudit:
+SELECT * FROM Conversation
+WHERE protocol IN ('MLS', 'MIXED')
+AND deleted_locally = 0
+AND (
+    type IS 'SELF'
+    OR EXISTS (
+        SELECT 1
+        FROM Member
+        WHERE Member.conversation IS Conversation.qualified_id
+        AND Member.user IS (SELECT SelfUser.id FROM SelfUser LIMIT 1)
+    )
+);
+
 selectActiveOneOnOneConversation:
 SELECT * FROM Conversation
 WHERE qualified_id = (SELECT active_one_on_one_conversation_id FROM User WHERE qualified_id = :user_id) AND deleted_locally = 0;
--- a/data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
+++ b/data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
@@ -323,6 +323,20 @@ SELECT * FROM Conversation WHERE mls_group_id = ? AND deleted_locally = 0;
 selectByGroupState:
 SELECT * FROM Conversation WHERE mls_group_state = ? AND (protocol IS 'MLS' OR protocol IS 'MIXED') AND deleted_locally = 0;
 
+selectActiveMLSConversationsForMembershipAudit:
+SELECT * FROM Conversation
+WHERE protocol IN ('MLS', 'MIXED')
+AND deleted_locally = 0
+AND (
+    type IS 'SELF'
+    OR EXISTS (
+        SELECT 1
+        FROM Member
+        WHERE Member.conversation IS Conversation.qualified_id
+        AND Member.user IS (SELECT SelfUser.id FROM SelfUser LIMIT 1)
+    )
+);
+
 selectActiveOneOnOneConversation:
 SELECT * FROM Conversation
 WHERE qualified_id = (SELECT active_one_on_one_conversation_id FROM User WHERE qualified_id = :user_id) AND deleted_locally = 0;
--- a/data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/conversation/ConversationDAO.kt
+++ b/data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/conversation/ConversationDAO.kt
@@ -105,6 +105,7 @@ interface ConversationDAO {
     suspend fun getConversationProtocolInfo(qualifiedID: QualifiedIDEntity): ConversationEntity.ProtocolInfo?
     suspend fun getConversationIdByGroupID(groupID: String): QualifiedIDEntity?
     suspend fun getConversationsByGroupState(groupState: ConversationEntity.GroupState): List<ConversationEntity>
+    suspend fun getActiveMLSConversationsForMembershipAudit(): List<ConversationEntity>
     suspend fun deleteConversationByQualifiedID(qualifiedID: QualifiedIDEntity): Boolean
     suspend fun setConversationDeletedLocally(qualifiedID: QualifiedIDEntity, deletedLocally: Boolean)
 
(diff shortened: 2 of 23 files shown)


MohamadJaara:
now that i am thinking about it checking every single conversation, it is overkill and not needed, all cases that will cause being in a new MLS group (external link, being added or group reset) will give the conv one of the pending states, so checking those only should do the tirck, wdyt?

MohamadJaara (changes requested):
just the comment about needing to scan every single conv when we onyl need this for pending stuff

sbakhtiarov:
> just the comment about needing to scan every single conv when we onyl need this for pending stuff

@MohamadJaara 
As far as I remember the was no pending_join conversation state in the MLS migration scenario. Checking all conversations is more reliable.

MohamadJaara:
> > just the comment about needing to scan every single conv when we onyl need this for pending stuff
> 
> @MohamadJaara As far as I remember the was no pending_join conversation state in the MLS migration scenario. Checking all conversations is more reliable.

hmmmm isn't it pending join or something if not this is an issue in its own thay should be fixed when we get that a conv updated protocol we must check CC if the same group ID is already established then make it established or mark pending otherwire 

i would realy like to avoid this full scan as much as we can

sbakhtiarov:
Yes, this issue was observed in this scenario:
Client was offline with no key packages during the Proteus → MLS migration.
After going online the conversation state was:

groupState = ESTABLISHED
epoch = 0
CoreCrypto conversation not found

This is why the ESTABLISHED conversations were added to the audit.

Garzas (changes requested):
A partial MLS registration can suppress the required post-registration slow sync.

clientRepository.registerMLSClient() persistently sets hasRegisteredMLSClient before markAuditRequiredAfterSlowSync() and the key-package upload have completed. If either of those later operations fails, RegisterMLSClientUseCase returns Left, so MLSClientManager does not clear the last slow-sync timestamp. On the next foreground run, however, the persisted registration flag is already true, causing the manager to skip the registration path entirely.

I reproduced this locally using the real RegisterMLSClientUseCaseImpl and MLSClientManagerImpl: the registration and deferred-audit marker succeed, the key-package upload fails, and the manager is invoked a second time. Registration and upload each happen exactly once, while clearLastSlowSyncCompletionInstant() is never called. The deferred audit therefore has no guaranteed real post-registration slow sync.

A restart does not reliably recover this either. SlowSyncRepository starts with an in-memory Pending state, while the persisted completion timestamp can make SlowSyncManager skip SlowSyncWorker and transition directly to Complete. Depending on initialization order, the audit observer can either remain deferred or mistake that Pending-to-Complete transition for the required post-registration sync and audit stale local data.

Could we make this workflow resumable, or persist a separate finalization state instead of treating hasRegisteredMLSClient as completion of the whole workflow? At minimum, scheduling the required slow sync should not depend on the later key-package upload succeeding. Please also add regression coverage for a failure after the registration flag has been persisted.

sbakhtiarov:
@MohamadJaara @Garzas 
I probably need to stop and re-think the solution. It is growing bigger on each iteration and every time there are new scenarios, failure flows and edge cases to fix.

PR #4340, opened 2026-07-24
fix: expose client id type to swift [WPB-27598]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-27598

## What changed

`ClientId` is now a concrete data class, so Swift gets `DataClientId` instead of `Any`.

The JSON format stays the same, and there’s a regression test for it.

## Why

Kotlin value classes are erased at the Objective-C boundary, which made client IDs show up as `Any` in iOS apps.

The generated iOS API now keeps the actual client ID type.


johnxnguyen (approved):
LGTM! 

FYI I found a few more similar objects that are type erased:
- `messageTimer` value on `Conversation`
- `teamId` value on `Conversation`
- `duration` on `SelfDeletionTimer`
- MLS group ids (various models)

I can create tickets for these, just thought I'd share already.

yamilmedina (approved):

Garzas (approved):

PR #4344, opened 2026-07-28
feat: multi ingress sso identity change [WPB-27414]

Garzas:
https://wearezeta.atlassian.net/browse/WPB-27414

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

- Retained account data could be reused after authenticating the same Wire `QualifiedID` through a different SSO identity provider.
- Updating an account without an IdP ID could overwrite the previously stored value with `null`.

### Causes

Accounts were identified only by their Wire `QualifiedID`. Kalium did not persist and compare the stable IdP ID used to initiate multi-ingress SSO authentication.

Additionally, the account upsert replaced `sso_identity_provider_id` unconditionally.

### Solutions

- Added the SSO identity provider ID to `StoreSessionParam`.
- Persisted the IdP ID in the global `Accounts` database.
- Added `SsoIdentityChanged` to `AddAuthenticatedUserUseCase.Result.Failure`.
- Compared the retained and newly authenticated IdP IDs before reusing local account data.
- Preserved the stored IdP ID when an account update does not provide one.
- Added unit and persistence regression tests.
- Updated the JVM and KMP ABI dumps.
- Added a consumer-facing changelog fragment.

### Compatibility

Adding the IdP ID to `StoreSessionParam` changes its constructor and `copy()` ABI. Previously compiled consumers using this class must be recompiled.

### Verification

- `AccountsDAOTest`
- `AddAuthenticatedUserUseCaseTest`
- Database migration verification
- `:logic:checkKotlinAbi`


saleniuk (approved):

data/persistence/src/commonMain/db_global/migrations/10.sqm
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE Accounts
+ADD COLUMN sso_identity_provider_id TEXT DEFAULT NULL;
```
sbakhtiarov, line 2: Need to double-check that this null will not break anything for existing users.
Will it mean that previous ssoIdentityProviderId will always be null and hasSsoIdentityChanged() will always return true?

Garzas: Yes, existing accounts will have a null IdP ID, so their first multi-ingress login after logout will be treated as an IdP change. This is intentional fail-closed behavior because the previously used IdP cannot be verified. Active accounts and non-multi-ingress logins are unaffected

sbakhtiarov: I guess it is fine then. I do not have a 100% understanding of this flow.

sbakhtiarov (approved):

PR #4352, opened 2026-07-30
feat: adminless events (WPB-25284)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25284

https://wearezeta.atlassian.net/browse/WPB-25284
----

# What's new in this PR?

Adding support for new events for "Prevent Adminless groups" feature.

## Issue
When a group loses all eligible admins, the backend sends an "adminless delete
reminder" event announcing that the group will be automatically deleted. This
pending-deletion state needs to be persisted and exposed on the conversation so
the client can surface it as a "ghost group" indicator in the conversation list
and inside the conversation.

## Causes
The previous approach persisted the reminder as an AdminlessDeleteReminder system message. That coupled the pending-deletion state to the message timeline, so it was lost on clearContent, produced no meaningful conversation
preview (it mapped to Unknown), and mixed a conversation-level fact into the per-message content model.

## Solutions

This PR replaces the system-message approach with a dedicated conversation-level
record:

Added an AdminlessGroupDelete table (conversation_id PK, deletion_timestamp,
ON DELETE CASCADE) and insertAdminlessGroupDelete (INSERT OR IGNORE),
exposed via ConversationDAO.insertAdminlessGroupDelete().


sbakhtiarov:
@MohamadJaara merged the WPB-25285 here to get rid of duplicated/removed code.

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/event/EventMapper.kt
```diff
@@ -623,11 +654,69 @@ internal class EventMapper(
...
-    ) = Event.Conversation.DeletedConversation(
+    ) = conversationDeleted(
         id = id,
         conversationId = deletedConversationDTO.qualifiedConversation.toModel(),
         senderUserId = deletedConversationDTO.qualifiedFrom.toModel(),
-        dateTime = Instant.parse(deletedConversationDTO.time)
+        dateTime = Instant.parse(deletedConversationDTO.time),
+    )
+
+    private fun conversationDeleted(
```
saleniuk: Just a nitpick but maybe it would be better to keep the same naming convention, as now we have "deletedConversation" and "conversationDeleted" functions 😅

logic/src/commonMain/kotlin/com/wire/kalium/logic/sync/receiver/ConversationEventReceiver.kt
```diff
@@ -97,6 +99,11 @@ internal class ConversationEventReceiverImpl(
                 Either.Right(Unit)
             }
 
+            is Event.Conversation.AdminlessDeleteReminder -> {
+                deleteConversationReminderEventHandler.handle(event)
+                Either.Right(Unit)
```
saleniuk, line 104: For new handlers we should already integrate errors into the handler itself, and not just always return `Either.Right`
Take a look at the comment from line 77 😄 

> // TODO: Make sure errors are accounted for by each handler.
> // onEvent now requires Either, so we can propagate errors,
> // but not all handlers are using it yet.

saleniuk (approved):

Garzas (approved):

PR #4356, opened 2026-07-31
feat: show last message for ghost group (WPB-25285)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25285
----

# What's new in this PR?

### Issues

When a group loses all eligible admins, the backend sends an "adminless delete
reminder" event announcing that the group will be automatically deleted. This
pending-deletion state needs to be persisted and exposed on the conversation so
the client can surface it as a "ghost group" indicator in the conversation list
and inside the conversation.

### Causes

- The previous approach persisted the reminder as an `AdminlessDeleteReminder`
  *system message*. That coupled the pending-deletion state to the message
  timeline, so it was lost on `clearContent`, produced no meaningful conversation
  preview (it mapped to `Unknown`), and mixed a conversation-level fact into the
  per-message content model.

### Solutions

This PR replaces the system-message approach with a dedicated conversation-level
record:

- Added an `AdminlessGroupDelete` table (`conversation_id` PK, `deletion_timestamp`,
  `ON DELETE CASCADE`) and `insertAdminlessGroupDelete` (`INSERT OR IGNORE`),
  exposed via `ConversationDAO.insertAdminlessGroupDelete()`.
- `DeleteConversationReminderEventHandler` now stores the deletion timestamp in
  that table (wrapped in `wrapStorageRequest`) instead of persisting a system
  message. It no longer depends on `PersistMessageUseCase`/`selfUserId`.
- Joined `AdminlessGroupDelete` into the `ConversationDetails` view and surfaced
  `adminlessGroupDeletionTimestamp` end-to-end: `ConversationViewEntity` →
  `ConversationDetailsWithEventsMapper` / `ConversationMapper` →
  `Conversation.adminlessGroupDeletionTimestamp` (domain).
- Removed the now-unused `AdminlessDeleteReminder` plumbing:
  `MessageContent.AdminlessDeleteReminder`, `MessageEntityContent.AdminlessDeleteReminder`,
  the `ADMINLESS_DELETE_REMINDER` content type, its insert query, the
  `MessageDetailsView` column, and the related mappings.
- Updated migration `142.sqm` to create the new table and rebuild the
  `ConversationDetails` / `ConversationDetailsWithEvents` views.

Because the record is conversation-scoped and no longer a message, clearing a
conversation's content preserves the pending-deletion state, and it is removed
only when the conversation itself is deleted (cascade).

### Notes / open questions

- `insertAdminlessGroupDelete` uses `INSERT OR IGNORE`, so the first timestamp
  wins and there is currently no update or clear path (only cascade on hard
  delete). This assumes the backend does not reschedule or cancel the deletion.
  If it can, we'll need an upsert and a clear-on-cancel handler.

### Dependencies

- [ ] Consumed by wire-android: https://github.com/wireapp/wire-android/pull/5127


data/persistence/src/commonMain/db_user/com/wire/kalium/persistence/Conversations.sq
```diff
@@ -556,7 +556,9 @@ clearContent {
 -- @NotifyCustomKey Asset
 -- @NotifyCustomKey Message
 DELETE FROM Asset WHERE key IN (SELECT asset_id FROM MessageAssetContent WHERE conversation_id = :conversationId);
-DELETE FROM Message WHERE conversation_id = :conversationId;
+DELETE FROM Message
+WHERE conversation_id = :conversationId
+    AND content_type != 'ADMINLESS_DELETE_REMINDER';
```
MohamadJaara, line 561: at this point does the ADMINLESS_DELETE_REMINDER need to be a reminder vs a flag in the conv table (wll_be_deleted_at) this way it pinned to the bottom, and does not effect messages or wdyt ?

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/GetPaginatedFlowOfConversationDetailsWithEventsBySearchQueryUseCase.kt
```diff
@@ -53,14 +55,30 @@ public class GetPaginatedFlowOfConversationDetailsWithEventsBySearchQueryUseCase
...
-            .flowOn(dispatcher.io)
+        val conversationsWithReminders = if (queryConfig.fromArchive) {
+            conversations
+        } else {
+            conversations.combine(conversationRepository.observeAdminlessDeleteReminderPreviews()) { pagingData, reminders ->
+                if (reminders.isEmpty()) {
+                    pagingData
+                } else {
+                    val remindersByConversationId = reminders.associateBy { it.conversationId }
+                    pagingData.map { conversation ->
```
MohamadJaara, line 74: i am not sure why do we need to observe these separately vs folding this data with the preview

MohamadJaara (changes requested):
few things need to be cleared before merging. mainly why making this as a message and observing it separately in the use case complicating the flow, when we can make it a column in the conv that X conversaion will be deleted

sbakhtiarov:
@MohamadJaara 
I was considering two approaches:
1. Only system message and all logic based on it.
2. New field per conversation (pending_deletion) with the deletion timestamp.

My decision was based on two things:
- This is a very rare scenario, 99.9% of users will never see this event/status. I am not sure it worth adding a new field for such scenario.
- This is a terminal scenario for the group. It will be deleted and there is no way to cancel this process and recover the group.

MohamadJaara:
there is a nother it can be its own table for this and when you open a conv you read the value from there and show the banner, what i am trying to avoid is this complex ordering of messages that can be simplified a lot

MohamadJaara (commented):
the second PR in the stack remove stuff that where added in the first PR maybe we can collabs both into one or also adjust the first one to remove the removed stuff, to make it easier to review the first one without guessing what will be removed

PR #4389, opened 2026-08-13
feat: Get crypto service report use case (WPB-27920)

ohassine:
<!--do not remove this marker, its needed to replace info when ticket title is updated -->
<!--jira-description-action-hidden-marker-start-->
https://wearezeta.atlassian.net/browse/WPB-27920
<!--jira-description-action-hidden-marker-end-->
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Description

Reports which JCA provider actually serves each cryptographic call site in kalium, so the Android debug screen can show it. Provider selection happens at runtime and varies by device, OEM and OS version, so it has to be read off a real handset.

CryptoServiceRegistry (core:cryptography) holds one record per call site; each site reports the instance the platform handed back:                                                 
                                                                                                                                                                                     
  - AESUtils — asset IV, asset key, asset cipher                                                                                                                                     
  - SecureRandom wrapper (androidMain, jvmMain) — database secret and random passwords, via getInstanceStrong()
  

Needs releases with:

### Testing

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/utils/CryptoServiceRegistry.kt
```diff
@@ -0,0 +1,59 @@
...
+     */
+    fun record(usage: CryptoUsage, record: CryptoServiceRecord) {
+        records = records + (usage to record)
+    }
+
+    fun recorded(): List<Pair<CryptoUsage, CryptoServiceRecord>> =
+        records.let { snapshot -> CryptoUsage.entries.mapNotNull { usage -> snapshot[usage]?.let { usage to it } } }
+
+    internal fun clear() {
+        records = emptyMap()
```
MohamadJaara: this is overkill and over-engineered; it is more than enough to have another function in the return the provider from SecureRandom and print that on the screen

MohamadJaara (changes requested):

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/utils/CryptoServiceRegistry.kt
ohassine: A second function isn't tied to what the app actually uses: if someone changes the implementation and doesn't update it, we end up showing wrong values

MohamadJaara: i will take it with a comment to remember updating that use case when updating the secure random over this, because it will never be updated
lets not add complexity for if something might be changed in the long future (which it will not)

yamilmedina (approved):

MohamadJaara (approved):

PR #4418, opened 2026-08-24
feat(crypto): migrate to Core Crypto v10 [WPB-26599]

MohamadJaara:
https://wearezeta.atlassian.net/browse/WPB-26599

## Summary

Migrates Kalium from Core Crypto 9.3 to the official Core Crypto 10.4 artifacts. The MLS, Proteus, and E2EI integrations now use the v10 APIs, and the legacy Core Crypto selection flag and obsolete PKI plumbing are removed.

## What changed

- Use the official Core Crypto artifacts for Android, JVM, and Apple targets.
- Adapt database setup, client IDs, credentials, key packages, MLS transport, and buffered-message replay to the v10 API.
- Replace the legacy ACME flow with Core Crypto's managed `X509CredentialAcquisition` flow and `PkiEnvironmentHooks`.
- Persist acquisition snapshots and credential-rotation checkpoints so interrupted enrollment can resume safely.
- Reconcile PKI trust anchors through Core Crypto and move revocation handling into its PKI environment.
- Update affected tests, public ABI files, documentation, and third-party notices.

## Breaking changes

`E2EIEnrollmentResult.Initialized` now contains only:

- `target`
- `oAuthClaims`
- `isNewClientRegistration`

The legacy ACME nonce, order, and authorization fields were removed because Core Crypto now owns that state.

`finalizeEnrollment` now expects the identity-provider ID token. Its `oAuthState` parameter remains for source compatibility but is no longer used.

## JavaScript

`core:cryptography` still compiles for JS, but Core Crypto is currently a platform stub: `coreCryptoCentral` is not implemented and the module's JS crypto tests are disabled. Browser/WASM bindings are intentionally deferred to a follow-up.


MohamadJaara:
@claude review

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/e2ei/E2EIRepository.kt
```diff
@@ -15,392 +15,885 @@
...
-                Either.Right(acmeMapper.fromDto(apiResponse, response))
-            })
+private class E2EICredentialAcquisitionWorkflow(
+    private val dependencies: E2EIRepositoryDependencies,
+    private val checkpointStore: E2EIRotationCheckpointStore,
+    private val pkiEnvironmentMutex: Mutex,
+    private val fetchTrustAnchors: suspend (CoreCryptoCentral, String) -> Either<E2EIFailure, Unit>,
+    private val fetchFederationCertificates: suspend (CoreCryptoCentral, String) -> Either<E2EIFailure, Unit>
+) {
+    suspend fun start(isNewClient: Boolean): Either<E2EIFailure, E2EIAuthenticationRequest> =
```
sbakhtiarov: Should we clear the checkpoint when start() fails to avoid blocking it permanently?

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/utils/CryptoMappers.kt
```diff
@@ -147,23 +132,57 @@ fun E2eiConversationState.toCryptography(): E2EIConversationState = when (this)
...
-fun DecryptedMessage.toBundle() = DecryptedMessageBundle(
-    message = message,
-    commitDelay = commitDelay?.toLong(),
-    senderClientId = senderClientId?.let { CryptoQualifiedClientId.fromEncodedString(it.copyBytes().decodeToString()) },
-    hasEpochChanged = hasEpochChanged,
-    identity = identity.toCryptography(),
-    crlNewDistributionPoints = crlNewDistributionPoints
-)
+fun DecryptedMessage.toBundle(): DecryptedMessageBundle = when (this) {
+    is DecryptedMessage.Text -> DecryptedMessageBundle(
```
Garzas: Nit/suggestion: for readability, could we preserve the Text / Commit / Proposal distinction here with a sealed interface? This would make the expected fields for each message type explicit and avoid using nullable fields as an implicit discriminator.

Garzas (approved):

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/CryptoCredential.kt
```diff
@@ -17,49 +17,32 @@
...
-    var authorizations: List<String>
-)
+/**
+ * A Core Crypto credential acquired independently of a client transaction.
+ *
+ * The underlying native type deliberately stays internal so higher layers cannot couple
+ * themselves to the Core Crypto API. A credential becomes persistent only after it is
+ * installed by Core Crypto.
+ */
+interface CryptoCredential {
```
typfel, line 27: The implementation of `E2EIClient`  has been deleted, should this file be renamed?

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/PkiTrustAnchors.kt
```diff
@@ -0,0 +1,228 @@
...
+)
+
+/**
+ * Plans a complete trust-anchor replacement without mutating the PKI environment.
+ *
+ * Core Crypto identifies a trust anchor by the SHA-256 fingerprint of the certificate's
+ * DER-encoded SubjectPublicKeyInfo. Calculating the same key here lets reconciliation avoid
+ * re-adding unchanged roots and remove roots that are no longer present in the backend bundle.
+ */
+internal fun planPkiTrustAnchorReconciliation(
```
typfel: What's the purpose of this?

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/PkiTrustAnchors.kt
```diff
@@ -0,0 +1,228 @@
...
+            .values
+            .toList(),
+        fingerprintsToRemove = currentByFingerprint
+            .filterKeys { it !in desiredByFingerprint }
+            .keys
+            .map { it.bytes.copyOf() }
+    )
+}
+
+internal fun pkiTrustAnchorFingerprint(pem: CertificateChain): ByteArray {
```
typfel: Who needs this?

core/cryptography/src/commonTest/kotlin/com/wire/kalium/cryptography/ProteusClientTest.kt
```diff
@@ -32,8 +32,7 @@ import kotlin.test.assertNotNull
 import kotlin.test.assertTrue
 
 @OptIn(ExperimentalCoroutinesApi::class)
-// After removing cryptobox, proteus is not implemented on iOS and JS
-@IgnoreJS
+// Proteus is not implemented on iOS after removing cryptobox.
```
typfel, line 35: Dont' iOS get proteus from core crypto?

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/E2EIClientPlatform.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+}
+
+internal class CryptoCredentialRefImpl(
+    credentialRef: CredentialRef,
+    private var ownsCredentialRef: Boolean = true
+) : CryptoCredentialRef {
+    private var nativeCredentialRef: CredentialRef? = credentialRef
+
+    internal fun borrow(): CredentialRef = checkNotNull(nativeCredentialRef) {
```
typfel: This borrow/takeOwnership seems do nothing but flip boolean, can it perhaps be removed?

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/E2EIClientPlatform.kt
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    override fun exportPem(): String = exportedPem
+
+    internal fun borrow(): Credential = checkNotNull(nativeCredential) {
+        "CryptoCredential has already been closed"
+    }
+
+    override fun close() {
```
typfel: You very rarely need to explicitly `close()` core crypto objects. They'll be freed automatically by the garbage collector.

sbakhtiarov (approved):

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
```diff
@@ -0,0 +1,504 @@
...
+        groupId: MLSGroupId,
+        users: List<CryptoQualifiedID>
+    ): Map<String, List<WireIdentity>> = groupId.toCrypto().useNative { nativeGroupId ->
+        users.useNativeMapped({ Uuid(it.value) }) { nativeUsers ->
+            coreCrypto.getUserIdentities(nativeGroupId, nativeUsers).toCryptographyAndClose()
+        }
+    }
+
+    override suspend fun <R> transaction(name: String, block: suspend (context: MlsCoreCryptoContext) -> R): R {
+        var pendingCredentialSelection: PendingCredentialSelection? = null
```
typfel: This seems unnecessarily complex I don't really understand why need `pendingCredentailSelection`

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
```diff
@@ -0,0 +1,504 @@
...
+                    cipherSuite = defaultCipherSuite,
+                    credentialType = CoreCredentialType.BASIC
+                ).takeNewest()
+            } else {
+                null
+            }
+            val activeCredential = x509Credential ?: basicCredential ?: coreCrypto.transaction { context ->
+                val basicCredential = Credential.basic(defaultCipherSuite, clientId)
+                try {
+                    context.addCredential(basicCredential)
```
typfel: There's some misunderstanding here. You only need to add a credential once typically on login. The code in this method searches for already added credentials and then adds them again this has not effect.

typfel (changes requested):
Ok, did first look through and the credential handling definitely needs some work and I it makes sense to fix that first before looking deeper at the rest of the changes.

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/CryptoCredential.kt
MohamadJaara: I renamed E2EIClient.kt to CryptoCredential and E2EIClientPlatform to CryptoCredentialPlatform these files wrap credentials, so the new names match their role

core/cryptography/src/commonTest/kotlin/com/wire/kalium/cryptography/ProteusClientTest.kt
MohamadJaara: the comment is old and got updated now, however i did try to enable the tests again but they failed during proteus init for some reason i can try to make them work in another PR/change

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/E2EIClientPlatform.kt
MohamadJaara: yep simplified it

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/PkiTrustAnchors.kt
MohamadJaara: This keeps the trust anchors in Core Crypto equal to the bundle returned by the backend.
I checked the current iOS implementation. It calls addTrustAnchor directly, but Core Crypto 10.4 changed that method from replacement to addition. It also returns an error when the same anchor is added twice. Kalium refreshes the anchors during periodic sync and credential acquisition, so a direct add would fail on repeat runs and would retain roots removed by the backend.
The reconciliation handles those cases with the APIs available in 10.4. The cleaner fix would be a replaceTrustAnchors operation in Core Crypto so each client does not need to calculate SPKI fingerprints itself.

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/PkiTrustAnchors.kt
MohamadJaara: pkiTrustAnchorFingerprint is exposed as internal mainly for tests. The calculation is still needed by reconciliation, but the standalone helper can be removed and folded into TrustAnchorFingerprint

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/CoreCryptoCentralImpl.kt
```diff
@@ -178,50 +178,42 @@ class CoreCryptoCentralImpl(
...
-                    )
-                } ?: X509CredentialAcquisition(requirePkiEnvironment(), nativeConfig)
-            } finally {
-                nativeConfig.destroy()
-            }
-            acquisition.finalizeCredential()
+            } ?: X509CredentialAcquisition(requirePkiEnvironment(), nativeConfig)
         } finally {
-            existingCredentialRef?.close()
+            nativeConfig.destroy()
```
typfel: Can be removed

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
```diff
@@ -54,16 +54,12 @@ class MLSClientImpl private constructor(
         if (isClosed) return
         isClosed = true
         try {
-            activeCredentialRef.close()
+            clientId.close()
```
typfel: Can be deleted

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
```diff
@@ -106,38 +99,25 @@ class MLSClientImpl private constructor(
...
-                            }
-                            pendingCredentialSelection = selection
-                        }
-                    )
+        var transactionCredential = activeCredentialRef
+        val result = coreCrypto.transaction(name) { context ->
+            block(
+                mlsCoreCryptoContext(
+                    context = context,
+                    selectedCredential = { transactionCredential },
```
typfel: Why do you want bind a credential to a transaction context? Why not inject or fetch the default credential when doing an operation which requires it like (creating a group, create a key package). Like this you limting yourself to only create one type key package per transaction.

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
```diff
@@ -106,38 +99,25 @@ class MLSClientImpl private constructor(
...
+
+        activeCredentialRef = transactionCredential
+        return result
     }
 
     private fun mlsCoreCryptoContext(
         context: CoreCryptoContext,
         selectedCredential: () -> CredentialRef,
-        onCredentialSelected: (PendingCredentialSelection) -> Unit
+        onCredentialSelected: (CredentialRef) -> Unit
```
typfel: What needs this callback?

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
typfel, line 301: Is this used?

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/CryptoTransactionContext.kt
```diff
@@ -47,7 +45,10 @@ interface MlsCoreCryptoContext {
      *
      * @return list of generated key packages. NOTE: can be more than the requested amount.
```
typfel: > NOTE: can be more than the requested amount.

This is no longer true it will be exactly the amount requested

core/cryptography/src/commonMain/kotlin/com/wire/kalium/cryptography/PkiTrustAnchors.kt
typfel: Let's talk about this. We are not supposed to constantly replace the trust anchor it defeats their purpose.

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
MohamadJaara: yes, by E2EIRepository after credential rotation

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
MohamadJaara: it is gone now

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/CoreCryptoCentralImpl.kt
```diff
@@ -0,0 +1,390 @@
...
+                    cipherSuite = config.cipherSuite.toCrypto(),
+                    displayName = config.displayName,
+                    clientId = nativeClientId,
+                    handle = config.handle,
+                    domain = config.clientId.userId.domain,
+                    team = config.teamId,
+                    validityPeriodSecs = validityPeriodSeconds.toULong()
+                )
+            } catch (throwable: Throwable) {
+                nativeClientId.close()
```
typfel: can be deleted

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/CoreCryptoCentralImpl.kt
```diff
@@ -0,0 +1,390 @@
...
+            val acquisition = try {
+                existingCredentialRef?.let {
+                    X509CredentialAcquisition.newFromCredentialRef(
+                        pkiEnvironment = requirePkiEnvironment(),
+                        config = nativeConfig,
+                        credentialRef = it.unwrap()
+                    )
+                } ?: X509CredentialAcquisition(requirePkiEnvironment(), nativeConfig)
+            } finally {
+                nativeConfig.destroy()
```
typfel: can be deleted

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/CoreCryptoCentralImpl.kt
```diff
@@ -0,0 +1,390 @@
...
+            }
+            acquisition.finalizeCredential()
+        } finally {
+            existingCredentialRef?.close()
```
typfel: can be deleted

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
```diff
@@ -0,0 +1,476 @@
...
+import com.wire.kalium.cryptography.utils.toCryptography
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+@Suppress("TooManyFunctions")
+class MLSClientImpl private constructor(
+    private val coreCrypto: CoreCrypto,
+    private val clientId: ClientId,
+    private val defaultCipherSuite: CipherSuite,
+    private var activeCredentialRef: CredentialRef?,
```
typfel: The `activeCredentialRef` adds quite some complexity in order to pass is to `getPublicKey()` but couldn't this be generated property derived from the `defaultCipherSuite?`

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/client/E2EIClientProvider.kt
```diff
@@ -22,157 +22,88 @@ import com.wire.kalium.common.error.E2EIFailure
...
-import com.wire.kalium.util.KaliumDispatcherImpl
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
-import kotlinx.coroutines.withContext
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.days
 import kotlin.time.Duration.Companion.seconds
 
+/** Builds the immutable input for Core Crypto v10's X.509 acquisition state machine. */
 internal interface E2EIClientProvider {
```
typfel: This name of this class no longer matches what it does. It no longer provides a `E2EIClient`

core/cryptography/src/coreCryptoMain/kotlin/com/wire/kalium/cryptography/MLSClientImpl.kt
```diff
@@ -413,33 +397,16 @@ private fun Map<Uuid, List<com.wire.crypto.WireIdentity>>.toCryptographyAndClose
...
-    }
-    val defaultCredential = try {
-        filter { it.type() == defaultType }.maxByOrNull { it.earliestValidity() }
-    } catch (throwable: Throwable) {
-        forEach { it.close() }
-        throw throwable
+    val defaultType = if (any { it.type() == CoreCredentialType.X509 }) {
+        CoreCredentialType.X509
+    } else {
+        CoreCredentialType.BASIC
```
typfel, line 407: Ah so much simpler without the try catch close

typfel (approved):

PR #4426, opened 2026-08-27
feat: fetch apps and collaborators (WPB-25747)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-25747

https://wearezeta.atlassian.net/browse/WPB-25747
----

# What's new in this PR?

### Issues

Apps are represented by users, but clients in teams with more than 2,000 members are not guaranteed to know about every team member. Consequently, the local apps database may be incomplete and app search can incorrectly return no results.

### Solutions

Add versioned support for:
- GET /teams/{tid}/collaborators from API v10.
- GET /teams/{tid}/apps from API v15.

Introduce SyncAppsUseCase, exposed through AppScope, which:
- Performs no synchronization below API v10 or when the self user has no team.
- Fetches collaborators on API v10–v14.
- Fetches both team apps and collaborators on API v15 and later.
- Hydrates collaborator profiles in batches of 500, keeps only app users, deduplicates the results, and upserts them into the local apps database.
- Tolerates the known 403 insufficient-permissions response from the collaborators endpoint so directly returned team apps remain available.
- Preserves the existing cache when other network or storage operations fail.

### DIFF
--- a/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/app/AppRepository.kt
+++ b/logic/src/commonMain/kotlin/com/wire/kalium/logic/data/app/AppRepository.kt
@@ -17,25 +17,44 @@
  */
 package com.wire.kalium.logic.data.app
 
+import com.wire.kalium.common.error.CoreFailure
+import com.wire.kalium.common.error.NetworkFailure
 import com.wire.kalium.common.error.StorageFailure
+import com.wire.kalium.common.error.wrapApiRequest
 import com.wire.kalium.common.error.wrapFlowStorageRequest
 import com.wire.kalium.common.error.wrapNullableFlowStorageRequest
 import com.wire.kalium.common.error.wrapStorageNullableRequest
+import com.wire.kalium.common.error.wrapStorageRequest
 import com.wire.kalium.common.functional.Either
+import com.wire.kalium.common.functional.flatMap
+import com.wire.kalium.common.functional.flatMapLeft
+import com.wire.kalium.common.functional.foldToEitherWhileRight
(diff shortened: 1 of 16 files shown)


yamilmedina (approved):

logic/src/commonMain/kotlin/com/wire/kalium/logic/data/app/AppRepository.kt
```diff
@@ -46,10 +65,55 @@ internal interface AppRepository {
...
+
+    override suspend fun syncApps(teamId: TeamId, includeTeamApps: Boolean): Either<CoreFailure, Unit> =
+        fetchTeamApps(teamId, includeTeamApps).flatMap { teamApps ->
+            fetchCollaborators(teamId).flatMap { collaboratorIds ->
+                val teamAppIds = teamApps.mapTo(mutableSetOf()) { it.id.value }
+                val qualifiedCollaboratorIds = collaboratorIds
+                    .asSequence()
+                    .filterNot(teamAppIds::contains)
+                    .distinct()
+                    .map { collaboratorId -> UserId(collaboratorId, selfUserId.domain) }
```
ohassine, line 84: Could this be incorrect for external users? We qualify all collaborator IDs with `selfUserId.domain`, but collaborators may belong to a different domain

sbakhtiarov: As far as I understand there are no federated collaborators support. Both server and iOS implementations do not use unqualified UserId.

ohassine (approved):

PR #4432, opened 2026-09-01
fix: handling federation conflict when creating MLS conversation (WPB-26348)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-26348

https://wearezeta.atlassian.net/browse/WPB-26348

----

# What's new in this PR?

### Issues

MLS group creation with users from non-federating backends returned a generic failure and could leave a local conversation in PENDING_CREATION.

### Causes (Optional)

API v13 incorrectly classified MLS HTTP 409 responses through the obsolete ConflictWithMissingUsers path. The federation conflict was also lost across Core Crypto transport and treated as recoverable.

### Solutions

Use the standard federation and MLS error parsers, preserve conflicting backend domains through Core Crypto, and map the conflict as terminal. Tombstone the pending local conversation and acknowledge pending recovery actions. Remove the unused ConflictWithMissingUsers code.


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/conversation/createconversation/GroupConversationCreator.kt
```diff
@@ -140,6 +144,18 @@ internal class GroupConversationCreatorImpl(
...
 
+    override suspend fun discardPendingMLSGroupCreation(conversationId: ConversationId): Boolean {
+        val wasDeleted = conversationRepository.setConversationDeletedLocally(conversationId, true).fold(
+            { failure ->
+                kaliumLogger.w("Failed to discard pending MLS conversation: $failure")
+                false
+            },
+            { true }
+        )
+        pendingActionsRepository.acknowledgePendingMLSGroupJoins(listOf(conversationId))
```
ohassine, line 155: should we call acknowledgePendingMLSGroupJoins even on failure ?

sbakhtiarov: Yes, this call removes local conversation and stops recovery attempts.

ohassine (approved):

yamilmedina (approved):

PR #4436, opened 2026-09-02
feat: handle meeting TZID, support v17 API [WPB-27614]

saleniuk:
https://wearezeta.atlassian.net/browse/WPB-27614

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Handling TZID field for meetings, to calculate recurrences properly.
Also, added development support for v17, minimum required API version to support Wire Meetings feature has been raised to 17 as it's the first one that contains everything that's required for it to work properly, and apart from adding TZID, also `trial` field has been removed in v17, so it's also done in the code.

Needs releases with:

- https://github.com/wireapp/kalium/pull/4434

#### Test Coverage (Optional)

#### How to Test

Create recurring meeting in a location where the time zone changes during the year (DST time) and check that the meeting occurrence times stay the same even after changing the DST time.

Here is an example - weekly meeting starting at 10AM Berlin time and also shown from the perspective of someone in Sydney, where, in the Southern Hemisphere, time changes throughout the year are the opposite of those in the Northern Hemisphere and also occur on different dates, so there are transitional periods between the time changes in these two locations.
When viewing the calendar in Sydney time, the event should not stay at the fixed time, user should explicitly see the time change from 6:00 PM → 7:00 PM → 8:00 PM across those three distinct seasonal phases.

| Occurrence Date | Berlin Local Time | Berlin Offset | Underlying UTC Time | Time Gap | Sydney Local Time | Sydney Offset | Alignment Phase |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| **Wed, Sept 30** | 10:00 AM | UTC+2 | 08:00 AM UTC | **8 hours** | **06:00 PM** | UTC+10 | **Initial Setup:** Berlin in summer time (CEST) / Sydney in winter time (AEST). |
| *Sun, Oct 4* | *—* | *—* | *—* | *—* | *—* | *—* | *Sydney springs forward to summer time (AEDT; Clocks +1 hour)* |
| **Wed, Oct 7** | 10:00 AM | UTC+2 | 08:00 AM UTC | **9 hours** | **07:00 PM** | UTC+11 | **Intermediate Phase:** Both cities simultaneously in summer alignment. |
| *Sun, Oct 25* | *—* | *—* | *—* | *—* | *—* | *—* | *Berlin falls back to winter time (CET; Clocks -1 hour)* |
| **Wed, Oct 28** | 10:00 AM | UTC+1 | **09:00 AM UTC** | **10 hours** | **08:00 PM** | UTC+11 | **Final Phase:** Berlin in winter time (CET) / Sydney in summer time (AEDT). |

### Attachments (Optional)

https://github.com/user-attachments/assets/eed24b2a-ce82-4fd7-8a77-8b6376f59434

----

----

----
##### References


data/persistence/src/commonMain/kotlin/com/wire/kalium/persistence/dao/meeting/MeetingOccurrencesGenerator.kt
```diff
@@ -133,7 +133,7 @@ object MeetingOccurrencesGenerator {
         var nextCandidateStart: Instant? = candidateStart
         while (minDateLimit != null && nextCandidateStart != null && nextCandidateStart + duration <= minDateLimit) {
             nextCandidateStart = if (recurrence != null) {
-                val advancedStart = nextCandidateStart.plusPeriod(recurrence.frequency, interval)
+                val advancedStart = nextCandidateStart.plusPeriod(recurrence.frequency, interval, TimeZone.of(tzid))
```
sbakhtiarov: Not 100% sure about this TimeZone.of(tzid). As far as I understand these time zone IDs are changing sometimes and we may have a IllegalTimeZoneException thrown here if the current version of the library does not have an updated time zone id. Could you check it? Maybe add a fallback to not stop the generation because of one meeting with unrecognized time zone?

saleniuk: `TimeZone.of(tzid)` receives the time zone from the system. For JS it’s covered by using `js-joda/timezone`, on iOS Apple keeps devices IANA TZID up to date, and Google also tries to keep it up to date, but they also said that mainline support for Android 11 and lower concluded in Q4 2025. We support Androids starting from 9, which means that 9, 10, 11 no longer get systematic updates and can fall behind - this is roughly 8.8% of our users.

 Now, `kotlinx-datetime` team is actively working on that exact issue - to instead use a TZID database that’s being updated every time the IANA database gets updated, and this could be used on older devices as well, but it’s still not released yet. They have only updated the docs, but this feature will probably be included in the upcoming release of this library.

We should wait and use this release because it’s the best option - covered by the well-known library that we already use and multiplatform, but while we wait, we can apply some temporary solution by using `ThreeTenBP` on Android, which also updates quite regularly to keep the IANA TZID up-to-date.

Apart from that, the app also needs to be secured in case of `IllegalTimeZoneException`, so for now it won’t generate recurrences if the given TZID cannot be found. We already have a worker that generates missing occurrences regularly and it also happens during the slow sync, so if the app gets updated and starts supporting given TZID then recurrences will be generated properly.

**TLDR**: unknown TZID secured, temporarily used `ThreeTenBP` library to use recent IANA TZID database, should be replaced with multiplatform `kotlinx-database-zoneinfo` when released

core/util/src/commonTest/kotlin/com/wire/kalium/util/TzidUtilTest.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Test
+    fun givenUnsupportedTimezone_whenAddingDays_thenReturnsNull() {
+        val result = TzidUtil.plusDaysOrNull(instant = Instant.parse("2026-03-28T09:00:00Z"), days = 1, tzid = "Unsupported/TZID")
+        assertNull(result)
+    }
+
+    @Test
+    fun givenTimezone_whenAddingDaysAcrossDaylightSavingChange_thenKeepsTheSameLocalTime() {
```
ohassine, line 44: Can we also add a test for another timezone, e.g. America/New_York, to make sure the DST handling isn't specific to Berlin?
```
@Test
fun givenNewYorkTimezone_whenAddingDaysAcrossDaylightSavingChange_thenKeepsTheSameLocalTime() {
    val result = TzidUtil.plusDaysOrNull(
        instant = Instant.parse("2026-03-08T10:00:00-05:00"),
        days = 1,
        tzid = "America/New_York"
    )

    assertEquals(
        Instant.parse("2026-03-09T10:00:00-04:00"),
        result
    )
}
```

saleniuk: Sure, I added `America/New_York` as well, and also added verification that if it's not the DST change date in the given TZID then local time should stay the same

ohassine (approved):

sbakhtiarov (approved):

PR #4452, opened 2026-09-08
fix: meeting members only from self-team self-domain [WPB-28245]

saleniuk:
https://wearezeta.atlassian.net/browse/WPB-28245

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Currently, Wire Meetings should allow adding only people from same team and same domain to the meeting, instead of all contacts. In this PR, the flag allowing to filter only locally known contacts is replaced with a flag allowing to return only people from the same team and domain.
Search queries are merged for simplicity - the ones with "ExcludingAConversation" and without are grouped into single query without any loss of performance, simply it doesn't filter out conversation members if this parameter is null. Also added filtering for the team and domain. Also remote search is filtered by team and domain if needed, according to the [docs](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/157254411/Use+case+search+users#Backend-interaction).

#### Test Coverage (Optional)

#### How to Test

Create a meeting and try to add some participants - there should be no guests nor federated, only team members.

----

----

----
##### References

### DIFF
--- a/changelog.d/search-use-case-default-arguments.changed.md
+++ b/changelog.d/search-use-case-default-arguments.changed.md
@@ -1,0 +1,9 @@
+`SearchUsersByNameUseCase` and `SearchUsersByHandleUseCase` replace
+`skipRemoteSearch` with `onlySelfTeamAndDomain`. When `true`, the new parameter
+restricts searches to the self user's team and domain instead of skipping remote
+search. Callers using the `skipRemoteSearch` named argument must update it to
+`onlySelfTeamAndDomain` and review the changed behavior, including calls that pass
+the Boolean positionally.
+
+`excludingMembersOfConversation` and `customDomain` now default to `null`,
+allowing callers to omit these arguments.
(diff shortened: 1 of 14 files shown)


sbakhtiarov:
If I understood it correctly the filtering is done on the client side (does server support this?). This may lead to client receiving only top 30 search results and filtering them out locally and not showing matching contacts if they are not within first 30 results.

sbakhtiarov (approved):

saleniuk:
> If I understood it correctly the filtering is done on the client side (does server support this?). This may lead to client receiving only top 30 search results and filtering them out locally and not showing matching contacts if they are not within first 30 results.

Yes, there is no way of asking backend to filter it out remotely. The docs state that filtering of partners should be done on client side. Also, the result is always combined from both local and remote search, so in case of searching only within team members, remote will return users not known locally only if there is a large team (>2k users), so there should usually be someone from these 2k users who matches the search query, or if the search query is so precise that applies only to a few users, then remote search should already return the desired ones.

ohassine (approved):

PR #4462, opened 2026-09-14
fix(apple): encrypt the local databases with SQLCipher

mtrossbach:
Hi Wire team,

while working with Kalium on iOS, we noticed that the global and the user databases are stored unencrypted on Apple. This PR encrypts them with SQLCipher, as on Android.

### Encrypt the databases (commit 1)

On Apple, the databases were never keyed: `CoreLogic` and the user storage passed `passphrase = null`, and the native driver ignored a passphrase anyway (`// TODO encrypt database using sqlcipher`).

- `databaseDriver` takes the passphrase and keys every connection in SQLiter's `onCreateConnection`, before SQLiter reads the file. That includes the reader connections of a WAL database.
- `CoreLogic` gives the global database its raw key from `SecurityHelper.globalDBKeyMaterial()`, and `PlatformUserStorageProperties` gets a `userDbSecretProvider`, both under a `DatabaseKeyLock` as on Android. Apple databases never had a legacy key, so there is nothing to rekey.
- The keys are the raw keys that `SecurityHelper` already creates and keeps in the Keychain, so SQLCipher skips its key derivation.

**Where SQLCipher comes from:** Kalium doesn't link SQLCipher on Apple. CoreCrypto's iOS library bundles it and exports the SQLite API, so in an app that uses Kalium, the SQLite that SQLiter calls is SQLCipher. We checked this in the iOS simulator: with CoreCrypto in the binary, `PRAGMA cipher_version` reports SQLCipher; without it, SQLiter runs on the system's SQLite. Plain SQLite would accept `PRAGMA key` and write the data unencrypted, so a connection is keyed only after `PRAGMA cipher_version` has confirmed SQLCipher; otherwise opening fails.

### Encrypt existing unencrypted databases (commit 2)

Existing installations have unencrypted files, which wouldn't open with a key. Before a driver opens a file with a key, `encryptPlaintextDatabase` looks at its first 16 bytes: an unencrypted file starts with the SQLite header, one that SQLCipher encrypted with a random salt. An unencrypted file is exported with `sqlcipher_export` into an encrypted copy, which is renamed over it.

The export runs with a rollback journal, which checkpoints a WAL database first, so no `-wal` file is left over. The unencrypted file stays intact until the rename, and a copy left behind by an interrupted run is deleted before the next one.

## Compatibility

- A changelog fragment, `apple-database-encryption.added.md`.
- ABI: the Apple `databaseDriver` takes an optional `passphrase`, and the Apple `PlatformUserStorageProperties` constructor a `userDbSecretProvider`; Kalium constructs it itself. `:logic:checkKotlinAbi` passes.
- Encryption follows `KaliumConfigs.shouldEncryptData()`, as on the other platforms.

## Not part of this PR

Databases in an App Group container, shared with an app extension: iOS terminates a suspended app that holds a lock on a file it doesn't recognize as SQLite, which is why CoreCrypto keeps the first 32 bytes of its keystore readable (`cipher_plaintext_header_size`). Kalium doesn't use App Group containers yet. Once it does, its databases will need the same.

## Testing

- `:data:persistence:iosSimulatorArm64Test`: 851 tests, 0 failures. The Apple test base now opens the databases with a raw key and in WAL mode, as the SDK does, so all DAO tests run encrypted. For that, the Apple tests of `data/persistence` link CoreCrypto, as apps do.
- New: `SqlCipherKeyTest` (key, wrong key, reader connections of a WAL database) and `PlaintextDatabaseMigrationTest` (unencrypted WAL database with its content and user version, an encrypted database left as it is, an interrupted run).
- `:logic` and `:domain:userstorage` compile for `iosSimulatorArm64`, and detekt passes.
- Not tested: an app on a device, and macOS.

Thanks for having a look.


Garzas:
Thanks for the PR and for adding SQLCipher support on Apple!

Do you already have a production iOS app using Kalium with unencrypted databases, or is this migration mainly intended for development/test installations?

If existing users need this migration, one scenario worth checking is the app and a notification extension opening the same database simultaneously after an update. Both could attempt the plaintext migration using the same `.encrypting` file. The current lock protects key retrieval, but not the migration itself. A process-local lock would also not serialize access between the app and its extension.

I haven’t reproduced this race, and it only applies if multiple SDK instances can access the same database. If all production installations will start with encrypted databases, this is mainly a development/test concern.

PR #4479, opened 2026-09-16
feat(drive): expose the pdf rendition of documents (WPB-28773)

ohassine:
<!--do not remove this marker, its needed to replace info when ticket title is updated -->
<!--jira-description-action-hidden-marker-start-->
https://wearezeta.atlassian.net/browse/WPB-28773
<!--jira-description-action-hidden-marker-end-->
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### What                                                                                                                                                                               
                                                                                                                                                                                     
  Makes the PDF rendition the backend generates for convertible documents (Word, PowerPoint, spreadsheets) available to clients, so a user who can't edit such a document can still  
  view it.                                                                                                                                                                           
                                                                                                                                                                                     
###  Why                                                                                                                                                                                
                                                                                                                                                                                     
  Previews[] on a node contains several renditions — image previews in different sizes and, for documents the backend can convert, a PDF of the whole file. The mapping kept only contentType.endsWith("jpg") entries, so the PDF rendition never reached the domain layer and Android had no way to display a document read-only.                                                                                                                                                           
                                                                                                                                                                                     
###   How                                                                                                                                                                                
                                                                                                                                                                                     
  - `NodePreview` carries contentType, and `Previews[]` is now mapped once, keeping every usable rendition instead of pre-filtering to thumbnails.                                       
  - Consumers select what they need through named extensions rather than re-deriving the rule:
 List<NodePreview>?.imagePreviewUrl()   // largest image rendition — thumbnails                                                                                                     
 List<NodePreview>?.pdfPreviewUrl()     // the PDF rendition — read-only display                                                                                                    
  - `Node.File` exposes both flattened URLs (previewUrl, pdfPreviewUrl), selected from that one list in toFileModel().                                                                 
  - New `GetPdfPreviewUrlUseCase` for callers that only hold an asset id (the Android message list builds attachments from the DB, not from a node listing).                           
  - `RefreshCellAssetStateUseCase` now picks the thumbnail by type. Two latent defects surfaced while doing so and are fixed here:                                                     
    - ` previews.maxBy { it.dimension }` threw on an empty list, and with mixed content could have handed a PDF URL to the image loader;                                                
    - the preview retry loop stopped at the first rendition of any kind, so a document whose PDF was generated before its thumbnail could end up with no thumbnail saved.            
                                                                                                                                                                                     
Needs releases with:

### Testing

#### Test Coverage (Optional)

#### How to Test

### Notes (Optional)

### Attachments (Optional)

----

----

----
##### References


saleniuk (approved):

domain/cells/src/commonMain/kotlin/com/wire/kalium/cells/domain/usecase/RefreshCellAssetStateUseCase.kt
```diff
@@ -182,8 +179,14 @@ internal fun NetworkFailure.isAssetNotFound(): Boolean {
...
 
+/**
+ * Whether the backend generates renditions for this kind of file.
+ *
+ * It deliberately does not look at [CellNode.previews]: renditions are generated asynchronously
+ * after upload, so a file that has just been received has none yet, and that is exactly the case
+ * the caller has to wait for.
+ */
 public fun CellNode.isPreviewSupported(): Boolean = when {
-    previews == null -> false
```
sbakhtiarov, line 186: Not sure about the changes here and in CellNodeDTO. 
Now these two scenarios do not stop the app from retrying preview downloading:
- errors (failed to generate previews)
- no previews (before the check previews == null meant this file type does not support preview generation and preview polling is not needed).

ohassine: - The "no previews" case not stopping is intentional, `previews == null` never meant "this type doesn't support previews," it also happens right after upload while renditions are still generating. Stopping there would break preview loading for freshly received files again.
- The "errors" case was a fair miss — fixed it now.

sbakhtiarov (approved):

PR #4482, opened 2026-09-17
feat: support multi-ingress SSO login (WPB-27414)

sbakhtiarov:
https://wearezeta.atlassian.net/browse/WPB-27414

https://wearezeta.atlassian.net/browse/WPB-27414
----

# What's new in this PR?

### Issues

Multi-ingress IdP-change detection only worked for the email/password login path. On SSO login the client had no way to know whether the environment supports multi-ingress, so it could not compare identity providers and fell back to blocking the user with an "account already in use" error.

Additionally, `AddAuthenticatedUserUseCase` reported `SsoIdentityChanged` whenever the incoming login supplied an IdP ID and the retained account had none. A stored `null` means "no login ever recorded an IdP" — every account row created before the column was added in `db_global/migrations/10.sqm` is `null` — so the first SSO re-login of any pre-existing retained account looked like an IdP change and offered to wipe the user's local data.

### Solutions

- Added `SystemSettingsApi` (`GET /system/settings`) and `SystemSettingsResponse`, exposing the backend's `ssoIdpChangeDetectionEnabled` flag.
- Added `TransientAuthenticatedNetworkContainer`: a short-lived authenticated client that uses the fresh SSO tokens to read that setting *before* a session is stored, and never refreshes or persists the credentials. `FetchPendingLoginSystemSettings` wraps it and closes it after the single call.
- `GetSSOLoginSessionUseCase.invoke` takes `checkIdpChangeDetection` and reports the result on `SSOLoginSessionResult.Success.isIdpChangeDetectionEnabled`, so callers can supply the IdP ID only when the backend supports the comparison.
- `SSOInitiateLoginResult.Success` now carries `identityProviderId` and `serverConfigId`, letting consumers correlate the SSO redirect with the IdP that started it.
- `hasSsoIdentityChanged` now reports a change only when **both** IdP IDs are known and differ. An account with no recorded IdP logs in normally and has its IdP recorded, so every subsequent change is detected.


logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/AddAuthenticatedUserUseCase.kt
```diff
@@ -110,8 +110,12 @@ public class AddAuthenticatedUserUseCase internal constructor(
...
 
     private fun hasSsoIdentityChanged(previousIdpId: String?, currentIdpId: String?): Boolean =
-        // The IdP ID is supplied only by the email-based multi-ingress flow.
-        currentIdpId != null && previousIdpId != currentIdpId
+        // Email SSO supplies an IdP directly; SSO-code login supplies one only when the backend enables detection.
+        // A null previous value means no login ever recorded an IdP for this account: the row predates the migration
+        // that added the column, or every login so far supplied none. A known IdP is never cleared afterwards
+        // (Accounts.sq insertOrReplace coalesces it), so null is never a deliberate "no IdP". With nothing to compare
+        // against we cannot claim a change; the login continues and records the IdP for later comparisons.
+        previousIdpId != null && currentIdpId != null && previousIdpId != currentIdpId
```
Garzas, line 118: This creates a security gap for accounts migrated with `previousIdpId == null`. On the first login after the migration, any IdP that returns the same `QualifiedID` will be accepted and the retained database and cryptographic material will be reused.

Unless Security/Product explicitly accepts this behavior, `null -> known` should still trigger `SsoIdentityChanged`. The current condition only protects accounts after an IdP has already been recorded.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/sso/FetchPendingLoginSystemSettings.kt
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package com.wire.kalium.logic.feature.auth.sso
+
+import com.wire.kalium.common.error.CoreFailure
+import com.wire.kalium.common.error.wrapApiRequest
+import com.wire.kalium.common.functional.Either
+import com.wire.kalium.common.functional.map
+import com.wire.kalium.network.api.model.SessionDTO
+import com.wire.kalium.network.networkContainer.TransientAuthenticatedNetworkContainer
```
Garzas, line 26: This currently breaks `LayerAccessRulesTest.useCasesShouldNotAccessNetworkLayerDirectly`. Both this file and `GetSSOLoginSessionUseCase.kt` depend directly on network-layer types.

Could we move the transient settings request behind a repository or data-source abstraction that returns a logic-layer model? The current Kalium PR has a failing `Gradle Run Tests` job because of this violation.

changelog.d/multi-ingress-sso-unknown-idp.fixed.md
```diff
@@ -0,0 +1,11 @@
+`AddAuthenticatedUserUseCase` no longer reports `SsoIdentityChanged` for a retained
+account that has no previously recorded SSO identity provider ID. An identity provider
+change is now reported only when both the stored and the incoming IdP ID are known and
+differ; a login against an account with no recorded IdP succeeds and records the IdP for
+later comparisons.
+
+  - ABI: no change
```
Garzas, line 7: `ABI: no change` does not match this PR. It changes the public `SSOInitiateLoginResult.Success` constructor and the signature of `GetSSOLoginSessionUseCase.invoke`.

The former is source and binary incompatible, while the latter is binary incompatible despite the default argument. Please either preserve the existing public signatures through compatible overloads/constructors or document the accepted breaking change and migration guidance.

logic/src/commonMain/kotlin/com/wire/kalium/logic/feature/auth/AddAuthenticatedUserUseCase.kt
saleniuk: I think similar - if we cannot compare then it's safer to not assume that they are the same and instead just treat is as a changed one 🤔

data/network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/SystemSettingsApi.kt
```diff
@@ -0,0 +1,44 @@
...
+
+interface SystemSettingsApi : BaseApi {
+    suspend fun settings(): NetworkResponse<SystemSettingsResponse>
+
+    companion object {
+        const val MIN_API_VERSION = 4
+    }
+}
+
+internal class SystemSettingsApiImpl(
```
saleniuk, line 34: This is done differently than the other APIs - we usually keep only interface in `network/api/base/` and implement proper versions in `network/api/v{version_number}`

sbakhtiarov: I am not sure about the best way to handle this one. This SystemSettingsApi is different from other API calls since it requires the authenticated network client (with bearer token), but the call is done before authenticated network container is created (during SSO login).
The original implementation was following existing pattern, but I had to pass the token as parameter directly to API function which I also did not like.

saleniuk (approved):

Garzas (approved):

